PR #25447, opened 2014-10-23
Bug 1087981 - [Message] Lazy init settings in notify.js r=schung

julienw:


try-server-hook:
[julienw Julien Wajsberg (julienw)](https://github.com/julienw) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=7d799d0f04ba)

apps/sms/js/notify.js
```diff
@@ -56,16 +73,22 @@
 
   var Notify = {
     ringtone: function notification_ringtone() {
-      if (settings['audio.volume.notification'] &&
-          settings['notification.ringtone']) {
-        ringtone(settings['notification.ringtone']);
-      }
+      return getSettings(
+        ['audio.volume.notification', 'notification.ringtone']
```
steveck-chung: Looks like we'll reuse these settings keys frequently. Maybe define these string to global scope?

apps/sms/js/notify.js
```diff
@@ -4,31 +4,48 @@
...
+
+  function getSettings(settings) {
+    return Promise.all(
+      settings.map(getSetting)
+    ).catch((e) => {
+      // catch and log errors
+      console.error('Error while retrieving settings', e.message, e);
+      return settings.map(() => null);
+    }).then((results) => {
+
```
steveck-chung: nit: Maybe we don't need space here

julienw: right

apps/sms/js/notify.js
julienw: actually we don't reuse them so much, we use them only when we receive a message and we're in the thread for this message (and maybe class0 messages too). That's why I choose to retrieve the setting only when we need it, because in these cases we don't mind some ms latency. I thought that adding an observer for these cases would lead to more code running than just retrieving when we need them.

What do you think?

steveck-chung: Sorry for the missleading, I just mean 'string', not the actual settings value. I agree that there is no need to cache the settings value here.

julienw: ok I get it, good idea, I'll do it :)

try-server-hook:
[julienw Julien Wajsberg (julienw)](https://github.com/julienw) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=27d7dd5fc762)

try-server-hook:
[julienw Julien Wajsberg (julienw)](https://github.com/julienw) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=51e633462df7)

julienw:
merged in 0368004d46ae09c3e3694d087da3c7c37d27139c

PR #25465, opened 2014-10-23
Bug 913903 - Navigator UI based on what contact does

hfiguiere:


hfiguiere:
With NGA this is irrelevant. Closing for good.

PR #25812, opened 2014-11-04
Bug 1090777 - No error message is displayed when importing VCard with so...

adrm:
...me non-UTF8 contacts


ADLR-es:
Marked as duplicated. Closing PR.

PR #25854, opened 2014-11-05
Bug 1087494 - [Gallery] Update Image Edit Crop Flow

punamdahiya:


autolander:

try-server-hook:
[punamdahiya punamdahiya](https://github.com/punamdahiya) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=dafc73bc5acf)

autolander:

try-server-hook:
[punamdahiya punamdahiya](https://github.com/punamdahiya) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=3f8260767693)

apps/gallery/index.html
```diff
@@ -14,6 +14,7 @@
     <!-- Reponsive design for tiny & large device -->
     <link rel="stylesheet" href="style/gallery_tablet.css" type="text/css" media="(min-height: 768px) and (min-width: 768px)" />
     <link rel="stylesheet" type="text/css" href="shared/style/status.css">
+    <link rel="stylesheet" type="text/css" href="shared/style/action_menu.css" />
```
punamdahiya: TBD: should be lazy loaded

autolander:

try-server-hook:
[punamdahiya punamdahiya](https://github.com/punamdahiya) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=99d4f137324a)

autolander:

try-server-hook:
[punamdahiya punamdahiya](https://github.com/punamdahiya) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=2c5a81d5a02d)

apps/gallery/js/ImageEditor.js
```diff
@@ -13,16 +13,20 @@ var editOptionButtons =
...
 $('edit-crop-button').onclick = setEditTool.bind(null, 'crop');
 $('edit-effect-button').onclick = setEditTool.bind(null, 'effect');
 $('edit-enhance-button').onclick = setEditTool.bind(null, 'enhance');
-$('edit-crop-none').onclick = undoCropHandler;
 $('edit-header').addEventListener('action', onCancelButton);
 $('edit-tool-apply-button').onclick = applyEditTool;
 $('edit-save-button').onclick = saveEditedImage;
+
 editOptionButtons.forEach(function(b) { b.onclick = editOptionsHandler; });
+editAspectMoreButtons.forEach(function(b) {b.onclick = editAspectMoreHandler;});
```
davidflanagan: You could also register a single handler on the menu or form element and let the click bubble up, then use e.target instead of this in the handler function. But I guess doing it this way more closely parallels the existing code, so is probably better in this case.

apps/gallery/js/ImageEditor.js
```diff
@@ -371,10 +400,14 @@ function setEditTool(editTool) {
...
-      imageEditor.edit(function() {
-        imageEditor.showCropOverlay();
-      });
+        // Initialize the source rectangle to full size of the original image
+        imageEditor.resetCropRegion();
+        // Show screen with crop overlay on entering crop edit tool
+        imageEditor.requestNewPreview(function() {
+          // Use saved crop region in editSettings to show crop overlay
+          imageEditor.showCropOverlay(
+            imageEditor.getCropOverlayRegion(savedEditSettings.crop));
```
davidflanagan: At this point editSettings and savedEditSettings are still the same, right? I'd think that it is better to use editSettings here, since you're not trying to go back and restore anything.

apps/gallery/js/ImageEditor.js
```diff
@@ -405,23 +439,13 @@ function cancelEditTool() {
...
-        imageEditor.forceSetCropRegion();
-      }
-
+      // Set source rectangle size from saved editSettings
+      // to go back to previous cropped size preview.
+      imageEditor.forceSetCropRegion();
+      // Discard changes and revert to previous crop sizes
+      imageEditor.requestNewPreview(function() {
+        imageEditor.hideCropOverlay();
+      });
```
davidflanagan: I haven't looked ahead yet to find the definition of requestNewPreview(), but at this point of the code, I don't understand why this case does not need a call to edit() and the other two cases do. What is actually displaying the new preview here?

apps/gallery/js/ImageEditor.js
```diff
@@ -716,7 +720,7 @@ function ImageEditor(imageBlob, container, edits, ready, croponly) {
   // The destination (preview) rectangle of canvas. Device pixels.
   this.dest = {};
   // Region displayed in crop overlay during drags. CSS pixels.
-  this.cropRegion = {};
+  this.cropOverlayRegion = {};
```
davidflanagan: I like this name for it.

apps/gallery/js/ImageEditor.js
```diff
@@ -776,10 +780,8 @@ function ImageEditor(imageBlob, container, edits, ready, croponly) {
     this.original.onload = function() {
       // Initialize the crop region to the full size of the original image
       self.resetCropRegion();
-      self.resetPreview();
-
       // Display an edited preview of it
-      self.edit(function() {
+      self.requestNewPreview(function() {
```
davidflanagan: Okay, so at this point it looks to me like requestNewPreview() must be calling edit() internally. That's not what I'd expect. Since edit() calls generateNewPreview() as needed, I'd expect that requestNewPreview would just set a flag saying that the current preview is out of date. Then, next time edit() is called, the preview will be created before being displayed. If you do it that way, then you'd just call requestNewPreview() before calling edit(), but wouldn't replace your edit() calls.  Of course, I haven't read all of this code yet.

apps/gallery/js/ImageEditor.js
```diff
@@ -867,14 +869,6 @@ ImageEditor.prototype.generateNewPreview = function(callback) {
   };
 };
 
-ImageEditor.prototype.resetPreview = function() {
-  if (this.preview.src) {
-    URL.revokeObjectURL(this.preview.src);
-    this.preview.removeAttribute('src');
-    $('edit-enhance-button').classList.add('disabled');
-  }
-};
```
davidflanagan: This resetPreview() is what I was expected requestNewPreview() was doing...

apps/gallery/js/ImageEditor.js
```diff
@@ -915,6 +908,22 @@ ImageEditor.prototype.resize = function() {
...
+    this.preview.removeAttribute('src');
+    $('edit-enhance-button').classList.add('disabled');
+  }
+
+  this.edit(function() {
+    if (callback) {
+      callback();
+    }
+  });
+};
```
davidflanagan: Okay, I see now that requestNewPreview() really is just the same as calling resetPreview and then calling edit, right?  In some places this saves you a line of code. But in others (like the case where the user cancels the edit) you end up with extra code because you could no longer just have a single edit() call after the switch statement and had to add edits to the exposure and filters case.  

I prefer having these as two separate functions because it allows us to invalidate the preview without forcing a new one to be created right away. If you called resetPreview() twice (for some reason) before calling edit() no extra work is done. But if you call requestNewPreview() twice, it redraws twice.

Not really a big deal. I slightly prefer having separate functions. If you do want to stick with this one function, though, I recommend you rename to displayNewPreview() to indicate that it does actually display the new preview.

apps/gallery/js/ImageEditor.js
```diff
@@ -915,6 +908,22 @@ ImageEditor.prototype.resize = function() {
...
+    URL.revokeObjectURL(this.preview.src);
+    this.preview.removeAttribute('src');
+    $('edit-enhance-button').classList.add('disabled');
+  }
+
+  this.edit(function() {
+    if (callback) {
+      callback();
+    }
+  });
```
davidflanagan: The nested function here does nothing. You can just write `this.edit(callback)`

apps/gallery/js/ImageEditor.js
```diff
@@ -1195,42 +1187,42 @@ ImageEditor.prototype.drawCropControls = function(handle) {
   context.beginPath();
```
davidflanagan: Answering my own question, I think that the canvas is using CSS pixels, so these lines widths are already implicitly multiplied by devicePixelRatio

apps/gallery/js/ImageEditor.js
```diff
@@ -1567,35 +1574,17 @@ ImageEditor.prototype.cropImage = function(callback) {
...
-  this.resetPreview();
-  // Adjust the image
-  var self = this;
-  this.edit(function() {
-    // Hide and reshow the crop overlay to reset it to match the new image size
-    self.hideCropOverlay();
-    self.showCropOverlay();
+  this.requestNewPreview(function() {
     if (callback) {
       callback();
```
davidflanagan: If the hide and show calls aren't necessary anymore then you don't need the `if (callback)` either and can just pass the callback directly to requestNewPreview()

apps/gallery/js/ImageEditor.js
```diff
@@ -1656,6 +1645,22 @@ ImageEditor.prototype.getCropRegion = function() {
   };
 };
 
+// Covert the cropped region in image pixels to object with left, right, top,
+// bottom properties in CSS Pixel to be used to show crop overlay
+ImageEditor.prototype.getCropOverlayRegion = function(region) {
+  var self = this;
```
davidflanagan: No need to use self here since there is no nested function.

apps/gallery/js/ImageEditor.js
```diff
@@ -1656,6 +1645,22 @@ ImageEditor.prototype.getCropRegion = function() {
   };
 };
 
+// Covert the cropped region in image pixels to object with left, right, top,
+// bottom properties in CSS Pixel to be used to show crop overlay
+ImageEditor.prototype.getCropOverlayRegion = function(region) {
```
davidflanagan: This would be clearer, I think, if you renamed the region argument to 'cropSettings'. Since the argument comes from editSettings.crop, this new name helps to distinguish it from the the region that is the return value.

apps/gallery/js/ImageEditor.js
```diff
@@ -1656,6 +1645,22 @@ ImageEditor.prototype.getCropRegion = function() {
...
+
+  var cropOverlay = {};
+  cropOverlay.left = Math.floor((region.x * self.scale) / dpr);
+  cropOverlay.top = Math.floor((region.y * self.scale) / dpr);
+  cropOverlay.right = Math.floor(((region.x + region.w) *
+                                 self.scale) / dpr);
+  cropOverlay.bottom = Math.floor(((region.y + region.h) *
+                                  self.scale) / dpr);
+  return cropOverlay;
+};
```
davidflanagan: Given that this function only gets called once, it seems like there ought to be some refactoring where it did not need to be a public method like this. But I can't think of how to do that cleanly, so a single-use function like this is the best we can do for now.

apps/gallery/locales/gallery.en-US.properties
```diff
@@ -100,11 +100,14 @@ back-button.ariaLabel = Back
...
 crop-options.ariaLabel = Crop
-crop-revert.ariaLabel = Revert
 crop-aspect-free.ariaLabel = Custom aspect ratio
 crop-aspect-portrait.ariaLabel = Portrait aspect ratio
 crop-aspect-landscape.ariaLabel = Landscape aspect ratio
 crop-aspect-square.ariaLabel = Square aspect ratio
+crop-aspect-more.ariaLabel = More aspect ratios
+crop-aspect-3x4.ariaLabel = 3:4
+crop-aspect-4x3.ariaLabel = 4:3
+crop-aspect-16x9.ariaLabel = 16:9
```
davidflanagan: For these ariaLabel strings, I suspect that it would be better to be explicit and say "3:4 aspect ratio", etc.

Do these strings also set the text for the menu items, or do you need to define 

crop-aspect-3x4=3:4

and the other strings separately? 

If you're going to make the ariaLabel strings more verbose than the visible strings, then you'll need separate properties, of course.

I suppose you might also localize these (both the strings and the aria labels) to say "3:4 portrait" and "4:3 landscape" and "16:9 landscape" since there isn't an icon that shows the orientation of the rectangle.

And finally, I'd guess that these new strings will mean that uplifting this to 2.2 is out of the question. I assume that the l10n deadline has passed, but I don't actually know that for sure.

apps/gallery/style/gallery.css
```diff
@@ -667,26 +667,27 @@ footer {
...
 }
 #edit-crop-aspect-landscape {
-  left: 60%;
+  left: 40%;
   background-image: url('images/iconaction_camera_crop3x2_30x30.png');
 }
 #edit-crop-aspect-square {
-  left: 80%;
+  left: 60%;
   background-image: url('images/iconaction_camera_crop1x1_30x30.png');
```
davidflanagan: You're going to change the images to use the icon font, right?

apps/gallery/index.html
```diff
@@ -209,11 +210,11 @@ <h1 id="edit-title" data-l10n-id="edit"></h1>
         </div>
 
         <div id="edit-crop-options" class="edit-options-bar hidden" role="radiogroup" data-l10n-id="crop-options">
-          <a id="edit-crop-none" class="button" data-icon="undo" role="button" data-l10n-id="crop-revert"></a>
           <a id="edit-crop-aspect-free" class="selected radio button" role="radio" aria-checked="true" data-l10n-id="crop-aspect-free"></a>
           <a id="edit-crop-aspect-portrait" class="radio button" role="radio" data-l10n-id="crop-aspect-portrait"></a>
           <a id="edit-crop-aspect-landscape" class="radio button" role="radio" data-l10n-id="crop-aspect-landscape"></a>
           <a id="edit-crop-aspect-square" class="radio button" role="radio" data-l10n-id="crop-aspect-square"></a>
+          <a id="edit-crop-aspect-more" class="radio button" role="radio" data-icon="more" role="button" data-l10n-id="crop-aspect-more"></a>
```
davidflanagan: This <a> element for the more button has two role="" attributes, which doesn't seem like it can be right.

punamdahiya: That's correct, it's a miss will remove role="button"

apps/gallery/js/ImageEditor.js
punamdahiya: requestNewPreview generates and displays new preview, and updates editSettings.crop in generateNewPreview https://github.com/punamdahiya/gaia/blob/Bug1087494/apps/gallery/js/ImageEditor.js#L829
So at this point values in editSettings and savedEditSettings are different and we need to use  savedEditSettings.crop to show crop overlay with cropped size from last edit.

apps/gallery/js/ImageEditor.js
punamdahiya: requestNewPreview is internally calling edit() and displays the new preview

apps/gallery/js/ImageEditor.js
punamdahiya: I agree, will update to keep these as two separate functions resetPreview and edit, so that resetPreview can be used without calling edit every time.

apps/gallery/js/ImageEditor.js
punamdahiya: thanks, will update to use callback directly

apps/gallery/js/ImageEditor.js
punamdahiya: Oops, will remove

apps/gallery/locales/gallery.en-US.properties
punamdahiya: rop-aspect-3x4.ariaLabel = 3:4 is setting both the text and ariaLabel of the menu item. I like the suggestion for both the strings and the aria labels to say "3:4 portrait" and "4:3 landscape" and "16:9 landscape" . 

If we need to make aria label even more verbose then we should use separate strings.

I believe you are right, string freeze is usually in effect after FL date, https://wiki.mozilla.org/Release_Management/B2G_Landing

apps/gallery/style/gallery.css
punamdahiya: Yes, working on that and will update patch with those changes

autolander:

try-server-hook:
[punamdahiya punamdahiya](https://github.com/punamdahiya) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=4bd627eff6e7)

autolander:

autolander:

autolander:

autolander:

autolander:

PR #25926, opened 2014-11-07
Bug 1084298 - [Messages] Decoupling the all inputs query logic from DOM tree structure

steveck-chung:
- Add map for each view for maintaining the data for rendering/selection


try-server-hook:
[steveck-chung steveck-chung](https://github.com/steveck-chung) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=d0265ee15a49)

apps/sms/js/thread_list_ui.js
```diff
@@ -857,13 +846,7 @@ var ThreadListUI = {
...
     var recordTime = +thread.timestamp;
 
-    // For legitimate in-memory thread objects, update the stored
-    // Thread instance with the newest data. This check prevents
-    // draft objects from inadvertently creating bogus thread
-    // objects.
-    if (Threads.has(thread.id)) {
-      Threads.set(thread.id, thread);
-    }
+    Threads.set(thread.id, thread);
```
steveck-chung: Since we will store all the threads(including draft) into `Threads`, I remove this checking to prevent the draft thread missing in threadlist view. Or you think it's more appropiate to set the draft into Threads while in saveDraft?

apps/sms/js/threads.js
```diff
@@ -61,7 +61,8 @@
       body: body,
       timestamp: new Date(record.timestamp),
       unreadCount: (options && options.unread) ? 1 : 0,
-      lastMessageType: record.type || 'sms'
+      lastMessageType: record.type || 'sms',
+      isDraft: true
```
julienw: maybe set it to false in the previous function?

apps/sms/js/threads.js
```diff
@@ -14,12 +14,12 @@
       this[key] = thread[key];
     }
 
-    this.messages = [];
+    this.messages = new Map();
```
julienw: my main problem with using a map is that we lose the ordering we have with arrays... While we don't care about ordering in JS right now, I think we really need it in the future: ordering needs to be done in JS rather than using the DOM.

So maybe we need both an array and a Map. That said I'm fine if we look at this issue later.

apps/sms/js/thread_list_ui.js
```diff
@@ -375,11 +367,6 @@ var ThreadListUI = {
         return list;
       }, []);
 
-      // That means that we've just removed some drafts
```
julienw: anyway if I can finish my db experiment we'll get rid of the existing drafts db :) so better not think too much about this case.

apps/sms/js/thread_list_ui.js
julienw: I do exactly the same here: https://github.com/mozilla-b2g/gaia/pull/28681/files#diff-e2424773c46c9d5f4ff27c9c20b1bb0dL851

(I changed the name of Threads.set).

So ok for me :D

apps/sms/js/thread_list_ui.js
```diff
@@ -140,14 +140,8 @@ var ThreadListUI = {
...
-  getAllInputs: function thlui_getAllInputs() {
-    if (this.container) {
-      return Array.prototype.slice.call(
-        this.container.querySelectorAll('input[type=checkbox]')
-      );
-    } else {
-      return [];
-    }
+  getIdIterator: function thlui_getIdIterator() {
+    return Threads.getAll().keys();
```
julienw: this is the only place we use getAll(); so maybe it's better to have a Threads.keys() instead?

julienw:
I just had a first quick look but I didn't look at it thoroughly yet (and didn't test on a device either).

From a quick look it looks quite good though :)

apps/sms/js/selection_handler.js
```diff
@@ -47,6 +47,26 @@ SelectionHandler.prototype = {
...
+      itemIdSet.add(id.toString());
+    }
+    return itemIdSet;
+  },
+
+  allSelected: function sel_allSelected() {
+    /*jshint unused: false */
+    var allItemCount = 0;
+    var selectedCount = this.selectedCount;
+    for(var id of this.getIdIterator()) {
```
julienw: nit: please put a space after `for`

apps/sms/js/selection_handler.js
```diff
@@ -89,15 +105,13 @@ SelectionHandler.prototype = {
   // and if all messages are checked : deselect them all.
   toggleCheckedAll: function sel_toggleCheckedAll() {
     var selected = this.selected.size;
-    var allInputs = this.getAllInputs();
-    var allSelected = (selected === allInputs.length);
+    var allInputs = this.inputIdSet;
+    var allSelected = (selected === allInputs.size);
```
julienw: in the future, I'd really like a special case for "all selected" case. Instead of keeping all ids, I'd like a simple boolean. This would also help with case where the user taps "select all" while we're still loading the list.

apps/sms/js/selection_handler.js
```diff
@@ -89,20 +109,18 @@ SelectionHandler.prototype = {
   // and if all messages are checked : deselect them all.
   toggleCheckedAll: function sel_toggleCheckedAll() {
     var selected = this.selected.size;
```
julienw: I don't really like this var name, maybe you can rename it to "selectedSize", or even get rid of it?

apps/sms/js/thread_ui.js
```diff
@@ -1497,6 +1492,7 @@ var ThreadUI = {
       }
       taskQueue.push(() => {
         if (!this._stopRenderingNextStep) {
+          messageMap.set(message.id, message);
```
julienw: shouldn't we simply use the existing Threads.registerMessage ?

BTW in `Threads.registerMessage` we could move the first line (`Thread.create(message)`) to the if block.

apps/sms/js/thread_ui.js
```diff
@@ -1864,7 +1860,9 @@ var ThreadUI = {
     }
 
     // Removing from DOM all messages to delete
+    var messageMap = Threads.active.messages;
     for (var i = 0, l = list.length; i < l; i++) {
+      messageMap.delete(+list[i]);
```
julienw: we should implement a Threads.unregisterMessage instead, to achieve good encapsulation.

(I swear I saw some code that already implemented this, but it's likely in some old uncommited PR :/)

apps/sms/js/thread_ui.js
```diff
@@ -2425,6 +2420,7 @@ var ThreadUI = {
     button.setAttribute('data-l10n-id', 'downloading-attachment');
 
     request.onsuccess = (function retrieveMMSSuccess() {
+      Threads.active.messages.delete(id);
```
julienw: same: we should use a new Threads.unregisterMessage method

apps/sms/js/thread_ui.js
```diff
@@ -1935,20 +1933,17 @@ var ThreadUI = {
     }
   },
 
-  checkInputs: function thui_checkInputs() {
+  updateSelectionStatus: function thui_updateSelectionStatus() {
```
julienw: just a note (because we won't change this here): this is called each time we add a new message when we're in edit mode, so this consumes a lot of CPU if we enter edit mode while the thread is still loading.

julienw: that's also why I'd like a boolean to represent the "select all" state :)

apps/sms/js/thread_ui.js
```diff
@@ -2497,6 +2493,7 @@ var ThreadUI = {
            this.onMessageSendRequestCompleted();
         }.bind(this)
       };
+      Threads.active.messages.delete(id);
```
julienw: same

apps/sms/js/threads.js
julienw: (for consistency)

julienw:
Found an issue: when selecting a "isDraft" thread, the "mark as read/unread" button should not be enabled, but it is with your patch.

apps/sms/js/thread_list_ui.js
steveck-chung: ya agree

apps/sms/js/thread_ui.js
steveck-chung: Although entering edit mode while the thread is still loading is a rare case, I agree it's still nice if we could refine this part.

julienw: not that rare in big threads :)

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=95b2983c1e330fafb8a2416c60cccc23ca954874

apps/sms/js/thread_ui.js
```diff
@@ -1864,7 +1859,10 @@ var ThreadUI = {
     }
 
     // Removing from DOM all messages to delete
+    var messages = Threads.active.messages;
     for (var i = 0, l = list.length; i < l; i++) {
+      var message = messages.get(+list[i]);
+      Threads.unregisterMessage(message);
```
steveck-chung: This looks not very neat since we will need to fetch the message by id everytime when we need to delete bunch of messages in same thread, but I'm not sure if it's necessary to have another method for this case.

julienw: let's assume that messages.get is fast enough

apps/sms/js/threads.js
```diff
@@ -86,12 +88,19 @@
...
-      var thread = Thread.create(message);
       var threadId = message.threadId;
+
       if (!this.has(threadId)) {
-        this.set(threadId, thread);
+        this.set(threadId, Thread.create(message));
+      }
+      this.get(threadId).messages.set(message.id, message);
+    },
+    unregisterMessage: function(message) {
```
julienw: how about taking the message id instead of the full message?

steveck-chung: But how can we delete the message by message ID only? we will still need threadId at least to fetch the thread contains the message(unless we assume it's active thread)

julienw: I thought you could use the same line you currently use before calling unregisterMessage but I missed that you use Threads.active which we could not do here.

Can we have another Map for messages ? This wouldn't take much more memory because we'd have the same "message" object.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1d91e6a25972816bb31b958f07bbdef2d731b5d3

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3dac08e04a899681da77ac603abbff5a1bcf6c95

apps/sms/js/threads.js
```diff
@@ -86,12 +89,22 @@
...
+      var messageId = +message.id;
+
       if (!this.has(threadId)) {
-        this.set(threadId, thread);
+        this.set(threadId, Thread.create(message));
+      }
+      this.get(threadId).messages.set(messageId, message);
+      messageMap.set(messageId, threadId);
+    },
+    unregisterMessage: function(id = +id) {
```
julienw: I'm sorry, this doesn't work... the right part is used only when the argument is not provided when you call the function...

julienw: just use "id = +id" on the first line :)

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=08fdfa5148f8a6499f8af05fa1b85cde37d899df

autolander:
http://docs.taskcluster.net/tools/task-graph-inspector/#wt-KalZoQSCP20IdUaXsyQ

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=bf2f465d315f7c250815b275686b04e963731235

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ad4b2eba8412acf1122b447d9d963814ce4f1093

PR #26617, opened 2014-12-04
Bug 1098168 - BaseIcon

alivedise:


apps/system/js/airplane_mode.js
```diff
@@ -83,6 +83,12 @@
 
     _start: function() {
       this._watchList = {};
+      LazyLoader.load(['js/airplane_mode_icon.js']).then(function() {
+        this.icon = new AirplaneModeIcon(this);
+        this.icon.start();
+      }.bind(this))['catch'](function(err) {
```
etiennesegonzac: nit: why note `.catch()`? (easier to read imo)

apps/system/js/base_icon.js
```diff
@@ -0,0 +1,199 @@
...
+      this.index = '';
+    }
+    this.amendProperty();
+    this.manager = manager;
+    this.publish('created');
+  };
+  BaseIcon.prototype = Object.create(BaseUI.prototype);
+  BaseIcon.prototype.constructor = BaseIcon;
+  BaseIcon.prototype.DEBUG = false;
+  BaseIcon.prototype.camalToDash = function(strings) {
```
etiennesegonzac: cam**e**lToDash ?

apps/system/js/statusbar.js
```diff
@@ -160,91 +74,105 @@ var StatusBar = {
...
+      role="listitem"></div>
+    <div id="statusbar-time" class="sb-icon-time" role="listitem"></div>`;
+  },
+  /* jshint ignore: end */
+
+  renderIcons: function() {
+    this.statusbarIconsMax.insertAdjacentHTML('afterbegin', this.iconView());
+    Service.register('getIcon', this);
+    Service.register('startIcon', this);
+    Service.register('stopIcon', this);
```
etiennesegonzac: Looks like nobody is using `startIcon`/`stopIcon` (and that they're not implemented), but still reading :)

apps/system/js/sound_manager.js
```diff
@@ -248,6 +298,9 @@
         self.CEAccumulatorTime = value;
       }
     });
+
+    Service.registerState('isHeadsetConnected', this);
```
etiennesegonzac: I do share the concerns about `Service` getting messy really quickly.

But, interestingly, I think `isHeadsetConnected` is a valid use case :)
It is a piece of state global to the phone that is of interest for multiple modules.

That said, I'll go nit pick on other `Services` uses now :)

apps/system/js/base_icon.js
```diff
@@ -0,0 +1,199 @@
...
+    if (ele) {
+      this.element = ele;
+      this.element.classList.add('active');
+      this.onrender && this.onrender();
+      this.UPDATE_ON_START && this.update();
+      this.debug('started early');
+    } else {
+      // XXX: To prevent icon ordering regression,
+      // we are waiting statusbar rendered and fetch the element here.
+      Service.request('Statusbar:getIcon', this.instanceID).then(
```
etiennesegonzac: I don't think we have a strong enough justification to expose `Statusbar.getIcon` through `Service`.
We're only using the queuing support of `service.js`, but apart from that the `Statusbar` doesn't really need to expose `getIcon` to the world.

I don't see any elegant solution to solve the DOM sharing between the statusbar and icons, but I'd like to settle on something that doesn't "pollute" `Service`.

My suggestion is to dispatch a `statusbarrendered` event at the end of `StatusBar#renderIcons` and to wait for this event instead of doing the `Service.request('Statusbar:getIcon')`.

We could even make a small private (to the `base_icon.js` IIFE) helper in order to end up with something like:
```
    waitForElement().then(function(element, early) {
      this.element = element;
      this.element.classList.toggle('active', early);
      this.onrender && this.onrender();
      this.UPDATE_ON_START && this.update();
    }.bind(this)).catch(function(error) {
      console.log(error);
    });
```
`waitForElement` would maintain a `_statusbarRendered` piece of state global to `base_icons.js` and
* resolve `(element, true)` if `_statusbarRendered` is `false` but `getElementById` returns something
* otherwise wait for the event (and update `_statusbarRendered`) then resolve with `(element, false)`
* and resolve `(element, false)` right away if `_statusbarRendered` is `true`

apps/system/test/unit/telephony_monitor_test.js
```diff
@@ -0,0 +1,41 @@
...
+require('/shared/test/unit/mocks/mock_navigator_moz_telephony.js');
+requireApp('system/js/service.js');
+requireApp('system/js/base_module.js');
+requireApp('system/js/telephony_monitor.js');
+
+suite('system/TelephonyMonitor', function() {
+  var subject;
+
+  setup(function() {
+    subject = BaseModule.instantiate('TelephonyMonitor',
```
etiennesegonzac: nit: I think there's a trailing space here :)
(our linter doesn't raise this anymore I think :/)

apps/system/js/download/download_manager.js
```diff
@@ -47,16 +49,59 @@ var DownloadManager = (function() {
         window.addEventListener('will-shutdown', function onShutdown() {
           ev.download.pause();
         });
+      })['catch'](function(err) {
```
etiennesegonzac: same, why not `.catch` ?

apps/system/js/media_recording.js
```diff
@@ -33,6 +33,12 @@
       this.container = document.getElementById('media-recording-status-list');
 
       window.addEventListener('mozChromeEvent', this);
+      LazyLoader.load(['js/recording_icon.js']).then(function() {
+        this.icon = new RecordingIcon(this);
+        this.icon.start();
+      }.bind(this)).catch(function(err) {
+        console.error(err);
```
etiennesegonzac: ocd: trailing space

apps/system/js/network_activity.js
```diff
@@ -0,0 +1,44 @@
...
+    name: 'NetworkActivity',
+    _start: function() {
+      if (this.icon) {
+        return;
+      }
+      LazyLoader.load(['js/network_activity_icon.js']).then(function() {
+        this.icon = new NetworkActivityIcon(this);
+        this._handle_visibilitychange();
+      }.bind(this)).catch(function(err) {
+        console.error(err);
```
etiennesegonzac: here too :)

apps/system/js/nfc_manager.js
```diff
@@ -92,6 +89,12 @@
     start: function nm_start() {
       this._debug('Starting NFC Manager');
       this._hwState = this.NFC_HW_STATE.OFF;
+      LazyLoader.load(['js/nfc_icon.js']).then(function() {
+        this.icon = new NfcIcon(this);
+        this.icon.start();
+      }.bind(this)).catch(function(err) {
+        console.error(err);
```
etiennesegonzac: here too :)

apps/system/js/statusbar.js
```diff
@@ -160,91 +74,105 @@ var StatusBar = {
...
-          conn.removeEventListener('voicechange', this);
-          conn.removeEventListener('datachange', this);
-        }
-      );
-    }
+    window.addEventListener('iconcreated', this);
+    window.addEventListener('iconshown', this);
+    window.addEventListener('iconhidden', this);
+    window.addEventListener('iconchanged', this);
+    window.addEventListener('iconwidthchanged', this);
```
etiennesegonzac: we should probably add basic unit testing for those

apps/system/test/unit/app_install_manager_test.js
```diff
@@ -228,6 +228,7 @@ suite('system/AppInstallManager >', function() {
   });
 
   teardown(function() {
+    Service.request.restore();
```
etiennesegonzac: nit: this shouldn't be needed

apps/system/test/unit/app_install_manager_test.js
```diff
@@ -683,8 +684,8 @@ suite('system/AppInstallManager >', function() {
           setup(function() {
             // reseting these mocks as we want to test only one call
             MockNotificationScreen.mTeardown();
-            MockStatusBar.mTeardown();
-
+            Service.request.restore();
```
etiennesegonzac: what are all those `Service.request.restore()` in `setup` functions doing?

etiennesegonzac:
I like the individual icons unit test but I think some things are worth unit-testing in a `base_icon_test.js`
* getting the icon element (testing whichever solution we end up with)
* methods exposed through `Service` and the fact that they are properly exposed
* dispatched events

apps/system/js/base_icon.js
```diff
@@ -0,0 +1,197 @@
...
+    this.displayed() ? this.show() : this.hide();
+  };
+  BaseIcon.prototype.enabled = function() {
+    return this._started;
+  };
+  BaseIcon.prototype.updateLabel = function(type, active) {
+    if (!this.element && !this.isVisible()) {
+      return;
+    }
+    this.element.setAttribute('aria-label', navigator.mozL10n.get((active ?
```
timdream: It would, if you change the string ID in the file from `id` to `id.ariaLabel` -- l10n.js will pick up l10nId change, read the `id` and update `aria-label` accordingly.

apps/system/js/download/download_manager.js
timdream: This is a `gjslint` workaround :'( . Maybe we should add a `// XXX workaround gjslint` here?

apps/system/js/base_icon.js
```diff
@@ -0,0 +1,199 @@
...
+   */
+  BaseIcon.prototype.UPDATE_ON_START = true;
+  BaseIcon.prototype._fetchElements = function() {
+    this.element = document.getElementById(this.instanceID);
+  };
+  BaseIcon.prototype.show = function(force) {
+    if (!this.element) {
+      return;
+    }
+    var hidden = this.element.hidden;
```
timdream: I wonder if we could keep the `isHidden` information in JS? Not sure if reading this property aggressively (since we will reach this line or the one in `hide()` every time `update()` is called) will cause any JS <--> C performance issue.

apps/system/js/base_icon.js
```diff
@@ -0,0 +1,199 @@
...
+  BaseIcon.prototype.handleEvent = function(evt) {
+    this.update();
+  };
+  /**
+   * this.update() will be controlled by the icon's manager actively.
+   * Override this function if your icon has more behavior than
+   * just showing/hiding.
+   *
+   * If update() is not overrided, you should at least override shouldDisplay()
+   * to let update know how to show/hide the icon.
```
timdream: Nice `BaseIcon` design :+1:

apps/system/js/base_icon.js
yzen: as @timdream said, when we use the id, we need to change propertyAriaLabel=... to property.ariaLabel=.... args work fine as well there.

apps/system/js/airplane_mode.js
```diff
@@ -83,6 +83,10 @@
 
     _start: function() {
       this._watchList = {};
+      LazyLoader.load(['js/airplane_mode_icon.js']).then(function() {
```
KevinGrandon: I generally agree with this, but I'm not so sure it's needed in LazyLoader cases. I do think that we should be able to trust LazyLoader to basically never error - since it's just loading files from disk, and these locations should be known. I'm sure there are people that disagree with me here though.

apps/system/test/unit/app_install_manager_test.js
alivedise: If my memory goes well, removing the restore will cause the test to run twice on its own and the second run will fail. It looks like the appInstallManager has a nested test route and if we don't clear the stub, you will "TypeError: Attempted to wrap request which is already stubbed"

apps/system/test/unit/app_install_manager_test.js
alivedise: removed

apps/system/js/statusbar.js
alivedise: done; iconwidthchanged is covered in update in icon width suite

apps/system/js/media_recording.js
alivedise: done

apps/system/js/network_activity.js
alivedise: done (why jshint does not tell me :/

apps/system/js/nfc_manager.js
alivedise: done

apps/system/js/airplane_mode.js
alivedise: Because gjslint complains...(but it does not complain all files :/)

apps/system/js/base_icon.js
alivedise: done!

apps/system/js/statusbar.js
alivedise: Ya, should be deprecated.

apps/system/js/base_icon.js
alivedise: My concern is someday the statusbar may be lazy loaded as well. It should not block the bootstrap process of the whole system app.

I could remove the getIcon service for now, but if someday we want to lazy load statusbar, we will need to find out a solution to cooperate icon with statusbar rendering well.

apps/system/test/unit/telephony_monitor_test.js
alivedise: done

apps/system/js/download/download_manager.js
alivedise: comment added

apps/system/js/base_icon.js
alivedise: Re-read your comment I found I see what you proposed: use event listener to replace the Service.request but still a promise in BaseIcon. I will go for it. Not sure why we need to care getElementById returns something but not get statusbarrendered (it means statusbar dispatched the event before the request for element)

try-server-hook:
[alivedise Alive.Kuo (alivedise)](https://github.com/alivedise) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=ba1a775f5d47)

apps/system/js/mute_icon.js
```diff
@@ -0,0 +1,25 @@
...
+  MuteIcon.prototype.name = 'MuteIcon';
+  MuteIcon.prototype.update = function() {
+    if (!this.element || !this.enabled()) {
+      return;
+    }
+    var muteState = this.manager.vibrationEnabled ? 'vibration' : 'mute';
+    if (muteState === 'vibration') {
+      this.element.classList.add('vibration');
+    } else {
+      this.element.classList.remove('vibration');
```
gmarty: nit: `this.element.classList.toggle('vibration', muteState === 'vibration');` is easy to read IMHO.

apps/system/js/base_icon.js
alivedise: done!

apps/system/js/mute_icon.js
alivedise: good catch

try-server-hook:
[alivedise Alive.Kuo (alivedise)](https://github.com/alivedise) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=582ace8f683e)

try-server-hook:
[alivedise Alive.Kuo (alivedise)](https://github.com/alivedise) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=20a0553850b6)

try-server-hook:
[alivedise Alive.Kuo (alivedise)](https://github.com/alivedise) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=6e65828d15e6)

apps/system/js/base_icon.js
```diff
@@ -0,0 +1,206 @@
...
+        // Statusbar is not launched or not rendered yet.
+        window.addEventListener('statusbariconrendered', function render() {
+          window.removeEventListener('statusbariconrendered', render);
+          element = document.getElementById(self.instanceID);
+          resolve(element);
+        });
+      } else {
+        resolve(element);
+      }
+    });
```
etiennesegonzac: :+1:

apps/system/js/operator_icon.js
```diff
@@ -0,0 +1,67 @@
...
+    l10nArgs.operator = operatorInfos.operator;
+
+    if (operatorInfos.region) {
+      l10nArgs.operator += ' ' + operatorInfos.region;
+    }
+
+    label.dataset.l10nArgs = JSON.stringify(l10nArgs);
+
+    label.dataset.l10nId = 'statusbarLabel';
+    label.textContent = _('statusbarLabel', l10nArgs);
```
zbraniecki: this line is not needed. and the above two you can replace with:
navigator.mozL10n.setAttributes(label, l10nId, l10nArgs); :)

try-server-hook:
[alivedise Alive.Kuo (alivedise)](https://github.com/alivedise) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=f799fa8f5d6e)

try-server-hook:
[alivedise Alive.Kuo (alivedise)](https://github.com/alivedise) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=7bf78325509d)

apps/system/test/unit/tethering_monitor_test.js
```diff
@@ -0,0 +1,84 @@
...
+    });
+
+    test('Should be enabled if there is any client on wifi', function() {
+      MockNavigatorSettings.mTriggerObservers('tethering.wifi.connectedClients',
+        { settingValue: 2 });
+      assert.isTrue(subject.connected);
+      assert.isTrue(subject.icon.update.called);
+    });
+  });
+});
```
KevinGrandon: nit: newline.

PR #26796, opened 2014-12-15
Bug 1111637 - Create a new test case for the switching to camera from gallery and back

nojunpark:


tests/python/gaia-ui-tests/gaiatest/tests/functional/gallery/manifest.ini
```diff
@@ -39,3 +39,11 @@ skip-if = device == "desktop" && os == "linux" && debug
 skip-if = device == "desktop" && os == "linux" && debug
 
 [test_gallery_share_picture.py]
+
+[test_gallery_switch_to_camera.py]
+# Bug 1111637 - Create a new test case for the switching to camera from gallery and back
+external = true
```
mwargers: Sorry, I'm only seeing this now. This is not an external = true test, is it?

PR #26836, opened 2014-12-17
Bug 947884 - multiple page vorbis comment

hfiguiere:


try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=11f7d3733008)

apps/music/js/metadata/core.js
```diff
@@ -99,7 +99,7 @@ var AudioMetadata = (function() {
     // the appropriate parser.  Start off with an 64kb chunk of data.
     // If the metadata is in that initial chunk we won't have to read again.
     return new Promise(function(resolve, reject) {
-      var headersize = Math.min(64 * 1024, blob.size);
+      var headersize = Math.min(256 * 1024, blob.size);
```
jimporter: Why is this change necessary? (And if it is, please update the comment just before this.)

apps/music/js/metadata/ogg.js
```diff
@@ -39,76 +44,127 @@ var OggMetadata = (function() {
...
   /**
    * Read the identification header of an Ogg container. It's always the first
    * page.
    *
-   * @param {BlobView} page The audio file being parsed.
+   * @param {BlobView} blobview The audio file being parsed.
    */
-  function readIdentificationHeader(page) {
-    var header = readPageHeader(page);
+  VorbisCommentParser.prototype.readIdentificationHeader = function(blobview) {
```
jimporter: I tend to prefer `VorbisCommentParser.prototype = { /* ... */ };` I think that's how we do it most everywhere else in the music app, although it's not absolutely essential that you change this, if you have a reason for needing to do it this way.

jimporter: Do we need to pass the `blobview` to this function, given that you've already passed it to this object's constructor?

apps/music/js/metadata/ogg.js
```diff
@@ -39,76 +44,127 @@ var OggMetadata = (function() {
...
+   * Read the Vorbis Comment. Deal with multiple pages. Parse the
+   * picture block too.
    *
    * @param {BlobView} page The audio file being parsed.
    * @param {Metadata} metadata The (partially filled-in) metadata object.
    * @return {Promise} A Promise that resolves with the completed metadata.
    */
-  function readCommentHeader(page, metadata) {
-    var header = readPageHeader(page);
+  VorbisCommentParser.prototype.readVorbisComment = function(blobview) {
```
jimporter: Ditto here; we can probably remove the `blobview` argument.

apps/music/js/metadata/ogg.js
```diff
@@ -39,76 +44,127 @@ var OggMetadata = (function() {
...
+    }.bind(this));
+  };
+
+  /**
+   * Read the comment header of an Ogg container.
+   *
+   * @param {BlobView} bloview The audio file being parsed.
+   * @param {Metadata} metadata The (partially filled-in) metadata object.
+   */
+  VorbisCommentParser.prototype.readCommentHeader = function(blobview) {
```
jimporter: Remove `blobview` here too?

apps/music/js/metadata/ogg.js
```diff
@@ -116,14 +172,22 @@ var OggMetadata = (function() {
...
     }
 
+    /*var version =*/ page.readUnsignedByte();
+    /*var type =*/ page.readUnsignedByte();
+
+    page.advance(8); // granule
+
+    /*var serial =*/ page.readUnsignedInt();
+    /*var seq =*/ page.readUnsignedInt();
+
```
jimporter: We probably don't need these changes...

apps/music/js/metadata/ogg.js
```diff
@@ -116,14 +172,22 @@ var OggMetadata = (function() {
    * @param {BlobView} page The audio file being parsed.
    * @return {Object} An object containing the page's segment table.
    */
-  function readPageHeader(page) {
+  VorbisCommentParser.prototype.readPageHeader = function(page) {
```
jimporter: Same thing about `page`; is it needed here?

apps/music/js/metadata/ogg.js
```diff
@@ -139,41 +203,52 @@ var OggMetadata = (function() {
    * @param {BlobView} page The audio file being parsed.
    * @param {Metadata} metadata The (partially filled-in) metadata object.
    */
-  function readAllComments(page, metadata) {
-    var vendor_string_length = page.readUnsignedInt(true);
-    page.advance(vendor_string_length); // skip libvorbis vendor string
+  VorbisCommentParser.prototype.readAllComments = function(page, num_comments) {
```
jimporter: And finally, is the `page` argument needed here?

apps/music/js/metadata/ogg.js
```diff
@@ -182,16 +257,70 @@ var OggMetadata = (function() {
...
+        index: page.tell(),
+        size: page.remaining()
+      }, comment_length);
     }
 
     var comment = page.readUTF8Text(comment_length);
+    return decodeComment(comment);
+  }
+
+  function readContinuedComment(leftovers, comment_length, page) {
```
jimporter: I wonder if we could merge this with `readComment` above, or add a wrapper around the two. That would simplify how we call this code, since there'd only be a single entry point. I know this is pretty similar to how it is now, but it might make it easier to understand in the future.

apps/music/js/metadata/core.js
hfiguiere: Argh the comment. Yeah this is necessary albeit arbitrary. multipage mean over 64KB. My sample had ~150KB art blob.

This remain a band-aid approach.

apps/music/js/metadata/ogg.js
hfiguiere: ah yeah. I actually wanted to check the serial and sequence for reading multiple pages. Sounds like I forgot.

apps/music/js/metadata/core.js
jimporter: Couldn't we just fetch the entire Ogg page once we get into the Ogg code?

apps/music/js/metadata/ogg.js
hfiguiere: I wanted to minize the diff, that why I didn't do  ```VorbisCommentParser.prototype = { /* ... */ }; ``` Maybe I will just do that.

hfiguiere: As for blobview, maybe I'll just remove it from the constructor and ```this```

apps/music/js/metadata/ogg.js
hfiguiere: Will actually document the fields in a comment and keep the ```advance(22)```

apps/music/js/metadata/ogg.js
jimporter: > I wanted to minize the diff

That's what `git diff -b` is for. :)

hfiguiere: I know that's what it is for. But then, ```git blame```. Anyway.

try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=fce04ab3b0d3)

apps/music/js/metadata/core.js
```diff
@@ -96,10 +96,10 @@ var AudioMetadata = (function() {
     }
 
     // Read the start of the file, figure out what kind it is, and call
-    // the appropriate parser.  Start off with an 64kb chunk of data.
+    // the appropriate parser.  Start off with an 256kb chunk of data.
     // If the metadata is in that initial chunk we won't have to read again.
     return new Promise(function(resolve, reject) {
-      var headersize = Math.min(64 * 1024, blob.size);
+      var headersize = Math.min(256 * 1024, blob.size);
```
hfiguiere: I haven't fixed this yet.
The problem is that BlobView doesn't allow loading more if needed. I found that 256kb was a decent compromise between more data and not too much for what we want to do. It is not perfect, far from it.

hfiguiere:
Added the missing tests.

try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=0b8de88e1a2e)

apps/music/js/metadata/core.js
jimporter: BlobView definitely allows loading more data: see https://github.com/mozilla-b2g/gaia/blob/master/shared/js/blobview.js#L93

hfiguiere: uh oh. nevermind. I'll look for solution that is better than that.

hfiguiere: that being this PR.

try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=f15691bdc5f0)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=35e7490a81f208429aeaa7e1b618fb002164e633

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=151537ff38714d5651674e3df46c7a9170523fa1

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=dbb5255beed9ed9771aa2d539f406b870f134930

apps/music/js/metadata/ogg.js
```diff
@@ -39,131 +43,194 @@ var OggMetadata = (function() {
...
+     * @return {Promise} A Promise that resolves with the completed metadata.
+     */
+    readVorbisComment: function(blobview) {
+      return new Promise(function(resolve, reject) {
+
+        this.done = function(metadata) {
+          LazyLoader.load('js/metadata/vorbis_picture.js').then(() => {
+            return VorbisPictureComment.parsePictureComment(metadata);
+          }).then(resolve);
+        }.bind(this);
```
jimporter: You can switch these to using arrow functions and drop the `.bind(this)`

apps/music/js/metadata/ogg.js
```diff
@@ -39,131 +43,194 @@ var OggMetadata = (function() {
...
-          case 3:
-            valid = fullpage.readBinaryText(6) === 'vorbis';
-            metadata.tag_format = 'vorbis';
-            break;
-          case 79:
-            valid = fullpage.readBinaryText(7) === 'pusTags';
-            metadata.tag_format = 'opus';
-            break;
+        // Look for a comment header from a supported codec
+        var first_byte = fullpage.readByte();
```
jimporter: You can't get rid of the try try block here. Any exception thrown in the callback to `BlobView.getMore` will get lost.

apps/music/js/metadata/ogg.js
```diff
@@ -39,131 +43,194 @@ var OggMetadata = (function() {
...
 
-    var sum = function(a, b) { return a + b; };
-    var comment_length = Array.reduce(header.segment_table, sum, 0);
+    /**
+     * Read the comment header of an Ogg container.
+     *
+     * @param {BlobView} bloview The audio file being parsed.
+     * @param {Metadata} metadata The (partially filled-in) metadata object.
+     */
+    readCommentHeader: function(blobview) {
```
jimporter: This function is asynchronous, right? It should be returning a Promise.

apps/music/js/metadata/ogg.js
```diff
@@ -39,131 +43,194 @@ var OggMetadata = (function() {
...
+    readVorbisComment: function(blobview) {
+      return new Promise(function(resolve, reject) {
+
+        this.done = function(metadata) {
+          LazyLoader.load('js/metadata/vorbis_picture.js').then(() => {
+            return VorbisPictureComment.parsePictureComment(metadata);
+          }).then(resolve);
+        }.bind(this);
+
+        this.readCommentHeader(blobview);
```
jimporter: Since this function should be returning a Promise, you should return the result of it so that the caller can wait for resolution of this inner Promise.

apps/music/js/metadata/ogg.js
```diff
@@ -39,131 +43,194 @@ var OggMetadata = (function() {
...
+    readCommentHeader: function(blobview) {
+      var metadata = {};
+      var header = this.readPageHeader(blobview);
 
-    return new Promise((resolve, reject) => {
-      page.getMore(page.index, comment_length, (fullpage, err) => {
+      var sum = function(a, b) { return a + b; };
+      this.page_length = Array.reduce(header.segment_table, sum, 0);
+
+      blobview.getMore(blobview.index, this.page_length,
```
jimporter: This should be wrapped in a Promise.

apps/music/js/metadata/ogg.js
```diff
@@ -39,131 +43,194 @@ var OggMetadata = (function() {
...
+          valid = fullpage.readBinaryText(6) === 'vorbis';
+          metadata.tag_format = 'vorbis';
+          break;
+        case 79:
+          valid = fullpage.readBinaryText(7) === 'pusTags';
+          metadata.tag_format = 'opus';
+          break;
+        }
+        if (!valid) {
+          throw new Error('malformed ogg comment packet');
```
jimporter: Since you're in the callback for `BlobView.getMore`, you can't throw here. You need to use `reject`, since exceptions can get lost here: https://github.com/mozilla-b2g/gaia/blob/master/shared/js/blobview.js#L67

apps/music/js/metadata/ogg.js
```diff
@@ -39,131 +43,194 @@ var OggMetadata = (function() {
...
+        }
+
+        var vendor_string_length = fullpage.readUnsignedInt(true);
+        fullpage.advance(vendor_string_length); // skip libvorbis vendor string
+
+        var num_comments = fullpage.readUnsignedInt(true);
+        var readLoop = function(fullpage, err) {
+          if(err) {
+            throw new Error(err);
```
jimporter: Ditto, use `reject` here.

apps/music/js/metadata/ogg.js
```diff
@@ -39,131 +43,194 @@ var OggMetadata = (function() {
...
-            reject('malformed ogg comment packet');
-            return;
+          blobview.advance(this.page_length);
+
+          if(!this.readAllComments(fullpage, num_comments, metadata)) {
+            var header = this.readPageHeader(blobview);
+            this.page_length = Array.reduce(header.segment_table, sum, 0);
+
+            blobview.getMore(blobview.index, this.page_length, readLoop);
           }
```
jimporter: You probably want something like `else { resolve(metadata); } ` here?

apps/music/js/metadata/ogg.js
```diff
@@ -39,131 +43,194 @@ var OggMetadata = (function() {
...
-          } else if (comment.field !== 'picture') {
-            // We already have a value, so append this new one.
-            metadata[comment.field] += ' / ' + comment.value;
+    /**
+     * Read all the comments in an Ogg container.
+     *
+     * @param {BlobView} page The audio file being parsed.
+     * @param {int} num_comments The number of comment to be parsed.
+     * @param {Metadata} metadata The (partially filled-in metadata object.
+     */
```
jimporter: Please document the return value here.

apps/music/js/metadata/ogg.js
jimporter: It should also probably be prefixed with `_`, since it's a private function.

apps/music/js/metadata/ogg.js
```diff
@@ -39,131 +43,194 @@ var OggMetadata = (function() {
...
-    if (capture_pattern !== 'OggS') {
-      throw new Error('malformed ogg page header');
-    }
+    /**
+     * Read the header for an Ogg page.
+     *
+     * @param {BlobView} page The audio file being parsed.
+     * @return {Object} An object containing the page's segment table.
+     */
+    readPageHeader: function(page) {
```
jimporter: This is a private function, right? If so, please prefix with `_`.

apps/music/js/metadata/ogg.js
```diff
@@ -39,131 +43,194 @@ var OggMetadata = (function() {
...
-            // We already have a value, so append this new one.
-            metadata[comment.field] += ' / ' + comment.value;
+    /**
+     * Read all the comments in an Ogg container.
+     *
+     * @param {BlobView} page The audio file being parsed.
+     * @param {int} num_comments The number of comment to be parsed.
+     * @param {Metadata} metadata The (partially filled-in metadata object.
+     */
+    readAllComments: function(page, num_comments, metadata) {
```
jimporter: Prefix this with `_` as in the other cases.

apps/music/js/metadata/ogg.js
```diff
@@ -172,16 +239,70 @@ var OggMetadata = (function() {
...
+    var view = BlobView.getFromArrayBuffer(newBuffer.buffer, 0,
+                                           newBuffer.byteLength,
+                                           true);
+    if (comment_length > view.remaining()) {
+      throw new EndOfPageError({
+        page: view,
+        index: 0,
+        size: newBuffer.byteLength
+      }, comment_length);
+    }
```
jimporter: This code seems awfully complex for what's going on. I think I'd prefer something that works like BlobView, but is smart enough to expand its buffer size until it fits the whole field. Then this function wouldn't even be necessary, and we could just use `readComment`. Essentially, `readComment`'s job would be to do its best to read a whole comment, but if it can't, it would indicate failure, and ask the caller to expand the buffer and try again.

apps/music/js/metadata/ogg.js
```diff
@@ -172,16 +239,70 @@ var OggMetadata = (function() {
    */
   function readComment(page) {
```
jimporter: This seems like it belongs inside `VorbisCommentParser` now...

apps/music/js/metadata/ogg.js
```diff
@@ -172,16 +239,70 @@ var OggMetadata = (function() {
...
+    }
+    return decodeComment(view.readUTF8Text(comment_length));
+  }
+
+  /**
+   * Decode the vorbis comment as coming from the buffer.
+   * @param {String} comment The comment data.
+   * @return {Object} a key value pair.
+   */
+  function decodeComment(comment) {
```
jimporter: This should go in `VorbisCommentParser` as well.

apps/music/js/metadata/ogg.js
hfiguiere: done

apps/music/js/metadata/ogg.js
hfiguiere: done. in other places too.

apps/music/js/metadata/ogg.js
hfiguiere: done

apps/music/js/metadata/ogg.js
hfiguiere: done

apps/music/js/metadata/ogg.js
hfiguiere: done

apps/music/js/metadata/ogg.js
hfiguiere: done

apps/music/js/metadata/ogg.js
hfiguiere: done

apps/music/js/metadata/ogg.js
hfiguiere: good point. But that involve moving this.done() around: it is cleaner it is not the promise resolve.

apps/music/js/metadata/ogg.js
hfiguiere: moved inside _readAllComments() since they are all private.

apps/music/js/metadata/ogg.js
hfiguiere: same

apps/music/js/metadata/ogg.js
hfiguiere: The problem is that we need to read the page header first before resuming the comment decoding. I don't think we can avoid this complexity.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b0ab2daf509fb72185cb20816082637b544631ea

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=26aec3368b1de11d051ca7a9706b9f43c5c297f9

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ce31f13b7d76915dab6a2ffea1e019dfe031fa32

apps/music/js/metadata/core.js
```diff
@@ -57,11 +57,9 @@ var AudioMetadata = (function() {
     }
 
     // Read the start of the file, figure out what kind it is, and call
-    // the appropriate parser.  Start off with an 64kb chunk of data.
-    // If the metadata is in that initial chunk we won't have to read again.
+    // the appropriate parser.
     return new Promise(function(resolve, reject) {
-      var headersize = Math.min(64 * 1024, blob.size);
-      BlobView.get(blob, 0, headersize, function(header, error) {
+      BlobView.get(blob, 0, blob.size, function(header, error) {
```
jimporter: Does this mean we're reading the entire audio file into memory? I think that could cause major memory issues for large files (FLACs, podcasts, etc).

apps/music/js/metadata/ogg.js
```diff
@@ -95,109 +129,199 @@ var OggMetadata = (function() {
...
           }
           if (!valid) {
-            reject('malformed ogg comment packet');
-            return;
+            reject(new Error('malformed ogg comment packet'));
           }
 
-          readAllComments(fullpage, metadata);
+          var vendor_string_length = fullpage.readUnsignedInt(true);
+          fullpage.advance(vendor_string_length);// skip libvorbis vendor string
```
jimporter: Nit: add a space before the comment.

apps/music/js/metadata/ogg.js
```diff
@@ -39,48 +43,78 @@ var OggMetadata = (function() {
...
+
+        var sum = function(a, b) { return a + b; };
+        this.page_length = Array.reduce(header.segment_table, sum, 0);
+
+        blobview.getMore(blobview.index, this.page_length, (fullpage, err) => {
+          if (err) {
+            reject(Error(err));
+          }
 
-        try {
```
jimporter: I'm a little worried about removing this try block. Are you sure absolutely nothing can throw an exception here? `.getMore()` will be async if we don't already have the block of bytes we want, and exceptions that escape this function will probably cause a hang.

jimporter: As an example, some of the `read` functions below could throw if we hit EOF, which could happen for invalid Ogg files.

apps/music/js/metadata/ogg.js
```diff
@@ -95,109 +129,199 @@ var OggMetadata = (function() {
...
 
-          LazyLoader.load('js/metadata/vorbis_picture.js').then(() => {
-            return VorbisPictureComment.parsePictureComment(metadata);
-          }).then(resolve);
-        } catch(e) {
-          reject(e);
-        }
+          var num_comments = fullpage.readUnsignedInt(true);
+          var readLoop = (fullpage, err) => {
+            if(err) {
```
jimporter: Nit: add a space after `if`.

apps/music/js/metadata/ogg.js
```diff
@@ -95,109 +129,199 @@ var OggMetadata = (function() {
...
+              reject(new Error(err));
+            }
+            blobview.advance(this.page_length);
+
+            if(!this._readAllComments(fullpage, num_comments, metadata)) {
```
jimporter: Nit: add a space after `if`.

apps/music/js/metadata/ogg.js
```diff
@@ -95,109 +129,199 @@ var OggMetadata = (function() {
...
+              try {
+                var header = this._readPageHeader(blobview);
+                this.page_length = Array.reduce(header.segment_table, sum, 0);
+              } catch(e) {
```
jimporter: Nit: space after `catch`? I see people do both in our codebase...

apps/music/js/metadata/core.js
hfiguiere: I am pretty sure that it doesn't. We already have a blob and AFAIK a slice is just a slice, not a copy of the buffer. I'll make sure, but there is no other clear solution to that. And it wasn't in the previous review because it is my test app that uncovered it: the unit test just take the whole blob.

jimporter: A Blob doesn't have its contents loaded into memory yet (at least, I sure hope it doesn't). However, BlobViews have an ArrayBuffer of the current slice in-memory: https://github.com/mozilla-b2g/gaia/blob/master/shared/js/blobview.js#L60

hfiguiere: I'll ask David, but if this is the case, this is just totally wrong. :-/

apps/music/js/metadata/ogg.js
hfiguiere: This make it not pass the linter... moving comment up.

jimporter: Wait... the linter actually demands inline comments not have a space before them? That's gross.

hfiguiere: no. it complains about lines too long

hfiguiere: and the linter doesn't enforce space before braces.... :-1:

apps/music/js/metadata/ogg.js
hfiguiere: not sure why it was removed - probably when I rebased and I didn't pay attention.... I'll add it back, it is safer.

apps/music/js/metadata/core.js
davidflanagan: Jim's right. This proposed change would read the entire blob into memory. The old code just reads the first 64k into memory

PR #26842, opened 2014-12-17
Bug 1098609 - Always display the menu based on the supported network type of the sim slot r=eragonj

crh0716:
https://bugzilla.mozilla.org/show_bug.cgi?id=1098609

### DIFF
--- a/apps/settings/js/carrier.js
+++ b/apps/settings/js/carrier.js
@@ -70,8 +70,7 @@ var CarrierSettings = (function(window, document, undefined) {
     }
 
     cs_addVoiceTypeChangeListeners();
-    cs_updateNetworkTypeLimitedItemsVisibility(
-      _mobileConnection.voice && _mobileConnection.voice.type);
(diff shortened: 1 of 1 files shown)


apps/settings/js/carrier.js
```diff
@@ -289,11 +286,25 @@ var CarrierSettings = (function(window, document, undefined) {
...
+      autoSelectOperatorItem.hidden = availableOperatorsHeader.hidden =
+      availableOperators.hidden = (mode !== 'gsm');
+      roamingPreferenceItem.hidden = (mode !== 'cdma');
+    }
 
-    roamingPreferenceItem.hidden =
-      (_networkTypeCategory[voiceType] !== 'cdma');
+    if (!voiceType) {
+      getSupportedNetworkInfo(conn, function(result) {
+        if (result.gsm || result.wcdma || result.lte) {
```
pgravel: I think this is a reasonable workaround for the issue.

PR #26847, opened 2014-12-17
Bug 1093388 - [Browser] Requires an internet connection error is overlap...

gmarty:
...ped by check settings in landscape mode


apps/system/style/net_error.css
```diff
@@ -96,21 +98,29 @@ form[role="dialog"][data-type="action"] > menu > button.icon-settings {
 
 #failed-logo {
   display: inline-block;
-  width: 13.5rem;
   height: 13.5rem;
+  width: 7.9rem;
```
mikehenrty: It seems a little strange that this doesn't max the pixels of the 1x image, ie 7.4rem. Perhaps you are using that extra space as padding?

gmarty: That's because the image is shown slightly bigger than it is (height is 135px rather than 125px), so the width is stretched too.
But actually, it makes more sense not to resize it in portrait mode.

try-server-hook:
[gmarty Guillaume C. Marty (gmarty)](https://github.com/gmarty) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=dc208f1d099f)

PR #26897, opened 2014-12-19
Bug 1042361 - [email] Avoid disruptive notification modes if a notification for the account already exists

jrburke:
Uses a soundFile that is just silent and a vibrationPattern of [1] for subsequent updates to new email notifications, since "silent" option not available in current supported mozbehaviors.

Good hack or great hack?

### DIFF
--- a/apps/email/js/sync.js
+++ b/apps/email/js/sync.js
@@ -45,7 +45,8 @@ define(function(require) {
       console.log('email: notifications not available');
       sendNotification = function() {};
     } else {
-      sendNotification = function(notificationId, title, body, iconUrl, data) {
+      sendNotification = function(notificationId, title, body,
+                                  iconUrl, data, behavior) {
         console.log('Notification sent for ' + notificationId);
 
         if (Notification.permission !== 'granted') {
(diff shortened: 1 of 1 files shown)


try-server-hook:
[jrburke James Burke (jrburke)](https://github.com/jrburke) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=909faefc9067)

jrburke:
Based on mhenretty's comment in the bug, I updated this to use an non-existent "url" for the sound file. This results in the same effect of no sound played. While I do not like sending malformed URLs, hopefully this is a temporary measure until the standardized silent option is implemented. 

Diff from previous state (.opus file also removed):

```patch
diff --git a/apps/email/js/sync.js b/apps/email/js/sync.js
index da0c502..110ec3b 100644
--- a/apps/email/js/sync.js
+++ b/apps/email/js/sync.js
@@ -298,10 +298,11 @@ define(function(require) {
 
             // If already have a notification, then do not bother with sound or
             // vibration for this update. Longer term, the notification standard
-            // will have a "silent" option, but using what is available for now.
+            // will have a "silent" option, but using a non-existent URL as
+            // suggested in bug 1042361 in the meantime.
             if (existingData && existingData.count) {
               behavior = {
-                soundFile: app.origin + '/sounds/silent.opus',
+                soundFile: 'does-not-exist-to-simulate-silent',
                 // Cannot use 0 since system/js/notifications.js explicitly
                 // ignores [0] values. [1] is good enough for this purpose.
                 vibrationPattern: [1]

```

jrburke:
Forgot to mention, tested on flame device and confirmed it still works, no sound played.

PR #26917, opened 2014-12-19
Bug 1113219, remove is_video_playing() and replace it with is_playing to...

mwargers:
... fix intermittent failure


geoelectric:
LGTM

PR #26996, opened 2014-12-24
Bug 1087933 - Allow to add recipients to a message being written in a thread

autra:
We redirect to the Composer with current subject, text, attachments and
contacts.

### DIFF
--- a/apps/sms/locales/sms.en-US.properties
+++ b/apps/sms/locales/sms.en-US.properties
@@ -37,6 +37,7 @@ selectMessages-label  = Select Messages
 settings              = Settings
 add-subject            = Add subject
 remove-subject         = Remove subject
+includeSomeoneElse-label    = Include someone else
 
 # New Message
 to                         = To:
--- a/apps/sms/test/marionette/conversation_test.js
+++ b/apps/sms/test/marionette/conversation_test.js
@@ -1,4 +1,4 @@
-/* global require, marionette, setup, suite, test */
+/* global require, marionette, setup, suite, test, __dirname */
 'use strict';
 
 var assert = require('chai').assert;
@@ -9,14 +9,25 @@ var Messages = require('./lib/messages.js');
 var InboxView = require('./lib/views/inbox/view');
 var Storage = require('./lib/storage.js');
 var Tools = require('./lib/views/shared/tools.js');
+var MessagesActivityCaller = require('./lib/messages_activity_caller.js');
+var Keyboard = require('../../../keyboard/test/marionette/lib/keyboard.js');
 
 marionette('Conversation Panel Tests', function() {
+  var apps = {};
+  apps[MessagesActivityCaller.ORIGIN] = __dirname + '/apps/activitycaller';
 
   var client = marionette.client({
-    desiredCapabilities: { raisesAccessibilityExceptions: false }
+    desiredCapabilities: { raisesAccessibilityExceptions: false },
+    profile: {
+      prefs: {
+        'focusmanager.testmode': true
+      },
+
+      apps: apps
+    }
   });
 
-  var messagesApp, storage;
+  var messagesApp, storage, activityCallerApp, keyboardApp;
 
   function assertIsDisplayed(element) {
     assert.isTrue(element.displayed(), 'Element should be displayed');
@@ -24,7 +35,9 @@ marionette('Conversation Panel Tests', function() {
 
   setup(function() {
     messagesApp = Messages.create(client);
+    keyboardApp = new Keyboard(client);
     storage = Storage.create(client);
+    activityCallerApp = MessagesActivityCaller.create(client);
 
     client.loader.getMockManager('sms').inject([
       'test_storages',
@@ -551,6 +564,120 @@ marionette('Conversation Panel Tests', function() {
           'Edit mode header should show correct number of messages'
         );
       });
+    });
+  });
+
+  suite('Add recipients from thread', function() {
+    var conversationView;
+    setup(function() {
+      var thread = ThreadGenerator.generate({
+        numberOfMessages: 1
+      });
+      storage.setMessagesStorage([thread], ThreadGenerator.uniqueMessageId);
+
+      messagesApp.launch();
+
+      var inboxView = new InboxView(client);
+      conversationView = inboxView.goToConversation(thread.id);
+    });
+
+    test('With empty composer: recipients should be kept',
+    function() {
+      var newMessageView = conversationView.includeSomeoneElse();
+      // this also ensure we are on Composer view
+      assert.equal(newMessageView.recipients, '+123',
+          'Recipients should be kept');
+      assert.equal(newMessageView.messageText, '', 'SMS text should be empty');
+
+      // add a recipient, a subject and send a mms to create a thread with
+      // several recipients
+      newMessageView.typeMessage('message2');
+      newMessageView.addNewRecipient('+234');
+
+      newMessageView.showSubject();
+      newMessageView.typeSubject('subject1');
+      conversationView = newMessageView.send();
+
+      newMessageView = conversationView.includeSomeoneElse();
+      var recipients = newMessageView.recipients;
+      assert.lengthOf(recipients, 2, 'There should be 2 recipients');
+      assert.include(recipients, '+123', '+123 should be in the recipients');
+      assert.include(recipients, '+234', '+234 should be in the recipients');
+      assert.equal(newMessageView.messageText, '', 'MMS text should be empty');
+    });
+
+    test('Sms being written: message and recipients should be kept',
+    function() {
+      conversationView.typeMessage('message2');
+      var newMessageView = conversationView.includeSomeoneElse();
+      assert.equal(newMessageView.recipients, '+123',
+        'Recipient should be +123');
+      assert.equal(newMessageView.messageText, 'message2',
+        'message text should be kept');
+    });
+
+    test('Mms being written: all info must be kept', function() {
+      conversationView.typeMessage('message2');
+      // add a subject
+      conversationView.showSubject();
+      conversationView.typeSubject('subject1');
+      var newMessageView = conversationView.includeSomeoneElse();
+
+      // assert that we kept all the infos
+      assert.equal(newMessageView.recipients, '+123',
+        'Recipient should be +123');
+      assert.equal(newMessageView.messageText, 'message2',
+        'Message text should be kept');
+      assert.equal(newMessageView.subject, 'subject1',
+        'Subject should be kept');
+
+      // let's add a recipients and send the message.
+      // we will have a thread with several recipients
+      newMessageView.addNewRecipient('+234');
+      conversationView = newMessageView.send();
+      // write another MMS
+      conversationView.typeMessage('message3');
+      conversationView.showSubject();
+      conversationView.typeSubject('subject2');
+      // add a blob
+      conversationView.addAttachment().choose('Messages Activity Caller')
+        .pickImage();
+      messagesApp.switchTo();
+      // take screenshot of current attachment to compare it later
+      var beforeScreenshot = conversationView
+        .takeComposerAttachmentScreenshot(0);
+
+      // redirect to new message to add recipients
+      newMessageView = conversationView.includeSomeoneElse();
+
+      // assert the content
+      var recipients = newMessageView.recipients;
+      assert.lengthOf(recipients, 2, 'There should be 2 recipients');
+      assert.include(recipients, '+123', '+123 should be in the recipients');
+      assert.include(recipients, '+234', '+234 should be in the recipients');
+      assert.equal(newMessageView.messageText, 'message3',
+        'Message text should be kept');
+      assert.equal(newMessageView.subject, 'subject2',
+        'Subject should be kept');
+
+      // hide the keyboard: when it is displayed, the attachment is hidden,
+      // which prevents the screenshot to be taken
+      client.switchToFrame();
+      var systemInputMgmt = client.loader.getAppClass(
+        'system', 'input_management'
+      );
+      systemInputMgmt.waitForKeyboardFrameDisplayed();
+      keyboardApp.switchTo();
+      keyboardApp.longPressSpaceBar(0.7);
+      client.switchToFrame();
+      // wait for the keyboard to be hidden
+      systemInputMgmt.waitForKeyboardFrameHidden();
+
+      messagesApp.switchTo();
+      assert.equal(
+        beforeScreenshot,
+        newMessageView.takeComposerAttachmentScreenshot(0)
+      );
     });
   });
 });
--- a/apps/sms/test/marionette/lib/views/conversation/view.js
+++ b/apps/sms/test/marionette/lib/views/conversation/view.js
@@ -37,7 +37,7 @@ ConversationView.prototype = {
 
   get toggleSelectionButtonTitle() {
     return this.accessors.toggleSelectionButton.text();
-  },   
+  },
 
   /**
    * Returns the messages in this conversation, following an optional filter.
@@ -179,6 +179,15 @@ ConversationView.prototype = {
     return this.composerAccessors.addAttachment();
   },
 
+  /**
+   * Take a screenshot of an attachment thumbnail in the composer.
+   * @param {Number} The attachment index
+   * @return {String} The screenshot
+   */
+  takeComposerAttachmentScreenshot: function(index) {
+    return this.composerAccessors.takeAttachmentScreenshot(index);
+  },
+
   typeSubject: function(subject) {
     this.composerAccessors.subjectInput.sendKeys(subject);
   },
@@ -191,6 +200,20 @@ ConversationView.prototype = {
     this.showOptions();
     this.menuAccessors.selectAppMenuOption('Add subject');
     this.client.helper.waitForElement(this.composerAccessors.subjectInput);
+  },
+
+  /**
+   * Launch the "include someone else" action.
+   * @returns {NewMessageView} the new message view we navigated to.
+   */
+  includeSomeoneElse: function() {
+    this.showOptions();
+    this.menuAccessors.selectAppMenuOption('Include someone else');
+
+    var NewMessageView = require('../new-message/view');
+    var newMessageView = new NewMessageView(this.client);
+    newMessageView.accessors.waitToAppear();
+    return newMessageView;
   },
 
   hideSubject: function() {
@@ -227,7 +250,7 @@ ConversationView.prototype = {
       node.scrollIntoView();
     });
     this.accessors.toggleSelectionButton.tap();
-  }, 
+  },
 
   isSubjectVisible: function() {
     var subjectInput = this.composerAccessors.subjectInput;
--- a/apps/sms/views/conversation/js/conversation.js
+++ b/apps/sms/views/conversation/js/conversation.js
@@ -1184,11 +1184,12 @@ var ConversationView = {
 
   /**
    * Navigates user to Composer or Thread panel with custom parameters.
-   * @param {Object} parameters Navigation parameters. `number` and `messageId`
-   * are mutually exclusive in the current implementation.
+   * @param {Object} parameters Navigation parameters. `number`, `messageId` and
+   * content are mutually exclusive in the current implementation.
    * @param {String} [parameters.number] Phone number or e-mail to send a
    * message to.
    * @param {Number} [parameters.messageId] Template message to resend.
+   * @param {Object} [parameters.content] Content to resend.
    * @returns {Promise} Promise that is resolved once navigation is completed.
    * Promise is rejected if no navigation happened, especially if the user did
    * not want to discard an existing message.
@@ -1197,13 +1198,17 @@ var ConversationView = {
     var navigateToComposer = () => {
       var draftCreatePromise;
 
+      var draft;
       if (parameters.messageId) {
         draftCreatePromise = this.storeDraftFromMessage(parameters.messageId);
       } else if (parameters.number) {
-        var draft = new Draft({
+        draft = new Draft({
           recipients: [parameters.number],
           type: Utils.isEmailAddress(parameters.number) ? 'mms' : 'sms'
         });
+        draftCreatePromise = Drafts.add(draft).store().then(() => draft.id);
+      } else if (parameters.content) {
+        draft = new Draft(parameters.content);
         draftCreatePromise = Drafts.add(draft).store().then(() => draft.id);
       } else {
         throw new TypeError('Unknown parameter');
@@ -1916,8 +1921,12 @@ var ConversationView = {
     }
     params.items.push(subjectItem);
 
-    // If we are on a thread, we can call to SelectMessages
+    // If we are on a thread, we can call to SelectMessages and add recipients
     if (Navigation.isCurrentPanel('thread')) {
+      params.items.push({
+        l10nId: 'includeSomeoneElse-label',
+        method: this.includeSomeoneElse.bind(this)
+      });
       params.items.push({
         l10nId: 'selectMessages-label',
         method: this.startEdit.bind(this)
@@ -1931,6 +1940,22 @@ var ConversationView = {
     });
 
     new OptionMenu(params).show();
+  },
+
+  /**
+   * Redirect to a new message prefilled with current composer content and
+   * recipients.
+   */
+  includeSomeoneElse: function() {
+    var content = {
+      recipients: this.activeThread.participants,
+      content: Compose.getContent(),
+      subject: Compose.getSubject(),
+      type: Compose.getType()
+    };
+    this.initiateNewMessage({ content });
+    // As we are creating a new draft, we can discard current Composer
+    Compose.clear();
   },
 
   startEdit: function conv_edit() {
--- a/apps/sms/views/conversation/test/unit/conversation_test.js
+++ b/apps/sms/views/conversation/test/unit/conversation_test.js
@@ -5099,6 +5099,20 @@ suite('conversation.js >', function() {
         }).then(done, done);
       });
 
+      test('initiate a new message with predefined content', function(done) {
+        var content = {
+          recipients: ['+123'],
+          content: 'content',
+          subject: 'subject',
+          type: 'sms'
+        };
+        ConversationView.initiateNewMessage({ content }).then(() => {
+          sinon.assert.calledWith(Navigation.toPanel, 'composer');
+          sinon.assert.calledWithMatch(Drafts.add, content);
+          sinon.assert.called(Drafts.store);
+        }).then(done, done);
+      });
+
       test('Rejects if called with a bad parameter', function(done) {
         ConversationView.initiateNewMessage({ test: 'test' }).then(
           () => new Error('Should not be resolved.'),
@@ -6305,8 +6319,11 @@ suite('conversation.js >', function() {
       test('should show option for adding subject', function() {
         assert.equal(options[0].l10nId, 'add-subject');
       });
+      test('should show option for adding recipients', function() {
+        assert.equal(options[1].l10nId, 'includeSomeoneElse-label');
+      });
       test('should show option for selecting messages', function() {
-        assert.equal(options[1].l10nId, 'selectMessages-label');
+        assert.equal(options[2].l10nId, 'selectMessages-label');
       });
     });
   });
--- a/apps/keyboard/test/marionette/dismiss_test.js
+++ b/apps/keyboard/test/marionette/dismiss_test.js
@@ -24,14 +24,6 @@ marionette('Dimiss the keyboard', function() {
     desiredCapabilities: { raisesAccessibilityExceptions: false }
   });
 
-  function longPressSpaceBar(time) {
-    var spaceBarSelector = '.keyboard-type-container[data-active]' +
-                           ' .keyboard-key[data-keycode="32"]';
-
-    var spaceBar = client.findElement(spaceBarSelector);
-    actions.longPress(spaceBar, time).perform();
-  }
-
   setup(function() {
     systemInputMgmt = client.loader.getAppClass('system', 'input_management');
     system = client.loader.getAppClass('system');
@@ -52,7 +44,7 @@ marionette('Dimiss the keyboard', function() {
 
   test('Longpressing the space bar should dismiss the keyboard', function() {
     // The time needed for dismiss is 500ms.
-    longPressSpaceBar(0.7);
+    keyboard.longPressSpaceBar(0.7);
 
     client.waitFor(() => {
       return !systemInputMgmt.keyboardFrameDisplayed();
@@ -62,7 +54,7 @@ marionette('Dimiss the keyboard', function() {
   });
 
   test('Longpressing for only 0.5 sec should not dimiss keyboard', function() {
-    longPressSpaceBar(0.5);
+    keyboard.longPressSpaceBar(0.5);
 
     var keyboardContainer = keyboard.currentPanel;
 
--- a/apps/keyboard/test/marionette/lib/keyboard.js
+++ b/apps/keyboard/test/marionette/lib/keyboard.js
@@ -25,6 +25,8 @@ Keyboard.Selectors = Object.freeze({
   returnKey:
     '.keyboard-type-container[data-active] ' +
     '.keyboard-key[data-l10n-id="returnKey2"]',
+  spaceBarKey: '.keyboard-type-container[data-active]' +
+    ' .keyboard-key[data-keycode="32"]',
   dismissSuggestionsKey:
     '.keyboard-type-container[data-active] ' +
     '.dismiss-suggestions-button',
@@ -110,6 +112,10 @@ Keyboard.prototype = {
 
   get returnKey() {
     return this.client.findElement(Keyboard.Selectors.returnKey);
+  },
+
+  get spaceBarKey() {
+    return this.client.findElement(Keyboard.Selectors.spaceBarKey);
   },
 
   get shiftKey() {
@@ -301,6 +307,10 @@ Keyboard.prototype = {
         .move(this.getKey(char)).wait(0.2).release().perform();
       }
     });
+  },
+
+  longPressSpaceBar: function(time) {
+    this.actions.longPress(this.spaceBarKey, time).perform();
   },
 
   switchTo: function() {
--- a/apps/sms/test/marionette/lib/views/new-message/view.js
+++ b/apps/sms/test/marionette/lib/views/new-message/view.js
@@ -112,6 +112,15 @@ NewMessageView.prototype = {
     return this.composerAccessors.addAttachment();
   },
 
+  /**
+   * Take a screenshot of an attachment thumbnail in the composer.
+   * @param {Number} the attachment index
+   * @return {String} The screenshot
+   */
+  takeComposerAttachmentScreenshot: function(index) {
+    return this.composerAccessors.takeAttachmentScreenshot(index);
+  },
+
   typeSubject: function(subject) {
     this.composerAccessors.subjectInput.sendKeys(subject);
   },
--- a/apps/sms/test/marionette/lib/views/shared/composer_accessors.js
+++ b/apps/sms/test/marionette/lib/views/shared/composer_accessors.js
@@ -56,6 +56,10 @@ ComposerAccessor.prototype = {
     return new ActivityChooserView(this.client);
   },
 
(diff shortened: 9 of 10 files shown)


lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=88f9328d5cf96abdd728959bec48dff4f1af8678

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ae82c01520fe937d2d8ae9efcf6d6d74a43565c6

apps/sms/views/conversation/js/conversation.js
```diff
@@ -723,6 +725,45 @@ var ConversationView = {
...
+      type: composer.type,
+      body: composer.body,
+      attachments: composer.content
+        .reduce(generateMmsAttachmentsForComposer, [])
+    });
+    Compose.focus();
+
+    if (recipients) {
+      for (var i in recipients) {
+        Utils.getContactDisplayInfo(
```
autra: I don't know if you remember, but you asked me to factorize that with how we get contacts for handleActivity. Actually I think it's not exactly the same, so I kept it like that. The difference is that I never have a contact. Btw is there a way to get the current MozContact of a conversation? It would be handy to directly pass it to the Composer instead of doing what I do here.

apps/sms/views/conversation/js/conversation.js
```diff
@@ -723,6 +725,45 @@ var ConversationView = {
...
+          Contacts.findByAddress.bind(Contacts), recipients[i]
+        ).then(function onData(data) {
+          if (data.type !== '') {
```
autra: And that's the way I came up with to know if I found a contact in the user DB or not. Not very satisfying imo. Do you have better?

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b4d914ff4444bdc4daa15d9bafa1108d7e929f78

apps/sms/views/conversation/js/conversation.js
```diff
@@ -721,6 +723,48 @@ var ConversationView = {
...
+      body: composer.body,
+      attachments: composer.content
+        .reduce(generateMmsAttachmentsForComposer, [])
+    });
+    Compose.focus();
+
+    if (recipients) {
+      var findByAddress = Contacts.findByAddress.bind(Contacts);
+      for (var i in recipients) {
+        Utils.getContactDisplayInfo(findByAddress, recipients[i])
```
autra: I don't know if you remember, but you asked me to factorize that with how we get contacts for handleActivity. Actually I think it's not exactly the same, so I kept it like that.

The difference is that in handleActivity when you don't have a activity.contact, it puts a 'manual' source. But here I don't have the contact even if sometimes there may be one in the user DB. So I want to really call findByAddress at some point.

Btw is there a way to get the current MozContact of a conversation? It would allow me to factorize with handleActivity.

apps/sms/views/conversation/js/conversation.js
```diff
@@ -721,6 +723,48 @@ var ConversationView = {
...
+        slides.push({
+          content: content.blob,
+          location: content.name
+        });
+      }
+      return slides;
+    }
+
+    function addRecipientsToConversation(data) {
+      if (data.type !== '') {
```
autra: That's the way I came up with to know if I found a contact in the user DB or not. Not very satisfying imo. Do you have better?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=eede3b99f9eff7b7298a03753fd41f6b07c33672

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e28766249a2e715d7858fe8f686318d70b4d326c

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=edabcfa3c65b946d6f2e62515636e265c86333f1

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a023918f12ea07c9f978f891217c29d190853e6c

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7877ad7e4b4ffd6c1cb8182b195f7593f75b085d

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1876,9 +1917,13 @@ var ConversationView = {
     }
     params.items.push(subjectItem);
 
-    // If we are on a thread, we can call to SelectMessages
+    // If we are on a thread, we can call to SelectMessages and add recipients
     if (Navigation.isCurrentPanel('thread')) {
       params.items.push({
+        l10nId: 'addRecipients-label',
+        method: this.addRecipients.bind(this)
```
julienw: I think you can have a look how you can leverage the new method `initiateNewMessage`, maybe adding a new parameter, or completing the "parameters.number" case. Or a new similar method (because the "focusComposer" boolean value would be different than what we have in `initiateNewMessage`), maybe `SwitchToNewMessage` (we call "New Message panel" this panel now, to distinguish it with the "composer" element -- we still use the 'composer' name when calling `toPanel ` though so we didn't change all places yet :) ).

apps/sms/views/conversation/js/conversation.js
```diff
@@ -739,6 +746,40 @@ var ConversationView = {
...
+      }
+      return slides;
+    }
+
+    Compose.fromMessage({
+      subject: composer.subject,
+      type: composer.type,
+      body: composer.body,
+      attachments: composer.content
+        .reduce(generateMmsAttachmentsForComposer, [])
```
julienw: I think it won't be necessary with the new approach of using drafts -- we save and recall attachments directly :)

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1923,6 +1968,21 @@ var ConversationView = {
...
+
+    // redirect to Composer panel and add all the contacts
+    return Navigation.toPanel('composer', {
+      composer: {
+        subject: Compose.getSubject(),
+        type: Compose.type,
+        body: Compose.getText(),
+        content: Compose.getContent()
+      },
+      recipients: Threads.active.participants
```
julienw: because we use URLs now, we can't use complex objects anymore, so instead we'll need to save as a draft and pass the draftId (look at initiateNewMessage for more information)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=35a5bbdb520f7d243c7ad3166e9d3d9b45ff1634

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c21ddcf932efe456cd949bcf802f54533a6be2a5

apps/sms/test/marionette/conversation_test.js
```diff
@@ -324,7 +336,7 @@ marionette('Conversation Panel Tests', function() {
       var inboxView = new InboxView(client);
       conversationView = inboxView.goToConversation(thread.id);
     });
-    
+
```
autra: Trailing spaces kill kittens, I've been told ;-)

autra: I've never done any fact-checking on that though, but better safe than sorry...

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6f8ee32c88922c02370a84798f8f3718d4603bea

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6021a22d2dfae7883fc2b0506c40f2ae4beaa7ad

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=18b02d1e3881ced0f8d8845a65fea53aabced47c

apps/sms/test/marionette/conversation_test.js
```diff
@@ -491,4 +503,83 @@ marionette('Conversation Panel Tests', function() {
...
+      });
+      messagesApp.launch();
+
+      storage.setMessagesStorage([thread], ThreadGenerator.uniqueMessageId);
+      // Set empty contacts store.
+      storage.setContactsStorage();
+      var inboxView = new InboxView(client);
+      conversationView = inboxView.goToConversation(thread.id);
+      messagesApp.Composer.messageInput.sendKeys('message2');
+    });
```
julienw: this setup seems to bring intermittent failures, see https://treeherder.mozilla.org/logviewer.html#?job_id=2935056&repo=gaia

apps/sms/test/marionette/conversation_test.js
```diff
@@ -491,4 +503,83 @@ marionette('Conversation Panel Tests', function() {
...
+      messagesApp.Composer.messageInput.sendKeys('message3');
+      messagesApp.showSubject();
+      messagesApp.Composer.subjectInput.sendKeys('subject2');
+      // add a blob
+      messagesApp.Composer.attachButton.tap();
+      messagesApp.Menu.selectSystemMenuOption('Messages Activity Caller');
+
+      console.log('here');
+      activityCallerApp.switchTo();
+      console.log('there');
```
julienw: you'll need to remove these console logs ;)

apps/sms/test/marionette/conversation_test.js
```diff
@@ -491,4 +503,83 @@ marionette('Conversation Panel Tests', function() {
...
+    test('Sms being written: Message and recipients should be kept',
+    function() {
+      conversationView.showOptions();
+      messagesApp.Menu.selectAppMenuOption('Add recipients');
+      // this also ensure we are on Composer view
+      messagesApp.getRecipient('+123');
+      client.waitFor(function() {
+        return messagesApp.Composer.messageInput.text() === 'message2';
+      });
+    });
```
julienw: so, in all new integration tests, we try to use the new way of writing them, using the "views" (so here it's conversationView) instead of using functions on messagesApp or accessors directly. This is more OO, while using messagesApp's functions is more procedural.

you can have a look at https://github.com/mozilla-b2g/gaia/blob/master/apps/sms/test/marionette/lib/views/conversation/view.js to see what's done already, and you can add new methods there.

For example you can look at showSubject and hideSubject that are close to what you want to do :)

julienw: note that in your case it would be a good idea to return a NewMessageView (https://github.com/mozilla-b2g/gaia/blob/master/apps/sms/test/marionette/lib/views/new-message/view.js) from this new method ;)

apps/sms/test/marionette/conversation_test.js
```diff
@@ -491,4 +503,83 @@ marionette('Conversation Panel Tests', function() {
...
+
+    test('Mms being written: all info must be kept', function() {
+      // add a subject
+      messagesApp.showSubject();
+      messagesApp.Composer.subjectInput.sendKeys('subject1');
```
julienw: you can use https://github.com/mozilla-b2g/gaia/blob/master/apps/sms/test/marionette/lib/views/conversation/view.js#L177 :)

apps/sms/test/marionette/conversation_test.js
```diff
@@ -491,4 +503,83 @@ marionette('Conversation Panel Tests', function() {
...
+      conversationView.showOptions();
+      messagesApp.Menu.selectAppMenuOption('Add recipients');
+      // this also ensure we are on Composer view
+      messagesApp.getRecipient('+123');
+      client.waitFor(function() {
+        return messagesApp.Composer.messageInput.text() === 'message2';
+      });
+      client.waitFor(function() {
+        return messagesApp.Composer.subjectInput.text() === 'subject1';
+      });
```
julienw: see https://github.com/mozilla-b2g/gaia/blob/master/apps/sms/test/marionette/lib/views/new-message/view.js#L50-L52

apps/sms/test/marionette/conversation_test.js
```diff
@@ -491,4 +503,83 @@ marionette('Conversation Panel Tests', function() {
...
+      messagesApp.showSubject();
+      messagesApp.Composer.subjectInput.sendKeys('subject1');
+      // click 'add Recipients'
+      conversationView.showOptions();
+      messagesApp.Menu.selectAppMenuOption('Add recipients');
+      // this also ensure we are on Composer view
+      messagesApp.getRecipient('+123');
+      client.waitFor(function() {
+        return messagesApp.Composer.messageInput.text() === 'message2';
+      });
```
julienw: https://github.com/mozilla-b2g/gaia/blob/master/apps/sms/test/marionette/lib/views/new-message/view.js#L46-L48 would be useful :)

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1910,9 +1920,16 @@ var ConversationView = {
     }
     params.items.push(subjectItem);
 
-    // If we are on a thread, we can call to SelectMessages
+    // If we are on a thread, we can call to SelectMessages and add recipients
     if (Navigation.isCurrentPanel('thread')) {
       params.items.push({
+        l10nId: 'addRecipients-label',
+        method: () => {
+          this.updateDraft();
```
julienw: shouldn't you call .then() here ?

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1910,9 +1920,16 @@ var ConversationView = {
...
+          this.initiateNewMessage({ draft: this.draft });
```
julienw: I haven't tried yet, but are you sure this works if draft is null ? this happens if hte message is empty. (it's quite a recent change though :) ).

julienw: note I'm not completely sold on this "save the draft, then reuse the draft to save another draft, and then delete the first one". I'll think of it over the night...

apps/sms/test/marionette/conversation_test.js
```diff
@@ -491,4 +503,83 @@ marionette('Conversation Panel Tests', function() {
...
+        return messagesApp.Composer.messageInput.text() === 'message3';
+      });
+      client.helper.waitForElement(
+        messagesApp.Selectors.Message.imgAttachment
+      );
+      client.waitFor(function() {
+        return messagesApp.Composer.subjectInput.text() === 'subject2';
+      });
+    });
+  });
```
julienw: I miss the case of adding a recipient with an empty message :) (I think this is failing with your current code)

apps/sms/views/conversation/js/conversation.js
julienw: OK now I think we should not do that; simply we should just create a new draft here, using Compose.getContent(), etc.

julienw: and we likely need to delete any current draft as you do already (but I'm not sure how we can prevent creating one in beforeLeave)

apps/sms/test/marionette/conversation_test.js
autra: So I guess line 292 and 331 should too, right? Or am I misunderstanding?

I'll try to make it wait for the element in goToConversation.

julienw: mmm I don't know it's failing here but not elsewhere.

That said, since bug 1200583 we changed how the storage work:
* you don't need to specify empty storages (so you can remove the contact storage)
* you need to set them _before_ launch.

maybe this will automagically fix the issue :)

autra: Before that, I need to *reproduce* the issue... which is not the case. I've not yet managed to make the integration test run in vagrant. I've tried Gasolin's foxbox, and now I'm trying with azasypkin's gaia-vagrant.

Btw I have the feeling `clang` is needed to run 'npm install' on gaia. Do you confirm that?

julienw: yes, you need node 4 and clang now :)

please ping me on IRC if you still have issues !

autra: > maybe this will automagically fix the issue :)

It has, apparently :-)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9325261681dc41c49f44d7aeeee96a3e206150a3

apps/sms/test/marionette/conversation_test.js
```diff
@@ -491,4 +503,101 @@ marionette('Conversation Panel Tests', function() {
...
     });
   });
+
+  suite('Add recipients from thread', function() {
+    var conversationView;
+    setup(function() {
+      var thread = ThreadGenerator.generate({
+        numberOfMessages: 1
+      });
+      storage.setMessagesStorage([thread], ThreadGenerator.uniqueMessageId);
```
autra: It seems I don't have any failure... But how do you check for intermittent failure on taskcluster? Does the automatic retry possibly hide them from the results? Do you know which Gij number it is by heart and you check manually?

apps/sms/test/marionette/conversation_test.js
```diff
@@ -491,4 +503,101 @@ marionette('Conversation Panel Tests', function() {
...
+        'There should be 2 recipients');
+      assert.include(newMessageView.recipients, '+123',
+        '+123 should be in the recipients');
+      assert.include(newMessageView.recipients, '+234',
+        '+234 should be in the recipients');
+      assert.equal(newMessageView.messageText, 'message3',
+        'Message text should be kept');
+      assert.equal(newMessageView.subject, 'subject2',
+        'Subject should be kept');
+      assert.ok(newMessageView.attachments, 'Attachment should be kept');
```
autra: So I try to use views everywhere, I hope it is what you expected. FWIW, I find this way of writing these tests a lot clearer and easier to parse (especially when the test switches view).

apps/sms/views/conversation/js/conversation.js
autra: Oh that's funny. github removes the "obsolete" marker when I restore a previous diff :-) I thought they would stay obsolete...

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1927,6 +1940,18 @@ var ConversationView = {
...
+  addRecipients: function() {
+    var draft = new Draft({
+      recipients: this.activeThread.participants,
+      content: Compose.getContent(),
+      subject: Compose.getSubject(),
+      type: Compose.getType()
+    });
+    this.initiateNewMessage({ draft });
+    // As we are creating a new draft, we can discard current Composer
+    Compose.clear();
```
autra: That's how I prevent `beforeLeave` to save a draft. Imo it makes sense, we are transferring the current composer to a draft, so we can clear it now. WDYT?

apps/sms/test/marionette/lib/messages.js
```diff
@@ -27,6 +27,7 @@ var ParticipantsAccessor = require('./views/participants/accessors');
     Message: {
       content: '.message-content > p:first-child',
       vcardAttachment: '[data-attachment-type="vcard"]',
+      imgAttachment: '[data-attachment-type="img"]',
```
autra: I think I should remove that.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4d16be67936bf7ce19c073c7ea0652fcba00f0f1

apps/sms/test/marionette/conversation_test.js
autra: Is it a good way to test my image attachment?

apps/sms/test/marionette/conversation_test.js
```diff
@@ -491,4 +503,101 @@ marionette('Conversation Panel Tests', function() {
...
+      newMessageView.typeSubject('subject1');
+      conversationView = newMessageView.send();
+
+      newMessageView = conversationView.addRecipients();
+      assert.equal(newMessageView.recipients.length, 2,
+        'There should be 2 recipients');
+      assert.include(newMessageView.recipients, '+123',
+        '+123 should be in the recipients');
+      assert.include(newMessageView.recipients, '+234',
+        '+234 should be in the recipients');
```
julienw: nit: it's quite costly to access .recipients, maybe you can cache it in a variable before asserting ?

Also you can use `assert.lengthOf` instead of `assert.equal(....length, `

apps/sms/test/marionette/conversation_test.js
julienw: More or less ;) When I think there could be an intermittent, I respin the right Gij (usually Gij16 or 17, but we need to check the logs to find the right one)

apps/sms/test/marionette/conversation_test.js
```diff
@@ -491,4 +503,101 @@ marionette('Conversation Panel Tests', function() {
...
+      // redirect to new message to add recipients
+      newMessageView = conversationView.addRecipients();
+
+      // assert the content
+      assert.equal(newMessageView.recipients.length, 2,
+        'There should be 2 recipients');
+      assert.include(newMessageView.recipients, '+123',
+        '+123 should be in the recipients');
+      assert.include(newMessageView.recipients, '+234',
+        '+234 should be in the recipients');
```
julienw: nit: same questions about "recipients"

apps/sms/test/marionette/conversation_test.js
julienw: > So I try to use views everywhere, I hope it is what you expected. FWIW, I find this way of writing these tests a lot clearer and easier to parse (especially when the test switches view).

yeah :)


> Is it a good way to test my image attachment?

So normally it would timeout if we have no attachment at all; but I'd be less worried if you would check the element's content.

For example you could use a screenshot of the element before calling "addRecipients" above, and take another one here, and check they're identical. See https://github.com/mozilla-b2g/gaia/blob/7420221a40c1a35bc8122ddd0813c1422a39c582/apps/sms/test/marionette/lib/views/conversation/message_accessors.js#L43 for an example in a Conversation Message, you can maybe do something similar in the composer_accessor ?

apps/sms/test/marionette/conversation_test.js
```diff
@@ -16,10 +17,20 @@ marionette('Conversation Panel Tests', function() {
...
     '/mocks/mock_navigator_moz_contacts.js'
   ];
+  var apps = {};
+  apps[MessagesActivityCaller.ORIGIN] = __dirname + '/apps/activitycaller';
 
-  var client = marionette.client();
+  var client = marionette.client({
+    profile: {
+      prefs: {
+        'focusmanager.testmode': true
```
julienw: can you explain what it is for ?

julienw: oh I see you just copied it from elsewhere. Maybe @azasypkin would know better ?

azasypkin: I'm wondering if we still need this, I remember when I was creating one of my first tests I've copied it from calendar tests. There was a comment like ```Do not require the B2G-desktop app window to have focus (as per the system window manager) in order to do focus-related things.```

Hey @JohanLorenzo do you know if\when we actually need it?

JohanLorenzo: I remember it was needed for every b2gdesktop test. Like we even [set it by default in the marionette python driver](https://dxr.mozilla.org/mozilla-central/source/testing/marionette/driver/marionette_driver/geckoinstance.py#179). I don't know if that's still mandatory with Mulet.

Also, IIRC, this pref was useful when you switch the the keyboard to type. This is done in these tests, so my guess is that it's still needed.

apps/sms/test/marionette/conversation_test.js
autra: Ok let's screenshot!

First one, so far so good
screen_attachment_before

Second one, nooooo !!
screen_attachment_after

Indeed, I focus the keyboard after redirecting to the new message view, and apparently, nothing is focused just after adding an attachment.

So I thought of 3 different ways to circumvent that
- Make the keyboard disappear. Is there a standard way to do that other than switching to keyboard iframe, long-tapping on the right spot and switching back to sms iframe?
- tap the empty `#messages-container`. This fails with the very unhelpful `Crash detected but error running stackwalk` message
- make the window bigger (at least vertically). But maybe we test a particular size on purpose? And is it even possible to specify the startup size?

Do you have another idea?

autra: Actually, enabling gecko logs for the second idea yields 
```bash 
1449314963991	Marionette	DEBUG	conn1 client <- {"error":"element not visible","message":"Element is not currently visible and may not be manipulated","stacktrace":null}
``` 
when I try to tap on the empty `#messages-container`. So 2 and 3 are related: the height of my viewport is not big enough

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=849edf30652518b0be3daa86602c8f4287a966ec

apps/sms/locales/sms.en-US.properties
```diff
@@ -37,6 +37,7 @@ selectMessages-label  = Select Messages
 settings              = Settings
 add-subject            = Add subject
 remove-subject         = Remove subject
+addRecipients-label    = Add Recipients
```
julienw: please change the string to "Include someone else" as said on the bug

apps/sms/test/marionette/lib/views/conversation/view.js
```diff
@@ -193,6 +199,16 @@ ConversationView.prototype = {
     this.client.helper.waitForElement(this.composerAccessors.subjectInput);
   },
 
+  addRecipients: function() {
```
julienw: nit: includeSomeoneElse

apps/sms/test/marionette/lib/views/conversation/view.js
```diff
@@ -179,6 +179,12 @@ ConversationView.prototype = {
     return this.composerAccessors.addAttachment();
   },
 
+  takeAttachmentScreenshot: function(index) {
```
julienw: not completely sure of the name, as we may think it's for attachments in messages from the list... Ideally I'd like a Composer view :) I'm sure @azasypkin has something like this in his head already ;)

In the mean time, how about using "takeComposerAttachmentScreenshot" ? it's quite long but at least it's meaningful?

julienw: also please add a small jsdoc

apps/sms/test/marionette/lib/views/conversation/view.js
julienw: also add a small jsdoc please

apps/sms/test/marionette/lib/views/conversation/view.js
```diff
@@ -193,6 +199,16 @@ ConversationView.prototype = {
     this.client.helper.waitForElement(this.composerAccessors.subjectInput);
   },
 
+  addRecipients: function() {
+    this.showOptions();
+    this.menuAccessors.selectAppMenuOption('Add recipients');
```
julienw: nit: will be "Include someone else" I guess ?

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1933,6 +1946,18 @@ var ConversationView = {
...
   },
 
+  addRecipients: function() {
+    var draft = new Draft({
+      recipients: this.activeThread.participants,
+      content: Compose.getContent(),
+      subject: Compose.getSubject(),
+      type: Compose.getType()
+    });
+    this.initiateNewMessage({ draft });
```
julienw: nit: here is a slightly different proposal, tell me what you think:

```javascript
var content = {
  recipients: this.activeThread.participants,
  content: Compose.getContent(),
  subject: Compose.getSubject(),
  type: Compose.getType()
};
this.initiateNewMessage({ content });
```

and in initiateNewMessage:
```javascript
  ... if (parameters.content) {
    var draft = new Draft(content);
    draftCreatePromise = Drafts.add(draft).store().then(() => draft.id);
  }
```

so quite similar but I think it's slightly simpler like this. We don't need to use "Draft" at all in `addRecipients` for example.

apps/sms/views/conversation/js/conversation.js
julienw: yeah I think it makes sense

apps/sms/views/conversation/test/unit/conversation_test.js
```diff
@@ -6295,7 +6295,10 @@ suite('conversation.js >', function() {
         assert.equal(options[0].l10nId, 'add-subject');
       });
       test('should show option for selecting messages', function() {
```
julienw: nit: the test name is wrong :)

apps/sms/views/conversation/test/unit/conversation_test.js
```diff
@@ -6295,7 +6295,10 @@ suite('conversation.js >', function() {
...
-        assert.equal(options[1].l10nId, 'selectMessages-label');
+        assert.equal(options[1].l10nId, 'addRecipients-label');
+      });
+      test('should show option for selecting messages', function() {
+        assert.equal(options[2].l10nId, 'selectMessages-label');
       });
     });
   });
```
julienw: nit: can you add some unit tests for the new option of initiateNewMessage ?

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1933,6 +1946,18 @@ var ConversationView = {
     new OptionMenu(params).show();
   },
 
+  addRecipients: function() {
```
julienw: nit: jsdoc please :)

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1197,14 +1197,23 @@ var ConversationView = {
...
       if (parameters.messageId) {
         draftCreatePromise = this.storeDraftFromMessage(parameters.messageId);
       } else if (parameters.number) {
-        var draft = new Draft({
+        draft = new Draft({
           recipients: [parameters.number],
           type: Utils.isEmailAddress(parameters.number) ? 'mms' : 'sms'
         });
         draftCreatePromise = Drafts.add(draft).store().then(() => draft.id);
+      } else if (parameters.draft) {
```
julienw: nit: please augment the jsdoc for initiateNewMessage for the new parameter (`content` if you follow my suggestion)

apps/sms/test/marionette/lib/views/conversation/view.js
azasypkin: > I'm sure @azasypkin has something like this in his head already ;)

Not really :) But I'd rather move this to ```composerAccessors``` and just reexpose here as we do for ```addAttachment```. Recently I was thinking about creating separate ```ComposeView``` that will use ```composeAccessors``` internally and we'll just need to call ```getComposeView``` on NewMessageView and ConversationView, so that we won't have to re-expose such methods for both NewMessage and Conversation - ComposeView will do it for us.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2b29437c1e1bc1fa973ac347d5408c337247364b

apps/sms/test/marionette/conversation_test.js
```diff
@@ -556,4 +567,106 @@ marionette('Conversation Panel Tests', function() {
...
+      assert.equal(newMessageView.messageText, 'message3',
+        'Message text should be kept');
+      assert.equal(newMessageView.subject, 'subject2',
+        'Subject should be kept');
+      // Tap the header to hide the keyboard: when it is displayed, the
+      // attachment is hidden, which prevents the screenshot to be taken
+      newMessageView.tapHeader();
+      client.waitFor(function() {
+        return beforeScreenshot ==
+          newMessageView.takeComposerAttachmentScreenshot(0);
```
autra: If you have better than that, I'll take it. Basically, I need to wait for the keyboard to disappear.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=499370e6d4fcec3912b34a51813c989012f901d6

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f9075957f7466a6113827b1467b8256a32d6a5d4

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=86b85b4318069f834abc82f526d85769f208162e

apps/sms/test/marionette/conversation_test.js
julienw: I have something for you :D
https://github.com/mozilla-b2g/gaia/blob/002fc618166f2887753b254a8b5e7d6195a86ed8/apps/keyboard/test/marionette/dismiss_test.js#L52-L61

Basically:
* I'd move `longPressSpaceBar` in lib/keyboard so that you could import it here (this is a little more explicit than tapHeader); but this is optional, as this is quite more work.
* it uses something from the system lib (note: `client.loader.getAppClass` is like a clever `require('../../../etc')`, defined in https://github.com/mozilla-b2g/gaia/blob/002fc618166f2887753b254a8b5e7d6195a86ed8/shared/test/integration/marionette_loader.js#L36-L43). Note that the test code here is quite obsolete, as now we have https://github.com/mozilla-b2g/gaia/blob/002fc618166f2887753b254a8b5e7d6195a86ed8/apps/system/test/marionette/lib/input_management.js#L62. Maybe you could add a `waitForKeyboardFrameHidden` just like we have `waitForKeyboardFrameDisplayed` already, and use it directly in this test.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=fadb6ce9dfd214a8c6586932649466ad2d788438

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8d832d45e4308c9319c0a56e539a4bbbfb5a2a6d

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d26838512a5026a1241be7080b094bd045b479ca

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a76f12c8065a0545b4451f0c745f4dc86ce37a75

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=40dff0ed80d7d7aaddc962b8cd011077c505fd34

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3938577aa9f9b5d070a4a3f2c68acf32b1c7c790

apps/sms/test/marionette/conversation_test.js
```diff
@@ -553,4 +566,114 @@ marionette('Conversation Panel Tests', function() {
...
+        'Subject should be kept');
+
+      // hide the keyboard: when it is displayed, the attachment is hidden,
+      // which prevents the screenshot to be taken
+      client.switchToFrame();
+      keyboardApp.switchTo();
+      keyboardApp.longPressSpaceBar(0.7);
+      client.switchToFrame();
+      // wait for the keyboard to be hidden
+      var systemInputMgmt = client.loader.getAppClass(
```
autra: is it ok to have a local var for that? As it is used only here, I thought I don't need to expose that.

apps/sms/test/marionette/conversation_test.js
```diff
@@ -553,4 +566,114 @@ marionette('Conversation Panel Tests', function() {
...
+        'system', 'input_management');
+      systemInputMgmt.waitForKeyboardFrameHidden();
+
+      client.apps.switchToApp('app://sms.gaiamobile.org');
```
julienw: nit: I think `messagesApp.switchTo()` works

apps/sms/test/marionette/conversation_test.js
julienw: yes why not, but also I wonder if we'd need waitForKeyboardFrameDisplayed before switching in line 666. Just to be intermittent-free.

autra: Yep, forgot that. I'm adding that now.

autra:
It seems some tests are failing, but they are unrelated to this right?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a41f747ca1319f643df855fcbe1cc10ec977cd8d

julienw:
very unlikely ... see current master: https://treeherder.mozilla.org/#/jobs?repo=gaia-master

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3673868a341aa6a557bcf4b0e06433b9bdfbce5b

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0839e3444c7faac60db851ee163f936d20099952

PR #27007, opened 2014-12-25
Bug 1115494 - [RTL] Share activity menu icons are not right-alined +autoland

anefzaoui:


apps/system/style/action_menu/action_menu_extended.css
```diff
@@ -136,3 +136,9 @@ form[role="dialog"][data-type="object"].visible {
     z-index: -1;
   }
 }
+
+/* RTL View */
+html[dir="rtl"] form[role="dialog"][data-type="action"] > menu > button.icon,
+html[dir="rtl"] form[role="dialog"][data-type="object"] > menu > button.icon {
+  background-position: right 0.5rem center;
+}
```
mikehenrty: nit: can we move this definition to under the non-rtl `button.icon` definition above [1]?

1.) https://github.com/mozilla-b2g/gaia/pull/27007/files#diff-276bfb67dfd3983bd8de40144517b791R33

anefzaoui: I think we better leave it there for consistency with other files that contain RTL definitions, we've always putted them at the bottom of the file to make finding them easier.
What do you think?

mikehenrty: Sounds good to me.

PR #27014, opened 2014-12-25
Bug 1091441 - [Messages] the thread view is flashing while loading if there are MMS

steveck-chung:
- Use promise.all to scroll after all the mms contents parsed and attached
- Fix the chunk bug

### DIFF
--- a/apps/sms/js/thread_ui.js
+++ b/apps/sms/js/thread_ui.js
@@ -1436,8 +1436,6 @@ var ThreadUI = {
           options.forEach((opt) => {
             opt.element.appendChild(this.createMmsContent(opt.slideArray));
           });
-          // Go to Bottom
-          this.scrollViewToBottom();
         });
       }
 
@@ -1716,7 +1714,13 @@ var ThreadUI = {
 
   showChunkOfMessages: function thui_showChunkOfMessages(number) {
     var elements = ThreadUI.container.getElementsByClassName('hidden');
-    for (var i = elements.length - 1; i >= 0; i--) {
+    var length = elements.length;
+
+    if (length === 0) {
+      return;
+    }
+
+    for (var i = length - 1; i >= length - number; i--) {
       var element = elements[i];
       element.classList.remove('hidden');
       if (element.tagName === 'HEADER') {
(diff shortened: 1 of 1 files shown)


steveck-chung:
I understand your opinion, but I don't get why we need `TaskRunner` here. What benefits you could get from executing the append promise one by one? Using Promise.all and scroll to bottom while everything ready should be reasonable.

julienw:
The issue is that when you use Promise.all, you wait for all promises before scrolling... that means that you don't scroll to the bottom right away to see the latest message.

On Flame I didn't see this was really an issue though, but I didn't try on Hamachi.

Aso, I think the pseudo code I outlined in my comment is more self-contained, I especially don't like this new property mmsPromises, that feels like another state we need to manage, and if we can avoid it, I think it's better.

But as I said, I prefer to land your patch than not land it, because it still improves the situation.

julienw:
I just feel it will be more difficult to maintain; I can't really justify more than what I said about the additional variable, sorry :/

Can you at least try my approach and see how it works?

try-server-hook:
[steveck-chung steveck-chung](https://github.com/steveck-chung) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=66bee7cd9671)

steveck-chung:
@julienw TaskRunner WIP: https://github.com/steveck-chung/gaia/compare/message-reduce-flashing-taskrunner?expand=1 The main issue for TaskRunner idea is we could not display the actual number of messages for first chunk(If we want to keep the idea of showFirstChunk) Maybe be can move the message index checking inside the appendMessage, but we may need additional flag for the case when rendering ended with total messages less then chunk size.

julienw:
here is the mail I sent to Steve about the Taskrunner WIP:


Remember that TaskRunner.push takes a function as parameter, not a
promise. The goal is to execute this function when the previous function
is finished.

So instead of :
```javascript
  this.taskRunner.push(this.appendMessage(message,/*hidden*/ true));
```
you can use :
```javascript
  this.taskRunner.push(() => this.appendMessage(message,/*hidden*/ true));
```

In showFirstChunk, you can just keep it like it is on master, because in
task_runner, we only put the function in the "then" callback. And it's
ok if a "then" callback does not return a promise or anything.

But you can also return `Promise.resolve()` if you want.

azasypkin:
unrelated let's rename ```TaskRunner``` to ```TaskQueue``` please :)

julienw:
yeah why not :)

PR #27023, opened 2014-12-27
Bug 1115799 : Check the menu dialog after adding to homescreen to verify that add to h...

nhirata:
...omescreen is not listed in the menu

### DIFF
--- a/tests/python/gaia-ui-tests/gaiatest/apps/homescreen/regions/bookmark_menu.py
+++ b/tests/python/gaia-ui-tests/gaiatest/apps/homescreen/regions/bookmark_menu.py
@@ -33,6 +33,10 @@ class BookmarkMenu(Base):
         Wait(self.marionette).until(lambda m: self.apps.displayed_app.name != self.name)
         self.apps.switch_to_displayed_app()
 
+    def is_add_to_home_button_not_in_dialog(self):
+        Wait(self.marionette).until(expected.element_not_present(
+            *self._add_bookmark_to_home_screen_dialog_button_locator))
+
     def type_bookmark_title(self, value):
         element = self.marionette.find_element(
             *self._bookmark_title_input_locator)
--- a/tests/python/gaia-ui-tests/gaiatest/tests/functional/browser/test_browser_bookmark.py
+++ b/tests/python/gaia-ui-tests/gaiatest/tests/functional/browser/test_browser_bookmark.py
@@ -34,6 +34,12 @@ class TestBrowserBookmark(GaiaTestCase):
         bookmark.type_bookmark_title(self.bookmark_title)
         bookmark.tap_add_bookmark_to_home_screen_dialog_button()
 
+        # tap menu button to test to see if the add to homescreen button is removed
+        browser.switch_to_chrome()
+        browser.tap_menu_button()
+        self._add_to_home=bookmark.is_add_to_home_button_not_in_dialog()
+        self.assertFalse(self._add_to_home, 'Add to Homescreen is not in menu options')
+
         # Switch to Home Screen to look for bookmark
         self.device.touch_home_button()
 


tests/python/gaia-ui-tests/gaiatest/tests/functional/browser/test_browser_bookmark.py
```diff
@@ -34,6 +34,12 @@ def test_browser_bookmark(self):
         bookmark.type_bookmark_title(self.bookmark_title)
         bookmark.tap_add_bookmark_to_home_screen_dialog_button()
 
+        # tap menu button to test to see if the add to homescreen button is removed
+        browser.switch_to_chrome()
+        browser.tap_menu_button()
+        self._add_to_home=bookmark.is_add_to_home_button_not_in_dialog()
```
JohanLorenzo: I think we can merge this line with the next one. `self._add_to_home` is not explicit enough to be worth defined.

tests/python/gaia-ui-tests/gaiatest/tests/functional/browser/test_browser_bookmark.py
```diff
@@ -34,6 +34,12 @@ def test_browser_bookmark(self):
...
+        self.assertFalse(self._add_to_home, 'Add to Homescreen is not in menu options')
```
JohanLorenzo: This message is only used [when the assertion fails](https://github.com/python/cpython/blob/2.7/Lib/unittest/case.py#L412). So it would be more clear to write: "Add to Homescreen should not be in menu options"

tests/python/gaia-ui-tests/gaiatest/apps/homescreen/regions/bookmark_menu.py
```diff
@@ -33,6 +33,10 @@ def tap_add_bookmark_to_home_screen_dialog_button(self):
         Wait(self.marionette).until(lambda m: self.apps.displayed_app.name != self.name)
         self.apps.switch_to_displayed_app()
 
+    def is_add_to_home_button_not_in_dialog(self):
```
JohanLorenzo: This function must return something. Used in `test_browser_bookmark.py`, it returns `None` which is considered to be [a false value](https://docs.python.org/2/library/stdtypes.html#truth-value-testing) by Python.

That's the reason why the test is passing. The problem is it can lead to a false positive. 

Here, I would use: 
``` Python
def is_add_to_home_button_in_dialog(self):
    return self.is_element_displayed(*self._add_bookmark_to_home_screen_dialog_button_locator)
```

tests/python/gaia-ui-tests/gaiatest/tests/functional/browser/test_browser_bookmark.py
```diff
@@ -34,6 +34,12 @@ def test_browser_bookmark(self):
         bookmark.type_bookmark_title(self.bookmark_title)
         bookmark.tap_add_bookmark_to_home_screen_dialog_button()
 
+        # tap menu button to test to see if the add to homescreen button is removed
```
viorelaioia: Maybe rephrase this comment: Tap menu button to check that Add to homescreen option is removed

tests/python/gaia-ui-tests/gaiatest/apps/homescreen/regions/bookmark_menu.py
viorelaioia: agree with Johan, this method should return True is the option is displayed and False if not.

tests/python/gaia-ui-tests/gaiatest/tests/functional/browser/test_browser_bookmark.py
viorelaioia: Agree, you should use <code>bookmark.is_add_to_home_button_not_in_dialog</code> method directly in the assert.

PR #27026, opened 2014-12-27
Bug 1109203 - Do not allow current_audio_channel to switch context

mwargers:


bebef1987:
http://jenkins1.qa.scl3.mozilla.com/job/flame-kk.ui.adhoc/542/

PR #27049, opened 2014-12-30
Bug 1116348 - Distinguish between images and videos in label text.

eeejay:


apps/gallery/js/thumbnail_item.js
```diff
@@ -38,8 +38,7 @@ function ThumbnailItem(fileData) {
 ThumbnailItem.formatter = new navigator.mozL10n.DateTimeFormat();
 
 ThumbnailItem.prototype.localize = function() {
-  var date = new Date(this.data.date);
-  var imageCreatedAt = navigator.mozL10n.get('image-created-at');
-  var label = ThumbnailItem.formatter.localeFormat(date, imageCreatedAt);
+  var label = MediaUtils.shortDescription(
```
davidflanagan: Would this method be better named getAriaLabel()? Or is the point that it should be a useful description even outside of accessibility contexts?

apps/gallery/js/thumbnail_item.js
```diff
@@ -38,8 +38,7 @@ function ThumbnailItem(fileData) {
...
+    this.data.date, !!this.data.metadata.video);
```
davidflanagan: Nit: the preferred style in this app for multi-line function calls is to put the first argument on the same line as the open paren, and then align the others with that.

```
var label = MediaUtils.shortDescription(this.data.date,
                                        !!this.data.metadata.video);
```

shared/js/media/media_utils.js
```diff
@@ -9,12 +9,20 @@
 var MediaUtils = {
    _: navigator.mozL10n.get,
 
+   _getDateTimeFormat: function() {
+    if (!this._dtf) {
+      this._dtf = new navigator.mozL10n.DateTimeFormat();
+    }
+
+    return this._dtf;
+   },
```
davidflanagan: Nit: It looks like the body of this function is indented only one space instead of two.

Consider adding this before this function:

```
_dtf: null, // Lazily initialized DateTimeFormat instance
```

Actually, I think it is now guaranteed that the gallery and video apps are always going to be creating a DateTimeFormat object, so maybe there isn't really any need to bother with this lazy initialization.  What if you just had

```
dtf: new navigator.mozL10n.DateTimeFormat()
```
Then just use this.dtf when needed?

Of course, then the order in which the modules were loaded would matter...

shared/js/media/media_utils.js
```diff
@@ -98,5 +106,12 @@ var MediaUtils = {
       return MediaUtils.binarySearch(array, element, comparator, from, mid);
     else
       return MediaUtils.binarySearch(array, element, comparator, mid + 1, to);
+  },
+
+  shortDescription: function(timestamp, video) {
+    var date = new Date(timestamp);
+    var dtf = this._getDateTimeFormat();
+    return dtf.localeFormat(
+      date, this._(video ? 'videoDescriptionShort' : 'imageDescriptionShort'));
```
davidflanagan: Same nit here about line wrapping.  And if it was me, I'd declare  var desc and do the this._ call and the conditional operator in the variable initalizer, to make the localeFormat call nice and clean.

shared/locales/media/media.en-US.properties
```diff
@@ -0,0 +1,14 @@
...
+byteUnit-MB = MB
+byteUnit-GB = GB
+byteUnit-TB = TB
+byteUnit-PB = PB
+byteUnit-EB = EB
+byteUnit-ZB = ZB
+byteUnit-YB = YB
+
+videoDescriptionShort = Video created at %I:%M %p, %B %e, %Y
+imageDescriptionShort = Image created at %I:%M %p, %B %e, %Y
```
davidflanagan: The word "created" is problematic. If the image or video was not captured by the camera, then we don't know when it was created. The date could just as well be when it was downloaded or when it was received (by sms or email, for example). To keep it simple you could just replace "created at" with "dated".  If you want to be more complicated, you could check whether the photo or video was stored under DCIM/, in which case you could use "Photo taken at" or "Video recorded at".

apps/gallery/js/thumbnail_item.js
eeejay: You mean MediaUtils.shortDescription? I named it so because in a future patch, we will perhaps have longer descriptions for videos/images in fullscreen. The fact that we set it via aria-label seems less relevant (it could just as easily be set on the alt or title attribute). I don't mind changing it though, if you think that would be more descriptive.

shared/js/media/media_utils.js
eeejay: I could look into that..

shared/js/media/media_utils.js
eeejay: Sounds good.

shared/locales/media/media.en-US.properties
eeejay: I just tested on iOS, and they don't use any verb at all. Simply "image, 12:45 pm, November 3rd 2012". That might be ambiguous, but at least it doesn't misinform.

try-server-hook:
[eeejay Eitan Isaacson (eeejay)](https://github.com/eeejay) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=d2b71adebd67)

apps/gallery/js/thumbnail_item.js
```diff
@@ -39,7 +39,9 @@ ThumbnailItem.formatter = new navigator.mozL10n.DateTimeFormat();
 
 ThumbnailItem.prototype.localize = function() {
   var date = new Date(this.data.date);
-  var imageCreatedAt = navigator.mozL10n.get('image-created-at');
-  var label = ThumbnailItem.formatter.localeFormat(date, imageCreatedAt);
+  var description = navigator.mozL10n.get(
+    !!this.data.metadata.video ?
+    'imageDescriptionShort' : 'videoDescriptionShort');
```
punamdahiya: This should be other way around i.e !!this.data.metadata.video ?
+ 'videoDescriptionShort' : 'imageDescriptionShort'

apps/gallery/locales/gallery.en-US.properties
```diff
@@ -91,5 +91,6 @@ preview = Preview
 # Header format used in thumbnail view
 # see http://pubs.opengroup.org/onlinepubs/007908799/xsh/strftime.html
 date-group-header = %B %Y
-# Date time format for each individual photo's description
-image-created-at = Image created at %I:%M %p, %B %e, %Y
+
+videoDescriptionShort = Video created at %I:%M %p, %B %e, %Y
+imageDescriptionShort = Image created at %I:%M %p, %B %e, %Y
```
punamdahiya: Are we going with word created or it should be updated to dated or nothing as recommended by you. See discussion https://github.com/mozilla-b2g/gaia/pull/27049#discussion-diff-22423639R14

apps/gallery/js/thumbnail_item.js
eeejay: doh.. had that locally and didn't stage in commit.

shared/locales/media/media.en-US.properties
eeejay: ok, chaning to 'dated'. good idea.

try-server-hook:
[eeejay Eitan Isaacson (eeejay)](https://github.com/eeejay) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=46bbe1c7dddf)

apps/gallery/js/thumbnail_item.js
```diff
@@ -39,7 +39,9 @@ ThumbnailItem.formatter = new navigator.mozL10n.DateTimeFormat();
 
 ThumbnailItem.prototype.localize = function() {
   var date = new Date(this.data.date);
-  var imageCreatedAt = navigator.mozL10n.get('image-created-at');
-  var label = ThumbnailItem.formatter.localeFormat(date, imageCreatedAt);
+  var description = navigator.mozL10n.get(
+    !this.data.metadata.video ?
+    'imageDescriptionShort' : 'videoDescriptionShort');
```
punamdahiya: nit: Sorry for missing this before, for readability we should put !this.data.metadata.video ? on the same line as the open parenthesis like
  var description = navigator.mozL10n.get(!this.data.metadata.video ?
    'imageDescriptionShort' : 'videoDescriptionShort');

try-server-hook:
[eeejay Eitan Isaacson (eeejay)](https://github.com/eeejay) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=365efceba815)

PR #27065, opened 2014-12-30
Bug 1101029 - Overlay iframe on top of front-most app. r=alive

ferjm:


try-server-hook:
[ferjm Fernando Jiménez Moreno (ferjm)](https://github.com/ferjm) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=0ecf445c3d9c)

try-server-hook:
[ferjm Fernando Jiménez Moreno (ferjm)](https://github.com/ferjm) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=8a9a8d9a72d8)

try-server-hook:
[ferjm Fernando Jiménez Moreno (ferjm)](https://github.com/ferjm) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=fbb921313344)

try-server-hook:
[ferjm Fernando Jiménez Moreno (ferjm)](https://github.com/ferjm) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=866fb2d8a218)

try-server-hook:
[ferjm Fernando Jiménez Moreno (ferjm)](https://github.com/ferjm) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=acff27ec2b50)

try-server-hook:
[ferjm Fernando Jiménez Moreno (ferjm)](https://github.com/ferjm) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=b6ba87b45db8)

try-server-hook:
[ferjm Fernando Jiménez Moreno (ferjm)](https://github.com/ferjm) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=c93b05cf0b70)

apps/system/js/global_overlay_window.js
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    var self = this;
+    if (!this.loaded) {
+      this.element.addEventListener('_loaded', function onLoaded() {
+        self.element.removeEventListener('_loaded', onLoaded);
+        // We allow passing the mouse clicks done in the overlay window
+        // to the overlayed content by setting mozpasspointerevents to true
+        // on the overlay window iframe. Check bug 796452 for more details.
+        self.browser.element.setAttribute('mozpasspointerevents' , 'true');
```
alivedise: Can't us set this attribute before the iframe is loaded?

alivedise: i.e., in browserFrame like inputMethod https://github.com/mozilla-b2g/gaia/blob/master/apps/system/js/browser_frame.js#L70

apps/system/js/global_overlay_window_manager.js
```diff
@@ -0,0 +1,111 @@
+/* global Service */
+
+'use strict';
+
+(function(exports) {
+  var GlobalOverlayWindowManager = function() {};
+  GlobalOverlayWindowManager.prototype = {
+    DEBUG: true,
```
alivedise: false

apps/system/js/global_overlay_window_manager.js
```diff
@@ -0,0 +1,111 @@
...
+          }.bind(this));
+          break;
+
+        case 'globaloverlayterminated':
+          if (this._instance == overlay) {
+            this._instance = null;
+          }
+          break;
+
+        case 'system-resize':
```
alivedise: Do you really need to be resized? If so, please follow attentionWindowManager - register hierarchy to HierarchyManager and do not stop immediate propagation.

apps/system/js/global_overlay_window_manager.js
```diff
@@ -0,0 +1,111 @@
...
+          if (this._instance != overlay) {
+            return;
+          }
+          this._instance.resize();
+          evt.stopImmediatePropagation();
```
alivedise: This is incorrect to me. Your UI is transparent and if you do this, the background app/dialog/attention will NOT be resized if you are opened.

apps/system/js/global_overlay_window_manager.js
```diff
@@ -0,0 +1,111 @@
+/* global Service */
+
+'use strict';
+
+(function(exports) {
+  var GlobalOverlayWindowManager = function() {};
+  GlobalOverlayWindowManager.prototype = {
+    DEBUG: true,
+    TRACE: false,
+    CLASS_NAME: 'GlobalOverlayWindowManager',
```
alivedise: EVENT_PREFIX: 'globaloverlaywindowmanager',

apps/system/js/global_overlay_window_manager.js
```diff
@@ -0,0 +1,111 @@
...
+
+    // We only allow one GlobalOverlayWindow instance at a time.
+    // Allowing multiple global overlay windows would lead us to weird UIs.
+    _instance: null,
+
+    publish: function(eventName, detail) {
+      this.debug('publishing: ', eventName);
+      var evt = new CustomEvent(eventName, { detail: detail });
```
alivedise: this.EVENT_PREFIX + eventName

apps/system/js/hierarchy_manager.js
```diff
@@ -58,6 +58,7 @@
      * @type {Array}
      */
     PRIORITIES: [
+      'OverlayWindowManager',
```
alivedise: Okay you have this. You need to tell HM when you are active and when you are inactive by publish activated/deactivated. However I am still not sure what's the desired behavior of GOWM - does it affect screen reader? does it need to hide the bottom windows for screen reader?

apps/system/style/zindex.css
```diff
@@ -334,3 +335,14 @@
...
+
+/**
+ * Global overlays are on top of everything, all the time.
+ */
+#screen > [data-z-index-level="global-overlays"] {
+  z-index: 2147483647;
+}
+
+#screen > [data-z-index-level="global-overlays"] > .appWindow.overlayWindow {
+  z-index: 2147483647;
```
alivedise: Be careful - that means you have chance to miss device menu and cannot shutdown/reboot your phone when there is an overlay window

apps/system/js/global_overlay_window.js
ferjm: I tried that, but for some reason that I don't understand, that doesn't work. If I set the attribute before the iframe is loaded, we don't pass the click events.

apps/system/js/global_overlay_window_manager.js
ferjm: I guessed that we need to be resized to wrap the hole screen. Otherwise, we might have an overlay window overlaying only part of the screen. Am I right?

apps/system/js/hierarchy_manager.js
ferjm: This is a good point.
Given that the only use case that we have so far for the GOWM is BuddyUp I would say that we shouldn't even allow placing the GOWM while the screen reader is active and we should hide it if it is already active. For BuddyUp we need the GOWM to draw the coordinates where the user should click. But screen reader is supposed to be for users that can't see the screen, so drawing on it makes no sense. Instead, BuddyUp should try a different approach to guide the user if the screen reader is active.
@rik what do you think?

apps/system/style/zindex.css
ferjm: This is also a good point. However we need this window to be on top of *any* window. It is rendered transparent and allowing to pass the pointer events, so unless the overlay window renders some UI that captures the pointer events, this should never happen. I agree that there's still a small chance that this happens though.

try-server-hook:
[ferjm Fernando Jiménez Moreno (ferjm)](https://github.com/ferjm) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=6c74fc9a4303)

apps/system/js/hierarchy_manager.js
ferjm: Thinking about this twice. I still believe that BuddyUp should use another approach if the screen reader is active. There is no point about showing visual feedback to a blind user. But still the screen reader should also be able to read the GOWM. That doesn't hurt and it would help future use cases different to the BuddyUp one.

try-server-hook:
[ferjm Fernando Jiménez Moreno (ferjm)](https://github.com/ferjm) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=005ccdfa8e9a)

apps/system/js/hierarchy_manager.js
ferjm: FWIW with the updated PR the screen reader can now read the GOW.

Rik: I don't think the screen reader issue is a problem for us. If a helpee can't see the screen, the helper will not point parts of the screen and just use text or voice to guide the helpee.

PR #27071, opened 2014-12-31
Bug 1105669 - Remove data connection checkings when enable interner sharing

shamenchens:


apps/system/js/internet_sharing.js
```diff
@@ -141,104 +138,15 @@
      * @memberof InternetSharing.prototype
      */
     internetSharingSettingsChangeHanlder: function(evt) {
```
EragonJ: nit: `internetSharingSettingsChangeHandler`

apps/system/locales/system.en-US.properties
```diff
@@ -559,13 +559,9 @@ search-or-enter-address=Search or enter address
 search-or-enter-address.placeholder=Search or enter address
 
 # Hotspot
-noSIMCardInHotspot2=SIM card is absent or inserted incorrectly, so turn off hotspot.
-noHotspotWhenAPMisOnWifiHotspot=Turn off airplane mode to use Wi-Fi hotspot.
-noHotspotWhenAPMisOnUsbHotspot=Turn off airplane mode to use USB tethering.
+noSIMCardInHotspot=SIM card is absent or inserted incorrectly, so turn off hotspot
+noHopspotWhenAPMisOn=Turn off airplane mode to use Wi-Fi hotspot
```
EragonJ: We may need to remove these two strings because there is no one using it anymore.

PR #27075, opened 2014-12-31
Bug 1060915 - Accept view-source: protocol

takenspc:
This patch makes the browser app accept view-source: URLs. 

It also unify scheme checking logic because there were regexp patterns which cannot handle scheme containing "-".


shared/js/url_helper.js
```diff
@@ -33,10 +31,10 @@ var UrlHelper = {
     var case1Reg = /^(\?)|(\?.+\s)/;
     // for cases, pure string
     var case2Reg = /[\?\.\s\:]/;
-    // for cases, data:uri
-    var case3Reg = /^(data\:)/;
+    // for cases, data:uri and view-source:uri
+    var case3Reg = /^(data|view-source)\:/;
     // for cases, only scheme but no domain provided
-    var case4Reg = /^\w+\:\/*$/;
+    var case4Reg = /^[a-z\u00a1-\uffff0-9-+]+\:\/*$/;
```
KevinGrandon: This looks like it's trying to match against the scheme. Instead of re-declaring a very similar regex here, can we use ```this.hasScheme()``` instead of the ```case4Reg.test()``` call in the next if statement?

PR #27082, opened 2014-12-31
Bug 1116780 - [Messages] We should not focus the composer after clicking...

julienw:
... a notification r=azasypkin


apps/sms/js/activity_handler.js
```diff
@@ -332,7 +334,9 @@ var ActivityHandler = {
         return;
```
julienw: miss a resolve here (and possibly in other places? need to check)

PR #27084, opened 2014-12-31
Bug 1113002 - The MMS app is closed unexpectly when is attached an secon...

julienw:
...d image from Gallery r=azasypkin

### DIFF
--- a/apps/sms/js/utils.js
+++ b/apps/sms/js/utils.js
@@ -654,6 +654,9 @@
             img, adjustedSize.width, adjustedSize.height
           );
 
+          // cleanup before the call to toDataURL as it will need more memory
+          cleanup();
+
           deferred.resolve({
             width: adjustedSize.width,
             height: adjustedSize.height,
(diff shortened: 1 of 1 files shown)


apps/sms/js/utils.js
```diff
@@ -653,6 +653,7 @@
           canvas = Utils.imageToCanvas(
             img, adjustedSize.width, adjustedSize.height
           );
+          img.src = '';
```
azasypkin: mmm, it looks like we do the same below in ```cleanup``` function that is called once this deferred resolved or rejected. Is there a reason why it can't wait for the ```cleanup``` function call?

julienw: I think one problem is that we call toDataURL before cleanup.

azasypkin: Okay, if the root cause of this issue is here then I'm fine, just try to call ```cleanup``` here instead and remove it from ```deferred.promise.then``` callback.

try-server-hook:
[julienw Julien Wajsberg (julienw)](https://github.com/julienw) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=28fa68591f7a)

PR #27111, opened 2015-01-05
Bug 1108984 - [Woodduck][Dialer]The "No missed message" info will be dis...

begeeben:
...played in "All" lab after deleting the missed call log in "All".
- Check if the call log is filtered before showing empty call log.

### DIFF
--- a/apps/communications/dialer/test/unit/call_log_test.js
+++ b/apps/communications/dialer/test/unit/call_log_test.js
@@ -790,9 +790,37 @@ suite('dialer/call_log', function() {
       sinon.assert.notCalled(CallLog.renderEmptyCallLog);
     });
 
-    test('renderEmptyCallLog should be called when there is no missedgroup',
+    test('renderEmptyCallLog should be not called when there is no ' +
+         'missedgroup and it is in all call log view',
     function() {
       this.sinon.spy(CallLog, 'renderEmptyCallLog');
+      CallLog.appendGroup(missedGroup);
+      CallLog.unfilter();
+      var container = CallLog.callLogContainer;
+      var checkboxes =
+        container.querySelectorAll('.missed-call input[type="checkbox"]');
+      for (var i = 0; i < checkboxes.length; i++) {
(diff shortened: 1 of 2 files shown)


apps/communications/dialer/test/unit/call_log_test.js
```diff
@@ -790,9 +790,19 @@ suite('dialer/call_log', function() {
...
+      var container = CallLog.callLogContainer;
+      var checkboxes =
+        container.querySelectorAll('.missed-call input[type="checkbox"]');
+      for (var i = 0; i < checkboxes.length; i++) {
+        var checkbox = checkboxes[i];
+        checkbox.checked = true;
+      }
       CallLog.deleteLogGroups();
       ConfirmDialog.executeYes();
       sinon.assert.calledOnce(CallLog.renderEmptyCallLog);
```
tamarahills: Would it be possible to change this test a bit so that it verifies that in addition to just the renderEmptyCallLog being called that the correct message for the noMissedCallsMsg is displayed in the right filter tab?  Otherwise, I think the test passes both with and without the change to call_log.js.

apps/communications/dialer/test/unit/call_log_test.js
```diff
@@ -790,11 +790,30 @@ suite('dialer/call_log', function() {
       sinon.assert.notCalled(CallLog.renderEmptyCallLog);
     });
 
-    test('renderEmptyCallLog should be called when there is no missedgroup',
+    test('renderEmptyCallLog should be not called when there is no ' +
+         'missedgroup and it is in all call log view',
     function() {
       this.sinon.spy(CallLog, 'renderEmptyCallLog');
       CallLog.deleteLogGroups();
```
tamarahills: Hi Yi-Fan,

Could we make this test look a bit more like the one below it where we append the missedGroup, (and leave it unfiltered/all vs. filtered) and the query for the missed call and set the checkbox to true and then delete(like you did in the test below).  I'm thinking that's closer to the failure case that you fixed.

PR #27116, opened 2015-01-05
Bug 1113499 - [Stingray][Smart-home] Open app with specified hash/URL

huchengtw-moz:


try-server-hook:
[huchengtw-moz John Hu (huchengtw-moz)](https://github.com/huchengtw-moz) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=f8323e72c424)

tv_apps/tv_shared/js/card_manager.js
```diff
@@ -553,8 +553,13 @@
         } else if (query.manifestURL && card.nativeApp &&
             card.nativeApp.manifestURL === query.manifestURL) {
 
-          found = card;
-          return true;
+          // The launchURL only happens to AppBookmark, we only need to check if
+          // launchURL is exactly the same in this case.
+          if (!(card instanceof AppBookmark) ||
```
dwi2: If I understand the comment correctly, this line should be 
```js
if ((card instanceof AppBookmark) ||
```
right?

dwi2: Oh I understood you wrong, this means we need to check launchURL also if card is AppBookmakr

rexboy7:
In fact we are planning to merge `AppBookmark` back to `Application` so I would suggest to move what you did back to `Application`.... but we have another bug  for that so it's up to you.

tv_apps/smart-system/js/app_window.js
```diff
@@ -1667,18 +1667,17 @@
...
-      if (iframe.src !== url) {
-        // Rewrite the URL of the app frame to the requested URL.
-        // XXX: We could ended opening URls not for the app frame
-        // in the app frame. But we don't care.
-        iframe.src = url;
-      }
+    // XXX: If the app is in foreground, it's too risky to change it's
+    // URL. We still change it because the home app is an overlay on top of any
+    // app.
+    var iframe = this.browser.element;
```
rexboy7: Should we prevent `homescreen` from being changed here?

tv_apps/tv_shared/js/card_manager.js
huchengtw-moz: Yes, you are correct... We need to check the launchURL only when the card is AppBookmark.

tv_apps/smart-system/js/app_window.js
huchengtw-moz: Nice catch. This means we should create a black list to prevent this behavior on some apps. I would prefer to use the simplest code here and add the black list when we really need one. Since this API is based on IAC, no 3rd-party app can access it. We can prevent it by programmers.

huchengtw-moz:
I had discussed with dwi2 the merge of AppBookmark and Application. He will handle that.

PR #27123, opened 2015-01-05
list of your playlists

errietta:


try-server-hook:
[errietta Errietta Kostala (errietta)](https://github.com/errietta) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=965d8bbacf65)

apps/music/js/ui/views/list_view.js
```diff
@@ -367,31 +414,22 @@ var ListView = {
...
-          }
-        }
+      case 'contextmenu':
+        var option = target.dataset.option;
 
-        if (target.id === 'views-list-search-clear') {
-          lv_resetSearch(this);
-          return;
+        if (option === 'title') {
+          var playlistName = prompt("Add to playlist?");
```
hfiguiere: we should make this string localizable since it is a user visible string.

errietta: @hfiguiere  I actually want to change this to be the way people wanted it so it asks to pick bluetooth or add to playlist, but I didn't know how to do that and also provide a list of playlists people can choose from instead of just text input!

errietta:
let's keep one PR for these chages, https://github.com/mozilla-b2g/gaia/pull/27125

PR #27128, opened 2015-01-05
Bug 1107714 - [Gallery] When setting a wallpaper in landscape, the wallpaper will expand from its normal size r=djf

punamdahiya:


try-server-hook:
[punamdahiya punamdahiya](https://github.com/punamdahiya) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=b4c014148fda)

apps/wallpaper/js/share.js
```diff
@@ -26,11 +26,14 @@ window.onload = function() {
...
-  var screenWidth = Math.ceil(screen.width * window.devicePixelRatio);
-  var screenHeight = Math.ceil(screen.height * window.devicePixelRatio);
+  var screenWidth;
+  var screenHeight;
 
   function startShare(request) {
+    // How big (in device pixels) is the screen?
+    screenWidth = Math.ceil(screen.width * window.devicePixelRatio);
+    screenHeight = Math.ceil(screen.height * window.devicePixelRatio);
+
```
davidflanagan: It makes me nervous to just move these assignments here, given the async nature of resize events.  Maybe it would be better to leave the initial assignments where they were, and also register a resize handler to detect when they change.

Or, since this is for wallpaper, we could always just take the longer dimension as the height and the shorter as the width:

```
screenWidth = Math.ceil(Math.min(screen.width, screen.height) * window.devicePixelRatio);
screenHeight = Math.ceil(Math.max(screen.width, screen.height) * window.devicePixelRatio);
```

apps/wallpaper/js/share.js
```diff
@@ -70,8 +73,8 @@ window.onload = function() {
 
         // Compute a scale that will make the image at least as big as
         // the screen in both dimensions.
-        var scalex = window.innerWidth / previewImage.width;
-        var scaley = window.innerHeight / previewImage.height;
+        var scalex = screen.width / previewImage.width;
+        var scaley = screen.height / previewImage.height;
```
davidflanagan: I'm not sure about changing window.innerWidth to screen.width here and elsewhere.

IIRC, the reason we use screen.width above is that when there is a soft home button (on nexus 4, e.g.) screen.height and window.innerHeight have different values. But we still want the wallpaper to be the full screen height.  I think that the rest of the computations are for manipulating the wallpaper preview image, and for those, using the window dimensions instead of the screen dimensions is correct.

I don't remember the details clearly. If you do a git blame on line 30, though, you might be able to figure out what bug it was when we started using screen dimensions instead of window dimensions, and that should explain why we're doing it.

try-server-hook:
[punamdahiya punamdahiya](https://github.com/punamdahiya) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=9048bec088f6)

apps/wallpaper/js/share.js
```diff
@@ -26,9 +26,11 @@ window.onload = function() {
...
   var previewImage = document.getElementById('previewImage');
 
-  // How big (in device pixels) is the screen?
-  var screenWidth = Math.ceil(screen.width * window.devicePixelRatio);
-  var screenHeight = Math.ceil(screen.height * window.devicePixelRatio);
+   // How big (in device pixels) is the screen?
+  var screenWidth = Math.ceil(Math.min(screen.width, screen.height) *
+                    window.devicePixelRatio);
+  var screenHeight = Math.ceil(Math.max(screen.width, screen.height) *
+                     window.devicePixelRatio);
```
punamdahiya: This fixes intermittent IndexSizeError  seen at Line141 due to incorrect screen.width and screen.height.

01-02 07:14:45.360 W/Wallpaper(10156): [JavaScript Error: "IndexSizeError: Index or size is negative or greater than the allowed amount" {file: "app://wallpaper.gaiamobile.org/js/share.js" line: 141}]

apps/wallpaper/js/share.js
```diff
@@ -26,9 +26,11 @@ window.onload = function() {
...
```
davidflanagan: Nit: please fix the indentation on lines 31 and 33 so that 'window.device' lines up with Math.min and Math.max

apps/wallpaper/js/share.js
```diff
@@ -117,6 +119,9 @@ window.onload = function() {
   }
 
   function scaleImage() {
+      // Disable set wallpaper button to prevent hammering and causing multiple
+      // calls to scaleImage.
+      setButton.disabled = true;
```
davidflanagan: Nice find. Thanks for fixing this.

One minor nit.  The line that creates the canvas doesn't have a canvas and looks like it is grouped with this disable code. If it was me, I'd move the createElement() call down three lines, and update the comment below so it talks about creating a canvas that is exactly the size of the screen in device pixels.

try-server-hook:
[punamdahiya punamdahiya](https://github.com/punamdahiya) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=98e5919054fd)

try-server-hook:
[punamdahiya punamdahiya](https://github.com/punamdahiya) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=f33a72bc17d7)

PR #27156, opened 2015-01-06
Bug 1113154 - Plivo sometimes raises a 404 error just after the call has...

JohanLorenzo:
... been created


try-server-hook:
[JohanLorenzo Johan Lorenzo (JohanLorenzo)](https://github.com/JohanLorenzo) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=b4b19285560f)

chirarobert:
@JohanLorenzo How does adding another case for the error number help us in this case? Even if the error is raised, wouldn't plivo still attempt to reconnect the call thus affecting the test that follows?

PR #27169, opened 2015-01-06
Bug 1118021 - FLAC art cover support

hfiguiere:


try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=74542346abf4)

hfiguiere:
Include bug 907929 that is already in r? in a separate PR.

try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=fc3369de11e4)

apps/music/js/metadata/flac.js
```diff
@@ -54,11 +88,9 @@ var FLACMetadata = (function() {
    */
   function findVorbisCommentBlock(blobview) {
```
jimporter: Since we're expecting to find multiple blocks now, I think we should switch to using a callback instead of a Promise. My general rule for Promises is to use them for something that's simply async, but if it's more like an event (e.g. it can happen multiple times), a callback is better. This would eliminate the need for us to call this function inside of the handler (see line 78).

Also, since this function is doing more than just finding the Vorbis Comment block, I think we should rename it to `getMetadataBlocks`.

apps/music/js/metadata/flac.js
```diff
@@ -54,11 +88,9 @@ var FLACMetadata = (function() {
...
     return readMetadataBlockHeader(blobview).then(function(block) {
-      // XXX: Support album art.
-      // See: http://flac.sourceforge.net/format.html#metadata_block_picture
 
       // Did we find a Vorbis comment block yet?
-      if (block.block_type === 4) {
+      if (block.block_type === 4 || block.block_type === 6) {
```
jimporter: We can probably eliminate this line and just have the callback filter out the metadata blocks it doesn't want.

apps/music/js/metadata/flac.js
hfiguiere: Will be simplier.

apps/music/js/metadata/flac.js
hfiguiere: Yes we can. And it will.

try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=f6d2bafb6190)

apps/music/js/metadata/flac.js
```diff
@@ -82,31 +110,31 @@ var FLACMetadata = (function() {
...
-      var header = blobview.readUnsignedByte();
-      var last = (header & 0x80) === 0x80;
-      var block_type = header & 0x7F;
-      var length = blobview.readUint24(false);
-
-      // Get the contents of this block, plus enough to read the next block's
-      // header if necessary.
-      blobview.getMore(blobview.viewOffset + blobview.index, length + 4,
-                       function(more, err) {
+  function readMetadataBlockHeader(blobview, callback) {
```
jimporter: I think we should leave this function how it was; turning it into a callback doesn't simplify things here.

apps/music/js/metadata/flac.js
```diff
@@ -36,37 +37,64 @@ var FLACMetadata = (function() {
...
+            metadata.picture.end += block.view.viewOffset;
+          });
+          has_picture = true;
+
+        }
+
+        if (!has_vorbis_comment || !has_picture) {
+          try {
+            block.view.advance(block.length - block.view.index);
+            findMetadataBlocks(block.view, processBlock);
```
jimporter: Instead of seeking and finding another metadata block here, let's do this in `findMetadataBlocks`.

apps/music/js/metadata/flac.js
```diff
@@ -36,37 +37,64 @@ var FLACMetadata = (function() {
...
-        return findVorbisCommentBlock(block.view);
-      }
-    });
+  function findMetadataBlocks(blobview, callback) {
+    readMetadataBlockHeader(
+      blobview, function(block, err) {
+        if (err || block.last) {
+          callback(null);
+        } else {
+          callback(block);
```
jimporter: We could change this to something like:

```
if (callback(block)) {
  block.view.advance(block.length - block.view.index);
  findMetadataBlocks(block.view, callback);
}
```

Then `processBlock` above will work more like a JS event listener: it'll get automatically called a few times until we're done; we won't need to explicitly start the search for another block inside `processBlock`.

apps/music/js/metadata/flac.js
hfiguiere: It does simplify its use though. So I'm good with it.

jimporter: I think it'll actually make it a little more complex to use once you've addressed my comments above.

apps/music/js/metadata/flac.js
jimporter: To elaborate a little more (and to explain why I like `readMetadataBlockHeader` the way it was), here's how I'd write this function. This also fixes an off-by-one bug that's my fault (We still need to process the last block, but then we need to stop *after* that one):

```
function findMetadataBlocks(blobview, callback) {
  readMetadataBlockHeader(blobview).then(function(block) {
    callback(block);
    if (block.last) {
      callback(null);
    } else {
      block.view.advance(block.length - block.view.index);
      findMetadataBlocks(block.view, callback);
    }
  }).catch(function(err) {
    console.error(err);
    callback(null);
  });
}
```

This has the nice side-effect that if any errors are thrown during our `then`, the `catch` will automatically pick them up, so we don't need to worry about adding a try block around `block.view.advance(...)`.

jimporter: You could also add some early-exit code to the above function if, say, `processBlock` returns false (to signify that we already have all the metadata we need).

hfiguiere: ok

apps/music/js/metadata/flac.js
hfiguiere: in that context, ok

apps/music/js/metadata/flac.js
hfiguiere: I had pondered that. done.

try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=a1c6a871311f)

try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=e696471d0a43)

try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=2dc8564eed97)

apps/music/js/metadata/flac.js
```diff
@@ -36,36 +37,55 @@ var FLACMetadata = (function() {
...
+      function processBlock(block) {
+        if (!block) {
+          resolve(metadata);
+          return false;
+        }
+
+        if (block.block_type === 4) {
+          readAllComments(block.view, metadata);
+          has_vorbis_comment = true;
+
```
jimporter: Nit: this blank line looks weird to me, but I have no especially strong feelings about it. Just wanted to mention it.

apps/music/js/metadata/flac.js
```diff
@@ -92,20 +112,21 @@ var FLACMetadata = (function() {
...
-          last: last,
-          block_type: block_type,
-          length: length,
-          view: more
+          resolve({
+            last: last,
+            block_type: block_type,
+            length: length,
+            view: more
+          });
```
jimporter: These are just whitespace changes, right? It might be worth reverting them to keep the diff smaller, but I don't feel strongly about it one way or the other.

apps/music/js/metadata/flac.js
hfiguiere: I can cut it loose.

apps/music/js/metadata/flac.js
hfiguiere: The whole function was indented weird. So yes it is a whitespace only change... :-/
But there was actual changes in there at one point - so I'm fine reverting it.

try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=b7eb75285803)

PR #27172, opened 2015-01-06
Bug 1118444 - Check against all navigator.languages values for possible supported locales.

Standard8:


try-server-hook:
[Standard8 Mark Banner (Standard8)](https://github.com/Standard8) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=22a28192cd2a)

try-server-hook:
[Standard8 Mark Banner (Standard8)](https://github.com/Standard8) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=86d5fa2264c1)

Standard8:
I just pushed a minor update to avoid an else after an early return.

apps/sharedtest/test/unit/l10n/lib/context/langnego_test.js
```diff
@@ -44,3 +44,20 @@ describe('Language negotiation with arguments', function() {
...
 });
+
+describe('Language negotiation with arguments and unsupported', function() {
+  var ctx;
+
+  beforeEach(function(done) {
+    ctx = new Context();
+    ctx.once(done);
+    ctx.registerLocales('en-US', ['pl']);
+    ctx.requestLocales('pl-PT', 'pl');
```
stasm: Can you use something else than `pl-PT` here? Maybe `de` or `fr` which I think are often used in other l10n.js tests.

shared/js/l10n.js
```diff
@@ -1348,12 +1348,21 @@
   // Getting ready
 
   function negotiate(available, requested, defaultLocale) {
-    if (available.indexOf(requested[0]) === -1 ||
-        requested[0] === defaultLocale) {
+    var supportedLocale;
+    // Find the first locale in the requested list that is supported.
+    for (var i = 0; i < requested.length; ++i) {
```
stasm: Is there any difference between `++i` and `i++` here?  It seems that in other parts of the code we use post-incrementing for loop counters.

Standard8: Probably not, in a past life I used to do that a lot, and haven't done recently, don't know why I did this time ;-) I'll change it.

apps/sharedtest/test/unit/l10n/lib/context/langnego_test.js
Standard8: pl seems to be the one used in this directory, and in the fixtures sub-directory, there's only pl.properties and en-US.properties.

I also took a look, and de/fr aren't so convenient as there accept languages start off with the locale they are actually for.

PR #27178, opened 2015-01-07
Bug 1112987 - After monkey test phone stay at browser or other apps, couldn't back to homescreen. R=etienne

songben423:
...dn't back to homescreen

[bug number  ] 1112987
[root cause  ] while app would swipe in by sheetTransition, it crashed at this time
[changes     ] add a flag about app crashed, if crashed this app would stop swiping in
[side effects] none
[self test   ] v2.1 self test is OK
[whether AOB ]
[reviewers   ] etienne@segonzac.info

### DIFF
--- a/apps/system/js/app_window.js
+++ b/apps/system/js/app_window.js
@@ -338,6 +338,7 @@
     if (this.browser) {
       return;
     }
+    this.isCrashed = false;
     this.debug(' ...revived!');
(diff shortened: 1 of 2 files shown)


try-server-hook:
[songben423 songben423](https://github.com/songben423) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=898de0843348)

PR #27193, opened 2015-01-07
Bug 1113851 - [Settings] Change time format to 24-hour, write gaia-ui te...

bebef1987:
...st

### DIFF
--- a/tests/python/gaia-ui-tests/gaiatest/tests/functional/settings/test_settings_change_time_format.py
+++ b/tests/python/gaia-ui-tests/gaiatest/tests/functional/settings/test_settings_change_time_format.py
@@ -1,0 +1,47 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from gaiatest import GaiaTestCase
+from gaiatest.apps.settings.app import Settings
+from gaiatest.apps.system.app import System
+from gaiatest.apps.lockscreen.app import LockScreen
+
+
+class TestSettingsChangeTimeFormat(GaiaTestCase):
+    def setUp(self):
+        GaiaTestCase.setUp(self)
+
+        # Set the time on the device
+        _seconds_since_epoch = self.marionette.execute_script("""
+                    var today = new Date();
+                    var yr = today.getFullYear();
+                    var mth = today.getMonth();
+                    var day = today.getDate();
+                    return new Date(yr, mth, day, 22, 0, 0).getTime();""")
+        self.data_layer.set_time(_seconds_since_epoch)
+
+    def test_settings_change_time_format(self):
+        """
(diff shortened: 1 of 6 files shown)


try-server-hook:
[bebef1987 Florin Strugariu (bebef1987)](https://github.com/bebef1987) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=b9715e9e5f04)

tests/python/gaia-ui-tests/gaiatest/tests/functional/settings/test_settings_change_time_format.py
```diff
@@ -0,0 +1,47 @@
...
+    def setUp(self):
+        GaiaTestCase.setUp(self)
+
+        # Set the time on the device
+        _seconds_since_epoch = self.marionette.execute_script("""
+                    var today = new Date();
+                    var yr = today.getFullYear();
+                    var mth = today.getMonth();
+                    var day = today.getDate();
+                    return new Date(yr, mth, day, 22, 0, 0).getTime();""")
```
geoelectric: Re: Robert's note, I don't think you want to count on the controlling system and the target having the same time. So doing all time operations in the context of the target makes sense here.

geoelectric:
So, codewise this looks pretty good.

Couple of concerns.

Biggest concern is that this has the side effect of changing the time on the target device, and doesn't change it back. I think we should save off and offset and restore it.

Setup:

1) read the current time. subtract from epoch of (today 10pm), which could be a positive or negative offset depending on when the test is run.

2) Set to today 10pm, as now.

Test:

Do the test as written.

Teardown:

Reset time to current - offset. 

That should restore things as closely as possible.

Second concern is that nothing is explicitly checking that the Date/Time is starting out at 12 hours. We do implicitly check via the status bar verification of 10:00 to start, so this is a more minor concern. If that error gets thrown, though, not sure it'll be obvious to someone triaging why it happened. 

It'd be more robust to save off the current format setting and set to 12 hour in setup, then run the test as written, then restore the original setting in teardown.

Final concern isn't an automation problem, it's a test problem. Nobody's checking the field on Date/Time that has the current time in it. That should also flip to 24 hours. It might be worth adding the check here, even though it's over and above the MozTrap test as written.

I'd like to hear back on at least the first concern before reviewing.

bebef1987:
make sure in the setup of each test that the ime and date of the phone are accurate. No need for any teardown steps here

chirarobert:
The test failed on treeherder. @bebef1987

try-server-hook:
[bebef1987 Florin Strugariu (bebef1987)](https://github.com/bebef1987) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=2c9f243cdc5f)

try-server-hook:
[chirarobert chirarobert](https://github.com/chirarobert) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=4476d42dfe84)

PR #27202, opened 2015-01-07
Bug 1115104 - [RTL][Homescreen] In RTL the homescreen icons/apps is not align properly

anefzaoui:


try-server-hook:
[anefzaoui Ahmed Nefzaoui (anefzaoui)](https://github.com/anefzaoui) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=7ce3c5b0dba0)

try-server-hook:
[anefzaoui Ahmed Nefzaoui (anefzaoui)](https://github.com/anefzaoui) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=4ba5fbca55b8)

try-server-hook:
[anefzaoui Ahmed Nefzaoui (anefzaoui)](https://github.com/anefzaoui) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=ff0acac8c99c)

shared/elements/gaia_grid/js/grid_view.js
```diff
@@ -467,11 +469,15 @@
           var lastItemInRow = this.items[idx - 1];
           step(lastItemInRow);
         }
-
+        var reversedPositions =
+              (windowWidth - this.layout.gridItemWidth) - (x * this.layout.gridItemWidth);
+        var decideXPosition = isRTL ?
+                  reversedPositions : x * this.layout.gridItemWidth;
```
Cwiiis: This should be in the same block as setCoordinates, otherwise we're doing extra work for nothing. I also think it would be better expressed like this:

var xPosition = x * this.layout.gridItemWidth;
if (isRTL) {
  xPosition = (windowWidth - this.layout.gridItemWidth) - xPosition;
}

shared/elements/gaia_grid/js/grid_view.js
```diff
@@ -467,11 +469,15 @@
           var lastItemInRow = this.items[idx - 1];
           step(lastItemInRow);
         }
-
```
Cwiiis: Don't remove this blank line please, it makes things hard to read if we squish them together too much :)

shared/elements/gaia_grid/js/grid_view.js
```diff
@@ -469,9 +471,14 @@
         }
 
         item.setPosition(idx);
+
```
Cwiiis: Looks like there's trailing spaces here - if you could get rid of those, us OCD folks would thank you :)

anefzaoui: No worries. Updating :)

try-server-hook:
[anefzaoui Ahmed Nefzaoui (anefzaoui)](https://github.com/anefzaoui) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=2cb637c06550)

try-server-hook:
[anefzaoui Ahmed Nefzaoui (anefzaoui)](https://github.com/anefzaoui) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=cfab6eb39c2a)

try-server-hook:
[anefzaoui Ahmed Nefzaoui (anefzaoui)](https://github.com/anefzaoui) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=cf968a572f3d)

try-server-hook:
[anefzaoui Ahmed Nefzaoui (anefzaoui)](https://github.com/anefzaoui) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=45e2f6db10e0)

shared/elements/gaia_grid/js/grid_view.js
```diff
@@ -378,8 +378,10 @@
         var item = new GaiaGrid.Placeholder();
         this.items.splice(idx + i, 0, item);
         item.setPosition(idx + i);
-        item.setCoordinates((coordinates[0] + i) * this.layout.gridItemWidth,
-                            this.layout.offsetY);
+
+        var coorX = (coordinates[0] + i) * this.layout.gridItemWidth;
+
+        item.setCoordinates(coorX, this.layout.offsetY);
```
Cwiiis: Nothing changes here? Remove this change if that's the case.

shared/elements/gaia_grid/js/grid_layout.js
```diff
@@ -171,7 +171,6 @@
             this.cols = parseInt(prop.value, 10);
             this.gridView.render();
           }
-
```
Cwiiis: Unnecessary change - I'm not against it, but better to change this when other changes are being made in this file than randomly here.

shared/elements/gaia_grid/js/grid_view.js
```diff
@@ -473,6 +477,8 @@
 
           // Insert placeholders to fill remaining space
           var remaining = this.layout.cols - x;
+          if(isRTL)
+            x = (windowWidth - this.layout.gridItemWidth) - x;
```
Cwiiis: Space after if and braces around following statement, to follow style with the rest of the file.

shared/elements/gaia_grid/js/items/group.js
```diff
@@ -183,10 +188,15 @@
 
           var itemVisible = (i - (index - nApps)) < COLLAPSED_GROUP_SIZE;
           if (!itemVisible) {
-            item.setCoordinates(x - width, y);
+            var xAxis = isRTL ? x + width : x - width;
```
Cwiiis: nit, call this xCoord or transformedX, or even just move the statement into the setCoordinates call and get rid of the variable altogether.

shared/elements/gaia_grid/js/items/group.js
```diff
@@ -183,10 +188,15 @@
...
+            item.setCoordinates(xAxis, y);
           } else {
             item.setCoordinates(x, y);
-            x += width;
+            if (isRTL) {
+              x -= width;
+            } else {
+              x += width;
+            }
```
Cwiiis: This would be nicer as 'x += isRTL ? -width: width;'

shared/elements/gaia_grid/style.css
```diff
@@ -374,3 +374,9 @@
 .edit-mode.hover-over-top {
   transform: translateY(0.8rem);
 }
+
+/* RTL View */
+:-moz-dir(rtl) .group .header .toggle {
+  left: 0;
+  right: unset;
```
Cwiiis: Interesting, wasn't aware of this syntax :) (or is it that any non-number will cause it to be unset?)

apps/verticalhome/test/marionette/grid_layout_test.js
```diff
@@ -0,0 +1,71 @@
...
+  });
+
+  test('apps grid is LTR in LTR language', function() {
+    var communicationManifestUrl =
+        'app://communications.gaiamobile.org/manifest.webapp';
+    var phoneIcon = home.getIcon(communicationManifestUrl, 'dialer').location();
+    var contactsIcon = home.getIcon(communicationManifestUrl, 'contacts')
+      .location();
+    var messageIcon = home.getIcon('app://sms.gaiamobile.org/manifest.webapp')
+      .location();
```
Cwiiis: This relies on us installing these apps and never changing their order. It would make more sense to just get the first two app icons and test that the first is aligned to the left, and that the second is to the right of the first.

apps/verticalhome/test/marionette/grid_layout_test.js
```diff
@@ -0,0 +1,71 @@
...
+      var phoneIcon = home.getIcon(communicationManifestUrl, 'dialer');
+      return phoneIcon.text() ===
+        home.localizedAppName('communications', 'dialer', 'ar');
+    });
+
+    var phoneIcon = home.getIcon(communicationManifestUrl, 'dialer').location();
+    var contactsIcon = home.getIcon(communicationManifestUrl, 'contacts')
+      .location();
+    var messageIcon = home.getIcon('app://sms.gaiamobile.org/manifest.webapp')
+      .location();
```
Cwiiis: Same here, just get the first two icons and test that the first is aligned to the right and the second is to the left of the first.

apps/verticalhome/test/marionette/grid_layout_test.js
Cwiiis: Side-note, 'aligned to the left' is probably best tested as x < (gridWidth/columns), to account for padding (and where I assume padding will never be so huge as to be greater than the grid item width).

shared/elements/gaia_grid/style.css
anefzaoui: AFAIK only unset behaves this way :)
https://developer.mozilla.org/en-US/docs/Web/CSS/unset
Best of all, applicable to all properties.

try-server-hook:
[anefzaoui Ahmed Nefzaoui (anefzaoui)](https://github.com/anefzaoui) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=cad7a9fd244d)

apps/verticalhome/test/marionette/grid_layout_test.js
anefzaoui: > Side-note, 'aligned to the left' is probably best tested as x < (gridWidth/columns), to account for padding (and where I assume padding will never be so huge as to be greater than the grid item width).

Actually here I test the order of the icons (respective to x coordinates) and not the fact that they are left-aligned. It seems sensible to me because that what matters to users anyway, and also because we don't care about padding this way. What do you think?

Cwiiis: Do we not care about the alignment though? You'll have to excuse if this is a basic question, I don't have much RTL experience - I'd have expected if things are going from right to left that they also align to the right - if it ordered RTL, but aligned to the left would it not feel a bit odd? Given your patch does this, I think it'd be sensible to test it - but if you disagree, I'd be fine with that.

try-server-hook:
[anefzaoui Ahmed Nefzaoui (anefzaoui)](https://github.com/anefzaoui) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=8222fbd29459)

try-server-hook:
[anefzaoui Ahmed Nefzaoui (anefzaoui)](https://github.com/anefzaoui) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=eb464e7af3b7)

anefzaoui:
> Do we not care about the alignment though? You'll have to excuse if this is a basic question, I don't have much RTL experience - I'd have expected if things are going from right to left that they also align to the right - if it ordered RTL, but aligned to the left would it not feel a bit odd? Given your patch does this, I think it'd be sensible to test it - but if you disagree, I'd be fine with that.

We don't really left or right aligne stuff, we translate icons to a certain number of pixels calculating in js. Your logic works fine in LTR, but as there is actually no padding, in RTL it begins to be hell : I would need to test appX > 2/3 * gridWidth, but that's actually not the case ! 2/3 * gridWidth is like 0.000001 px bigger than appX (rounding error inside) :) This will make us do some very bad hacks to make this logic work in my opinion. So that's why I kept only the check on ordering here (in LTR as well for the sake of consistency). Please let me know if you have a better idea

Cwiiis:
Instead of testing appX > 2/3 * gridWidth, you could just assume that column width is 3 and test that appX > gridWidth/2 - that would be a reasonable sanity check that it's at least not left-aligned.

try-server-hook:
[anefzaoui Ahmed Nefzaoui (anefzaoui)](https://github.com/anefzaoui) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=db397e8cd2f4)

apps/verticalhome/test/marionette/grid_layout_test.js
```diff
@@ -0,0 +1,75 @@
...
+    var app2Location = home.getNthIcon(2).location();
+    var app3Location = home.getNthIcon(3).location();
+    var gridWidth = client.helper.waitForElement(home.Selectors.grid).size()
+      .width;
+
+    // check that they are all on the same line
+    assert.equal(app1Location.y, app2Location.y,
+                 'First 2 apps should be on the same line.');
+    assert.equal(app2Location.y, app3Location.y,
+                 'Last 2 apps should be on the same line.');
```
Cwiiis: It's nit picking, but I'd reword this not to say 'last 2', which is a little misleading - perhaps we could just assert.ok(app1Location.y == app2Location.y == app3.Location.y, 'First three apps should be on the same line')? I don't like the three-way equivalence check though either... So maybe just reword '2nd and 3rd apps should be on the same line' ? I'm fine with however you'd like to do this, including leaving as is - it's pretty clear what it's checking all the same :)

shared/elements/gaia_grid/js/grid_view.js
```diff
@@ -467,6 +469,9 @@
 
           // Insert placeholders to fill remaining space
           var remaining = this.layout.cols - x;
+          if (isRTL) {
+            x = (windowWidth - this.layout.gridItemWidth) - x;
```
Cwiiis: I hadn't thought of this before, but this assumes that the grid will take the whole window (which it will in all of our use-cases, but still) - this could be expressed as (this.layout.gridItemWidth * (this.layout.cols - 1)) - x; Alternatively, change windowWidth to gridWidth and calculate that above instead of getting the window width (there may already be a layout function for this, I don't recall off-hand).

shared/elements/gaia_grid/js/items/group.js
```diff
@@ -36,6 +36,8 @@
     this.pendingCollapse = this.detail.collapsed = !!this.detail.collapsed;
   }
 
+  var windowWidth = window.innerWidth;
+
```
Cwiiis: Same concern here about conflating window-width and grid-width. If GridLayout doesn't have a gridWidth function, it might be nice to add one.

try-server-hook:
[anefzaoui Ahmed Nefzaoui (anefzaoui)](https://github.com/anefzaoui) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=c95be13cc585)

try-server-hook:
[anefzaoui Ahmed Nefzaoui (anefzaoui)](https://github.com/anefzaoui) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=42a9d1f6877d)

PR #27214, opened 2015-01-07
Bug 1116444: handle files that have invalid EXIF orientation values

davidflanagan:
There's no need to treat invalid EXIF orientation values as an error. We can just ignore them and use the default orientation. This will make the parser more robust.


try-server-hook:
[davidflanagan David Flanagan (davidflanagan)](https://github.com/davidflanagan) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=ca0fb6ba2f92)

PR #27245, opened 2015-01-08
Bug 1118006 - Make the long-press delay shorter in edit mode. r=kgrandon

Cwiiis:


shared/elements/gaia_grid/js/grid_dragdrop.js
```diff
@@ -698,30 +732,45 @@
...
+                pageY: touch.pageY
+              };
+
+              this.positionIcon();
+
+              if (!this.isScrolling) {
+                this.scrollIfNeeded();
+              }
+            } else if (this.longPressTimeout &&
+                       (Math.abs(touch.pageX - this.touchStart.x) >
```
KevinGrandon: This looks funky to me. I wonder if we could define a helper?
```
 } else if (this.longPressTimeout && this.touchWithinThreshold(touch)) {
```

crdlc: +1

shared/elements/gaia_grid/js/grid_dragdrop.js
```diff
@@ -659,14 +677,32 @@
...
 
-          case 'contextmenu':
-            if (this.icon || this.canceled) {
+            if (e.touches.length > 1) {
+              this.canceled = true;
               return;
             }
 
+            if (this.inEditMode) {
+              this.touchStart.x = e.touches[0].pageX;
```
crdlc: Maybe we could save a reference to e.touches[0] and also a reference to x and y to reuse here https://github.com/mozilla-b2g/gaia/pull/27245/files#diff-18ee9a6d12712069d19e901b9ef493c2R695

shared/elements/gaia_grid/js/grid_dragdrop.js
```diff
@@ -627,24 +650,19 @@
...
-      this.container.removeEventListener('touchend', this);
-      window.removeEventListener('touchcancel', this);
-    },
-
-    addDragHandlers: function() {
-      this.container.addEventListener('touchmove', this);
-      this.container.addEventListener('touchend', this);
-      window.addEventListener('touchcancel', this);
+    cancelLongPressTimeout: function() {
+      if (this.longPressTimeout) {
```
crdlc: perhaps this.longPressTimeout !== null is safer, what do you think? thx

shared/elements/gaia_grid/js/grid_dragdrop.js
Cwiiis: As it only gets used once, this doesn't really save any space - touchStart ends up being that reference. Will leave this as is I think.

shared/elements/gaia_grid/js/grid_dragdrop.js
Cwiiis: Never thought about it before - I don't think you can have a timeout that tests true, but I'm not certain - that's definitely safer and I think it reads better too, will do.

try-server-hook:
[Cwiiis Chris Lord (Cwiiis)](https://github.com/Cwiiis) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=0b6cd1ffc5c9)

try-server-hook:
[Cwiiis Chris Lord (Cwiiis)](https://github.com/Cwiiis) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=0a3301ae5f74)

try-server-hook:
[Cwiiis Chris Lord (Cwiiis)](https://github.com/Cwiiis) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=470249450bdf)

try-server-hook:
[Cwiiis Chris Lord (Cwiiis)](https://github.com/Cwiiis) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=734901b75a44)

try-server-hook:
[Cwiiis Chris Lord (Cwiiis)](https://github.com/Cwiiis) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=a3355d5ce2fb)

try-server-hook:
[Cwiiis Chris Lord (Cwiiis)](https://github.com/Cwiiis) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=4770555f901d)

shared/elements/gaia_grid/js/grid_view.js
```diff
@@ -566,7 +566,9 @@
       // All the children of this element are absolutely positioned and then
       // transformed, so manually set a height for the convenience of
       // embedders.
-      this.element.style.height = this.layout.offsetY + 'px';
+      var padding =
```
KevinGrandon: Hmm, seems a bit fragile. Could we possibly get this same value from something like: ```this.element.getBoundingClientRect().bottom``` ? (Or even the parent of the gaia-grid element if not?

shared/elements/gaia_grid/js/grid_dragdrop.js
```diff
@@ -7,6 +7,16 @@
...
 
+  /* This is an alternative delay used to initiate icon movement when in edit
+   * mode. The normal contextmenu delay is too long once we're in edit mode
+   * and causes user confusion, but a short delay globally would cause too many
+   * accidental triggers of edit mode.
+   */
+  const EDIT_LONG_PRESS_DELAY = 100;
+
+  /* The movement threshold to use for the above synthetic long-press. */
+  const EDIT_LONG_PRESS_THRESHOLD = 5;
```
KevinGrandon: Should we multiple this by ```window.devicePixelRatio```? I have a feeling it would be very easy to go outside of this threshold on a @2.25x and higher device.

apps/verticalhome/style/css/app.css
```diff
@@ -15,9 +15,10 @@ body {
 
 gaia-grid {
   display: block;
-  height: 100%;
+  /* Minus the bottom margin and the search-box height */
+  min-height: calc(100% - 5rem - 5rem);
```
KevinGrandon: Are these css changes needed for this patch?

KevinGrandon:
Hmm, so here's one case that I noticed when pretending to be a monkey. It seems that it's possible to get the icon stuck in some cases when issuing multiple touches. This screenshot was taken with no fingers touching the screen. Still trying to narrow down a solid STR, but it has something to do with multiple touches.

shared/elements/gaia_grid/js/grid_view.js
```diff
@@ -544,13 +544,13 @@
...
             item.render();
+          }
 
-            if (item.detail.type === 'divider') {
-              if (oddDivider) {
-                item.element.classList.add('odd');
-              } else {
-                item.element.classList.remove('odd');
-              }
+          if (item.detail.type === 'divider') {
```
crdlc: if you set oddDivider to null in the initialization you can check here if (oddDivider !== null) instead of item.detail.type === 'divider' because it is a bit slower. if oddDivider is true or false means this is a divider. But just an idea, do it only if you consider good idea

crdlc: Sorry forget this, the previous condition is over items[i] so please keep you code

shared/elements/gaia_grid/js/grid_dragdrop.js
Cwiiis: Good idea, hadn't considered this.

apps/verticalhome/style/css/app.css
Cwiiis: Yes - the height does nothing as we set this manually in grid_view, but if we don't set min-height, the overflow: hidden means we end up clipping items that get dragged below the container (which is also the reason for changing margin-bottom to padding-bottom)

Cwiiis:
It doesn't look like it was this patch, but at some point, this.canceled has started to be ignored? That'd explain why multiple touches can end up in an inconsistent state. There's another inconsistent state that can happen via tapping that I'm fixing too.

shared/elements/gaia_grid/js/grid_view.js
Cwiiis: We explicitly size the element, getBoundingClientRect would just return whatever height we last set (we could unset and re-set it, but then it'd just return the min-height + padding, which isn't what we want). This will reliably return the style's padding-bottom... I don't really like it, but short of introducing an explicit bottom-padding function to GaiaGrid, I'm not sure the best way of going about this :(

try-server-hook:
[Cwiiis Chris Lord (Cwiiis)](https://github.com/Cwiiis) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=fa6b3db56505)

try-server-hook:
[KevinGrandon Kevin Grandon (KevinGrandon)](https://github.com/KevinGrandon) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=2c3c45568db2)

try-server-hook:
[Cwiiis Chris Lord (Cwiiis)](https://github.com/Cwiiis) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=c5dd6f50d147)

try-server-hook:
[Cwiiis Chris Lord (Cwiiis)](https://github.com/Cwiiis) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=fea95f17fe3a)

apps/sharedtest/test/unit/elements/gaia_grid/grid_dragdrop_test.js
```diff
@@ -100,6 +100,7 @@ suite('GaiaGrid > DragDrop', function() {
       stopImmediatePropagation: function() {},
       preventDefault: function() {}
     });
+    this.sinon.clock.tick(20);
```
KevinGrandon: nit: the ```20``` makes me squirm a bit as I think it would be fairly easy to forget where this is coming from in the future. I might opt to export that value on GridDragDrop, then access that in the test. I think that would be more clear, but if you just want to add a comment or something that might work as well.

try-server-hook:
[Cwiiis Chris Lord (Cwiiis)](https://github.com/Cwiiis) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=e3ba9674a975)

PR #27274, opened 2015-01-09
Bug 1119738 - [Contacts] Pick vcard Activity should not return Facebook ...

jmcanterafonseca:
...data

### DIFF
--- a/apps/communications/contacts/js/activities.js
+++ b/apps/communications/contacts/js/activities.js
@@ -1,5 +1,5 @@
 /* globals ConfirmDialog, Contacts, LazyLoader, utils, ActionMenu,
-   VCardReader, ContactToVcardBlob, VcardFilename */
+   VCardReader, ContactToVcardBlob, VcardFilename, fb */
 /* exported ActivityHandler */
 
 'use strict';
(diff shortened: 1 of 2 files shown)


try-server-hook:
[jmcanterafonseca Jose M. Cantera (jmcanterafonseca)](https://github.com/jmcanterafonseca) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=28c6ad742a5c)

arcturus:
In general, the code looks pretty good, but was wondering if it's necessary that we add another facebook dependencies to the activities file.
Can we move this dependency to a class that already contains the FB dependency, and make this 'selection' a bit more agnostic.

I mean, something like let the list, or the list model to decide if you can select an element. And if not giving a 'reason'. So later on we could reuse this in different cases to allow or disallow (better said) the selection of some elements.

apps/communications/contacts/js/activities.js
```diff
@@ -233,6 +240,10 @@ var ActivityHandler = {
     }
 
     if (this.activityDataType.indexOf('text/vcard') !== -1) {
+      if (fb.isFbContact(theContact)) {
+        Contacts.confirmDialog(null, 'facebook-export-forbidden', dismiss);
```
arcturus: Was trying it out on the phone, and sounds a bit weird to me:
'The contact cannot be shared due to Facebook constraint.'

Perhaps we need a better wording, not a native speaker, but saying that we have a 'constraint', and not saying which one sounds a bit weird. I would rather say something like 'Facebook contacts cannot be exported'

jmcanterafonseca: closing this PR as we are going to implement it as suggested by Francisco. ongoing work is being done at PR #27419

PR #27285, opened 2015-01-09
1115796 add langpacks to l10n

zbraniecki:
https://bugzilla.mozilla.org/show_bug.cgi?id=1115796


try-server-hook:
[zbraniecki Zibi Braniecki (zbraniecki)](https://github.com/zbraniecki) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=96ec12e7d547)

shared/js/l10n.js
```diff
@@ -1172,10 +1175,22 @@
 
       switch (type) {
         case 'json':
-          io.loadJSON(path, onJSONLoaded, sync);
+          if (source === 'bundle') {
```
stasm: How about `app` instead of `bundle`?  At first, I thought this was referring to _localization bundles_, i.e. langpacks.

shared/js/l10n.js
```diff
@@ -1374,11 +1390,14 @@
   Context.prototype.registerLocales = function(defLocale, available) {
     /* jshint boss:true */
     this.availableLocales = [this.defaultLocale = defLocale];
+    this.localeSources[defLocale] =
+      (available && (defLocale in available)) ? available[defLocale] : 'bundle';
 
     if (available) {
-      for (var i = 0, loc; loc = available[i]; i++) {
+      for (var loc in available) {
```
stasm: I recall at least one case when `Object.prototype` was extended in the Cost Control app, which at the time broke our code like this.  I think it's safer to use the `var i = 0` idiom for looping here, even if I'm not a fan of it at all.

shared/js/l10n.js
```diff
@@ -1492,6 +1511,7 @@
         return getDirection(navigator.mozL10n.ctx.supportedLocales[0]);
       }
     },
+    gaiaVersion: '2.2',
```
stasm: In https://bugzilla.mozilla.org/show_bug.cgi?id=1115797 and https://bugzilla.mozilla.org/show_bug.cgi?id=1115798 I was able to work around having to hardcode the version of Gaia by making use of the `deviceinfo.os` setting. That worked well because in both cases we're in the Settings app which has the right privileges to access settings (as the name would imply :)).

I'm not sure how to work around this here, though.  I don't think we'd want to add a dependency on mozSetting in l10n.js (in fact, we worked hard to remove it in the past).

I currently see two solution:

1. On buildtime, put the value of the `deviceinfo.os` setting in a `meta` element in each l10n-enabled HTML file.

2. When `getAdditionalLanguages` returns, save the most recent version number for each locale and use it for language negotation _and_ resource fetching.  This is based on the assumption that a user running 2.2 will be only taken to the Marketplace listing of langpacks for 2.2, so the versions will match initially.  When the user updates Gaia to 2.3 however, they will also need to update the langpack, or the translations will be outdated.  Maybe there's a way to push new langpacks for 2.3 on system update?  In any case, this approach avoids the problem of versioning for now.

@zbraniecki, what are your thoughts on this?

shared/js/l10n.js
```diff
@@ -1607,20 +1626,32 @@
...
+    if (navigator.mozApps && navigator.mozApps.getAdditionalLanguages) {
+      navigator.mozApps.getAdditionalLanguages().then(
+        registerLocales.bind(this));
+    } else {
+      registerLocales.call(this);
     }
     return initLocale.call(this);
   }
 
+  function registerLocales(extraLangs) {
```
stasm: Maybe make `registerLocales` take `availableLanguages` and `defaultLanguage` as arguments?

shared/js/l10n.js
```diff
@@ -1607,20 +1626,32 @@
...
+    var languages = extendLanguagesList(
+      meta.availableLanguages,
+      extraLangs || {});
+    this.ctx.registerLocales(meta.defaultLanguage, languages);
+
+    meta = {};
```
stasm: Do you need to clear it here?

build/l10n.js
```diff
@@ -70,6 +70,8 @@
     if (!containsFetchableLocale) {
       document.documentElement.dataset.noCompleteBug = true;
     }
+
+    L10n.registerLocales.call(this);
```
stasm: On buildtime, can we simply call the regular `this.ctx.registerLocales` here?

shared/js/l10n.js
stasm: Ah, okay, I get it now. `available` is no longer an array;  it's an object with lang codes as keys and `bundle` or `extra` as values. I'm not sure if I like this change.  Maybe we should have two methods: `registerLocales` for local languages and `registerAdditionalLanguages` for langpacks?

shared/js/l10n.js
stasm: One more thought about the second point above:  I think it gives us time to work out a good solution for versioning in 2.3.  Worst case scenario, we'll put `deviceinfo.os` in a `meta` in 2.3 or we'll hardcode it in l10n.js and we will disable 2.2 langpacks in 2.3 apps this way, if we decide that's what we want to do.

shared/js/l10n.js
stasm: Or maybe have `registerLocales` accept three arguments: `defaultLanguage`, `availableLanguages` and (optionally)`additionalLanguages`.

build/l10n.js
zbraniecki: I would have to access meta.defaultLanguage|availableLanguages from build/l10n.js. Wanted to avoid having to expose that.

shared/js/l10n.js
zbraniecki: Yeah, otherwise at build time we will have the same meta across documents.

shared/js/l10n.js
zbraniecki: So, I'm not against it in the current incarnation, but I'd like to make langpacks a bindings feature that library is unaware of. I'd love it to be that only bindings and IO know about different sources of data. 

Do you agree?

shared/js/l10n.js
zbraniecki: one other option suggested by @fabricedesre is to use navigator.userAgent and map Gecko37 to 2.2.

try-server-hook:
[zbraniecki Zibi Braniecki (zbraniecki)](https://github.com/zbraniecki) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=380713e6a311)

shared/js/l10n.js
stasm: Sounds like we should clear it in buildtime then?

build/l10n.js
stasm: Alternatively, you should be able to declare `meta` as local to the `initResoures` function, since its whole logic is sync now. This would also let you avoid having to clear `meta`.

shared/js/l10n.js
stasm: Sounds like a good compromise:  we'll use this information only when mozApps is available anyways, so we're already Gecko-specific.

shared/js/l10n.js
stasm: (I take it that this is a reply to my other comment about `ctx.registerLocales`.)

Yes, I agree.  I recall a conversation we had in November about the core library not having any IO methods and instead using IO methods provided by the platform.  I think this would require deeper changes in how we define our IO module right now (e.g. each IO request would need to know the language of the resource).  It also seems that in your current patch you're making changes in the Locale and Context classes to differentiate between the `app` and `extra` sources.  Would you like to move this into the Bindings code?

build/l10n.js
zbraniecki: love it!

shared/js/l10n.js
zbraniecki: no. 2.0 branch is fairly Gaia specific anyways. But I'd like to make it a goal for 3.0.

try-server-hook:
[zbraniecki Zibi Braniecki (zbraniecki)](https://github.com/zbraniecki) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=aa505092bb47)

try-server-hook:
[zbraniecki Zibi Braniecki (zbraniecki)](https://github.com/zbraniecki) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=0e544cb3e75d)

try-server-hook:
[zbraniecki Zibi Braniecki (zbraniecki)](https://github.com/zbraniecki) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=a611f095144a)

shared/js/l10n.js
```diff
@@ -1454,6 +1511,25 @@
     attributeFilter: ['data-l10n-id', 'data-l10n-args']
   };
 
+  var gaiaGeckoVersionMap = {
+    '37.0': '2.2'
+  };
+
+  function setGaiaVersion() {
```
stasm: `getGaiaVersion`?

stasm: Ah, no! nvm, I get it :)

shared/js/l10n.js
```diff
@@ -1454,6 +1511,25 @@
     attributeFilter: ['data-l10n-id', 'data-l10n-args']
   };
 
+  var gaiaGeckoVersionMap = {
```
stasm: Rename to `Gecko2GaiaVersions`?

shared/js/l10n.js
stasm: Actually, I still think `getGaiaVersion` would be more appropriate here.

shared/js/l10n.js
```diff
@@ -1607,27 +1684,32 @@
...
-      // if there was no availableLanguages meta,
-      // register the default locale only
-      this.ctx.registerLocales(this.ctx.defaultLocale);
+    if (navigator.mozApps && navigator.mozApps.getAdditionalLanguages) {
+      navigator.mozApps.getAdditionalLanguages().then(
+        registerLocales.bind(this, meta));
+    } else {
+      registerLocales.call(this, meta);
     }
-    return initLocale.call(this);
```
stasm: Is there any benefit of removing this line from here?  Could we bring it back and remove the new `registerLocales` function instead?

shared/js/l10n.js
```diff
@@ -1607,27 +1684,32 @@
...
+  }
+
+  function registerLocales(meta, extraLangs) {
```
stasm: As I said above, maybe remove this helper function?  It looks like it's only doing two things, and `initLocale` can be safely called in `initResources`.

shared/js/l10n.js
```diff
@@ -1607,27 +1684,32 @@
...
     });
+    return langs;
   }
 
-  function onMetaInjected(node) {
-    if (this.ctx.availableLocales.length) {
-      return;
-    }
-
+  function onMetaInjected(node, meta) {
```
stasm: I'd consider inverting the order of arguments here;  the least variable argument should come first.

shared/js/l10n.js
```diff
@@ -1371,17 +1387,59 @@
...
+      }
+    }
+    return null;
+  }
 
-    if (available) {
-      for (var i = 0, loc; loc = available[i]; i++) {
-        if (this.availableLocales.indexOf(loc) === -1) {
-          this.availableLocales.push(loc);
+  function extendLanguagesList(availableLangs, extraLangs) {
```
stasm: Could you add a comment with examples of `availableLangs` and `extraLangs` and explain why they're of different types, please?

shared/js/l10n.js
```diff
@@ -1371,17 +1387,59 @@
...
+    var loc, lp;
+    var langs = {};
+
+    if (availableLangs) {
+      for (loc in availableLangs) {
```
stasm: IIUC, `availableLangs` is an array;  maybe use a regular for loop here?

stasm: OK, I now see `splitAvailableLanguagesString`;  `availableLangs` is an object with lang codes and versions.  This changes the Context API.  If we do this, I think we should release a new minor version: 2.1.x or even 2.2.x (to match the Gaia version). I wonder if maybe instead we should either:

 - make `availableLangs` polymorphic: an array for legacy code and an object with version for new code, or
 - accept it only as an array and map `splitAvailableLanguagesString` over its elements;  this would make it inconsistent with the type signature of `extraLangs`, so I'm not sure this is a good idea at all.

shared/js/l10n.js
```diff
@@ -1172,10 +1175,22 @@
...
+            navigator.mozApps.getLocalizationResource(
+              this.id, gaiaVersion, path, 'json')
+                .then(onJSONLoaded.bind(this, null), onL10nLoaded);
+          } else {
+            io.loadJSON(path, onJSONLoaded, sync);
+          }
           break;
         case 'properties':
-          io.load(path, onPropLoaded, sync);
+          if (source === 'extra') {
```
stasm: There's a little bit of code duplication between `case 'json'` and `case 'properties'`, with the `source` check and calling two different IO methods.  My suggestion would be to create an object like so:

```
var bindingsIO = {
  extra: function(id, ver, path, type, callback, errback) {
    navigator.mozApps.getLocalizationResource(id, ver, type).then (callback, errback));
  },
  app: function(id, ver, path, type, callback, errback, sync) {
    // check type here
    io.loadJSON(path, callback, sync);
  },
};
```
…and then use it like so:

```
case 'json':
  bindingsIO[source](this.id, gaiaVersion, path, type, onJSONLoaded.bind(this, null), onL10nLoaded);
  break;
```

try-server-hook:
[zbraniecki Zibi Braniecki (zbraniecki)](https://github.com/zbraniecki) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=dbb2b633dcc5)

shared/js/l10n.js
zbraniecki: That's really a consistency discussion. I want to keep all onElementInjected in bindings have similar signature, and from that perspective meta is an additional extra param. I prefer to keep it that way.

shared/js/l10n.js
zbraniecki: I like it.

I even started thinking about maybe moving this to io? This way IO would start being part of the bindings where it really belongs and we'd have langpacks be an artifact of the environment with library intact.
I'm ok with leaving this for 3.0 because that would require io to manage localeSources and know which language it is trying to load, which right now belongs to Locale.

What do you think?

try-server-hook:
[zbraniecki Zibi Braniecki (zbraniecki)](https://github.com/zbraniecki) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=33e184f27d0e)

shared/js/l10n.js
zbraniecki: well, no. Because in case of LangpackAPI use we are becoming asynchronous so initLocale must wait for registerLocales which waits to getAdditionalLanguages.

Removing the new registerLocales will mean that all its code (with the stuff for merging available and extra langs) will be moved to initResources growing this function a lot. Do you want it?

try-server-hook:
[zbraniecki Zibi Braniecki (zbraniecki)](https://github.com/zbraniecki) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=4a0fb07c261b)

shared/js/l10n.js
```diff
@@ -1161,19 +1178,27 @@
...
       var resLink = decodeURI(ctx.resLinks[i]);
       var path = resLink.replace('{locale}', idToFetch);
       var type = path.substr(path.lastIndexOf('.') + 1);
 
+      var cb = null;
+      var ioType = 'text';
       switch (type) {
         case 'json':
-          io.loadJSON(path, onJSONLoaded, sync);
+          ioType = 'json';
```
stasm: With this code, you're assigning to `ioType` twice in every loop pass if the type if `json`.  Can we just pass `type` here and if needed change `properties` to `text` in `bindingsIO.extra`?

stasm: I also think it's okay to simply call `bindingsIO[source]` under each `case`, with different arguments.

shared/js/l10n.js
stasm: Ah, you're right, I didn't consider the fact that `getAdditionalLanguages` was async.  Now that you move most of the `available`/`extra` logic into the helper `registerLocales`, it makes a lot of sense to keep it around :)  Let's not make `initResources` any bigger.

Would it make sense to split this into smaller pieces maybe?  Like so:

```
if (navigator.mozApps && navigator.mozApps.getAdditionalLanguages) {
  navigator.mozApps.getAdditionalLanguages().then(
    extendLocales.bind(null, meta)).then(
      Function.prototype.apply.bind(this.ctx.registerLocales, this.ctx)).then(
        initLocale.bind(this));
} else {
  this.ctx.registerLocales(meta.defaultLocales, Object.keys(meta.availableLocales));
  initLocale.call(this);
}
```

…where `extendLocales` is your `registerLocales` function from line 1674 without its two last lines, returning `[defaultLanguage, availableLanguages]`.

try-server-hook:
[zbraniecki Zibi Braniecki (zbraniecki)](https://github.com/zbraniecki) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=cf2e3eda1f33)

try-server-hook:
[zbraniecki Zibi Braniecki (zbraniecki)](https://github.com/zbraniecki) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=5f685a149734)

shared/js/l10n.js
zbraniecki: What's the goal of that? We're getting a less readable code, but I'm not sure what are we gaining.

stasm: It separates `extendLocales` from `ctx.registerLocales` from `initLocale`.  It makes `extendLocales` pure which makes it easier to understand and test.  It doesn't conflate registration with initialization, which are two logically separate steps.

shared/js/l10n.js
```diff
@@ -1161,19 +1181,25 @@
...
     var idToFetch = this.isPseudo ? ctx.defaultLocale : this.id;
+    var source = navigator.mozL10n._config.localeSources[this.id] || 'app';
+    var gaiaVersion = navigator.mozL10n._config.gaiaVersion;
+
     for (var i = 0; i < ctx.resLinks.length; i++) {
       var resLink = decodeURI(ctx.resLinks[i]);
       var path = resLink.replace('{locale}', idToFetch);
       var type = path.substr(path.lastIndexOf('.') + 1);
 
+      var cb = null;
```
stasm: nit: just leave this undefined as `var cb;`?

shared/js/l10n.js
stasm: OK, I think we might disagree here, but let's go with your argument of consistency.

shared/js/l10n.js
zbraniecki: Unfortunately that pushes us way beyond firstPaint because of the amount of async your code causes (4 instead of 1).

Even if that was not the case, I believe that your code would still be slower because of the amount of promises and since it is a hot-path in runtime, I'd be against it.

Even if that was not the case, your concept of what is easy to understand somehow ignores pushing for Function.prototype.apply.bind monsters which I'd like to avoid at all cost.

try-server-hook:
[zbraniecki Zibi Braniecki (zbraniecki)](https://github.com/zbraniecki) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=3226e74a5c63)

try-server-hook:
[zbraniecki Zibi Braniecki (zbraniecki)](https://github.com/zbraniecki) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=78f99f007d2d)

try-server-hook:
[zbraniecki Zibi Braniecki (zbraniecki)](https://github.com/zbraniecki) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=8d554efc5be8)

PR #27286, opened 2015-01-09
Bug 1109422 - [AUM] Updates to support monthly active users

marshall:
Updates made to support monthly active users calculation for the Metrics team:
https://bugzilla.mozilla.org/show_bug.cgi?id=1109422

### DIFF
--- a/apps/ftu/js/navigation.js
+++ b/apps/ftu/js/navigation.js
@@ -241,6 +241,18 @@ var Navigation = {
         var improve = document.getElementById('browser_os_improve');
         navigator.mozL10n.setAttributes(improve, 'helpImprove2',
           getLocalizedLink('helpImprove'));
+
+        // Initialize the share checkbox according to the preset value
+        // of debug.performance_data.shared
+        var sharePerformance = document.getElementById('share-performance');
+        var settingName = sharePerformance.name;
+        var settings = navigator.mozSettings;
+        var req = settings && settings.createLock().get(settingName);
+        if (req) {
+          req.onsuccess = function() {
+            sharePerformance.checked = req.result[settingName] || false;
+          };
+        }
         break;
       case '#browser_privacy':
         UIManager.mainTitle.setAttribute('data-l10n-id', 'aboutBrowser');
--- a/apps/system/js/app_usage_metrics.js
+++ b/apps/system/js/app_usage_metrics.js
@@ -39,7 +39,8 @@
  *  normal app termination from abnormal and I can't figure out any way
  *  to tell when an app has crashed.
  */
-/* global asyncStorage, SettingsListener, performance */
+/* global asyncStorage, SettingsListener, performance, SIMSlotManager,
+          MobileOperator */
 (function(exports) {
   'use strict';
 
@@ -49,7 +50,7 @@
 
   // This is the asyncStorage key we use to persist our app usage data so
   // that it survives across device restarts.
-  const PERSISTENCE_KEY = 'metrics.app_usage.data';
+  const PERSISTENCE_KEY = 'metrics.app_usage.data.v2';
 
   // This is the asyncStorage key we use to persist our device ID
   const DEVICE_ID_KEY = 'metrics.app_usage.deviceID';
@@ -88,7 +89,6 @@
     ATTENTIONOPENED,
     ATTENTIONCLOSED
   ];
-
 
   // This AppUsageMetrics() constructor is the value we export from
   // this module. This constructor does no initialization itself: that
@@ -135,7 +135,7 @@
 
   // How often do we try to send the reports
   // Can be overridden with metrics.appusage.reportInterval setting.
-  AUM.REPORT_INTERVAL = 24 * 60 * 60 * 1000;  // 1 day
+  AUM.REPORT_INTERVAL = 14 * 24 * 60 * 60 * 1000;  // 2 weeks
 
   // If the telemetry server does not respond within this amount of time
   // just give up and try again later.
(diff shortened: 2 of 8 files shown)


try-server-hook:
[marshall Marshall Culpepper (marshall)](https://github.com/marshall) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=bf8c3d27ce63)

apps/system/js/app_usage_metrics.js
```diff
@@ -558,18 +538,18 @@
...
         // If so, then send it if the retry interval has elapsed
-        if (absoluteTime - this.lastFailedTransmission > AUM.RETRY_INTERVAL) {
+        if (relativeTime - this.lastFailedTransmission > AUM.RETRY_INTERVAL) {
           this.transmit();
         }
       }
       // Otherwise, if we have not failed to transmit, then send it if the
       // reporting interval has elapsed.
-      else if (absoluteTime - this.metrics.startTime() > AUM.REPORT_INTERVAL) {
+      else if (totalElapsed > AUM.REPORT_INTERVAL) {
```
davidflanagan: This will only send the report after two weeks of the user having the phone turned on. If a user turns off their phone every night and only has it on for 12 hours a day, then their reports will only be sent every 4 weeks.

I think you are going to have to use absolute time for at least some of this module.

apps/system/js/app_usage_metrics.js
```diff
@@ -613,19 +592,77 @@
...
+        simInfo.network.mnc = voiceNetwork.mnc;
+        simInfo.network.mcc = voiceNetwork.mcc;
+      }
+
+      if (iccInfo) {
+        simInfo.icc = {
+          mnc: iccInfo.mnc,
+          mcc: iccInfo.mcc,
+          spn: iccInfo.spn
+        };
```
davidflanagan: Are there privacy implications here with this data?  I don't know much about sims... Is mnc like the US area code?  

I gather that mcc and spn pretty much just identify the carrier, and that seems okay. I'm not sure about mnc, though.

apps/system/js/app_usage_metrics.js
```diff
@@ -761,18 +814,21 @@
...
-      oldusage.uninstalls += newusage.uninstalls;
-
-      for (var url in newusage.activities) {
-        var newcount = newusage.activities[url];
-        var oldcount = oldusage.activities[url] || 0;
-        oldusage.activities[url] = oldcount + newcount;
+      var newdays = newbatch.data.apps[app];
+      for (var day in newdays) {
+        var newusage = newdays[day];
+        var oldusage = this.getAppUsage(app, day);
```
davidflanagan: I think day is a timestamp here, like '20150112', right? But getAppUsage() expects a Date object as the second argument, not a timestamp.  If I'm right it means that there are not tests for this case, or that you haven't run them yet.

apps/system/js/app_usage_metrics.js
```diff
@@ -613,19 +592,77 @@
...
-      data.deviceinfo = deviceinfo;
+    AUM.getSettings(deviceInfoQuery, function(deviceInfo) {
+      data.deviceinfo = deviceInfo;
+      data.simInfo = getSIMInfo();
+
       // Now transmit the data
       send(data);
     });
 
+    function getSIMInfo() {
```
albertopq: Wouldn't make sense to move this method to the SIMSlotManager? Then we can add some unit tests to it :)

marshall: I like the idea, but the format of the return object here is somewhat specific (meant to be sent to a server as part of a JSON payload). Ideally though we could share the code between this and ftu_ping.js ..

try-server-hook:
[marshall Marshall Culpepper (marshall)](https://github.com/marshall) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=6a3e91dd8412)

apps/system/js/app_usage_metrics.js
marshall: Not that I know of, we're also providing this data in the FTU ping. MCC actually identifies the country ('mobile country code') and MNC identifies the network

apps/system/js/app_usage_metrics.js
albertopq: Can't we return something generic (most of the info is taken from navigator.mozIccManager.getIccById(conn.iccId);
), that this module will format into the specifics we need to send to the server? It looks weird to me a private function here, that is 50 lines long and is that related to SIM.  One more reason to be abstracted, if we need to share it with another module. Just an opinion, I don't really know much about this module :)

marshall: Yeah, that's a fair point. There are basically two buckets of data here: the ICC network (MCC/MNC/SPN registered on the SIM card), and the registered voice network (what the radio is physically connected to, could be roaming for example). Both of these data buckets could be provided as high level getters in SIMSlot, then formatted in this service..

apps/system/js/app_usage_metrics.js
```diff
@@ -613,19 +613,76 @@
...
+      'deviceinfo.software': 'unknown',
+      'deviceinfo.update_channel': 'unknown'
     };
 
     // Query the settings db to get some more device-specific information
-    AUM.getSettings(deviceInfoQuery, function(deviceinfo) {
-      data.deviceinfo = deviceinfo;
+    AUM.getSettings(deviceInfoQuery, function(deviceInfo) {
+      data.deviceinfo = deviceInfo;
+      data.simInfo = getSIMInfo();
```
davidflanagan: Your data object here has one property with lowercase info and one with capitalized Info.  Did you mean to type data.deviceInfo here?

apps/system/js/app_usage_metrics.js
```diff
@@ -613,19 +613,76 @@
...
+        icc: null
+      };
+
+      if (SIMSlotManager.noSIMCardConnectedToNetwork()) {
+        // No connected SIMs
+        return simInfo;
+      }
+
+      var slots = SIMSlotManager.getSlots().filter(function(slot) {
+        return !slot.isAbsent() && !slot.isLocked();
```
davidflanagan: I don't know anything about SIMSlotManager, but have you tested this on a dual sim device with only one card in it?  Do you get an array of two elements where one of the elements is null?  How about if the one card is in slot #2? Can that happen, and if so, do you need to put |slot &&| before the !slot.isAbsent()?

apps/system/test/unit/app_usage_metrics_test.js
```diff
@@ -49,11 +55,15 @@ suite('AppUsageMetrics:', function() {
       set: setNavigatorOnLine
     });
 
+    realPerformanceNow = window.performance.now;
+    window.performance.now = function() { return Date.now(); };
```
davidflanagan: I couldn't remember why this was necessary. Might be worth adding a comment that this is here because the test uses a fake clock.  And the fake performance.now could be made more realistic by setting taking Date.now() in the suiteSetup and subtracting it from the returned value. Doesn't actually matter, though.

apps/system/js/app_usage_metrics.js
davidflanagan: I see that this has been fixed in getAppUsage()...

apps/system/js/app_usage_metrics.js
davidflanagan: I've looked it up and agree that this is not privacy infringing.

apps/system/js/app_usage_metrics.js
```diff
@@ -671,20 +728,35 @@
...
         invocations: 0,
         installs: 0,
         uninstalls: 0,
         activities: {}
       };
-      this.data.apps[app] = usage;
     }
-    return usage;
+    return dayUsage;
+  };
```
davidflanagan: My intuition about this is that this data would be more usefully organized first by date and then by app. But you have the app url as the top-level property and then have dates below that.  I don't think it matters much, and if the server side is already being coded to expect this organization, then I'd leave it alone. But if not, at least consider changing data.apps[app][day] to data.days[day][app].

marshall: I had done that originally, but I realized that this way of organizing data decreases the amount of duplicate data in the payload, so instead of seeing..

```json
{"apps": {
  "20150202":  {
    "A_REALLY_LONG_APPURL": {}
  },
  "20150203": {
    "A_REALLY_LONG_APP_URL": {}
  }
}
```

we'd see 
```json
{"apps": {
  "A_REALLY_LONG_APPURL": {
    "20150202": {},
    "20150203": {}
  }
}
```

autolander:

try-server-hook:
[marshall Marshall Culpepper (marshall)](https://github.com/marshall) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=517b2225e5e4)

apps/ftu/js/navigation.js
```diff
@@ -241,6 +241,18 @@ var Navigation = {
...
         navigator.mozL10n.setAttributes(improve, 'helpImprove2',
           getLocalizedLink('helpImprove'));
+
+        // Initialize the share checkbox according to the preset value
+        // of debug.performance_data.shared
+        var sharePerformance = document.getElementById('share-performance');
+        var setting = sharePerformance.name;
+        var settings = navigator.mozSettings;
+        var req = settings && settings.createLock().get(setting);
+        if (req) {
```
fcampo: not really needed imo, if req is null we won't get into onsuccess anyway.

apps/ftu/js/navigation.js
```diff
@@ -241,6 +241,18 @@ var Navigation = {
         var improve = document.getElementById('browser_os_improve');
         navigator.mozL10n.setAttributes(improve, 'helpImprove2',
           getLocalizedLink('helpImprove'));
+
+        // Initialize the share checkbox according to the preset value
+        // of debug.performance_data.shared
+        var sharePerformance = document.getElementById('share-performance');
+        var setting = sharePerformance.name;
```
fcampo: rather name it `settingName` instead please.

apps/ftu/js/navigation.js
```diff
@@ -241,6 +241,18 @@ var Navigation = {
...
+        var settings = navigator.mozSettings;
```
fcampo: we can make the `settings` var in line 71 [https://github.com/marshall/gaia/blob/bug1109422_mauChanges/apps/ftu/js/navigation.js#L71] global (out of the object), so we can avoid create and initialize again.

autolander:

try-server-hook:
[marshall Marshall Culpepper (marshall)](https://github.com/marshall) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=c3f0788a5c47)

try-server-hook:
[marshall Marshall Culpepper (marshall)](https://github.com/marshall) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=ca987dc6dfca)

try-server-hook:
[marshall Marshall Culpepper (marshall)](https://github.com/marshall) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=9a936f53f386)

PR #27319, opened 2015-01-12
Bug 1120381 - [Stingray][AppDeck] Scroll to center button on vertical mi...

rexboy7:
...ddle when navigating gridView

### DIFF
--- a/tv_apps/app-deck/js/app_deck.js
+++ b/tv_apps/app-deck/js/app_deck.js
@@ -139,10 +139,21 @@
           this._spatialNavigator.focus(elem);
         }
         elem.focus();
+        // Locate focused button to vertical middle of grid view. This way
+        // buttons one row above would always slightly be on bottom of the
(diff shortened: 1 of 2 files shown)


tv_apps/app-deck/js/app_deck.js
```diff
@@ -139,12 +139,19 @@
           this._spatialNavigator.focus(elem);
         }
         elem.focus();
+        this._scrollTo(elem);
```
dwi2: I'll like to have some comment here. Because code reader might get confused when they read through here. What you said on https://bugzilla.mozilla.org/show_bug.cgi?id=1120381#c1 is suitable here I think.

try-server-hook:
[rexboy7 Rex KM Lee (rexboy7)](https://github.com/rexboy7) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=3c58f3e35156)

PR #27330, opened 2015-01-12
Bug 1120612 - Ringtones app doesn't release audio channel when launching...

jimporter:
... child activities

### DIFF
--- a/apps/ringtones/js/tone_player.js
+++ b/apps/ringtones/js/tone_player.js
@@ -38,8 +38,7 @@ function TonePlayer() {
 
   window.addEventListener('visibilitychange', function() {
     if (document.hidden) {
-      this.stop();
-      this._setExclusiveMode(false);
+      this.stop(true);
     }
   }.bind(this));
 }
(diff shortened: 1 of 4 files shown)


try-server-hook:
[jimporter Jim Porter (jimporter)](https://github.com/jimporter) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=0082b59494fa)

try-server-hook:
[jimporter Jim Porter (jimporter)](https://github.com/jimporter) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=f27c7ee2ef82)

apps/ringtones/js/tone_player.js
```diff
@@ -93,11 +92,18 @@ TonePlayer.prototype = {
...
+   * if we had validated the tone (we need that information when handling a pick
+   * activity).
+   *
+   * @param {Boolean} close true if we should also deactivate "exclusive mode"
+   *   (read: close the audio channel associated with the tones).
    */
-  stop: function() {
+  stop: function(close) {
     this._player.pause();
+    if (close) {
```
davidflanagan: If I'm understanding your code correctly, I think you already call setExclusiveMode(true) if this tone is played again. So IIUC, you could just unconditionally call setExclusiveMode(false) here, and then you wouldn't need to add the argument.  And that would simplify your patch a lot.

jimporter: If I do that, it closes the audio channel and immediately lets background audio start playing again. My original patch did as you say, but I talked to Tif and she wants the existing behavior (which includes pausing the preview - but *not* letting background audio resume - when you tap the "..." in the ringtones manager).

PR #27350, opened 2015-01-13
 Bug 1116780 - [Messages] We should not focus the composer after clicking a notification r=julien

azasypkin:


apps/sms/js/activity_handler.js
```diff
@@ -21,8 +21,6 @@ const ActivityDataType = {
 };
 
 var ActivityHandler = {
-  isLocked: false,
```
azasypkin: Couldn't look at it anymore :)

try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=6a398ee3767c)

apps/sms/js/activity_handler.js
```diff
@@ -229,25 +229,24 @@ var ActivityHandler = {
     //Validate if message still exists before opening message thread
     //See issue https://bugzilla.mozilla.org/show_bug.cgi?id=837029
     if (!message) {
-      return;
+      return Promise.reject(new Error('Message is not defined!'));
     }
 
-    MessageManager.getMessage(message.id).then((message) => {
+    return MessageManager.getMessage(message.id).then((message) => {
```
azasypkin: I don't like that I've added too much ```return``` statements here, but I feel that's more correct at the same time

apps/sms/js/activity_handler.js
```diff
@@ -294,70 +293,56 @@ var ActivityHandler = {
...
     if (Compose.isEmpty()) {
-      this.launchComposer(activity);
-    } else {
-      // ask user how should we do
-      ActivityHandler.displayUnsentConfirmation(activity);
+      return this.launchComposer(activity);
     }
+
+    // ask user how should we do
+    ActivityHandler.displayUnsentConfirmation(activity);
```
azasypkin: If Jenny and Fang agree to change confirm dialog styling, I'd probably get rid of this method in favor of plain ```Utils.confirm``` and then I can return its promise-based result too :)

apps/sms/js/activity_handler.js
```diff
@@ -294,70 +293,56 @@ var ActivityHandler = {
...
+   */
+  toView: function ah_toView(message) {
     if (!message) {
-      return;
+      return Promise.reject(new Error('Message is not defined!'));
     }
 
-    this.isLocked = false;
-    var threadId = message.threadId ? message.threadId : null;
-    var body = message.body || '';
```
azasypkin: it's handled in ```triggerNewMessage``` anyway

apps/sms/js/activity_handler.js
julienw: yeah, good move

apps/sms/js/activity_handler.js
```diff
@@ -229,25 +229,24 @@ var ActivityHandler = {
...
       }
 
       if (Compose.isEmpty()) {
-        ActivityHandler.toView(message);
-        return;
+        return ActivityHandler.toView(message);
       }
 
-      Utils.confirm('discard-new-message').then(() => {
+      return Utils.confirm('discard-new-message').then(() => {
```
julienw: I'm not sure of this "return", or maybe you should catch the possible rejection from "confirm" ?

apps/sms/js/activity_handler.js
julienw: that said, given it's a blocker, I'm not sure :/

apps/sms/js/activity_handler.js
```diff
@@ -125,21 +115,31 @@ var ActivityHandler = {
...
-    // case no contact and no thread id: gobble the error
-    .catch(() => {})
-    // finally call toView whatever contact and threadId we have.
-    .then( () => this.toView(viewInfo) );
+    var viewInfoPromise;
+    if (viewInfo.number) {
+      // try to get a thread from number
+      // if no thread, promise is rejected and we try to find a contact
+      viewInfoPromise = MessageManager.findThreadFromNumber(viewInfo.number).
+        then(function onResolve(threadId) {
```
julienw: nit: so you prefer to put the "then" on the next line, why not; but I think it's not really readable that the 2 functions are not aligned

apps/sms/js/activity_handler.js
```diff
@@ -125,21 +115,31 @@ var ActivityHandler = {
...
+      )
+      // case no contact and no thread id: gobble the error
+      .catch(() => {});
+    }
+
+    return (viewInfoPromise || Promise.resolve()).
+      then(() => this.toView(viewInfo)).
+      then(() => {
+        // Focus on Compose only in case we navigated to Thread panel, as for
+        // Compose all focus-related stuff is handled by ThreadUI itself
```
julienw: then, why can't we do the same for thread ?

I know we want to do this only when we come from an activity, so maybe we can pass another property in the "args" argument to toPanel? like "focusComposer: true" ?

apps/sms/js/activity_handler.js
julienw: why not but I don't see what you do with this returned promise?

apps/sms/js/activity_handler.js
```diff
@@ -229,25 +229,24 @@ var ActivityHandler = {
     //Validate if message still exists before opening message thread
     //See issue https://bugzilla.mozilla.org/show_bug.cgi?id=837029
     if (!message) {
-      return;
+      return Promise.reject(new Error('Message is not defined!'));
```
julienw: I think I'd simply return a Promise.resolve(). Returning a rejected promise is just like throwing an error. And I don't think it's appropriate here. However we might want to have a console.error.

apps/sms/js/activity_handler.js
```diff
@@ -281,7 +280,7 @@ var ActivityHandler = {
 
   // Launch the UI properly
   launchComposer: function ah_launchComposer(activity) {
-    Navigation.toPanel('composer', { activity: activity });
+    return Navigation.toPanel('composer', { activity: activity });
```
julienw: be careful as you really change the semantic here; before we were returning right away, now we still return right away :) but if we use the promise we'll wait until the end of the transition, which takes time. So yeah, I'm not against this, but just be careful.

apps/sms/js/activity_handler.js
```diff
@@ -294,70 +293,56 @@ var ActivityHandler = {
       };
 
     if (Compose.isEmpty()) {
-      this.launchComposer(activity);
-    } else {
-      // ask user how should we do
-      ActivityHandler.displayUnsentConfirmation(activity);
+      return this.launchComposer(activity);
```
julienw: I usually don't like having a return in the middle of a function, but the function is small here so why not.

apps/sms/js/activity_handler.js
julienw: good for me

apps/sms/js/activity_handler.js
```diff
@@ -294,70 +293,56 @@ var ActivityHandler = {
...
+   * number to pre-populate the recipients list with, "body" is an optional body
+   * to preset the compose input with, "contact" is an optional MozContact
+   * instance, "threadId" is an optional threadId corresponding to a new or
+   * existing thread.
+   * @returns {Promise}
+   */
+  toView: function ah_toView(message) {
     if (!message) {
-      return;
+      return Promise.reject(new Error('Message is not defined!'));
```
julienw: same comment as earlier

apps/sms/js/activity_handler.js
```diff
@@ -294,70 +293,56 @@ var ActivityHandler = {
...
-        return;
-      }
-
-      Navigation.toPanel('thread', { id: threadId }).then(Compose.focus);
+      deferred.resolve(!message.threadId ?
+        ActivityHandler.triggerNewMessage(
+          message.body, message.number, message.contact
+        ) :
+        Navigation.toPanel('thread', { id: message.threadId })
+      );
```
julienw: please, just use a if :)

apps/sms/js/activity_handler.js
julienw: right, weird.

apps/sms/js/activity_handler.js
azasypkin: Ah, both work for me, will return it back since it's more readable for you :)

apps/sms/js/activity_handler.js
azasypkin: Sure we can, I had this in mind too, just tell me what approach you like more :)

apps/sms/js/activity_handler.js
azasypkin: I use it to make unit tests simpler, I hate doing like this though

apps/sms/js/activity_handler.js
azasypkin: > I think I'd simply return a Promise.resolve(). Returning a rejected promise is just like throwing an error.

But if message is not defined it's definitely something not expected, isn't it? I see that we have only two places where we call this method: on sms received - if message is undefined there, then we'll get exception before we call this one, and the second one is when we handle notification click where ```message``` is always defined
```
ActivityHandler.handleMessageNotification({
  id: params.id,
  threadId: params.threadId
});
```

So maybe we can just get rid of this ```if``` - I don't really think it's useful at all?

apps/sms/js/activity_handler.js
azasypkin: Yeah, will carefully analyze it one more time

apps/sms/js/activity_handler.js
azasypkin: I didn't like it earlier, but now I like early return :) So that I can stop reading the function for the use case I was following once I see return and can be sure that the rest of function is for another use case.

apps/sms/js/activity_handler.js
azasypkin: Same as I asked for previous comment, maybe we can just get rid of it entirely? I see 4 places where ```toView``` is called and for all of them ```message``` should be defined or it just fails before :)

apps/sms/js/activity_handler.js
azasypkin: I knew you'll say that! :) Okay, will use ```if\else``` here

apps/sms/js/activity_handler.js
azasypkin: Okay, let's make everyone happy here, we have old dedicated bug for this cleanup anyway - https://github.com/mozilla-b2g/gaia/pull/27394 :)

try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=50a4e6b3376d)

apps/sms/js/activity_handler.js
azasypkin: Here is PR so that you can choose what you like more (now I like the new one more :) ) - https://github.com/mozilla-b2g/gaia/pull/27395

apps/sms/js/activity_handler.js
julienw: this if is not useful at all, you're right, let's remove it

apps/sms/js/activity_handler.js
julienw: yeah it's a case by case choice; in big functions it's not good, but big functions should not happen either. :)

apps/sms/js/activity_handler.js
julienw: yeah, you're right

I also tried to check the case of clicking on the notification for a message that has been deleted -- but this should not happen now because we remove the notification when we enter or delete a thread :) (unless we have another bug somewhere, so we can keep the check -- I mean the getMessage call in handleMessageNotification)

apps/sms/js/activity_handler.js
julienw: me too !

azasypkin:
Obsoleted in favour of https://github.com/mozilla-b2g/gaia/pull/27395

PR #27369, opened 2015-01-14
Bug 1118054 - Create test to catch regression in bug 1117981: Lockscreen doesn't show updated time after being woken up from sleep

nojunpark:


try-server-hook:
[npark-mozilla No-Jun Park (npark-mozilla)](https://github.com/npark-mozilla) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=e3ddfd8d3e4e)

try-server-hook:
[npark-mozilla No-Jun Park (npark-mozilla)](https://github.com/npark-mozilla) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=969c3efbb14c)

try-server-hook:
[npark-mozilla No-Jun Park (npark-mozilla)](https://github.com/npark-mozilla) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=4e45e90041ad)

npark-mozilla:
I commented out line 15, but i'm thinking now perhaps I should re-enable it, thoughts?

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/manifest.ini
```diff
@@ -27,3 +27,5 @@ carrier = true
 [test_lockscreen_notification.py]
 
 [test_lockscreen_wake_with_notification.py]
+
+[test_lockscreen_time_check.py]
```
JohanLorenzo: Missing new line.

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
```diff
@@ -0,0 +1,45 @@
...
+from gaiatest.apps.lockscreen.app import LockScreen
+
+
+class TestLockScreen(GaiaTestCase):
+
+    _seconds_since_epoch = 1357043430  # the time is 12:30 GMT
+
+    def setUp(self):
+        GaiaTestCase.setUp(self)
+        #self.connect_to_local_area_network()
```
JohanLorenzo: Commented code.

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def test_lockscreen_time_check(self):
+
+        # disable time/timezone automatic update
+        self.data_layer.set_setting('time.clock.automatic-update.enabled', False)
+        self.data_layer.set_setting('time.timezone.automatic-update.enabled', False)
+
+        # set time manually to 12:30
+        self.data_layer.set_time(self._seconds_since_epoch * 1000)
+        self.data_layer.set_setting('time.timezone', 'Atlantic/Reykjavik')
```
JohanLorenzo: It seems more prudent to me to change the timezone before the time because it might lead to intermittent failures. Let's say we're in PST, we change the time to 12:30PM and the change is immediately taken into account. Then we change it to UTC => It'll be 8:30PM.

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
```diff
@@ -0,0 +1,45 @@
...
+
+        self.device.lock()
+        lock_screen = LockScreen(self.marionette)
+
+        #check the time on the lockscreen and unlock.  Below check gives 10 minutes of windows
+        #on desktop b2g, it usually takes 8 minutes to get to this point
+        self.assertTrue('12:3' in lock_screen.time)
```
JohanLorenzo: On my B2G desktop, it takes 9 seconds to get there, and then the test fails. How about setting the time in Settings like a user would do?

npark-mozilla: hmm, I should try that, this might take a bit of a time, but i guess that makes it more faithfully reproducing the steps in the bug.  I am not sure how the try server tests work, because by the time it reaches line 32, lockscreen time was showing 12:38.

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
```diff
@@ -0,0 +1,45 @@
...
+        # set the time to be set automatically
+        self.data_layer.set_setting('time.clock.automatic-update.enabled', True)
+        self.data_layer.set_setting('time.timezone.automatic-update.enabled', True)
+        self.device.lock()
+
+        #wait until device is off and turn back on
+        Wait(self.marionette, timeout=20).until(
+            lambda m: self.device.is_screen_enabled == False)
+        self.device.turn_screen_on()
+        self.assertFalse('12:3' in lock_screen.time)
```
mwargers: So this would fail if the automatic time is actually 12:3?

try-server-hook:
[npark-mozilla No-Jun Park (npark-mozilla)](https://github.com/npark-mozilla) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=456dc2ceb40f)

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
npark-mozilla: it would fail if the new time is actually 12:3x.  but if this test is run on some really odd time, there is a chance that this test will fail.  I am thinking of ways to read the current time, then change it.  However, right now Bug 1122503 is blocking me from submitting this test.  (found it during the script writing process)

autolander:

try-server-hook:
[npark-mozilla No-Jun Park (npark-mozilla)](https://github.com/npark-mozilla) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=34f14e07f4cd)

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/date_and_time.py
```diff
@@ -21,3 +29,59 @@ def __init__(self, marionette):
     def select_time_format(self, time_format):
         self.marionette.find_element(*self._24h_selector_locator).tap()
         self.select(time_format)
+
+    def set_automatic_time_update(self, state):
+        """
```
JohanLorenzo: I don't think this documentation adds much value to the name of the function. Looks like we can remove it to me.

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/date_and_time.py
```diff
@@ -21,3 +29,59 @@ def __init__(self, marionette):
...
+                li.tap()
+                break
+        else:
+            raise Exception("Element '%s' could not be found in select wrapper" % region)
+        close_button.tap()
+        self.marionette.switch_to_frame(self.apps.displayed_app.frame)
+
+    @property
+    def is_autotime_enabled(self):
+        if self.marionette.find_element(*self._autotime_enabled_locator).get_attribute('data-state') == \
```
JohanLorenzo:
``` Python
return self.marionette.find_element(*self._autotime_enabled_locator).get_attribute('data-state') == 'auto'
```

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/date_and_time.py
```diff
@@ -21,3 +29,59 @@ def __init__(self, marionette):
...
+        """
+        Setes the time region
+        :param region: Must match the region shown in the list
+        :return:
+        """
+
+        self.marionette.find_element(*self._timezone_region_locator).tap()
+        self.marionette.switch_to_frame()
+        self.wait_for_condition(
+            lambda m: len(self.marionette.find_elements(By.CSS_SELECTOR, '.value-selector-container li')) > 0)
```
JohanLorenzo: I would put this selector upper in the class, with the others.

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/date_and_time.py
```diff
@@ -21,3 +29,59 @@ def __init__(self, marionette):
...
+
+        options = self.marionette.find_elements(By.CSS_SELECTOR, '.value-selector-container li')
```
JohanLorenzo: I would put this selector upper in the class, with the others.

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/date_and_time.py
```diff
@@ -21,3 +29,59 @@ def __init__(self, marionette):
...
+        close_button = self.marionette.find_element(By.CSS_SELECTOR, 'button.value-option-confirm')
```
JohanLorenzo: I would put this selector upper in the class, with the others.

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/date_and_time.py
```diff
@@ -21,3 +29,59 @@ def __init__(self, marionette):
...
+
+        # loop options until we find the match
+        for li in options:
+            if region in li.text:
+                li.tap()
+                break
+        else:
```
JohanLorenzo: Oh I didn't know about the for else in Python. That's neat! :+1:

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/date_and_time.py
```diff
@@ -21,3 +29,59 @@ def __init__(self, marionette):
...
+        """
+        toggles 'Set Time Automatically' flag
+        """
+
+        Wait(self.marionette).until(expected.element_displayed(
+            Wait(self.marionette).until(
+                expected.element_present(*self._autotime_enabled_locator))))
+
+        if not state:
+            if self.is_autotime_enabled is True:
```
JohanLorenzo: Having this check makes this function a bit too smart. The a function in the page class should not behave differently depending on the state of the UI. I would remove this if.

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/date_and_time.py
```diff
@@ -21,3 +29,59 @@ def __init__(self, marionette):
...
+                self.marionette.find_element(*self._autotime_enabled_switch_locator).tap()
+            self.wait_for_condition(lambda m: self.is_autotime_enabled is False)
+        else:
+            if self.is_autotime_enabled is False:
```
JohanLorenzo: Same as above, I would remove this if too.

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/date_and_time.py
```diff
@@ -21,3 +29,59 @@ def __init__(self, marionette):
...
+                self.marionette.find_element(*self._autotime_enabled_switch_locator).tap()
+            self.wait_for_condition(lambda m: self.is_autotime_enabled is True)
+
+    def set_region(self, region):
+        """
```
JohanLorenzo: I don't think this documentation adds much value to the name of the function. Looks like we can remove it to me.

tests/python/gaia-ui-tests/gaiatest/apps/system/regions/time_picker.py
```diff
@@ -21,48 +17,57 @@
...
+    _hour24_picker_locator = (By.CSS_SELECTOR, '.value-picker-hour24-state > div')
 
     def __init__(self, marionette):
         Base.__init__(self, marionette)
         self.marionette.switch_to_frame()
-        Wait(self.marionette).until(expected.element_displayed(*self._hour_picker_locator))
+        self.wait_for_element_displayed(*self._current_element(*self._hour_picker_locator))
 
-        # TODO: wait for the time picker to fade in Bug 1038186
+        # TODO: wait for the time piker to fade in Bug 1038186
```
JohanLorenzo: Isn't picker the right word here?

tests/python/gaia-ui-tests/gaiatest/apps/system/regions/time_picker.py
```diff
@@ -21,48 +17,57 @@
...
         time.sleep(2)
 
     def tap_done(self):
-        time_picker = self.marionette.find_element(*self._time_picker_locator)
         self.marionette.find_element(*self._done_button_locator).tap()
-        Wait(self.marionette).until(expected.element_not_displayed(time_picker))
-        # TODO: wait for the time picker to fade out Bug 1038186
+        self.wait_for_element_not_displayed(*self._time_picker_locator)
+        # TODO: wait for the time piker to fade out Bug 1038186
```
JohanLorenzo: Isn't picker the right word here?

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
```diff
@@ -0,0 +1,75 @@
...
+        Wait(self.marionette, timeout=20).until(
+            lambda m: not self.device.is_screen_enabled)
+        self.device.turn_screen_on()
+
+        # Check it reverted to the correct time.
+        # Allow 10 minutes difference max (since treeherder tests take longer)
+        self.assertTrue(self.get_time_difference(old_time, lock_screen.time) < 10)
+
+
+    def get_time_difference(self, time_a, lockscreen_time):
```
JohanLorenzo: Why don't we use [datetime.timedelta](https://docs.python.org/2/library/datetime.html#datetime.timedelta), here?

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/manifest.ini
```diff
@@ -26,3 +26,5 @@ carrier = true
...
```
JohanLorenzo: Like you said on the stand up this morning, ```skip-if = device == "desktop" ``` would be needed here as we can't change the time at will on b2gdesktop.

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
```diff
@@ -0,0 +1,75 @@
...
+        self.marionette.switch_to_frame()
+        self.device.lock()
+
+        # wait until device is off and turn back on to check that the time is changed
+        Wait(self.marionette, timeout=20).until(
+            lambda m: not self.device.is_screen_enabled)
+        self.device.turn_screen_on()
+
+        # Check it reverted to the correct time.
+        # Allow 10 minutes difference max (since treeherder tests take longer)
```
JohanLorenzo: As this test won't run on Treeherder, this comment will need to be updated.

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/date_and_time.py
npark-mozilla: But this method has to both enable and disable the switch depending on the 'state' value, and since I need to enable/disable it on the same script, it has to know whether it is already in the state that I want to put it in.  In the end, what matters is that it is in the correct state.  If I can't put it in the date_and_time.py, should I move this method out to the test script?

tests/python/gaia-ui-tests/gaiatest/apps/system/regions/time_picker.py
npark-mozilla: updated

tests/python/gaia-ui-tests/gaiatest/apps/system/regions/time_picker.py
npark-mozilla: updated

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/date_and_time.py
npark-mozilla: removed

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/date_and_time.py
npark-mozilla: same comment as above

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/date_and_time.py
npark-mozilla: moved

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/date_and_time.py
npark-mozilla: moved

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/date_and_time.py
npark-mozilla: moved

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/date_and_time.py
npark-mozilla: changed

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/date_and_time.py
npark-mozilla: removed

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/date_and_time.py
JohanLorenzo: Having this `if` on `state` looks fine to me. That's ` if self.is_autotime_enabled is True ` which is the problem to me.

That is to say, it can lead to a false positive. For instance: The tests deactivates autotime, but due to a kernel bug, autotime was already deactivated, so the test won't report an error until you try to activate autotime 10 lines later in the test file. This will make the test hard to debug or worse, if a test doesn't try to activate autotime, lead to an actual false positive.

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/manifest.ini
npark-mozilla: added

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
npark-mozilla: oh, I specifically wanted to compare against the time that is "shown" on the lockscreen vs. the time I see it in settings, and the annoying thing is, in settings, it shows in HH:MM: AM/PM format only, while the lockscreen does NOT show AM/PM but still shows in 12 Hr format.  (so you don't know whether it is 1 AM or 1 PM)  So I wrote the method to get the delta, assuming this test does not take more than 12 hours.

npark-mozilla: Plus I had to extract the numbers from the string literal anyways

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/date_and_time.py
```diff
@@ -21,3 +29,59 @@ def __init__(self, marionette):
...
+        # loop options until we find the match
+        for li in options:
+            if region in li.text:
+                li.tap()
+                break
+        else:
+            raise Exception("Element '%s' could not be found in select wrapper" % region)
+        close_button.tap()
+        self.marionette.switch_to_frame(self.apps.displayed_app.frame)
+
```
mwargers: Isn't this what self.apps.switch_to_displayed_app()) is for?
http://mxr.mozilla.org/gaia/source/tests/python/gaia-ui-tests/gaiatest/gaia_test.py#96

tests/python/gaia-ui-tests/gaiatest/apps/system/regions/time_picker.py
```diff
@@ -21,48 +17,57 @@
...
+
+    _hour_picker_locator = (By.CSS_SELECTOR, '.value-picker-hours > div')
+    _minutes_picker_locator = (By.CSS_SELECTOR, '.value-picker-minutes > div')
+    _hour24_picker_locator = (By.CSS_SELECTOR, '.value-picker-hour24-state > div')
 
     def __init__(self, marionette):
         Base.__init__(self, marionette)
         self.marionette.switch_to_frame()
-        Wait(self.marionette).until(expected.element_displayed(*self._hour_picker_locator))
+        self.wait_for_element_displayed(*self._current_element(*self._hour_picker_locator))
```
mwargers: We're supposed to use the Wait() methods, so you shouldn't change this to wait_for_element_displayed

tests/python/gaia-ui-tests/gaiatest/apps/system/regions/time_picker.py
```diff
@@ -21,48 +17,57 @@
...
-        # TODO: wait for the time picker to fade in Bug 1038186
+        # TODO: wait for the time piker to fade in Bug 1038186
         time.sleep(2)
 
     def tap_done(self):
-        time_picker = self.marionette.find_element(*self._time_picker_locator)
         self.marionette.find_element(*self._done_button_locator).tap()
-        Wait(self.marionette).until(expected.element_not_displayed(time_picker))
-        # TODO: wait for the time picker to fade out Bug 1038186
+        self.wait_for_element_not_displayed(*self._time_picker_locator)
```
mwargers: Why change this? We're supposed to use the Wait() methods, no?

tests/python/gaia-ui-tests/gaiatest/apps/system/regions/time_picker.py
```diff
@@ -21,48 +17,57 @@
...
-        minute = current.text
-        next = current.find_element(By.XPATH, 'following-sibling::*')
-        # TODO: Bug 1129907 - Unable to use precise actions to select timepicker values in Gaia
-        # TODO: Bug 1031456 - invoking js event without release() loses context
-        Actions(self.marionette).press(next).move(current).perform()
-        Wait(self.marionette).until(lambda m: self.minute != minute)
+    def spin_minute(self):
+        old_minute = self.minute
+        self._flick_menu_up(self._minutes_picker_locator)
+        self.wait_for_condition(lambda m: self.minute != old_minute)
```
mwargers: You need to use Wait() here.

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/date_and_time.py
npark-mozilla: Aha, I see what you meant.  so you're saying that when I tap the autotime, the tester has to be aware of the state of the autotime switch.  Initially what I meant to do was to enable it regardless of the current state of the autotime, because my goal is have it enabled regardless.  Make sense, that logic should be apparent in the test script.  I'll make the change.

tests/python/gaia-ui-tests/gaiatest/apps/system/regions/time_picker.py
mwargers: Johan explained me this object inherits from the Base object and then it's fine to use wait_for_element_displayed apparently.

tests/python/gaia-ui-tests/gaiatest/apps/system/regions/time_picker.py
npark-mozilla: ah, i was pasting over the current code in master because I reverted the change.  I'll change it to Wait() call.

mwargers: Same here, I guess it's fine.

tests/python/gaia-ui-tests/gaiatest/apps/system/regions/time_picker.py
mwargers: Same here, I guess it's fine.

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
```diff
@@ -0,0 +1,75 @@
...
+        self.marionette.switch_to_frame()
+        self.device.lock()
+        lock_screen = LockScreen(self.marionette)
+        self.assertTrue(self.get_time_difference(new_time, lock_screen.time) < 5)
+
+        # configure to set the time automatically (this will revert the timezone change), then lock screen
+        lock_screen.switch_to_frame()
+        lock_screen.unlock()
+        self.marionette.switch_to_frame(self.apps.displayed_app.frame)
+        datetime.set_automatic_time_update(True)
```
mwargers: self.apps.switch_to_displayed_app() ?

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
JohanLorenzo: Can't we convert this time to a python time, like we do [in the music app](https://github.com/mozilla-b2g/gaia/blob/master/tests/python/gaia-ui-tests/gaiatest/apps/music/regions/player_view.py#L62)?

npark-mozilla: It's slightly different, because in the music app we're getting the elapsed time (starts from 0:00), but in this case I need the difference between two times, and having AM/PM complicated things slightly.  for example, 12:59 to 1:02 is 3 minutes, but you have to make sure 12 = 0

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
npark-mozilla: ah, fixed

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
npark-mozilla: updated

autolander:

try-server-hook:
[npark-mozilla No-Jun Park (npark-mozilla)](https://github.com/npark-mozilla) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=c3ccfbf42841)

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
```diff
@@ -0,0 +1,80 @@
...
+        # Auto time update is by default set to true, turn it off to make region change
+        datetime.change_automatic_time_update(False)
+
+        # record time and change the region
+        old_time = datetime.get_current_time_text
+        datetime.set_region('Atlantic Ocean')
+
+        # get the displayed time after the region change
+        new_time = datetime.get_current_time_text
+        self.assertFalse(new_time == old_time)
```
JohanLorenzo: self.assertNotEqual

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
```diff
@@ -0,0 +1,80 @@
...
+        self.device.lock()
+
+        # wait until device is off and turn back on to check that the time is changed
+        Wait(self.marionette, timeout=20).until(
+            lambda m: not self.device.is_screen_enabled)
+        self.device.turn_screen_on()
+
+        # Check it reverted to the correct time, and compare it with the previously shown time
+        # Allow 5 minutes difference max
+        self.assertTrue(self.get_time_difference(old_time, lock_screen.time) < 5)
```
JohanLorenzo: self.assertLessEqual

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
```diff
@@ -0,0 +1,80 @@
...
+
+        # get the displayed time after the region change
+        new_time = datetime.get_current_time_text
+        self.assertFalse(new_time == old_time)
+
+        # lock screen and check time on the lockscreen
+        self.marionette.switch_to_frame()
+        self.device.lock()
+        lock_screen = LockScreen(self.marionette)
+        self.assertTrue(self.get_time_difference(new_time, lock_screen.time) < 1)
```
JohanLorenzo: self.assertLessEqual

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
JohanLorenzo:
``` Python
# in Lockscreen 
    @property
    def time(self):
        return time.strptime(self.marionette.find_element(*self._time_locator).text, '%M:%S %p')
        # AM/PM is hidden in a span, we would be able to get it

# Same thing in settings/date time

# in test_lockscreen_time_check.py 
old_time = datetime.time
new_time = datetime.time
delta = new_time - lockscreen.time
self.assertLessEqual(delta.total_seconds(), 5)
```

Do you think this is doable?

autolander:

autolander:

try-server-hook:
[npark-mozilla No-Jun Park (npark-mozilla)](https://github.com/npark-mozilla) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=4457535f180d)

try-server-hook:
[npark-mozilla No-Jun Park (npark-mozilla)](https://github.com/npark-mozilla) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=b89cd781ae25)

npark-mozilla:
hmm, i think i screwed up something during the merge, and now I have missing file.  (and I can't push that missing file).  I'll create another pull request

autolander:

autolander:

autolander:

npark-mozilla:
actually it just pushed my commits, guess it was delayed

autolander:

npark-mozilla:
Commits squashed.  sorry about the crazy long commit message.

JohanLorenzo:
You can edit the commit message to delete all the garbage: `git commit --amend`

autolander:

autolander:

npark-mozilla:
thanks!, for some reason I can't change the author but glad to got rid of the 50 page commit message.

autolander:

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class TestLockScreen(GaiaTestCase):
+
+    def setUp(self):
+        GaiaTestCase.setUp(self)
+        self.connect_to_local_area_network()
+
+    def test_lockscreen_time_check(self):
+
```
mwargers: Perhaps add a comment that references the bug number that this test is covering?

npark-mozilla: ahh, can't believe i missed that, (or i may have overwritten it)

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
```diff
@@ -0,0 +1,80 @@
...
+from gaiatest.apps.settings.app import Settings
+from gaiatest import GaiaTestCase
+from gaiatest.apps.lockscreen.app import LockScreen
+
+
+class TestLockScreen(GaiaTestCase):
+
+    def setUp(self):
+        GaiaTestCase.setUp(self)
+        self.connect_to_local_area_network()
```
mwargers: You don't need that, do you?

autolander:

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
npark-mozilla: actually i do, otherwise I can't get the automatic time update from internet.  (the switch won't be visible)

try-server-hook:
[npark-mozilla No-Jun Park (npark-mozilla)](https://github.com/npark-mozilla) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=282055f4b622)

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/manifest.ini
mwargers: You also need to add  ```carrier = true ```, because the automatic time setting depends on a carrier connection.

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
```diff
@@ -0,0 +1,80 @@
...
+        # Auto time update is by default set to true, turn it off to make region change
+        datetime.change_automatic_time_update(False)
+
+        # record time and change the region
+        old_time = datetime.get_current_time_text
+        datetime.set_region('Atlantic Ocean')
+
+        # get the displayed time after the region change
+        new_time = datetime.get_current_time_text
+        self.assertNotEqual(new_time, old_time)
```
mwargers: That assumes that you're not running this test in a timezone that is in 'Atlantic Ocean'. I guess that is practically spoken no problem.

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/date_and_time.py
```diff
@@ -21,3 +30,45 @@ def __init__(self, marionette):
...
+            Wait(self.marionette).until(
+                expected.element_present(*self._autotime_enabled_locator))))
+
+        self.marionette.find_element(*self._autotime_enabled_switch_locator).tap()
+
+        if not state:
+            self.wait_for_condition(lambda m: self.is_autotime_enabled is False)
+        else:
+            self.wait_for_condition(lambda m: self.is_autotime_enabled is True)
+
```
mwargers: Why not call this function toggle_automatic_time_update() and then check that at the end the old self.is_autotime_enabled != new self.is_autotime_enabled?
That way, you don't need a function argument at all. During the test, you should know the state of self.is_autotime_enabled anyway. You can do that check in the test file itself.

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
```diff
@@ -0,0 +1,80 @@
...
+        Wait(self.marionette, timeout=20).until(
+            lambda m: not self.device.is_screen_enabled)
+        self.device.turn_screen_on()
+
+        # Check it reverted to the correct time, and compare it with the previously shown time
+        # Allow 5 minutes difference max
+        self.assertLessEqual(self.get_time_difference(old_time, lock_screen.time), 5)
+
+
+    def get_time_difference(self, time_a, lockscreen_time):
```
mwargers: Minor nit, wouldn't it be better to name lockscreen_time time_b?

mwargers: Or perhaps better yet, old_time and new_time, since it seems to be necessary for this function that the first argument has the older time and the 2nd one the newer time.

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/date_and_time.py
```diff
@@ -5,12 +5,21 @@
 from marionette.by import By
 from marionette import expected
 from marionette import Wait
+
 from gaiatest.apps.base import Base
 
 
 class DateAndTime(Base):
     _24h_selector_locator = (By.CSS_SELECTOR, 'select.time-format-time')
+    _dateandtime_menu_item_locator = (By.ID, 'menuItem-dateAndTime')
```
mwargers: You're not using this locator.

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/date_and_time.py
```diff
@@ -21,3 +30,45 @@ def __init__(self, marionette):
...
+        close_button = self.marionette.find_element(*self._timezone_confirm_button_locator)
+
+        # loop options until we find the match
+        for li in options:
+            if region in li.text:
+                li.tap()
+                break
+        else:
+            raise Exception("Element '%s' could not be found in select wrapper" % region)
+        close_button.tap()
```
mwargers: Nit, perhaps add a line break here before close_button.tap()?

mwargers: And perhaps move the 
+ close_button = self.marionette.find_element(*self._timezone_confirm_button_locator)
call just before it?

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/manifest.ini
npark-mozilla: Actually I can get the automatic setting via wifi, so the phone does not necessarily have the sim card

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
npark-mozilla: Actually initially its variable was old_time and new_time, but then it wasn't apparent in the parameter name that lockscreen time (has no AM/PM) and the time from the setting (has AM/PM) has different format, I wanted to make sure that was clear in the declaration

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/date_and_time.py
npark-mozilla: good idea, will do

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/date_and_time.py
npark-mozilla: removed

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/date_and_time.py
npark-mozilla: good idea, will do that

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
npark-mozilla: yeah I had to actually think about that :)

autolander:

try-server-hook:
[npark-mozilla No-Jun Park (npark-mozilla)](https://github.com/npark-mozilla) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=c587cf0f227a)

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
```diff
@@ -0,0 +1,84 @@
...
+        # wait until device is off and turn back on to check that the time is changed
+        Wait(self.marionette, timeout=20).until(
+            lambda m: not self.device.is_screen_enabled)
+        self.device.turn_screen_on()
+
+        # Check it reverted to the correct time, and compare it with the previously shown time
+        # Allow 5 minutes difference max
+        self.assertLessEqual(self.get_time_difference(old_time, lock_screen.time), 5)
+
+    def get_time_difference(self, old_time, new_lockscreen_time):
```
JohanLorenzo: The following comment got lost in the PR:

``` Python
# in Lockscreen 
    @property
    def time(self):
        return time.strptime(self.marionette.find_element(*self._time_locator).text, '%M:%S %p')
        # AM/PM is hidden in a span, we would be able to get it

# Same thing in settings/date time

# in test_lockscreen_time_check.py 
new_time = datetime.time
delta = new_time - lockscreen.time
self.assertLessEqual(delta.total_seconds(), 5)
```

Do you think this would be doable?

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
JohanLorenzo: Adding a comment in the code would probably be a good idea, then :)

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/manifest.ini
JohanLorenzo: How about `wifi = true`?

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
npark-mozilla: ah, i was looking for this comment, and thought you deleted them.  the thing is that I still need the roundabout way of computing the difference, because in case new_time = 1:01 and lockscreen_time = 12:59, the difference is 2 min, but simple subtraction would give me 11:58.  Hence I have to convert them to minutes, check whether it's 12 o'clock, and do subtraction.  In order to do this, I find it simpler for me to extract the numbers from the string, rather than using the time format (well, at least easier to see the logic when you read)

This is all because the lockscreen won't display AM/PM.  very annoying.  actually, it should be a bug really.

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
npark-mozilla: ah, you're right. added

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/manifest.ini
npark-mozilla: true, we are guaranteed to have wifi connection in the lab i guess?

autolander:

try-server-hook:
[npark-mozilla No-Jun Park (npark-mozilla)](https://github.com/npark-mozilla) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=68f7846c49b0)

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
JohanLorenzo: AM/PM was present before 2.0 and was removed by UX: https://bugzilla.mozilla.org/show_bug.cgi?id=950884#c33

As the lockscreen doesn't display AM/PM, I would prefer to have this simple hack in the lockscreen, rather than reimplementing something which is already handled by the Python standard lib.
``` Python
    @property
    def time(self):
        # The lockscreen doesn't display AM/PM since 2.0. It's still present behind a hidden span though.
        lockscreen_time = self.marionette.execute_script("return document.getElementById('%s').innerHTML" % self._time_locator[1])
        lockscreen_time = lockscreen_time.replace('<span>', '').replace('</span>', '')
        return time.strptime(lockscreen_time, '%H:%M %p')
```

try-server-hook:
[npark-mozilla No-Jun Park (npark-mozilla)](https://github.com/npark-mozilla) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=cf4d7de9f96f)

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
npark-mozilla: What i did was to still return it as string literals, and convert it to datetime struct within the get_time_difference method, since once converted to time.struct, it is harder to modify the value, might negatively affect other tests.  Still, the method is lot cleaner because manual checks are not needed.

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
```diff
@@ -0,0 +1,75 @@
...
+        # wait until device is off and turn back on to check that the time is changed
+        Wait(self.marionette, timeout=20).until(
+            lambda m: not self.device.is_screen_enabled)
+        self.device.turn_screen_on()
+
+        # Check it reverted to the correct time, and compare it with the previously shown time
+        # Allow 5 minutes difference max
+        self.assertLessEqual(self.get_time_difference(old_time, lock_screen.time), 240)
+
+    def get_time_difference(self, old_time, new_time):
```
JohanLorenzo: If you move the conversion to the page class (like we do in Music), we don't need to define this method anymore.

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
```diff
@@ -0,0 +1,75 @@
...
+        from the text values, get time difference, returns in seconds.
+        both should be in time struct format
+        old_time: time from settings
+        new_time: time shown on lockscreen
+        """
+        dt_old_time = datetime.strptime(old_time, '%I:%M %p')
+        dt_new_time = datetime.strptime(new_time, '%I:%M %p')
+
+        difference = dt_new_time - dt_old_time
+        return difference.seconds # this also works when new_time is on next day
```
JohanLorenzo: I think you return only the number of seconds and not the total amount of seconds. This will get the test to always pass. You need to use [total_seconds()](https://docs.python.org/2/library/datetime.html#datetime.timedelta.total_seconds) here.

npark-mozilla: ah, just double checked, seconds give me the difference in seconds, max of 24 hours.  total_seconds() adds the day difference to it.  since we are not providing date, i think seconds works better in this case.  For example, 12:00 AM - 11:59PM will give me the delta of (-1, 60), and 11:59PM - 12:00AM will give me (0,86340).

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
npark-mozilla: i thought about that, but it's odd to move it to settings.dateandtime class, and it's also not exactly lockscreen related action, and getting the time difference is (so far) unique to this test script, so I couldn't find a better place to put it other than the test script itself.

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
JohanLorenzo: Oh my bad, I misread the documentation. You're absolutely right, let's keep it the current way then.

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/test_lockscreen_time_check.py
JohanLorenzo: Sorry, I mean: if we put ` datetime.strptime(old_time, '%I:%M %p') ` in the different page classes, we would just have to put `dt_new_time - dt_old_time` in the test itself without defining `get_time_difference` anymore. What do you think about that?

npark-mozilla: hmm, tempting, what i'll do is then leave the current string literal returning method, and create a new one that returns datetime format.  that should do it i think without breaking existing tests

try-server-hook:
[npark-mozilla No-Jun Park (npark-mozilla)](https://github.com/npark-mozilla) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=48b87662e23b)

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/date_and_time.py
patrickcalison: I happy.

tests/python/gaia-ui-tests/gaiatest/tests/functional/lockscreen/manifest.ini
mwargers: Interesting, I tried this myself. I removed my sim cards, but I couldn't get the automatic setting via wifi. In fact, the whole Automatic setting wasn't visible in the UI.

try-server-hook:
[npark-mozilla No-Jun Park (npark-mozilla)](https://github.com/npark-mozilla) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=0ba667989d4a)

try-server-hook:
[npark-mozilla No-Jun Park (npark-mozilla)](https://github.com/npark-mozilla) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=8f66b3183c77)

try-server-hook:
[npark-mozilla No-Jun Park (npark-mozilla)](https://github.com/npark-mozilla) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=8220972ee556)

try-server-hook:
[npark-mozilla No-Jun Park (npark-mozilla)](https://github.com/npark-mozilla) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=b00ceb2ea51d)

try-server-hook:
[npark-mozilla No-Jun Park (npark-mozilla)](https://github.com/npark-mozilla) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=86a5127a5762)

PR #27388, opened 2015-01-14
Bug 1079557 - [Calendar] launch time improvement r=gaye

millermedeiros:
got `content-interactive` from ~1850ms to ~1180ms
- bundled CSS/JS;
- changed the initialization order;
- replaced the module loader from alameda to curl.js; (there is something weird with our dependency tree that slows down alameda..)
- simplified some methods;
- reduced/batched reflows;
- faster `nextTick` implementation by using native `Promise` instead of `postMessage`;


try-server-hook:
[millermedeiros Miller Medeiros (millermedeiros)](https://github.com/millermedeiros) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=102a234790c4)

apps/calendar/build/build.js
```diff
@@ -29,11 +29,12 @@ exports.execute = function(options) {
   utils.ensureFolderExists(utils.getFile(options.STAGE_APP_DIR));
 
   dump('Will run rjs optimizer...\n');
-  var optimize = new Promise((accept, reject) => {
+  var optimizejs = new Promise((accept, reject) => {
     requirejs.optimize([config.path], accept, reject);
   });
 
-  optimize.then(() => {
+  optimizejs.then((log) => {
```
gaye: No need for parens :)

gaye: Also AFAICT the changes here are just logging the result of the call to `requirejs.optimize()`

apps/calendar/build/calendar.build.js
```diff
@@ -17,15 +18,34 @@
 
   modules: [
     {
-      name: 'ext/alameda',
+      create: true,
```
gaye: What does create mean?

apps/calendar/build/calendar.build.js
```diff
@@ -17,15 +18,34 @@
...
-        'debug',
-        'next_tick'
+        'views/errors',
+        'views/month',
+        'views/month_day_agenda',
+        'views/time_header'
       ]
     },
     {
+      create: true,
```
gaye: Was there any improvement from only making two bundles?

apps/calendar/js/app.js
```diff
@@ -53,9 +53,16 @@ module.exports = {
...
 
+    var loadedSubViews = false;
+
     this._pendingManager.oncomplete = function onpending() {
       document.body.classList.remove(pendingClass);
       performance.pendingReady();
+      // start loading sub-views as soon as possible
+      if (!loadedSubViews) {
+        loadedSubViews = true;
+        window.require(['lazy_loaded', 'css!lazy_loaded']);
```
gaye: No need to grab `require` off of `window` here.

apps/calendar/js/app.js
```diff
@@ -259,12 +277,17 @@ module.exports = {
...
     this._routes();
 
-    var recurringEventsController = new RecurringEventsController(this);
-    this.observePendingObject(recurringEventsController);
-    recurringEventsController.observe();
-    this.recurringEventsController = recurringEventsController;
-
     nextTick(() => this.view('Errors'));
+
+    // Restart the calendar when the timezone changes.
```
gaye: Hrm why here instead of `main`?

apps/calendar/js/app.js
```diff
@@ -304,31 +327,24 @@ module.exports = {
...
-    navigator.mozL10n.once(next);
     this.db.load(() => {
-      next();
+      this._initControllers();
       // it should only start listening for month change after we have the
       // calendars data, otherwise we might display events from calendars that
       // are not visible. this also makes sure we load the calendars as soon as
       // possible
-      this.store('Calendar').all(() => dayObserver.init());
+      this.store('Calendar').all(() => {
```
gaye: What changed here?

apps/calendar/js/app.js
```diff
@@ -374,7 +390,7 @@ module.exports = {
 
     var snake = snakeCase(name);
     debug('Will try to load view', name);
-    require([ 'views/' + snake ], (aView) => {
+    window.require([ 'views/' + snake ], (aView) => {
```
gaye: Why `window.require`?

apps/calendar/js/main.js
```diff
@@ -1,25 +1,11 @@
...
-
-    window.addEventListener('load', function onLoad() {
-      window.removeEventListener('load', onLoad);
-      return accept();
-    });
-  });
-}
-
-console.log('Will configure rjs...');
+window.require = window.require || window.curl;
```
gaye: Will `window.require` ever already be defined when this is called?

apps/calendar/js/main.js
```diff
@@ -34,30 +20,26 @@ require.config({
...
+    'views/modify_event': 'delayed',
+    'views/settings': 'delayed',
+    'views/view_event': 'delayed'
+  }
+});
 
-    app.init();
-  });
+define(function(require) {
+  'use strict';
```
gaye: <<

apps/calendar/js/main.js
```diff
@@ -34,30 +20,26 @@ require.config({
...
+  var app = require('app');
+  app.init();
 });
 
-}());
+require(['main']);
```
gaye: I think it's a little bit sloppy to have all of this executed in the global space. Can we confine ourselves to a self-executing anonymous function block like we did before?

apps/calendar/js/next_tick.js
```diff
@@ -1,29 +1,14 @@
...
 
 /**
  * Very similar to node's nextTick.
- * Much faster then setTimeout.
+ * Much faster than setTimeout or window.postMessage
  */
 module.exports = function(callback) {
-  nextTickStack.push(callback);
-  window.postMessage(NEXT_TICK, '*');
+  resolved.then(callback);
```
gaye: Can we make a jsperf with this just so we know for sure?

apps/calendar/js/utils/ordered_map.js
```diff
@@ -1,110 +0,0 @@
-define(function(require, exports, module) {
```
gaye: Wow I didn't know we weren't using this anymore lol

apps/calendar/build/build.js
millermedeiros: yes.. before I had separate targets for JS/CSS that's why I renamed the `optimize` to `optimizejs`..

do you think the log is a good thing or not necessary? (looks like a `build.txt` file is also saved on the `build_stage/calendar` folder)

apps/calendar/build/calendar.build.js
millermedeiros: `create` means that it creates a new module/file named `bundle`.. decided to use this name instead `ext/curl` since it makes more sense.

apps/calendar/build/calendar.build.js
millermedeiros: yes there was. easier to remove the shared dependencies and since we call `require(['lazy_loaded'])` as soon as `app-loaded` it doesn't make any perceived difference to the user (don't wait any click to happen before triggering the load)

apps/calendar/js/app.js
millermedeiros: yes, there is because it's using `curl` and not `alameda`.

apps/calendar/js/app.js
millermedeiros: makes more sense here since this is only needed after the app is initialized. it also made it easier to measure how longs things were taking and to toggle the behavior.

apps/calendar/js/app.js
millermedeiros: nothing.. I'll revert to old style.. but during the development I tried a lot of different things here and even had a `window.performance.mark('calendar.all')` here to get a better idea of how long things were taking..

I removed all the `window.performance.mark` calls because I don't know if that would cause any conflicts with datazilla... maybe we should add them back?

millermedeiros: see the old mark calls: https://github.com/millermedeiros/gaia/commit/b55d0ee024e51fce9e25188b2529fce2c9c40718

apps/calendar/js/app.js
millermedeiros: because we are using `curl.js` instead of `alameda` or `requirejs`

apps/calendar/js/main.js
millermedeiros: probably not, but I kept it like this because I was testing with `alameda` and `curl.js`.. `alameda` exports a global `require` function, while `curl` doesn't.

apps/calendar/js/main.js
millermedeiros: ok

apps/calendar/js/next_tick.js
millermedeiros: 100% sure, it's over 100x faster: http://jsperf.com/test002/8 - I even talked to dbaron about it and also created a blog post: http://blog.millermedeiros.com/promise-nexttick/

this change improved the startup perf in ~80ms.

apps/calendar/js/utils/ordered_map.js
millermedeiros: I think we stopped using this after the timeController/dayObserver refactor

apps/calendar/js/views/calendar_colors.js
```diff
@@ -1,229 +0,0 @@
-define(function(require, exports, module) {
```
millermedeiros: so good to see this file disappear!!!

apps/calendar/js/views/modify_account.js
```diff
@@ -8,7 +8,6 @@ var Presets = require('presets');
 var URI = require('utils/uri');
 var View = require('view');
 
-require('css!modify_account_view');
```
millermedeiros: I decided to bundle all the lazy loaded styles into a single stylesheet since that was faster/simpler.

apps/calendar/build/build.js
gaye: Probably not necessary? We're never really debugging this

apps/calendar/build/calendar.build.js
gaye: Hrm would it not do that by default like it was before?

apps/calendar/js/app.js
gaye: If there was a difference, then wouldn't you have to change all of the calls to require?

apps/calendar/js/main.js
gaye: So then just `window.require = window.curl;`?

millermedeiros: my idea here was that I was testing with both curl and alameda.. doing like this the code runs with any of the libs (unit tests still uses alameda..) better to keep it like this.

apps/calendar/js/app.js
millermedeiros: because the local `require` is synchronous (it's part of the AMD spec).. the async `require` is not part of the AMD spec, so other libs don't implement it.

apps/calendar/build/calendar.build.js
millermedeiros: before it was just compiling the `ext/alameda` and `include`'ing the other modules.. now it creates a new file called `bundle.js` which makes way more sense than having a bunch of stuff that is not part of `alameda` inside the `ext/alameda.js` file

try-server-hook:
[millermedeiros Miller Medeiros (millermedeiros)](https://github.com/millermedeiros) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=bcc47e820004)

apps/calendar/build/calendar.build.js
gaye: Okay fair enough

apps/calendar/js/app.js
gaye: ... So then `window.require` is not the same function as `require` after this changeset?

apps/calendar/js/ext/curl.js
```diff
@@ -0,0 +1,1367 @@
+/** @license MIT License (c) copyright 2010-2013 B Cavalier & J Hann */
```
gaye: Shall we go ahead and add this to `package.json`?

apps/calendar/js/main.js
```diff
@@ -1,25 +1,15 @@
-(function() {
+(function(){
```
gaye: Nit: space between `)` and `{`

apps/calendar/js/main.js
```diff
@@ -34,30 +24,24 @@ require.config({
...
-    // We do this on a timer because this event may fire
-    // many times. Refreshing the url of the calendar frequently
-    // can result in crashes so we attempt to do this only after
-    // the user has completed their selection.
-    debug('Will listen for timezone change...');
-    window.addEventListener('moztimechange', () => {
-      debug('Noticed timezone change!');
-      nextTick(() => app.forceRestart(), app._mozTimeRefreshTimeout);
-    });
+// first require.config call is used by r.js optimizer, so we do this second
```
gaye: So the r.js optimizer ignores subsequent calls to `config`?

apps/calendar/js/main.js
```diff
@@ -34,30 +24,24 @@ require.config({
...
+    'views/modify_account': 'lazy_loaded',
+    'views/modify_event': 'lazy_loaded',
+    'views/settings': 'lazy_loaded',
+    'views/view_event': 'lazy_loaded'
+  }
+});
 
-    app.init();
-  });
+require(['app'], function(app) {
```
gaye: Nit: Can use nice one line lambda here

apps/calendar/js/next_tick.js
gaye: Cool thanks for that!

apps/calendar/js/views/calendar_colors.js
gaye: +1

apps/calendar/js/views/month.js
```diff
@@ -120,9 +120,7 @@ Month.prototype = {
       this.currentFrame.deactivate();
     }
 
-    this._getFrame(this._previousTime());
```
gaye: What happened here?

apps/calendar/js/views/view_event.js
```diff
@@ -83,10 +82,8 @@ ViewEvent.prototype = {
     this.setContent('location', model.location);
 
     if (this.originalCalendar) {
-      // Set calendar color.
-      this.element
-        .querySelector('section[data-type="list"]')
-        .className = 'calendar-id-' + model.calendarId;
+      this.element.querySelector('.icon-calendar-dot').style.color =
```
gaye: Perhaps https://developer.mozilla.org/en-US/docs/Web/API/Element.getElementsByClassName for clarity?

apps/calendar/js/main.js
millermedeiros: yes, r.js only grabs the first `require.config` call.

apps/calendar/js/views/month.js
millermedeiros: reduced to a single "frame" since now the busytimes are only loaded/rendered after the view is activated. gets some extra ms.

apps/calendar/js/views/view_event.js
millermedeiros: I think `querySelector` is a better option since it returns a single element and it's clear enough.

apps/calendar/js/app.js
millermedeiros: yes, `window.require !== require`

try-server-hook:
[millermedeiros Miller Medeiros (millermedeiros)](https://github.com/millermedeiros) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=4f810f621fbb)

apps/calendar/js/ext/curl.js
millermedeiros: done

apps/calendar/js/app.js
gaye: Oh I thought we'd deleted alameda. So now we're using the alameda implementation of async require and the curl implementation of sync require?

millermedeiros: workers and unit tests still uses alameda (curl can't load scripts dynamically inside a web worker)

gaye: Ok makes sense. Would you mind adding a comment somewhere explaining all of this?

try-server-hook:
[millermedeiros Miller Medeiros (millermedeiros)](https://github.com/millermedeiros) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=893ff4727837)

PR #27408, opened 2015-01-15
Bug 1121301 - Remove the event envelope sending commands which are handled by MODEM.

weilonge:


frsela:
Are handled by MODEM in ALL modems? or only some vendors. I mean, is this a generic solution?

apps/system/js/icc_events.js
```diff
@@ -60,22 +60,14 @@ var icc_events = {
...
         iccEventSet[iccMan.STK_EVENT_TYPE_MT_CALL]) {
-        self.downloadEvent(iccEventSet[iccMan.STK_EVENT_TYPE_MT_CALL], {
-          eventType: icc._iccManager.STK_EVENT_TYPE_MT_CALL,
-          number: call.id ? call.id.number : call.number,
-          isIssuedByRemote: outgoing,
-          error: null
-        });
+        // Remove the event envelope of STK_EVENT_TYPE_MT_CALL
+        // which will be handled by MODEM
       }
```
frsela: If not used, you can remove the if clause

apps/system/js/icc_events.js
```diff
@@ -60,22 +60,14 @@ var icc_events = {
...
           iccEventSet[iccMan.STK_EVENT_TYPE_CALL_DISCONNECTED]) {
-          self.downloadEvent(
-            iccEventSet[iccMan.STK_EVENT_TYPE_CALL_DISCONNECTED], {
-            eventType: icc._iccManager.STK_EVENT_TYPE_CALL_DISCONNECTED,
-            number: call.id ? call.id.number : call.number,
-            error: err
-          });
+          // Remove the event envelope of STK_EVENT_TYPE_CALL_DISCONNECTED
+          // which will be handled by MODEM
         }
```
frsela: The same, now the if clause is not used anymore, delete it

apps/system/js/icc_events.js
```diff
@@ -84,25 +76,16 @@ var icc_events = {
           case 'connected':
             if (iccEventSet &&
               iccEventSet[iccMan.STK_EVENT_TYPE_CALL_CONNECTED]) {
-              self.downloadEvent(
-                iccEventSet[iccMan.STK_EVENT_TYPE_CALL_CONNECTED], {
-                eventType: icc._iccManager.STK_EVENT_TYPE_CALL_CONNECTED,
-                number: call.id ? call.id.number : call.number,
-                isIssuedByRemote: outgoing
-              });
```
frsela: In this case, case 'connected' is a dummy one, remove it too

apps/system/js/icc_events.js
```diff
@@ -84,25 +76,16 @@ var icc_events = {
...
             call.removeEventListener('statechange', callStateChange);
             if (iccEventSet &&
               iccEventSet[iccMan.STK_EVENT_TYPE_CALL_DISCONNECTED]) {
-              self.downloadEvent(
-                iccEventSet[iccMan.STK_EVENT_TYPE_CALL_DISCONNECTED], {
-                eventType: icc._iccManager.STK_EVENT_TYPE_CALL_DISCONNECTED,
-                number: call.id ? call.id.number : call.number,
-                isIssuedByRemote: outgoing,
-                error: null
-              });
```
frsela: Ups, these same for disconnected one, you can remove the switch !

apps/system/js/icc_events.js
```diff
@@ -213,8 +196,8 @@ var icc_events = {
         break;
       case icc._iccManager.STK_EVENT_TYPE_LOCATION_STATUS:
         DUMP('icc_events_register - Location changes event');
-        conn.addEventListener('voicechange',
-          this.register_icc_event_voicechange);
```
frsela: The same

frsela:
Another question, if managed by modem, I suppose we never will receive that EVENT_TYPE_*_STATUS ...

In that case all unused code should be removed, but add an explaining comment at the top of the file in order to explain why some events are not manage here ;)

Thank you !

weilonge:
This is not a generic issue, and this customization is only for a specific MODEM chip.
There is a table (at bug 1121301 comment 8) which describes what kind of event should be handled by MODEM (MD).
The event envelope of following event is removed:
Event: MT call                     MD
Event: Call connected              MD
Event: Call disconnected           MD
Event: Location status             MD

But some others (ex: STK_EVENT_TYPE_USER_ACTIVITY) are still handled by AP(Gaia).
That's why Gaia still have to receive EVENT_TYPE_*_STATUS even Gaia don't handle every event.

I did not remove all handler codes for call and location statue because it's still possible that the partner will customize the behavior of all setup event by their own.
If we remove all codes here, the partner would be difficult to see the whole picture when they try to customize it. (the code maintenance by partner is different so our git history will be lost. )

Thanks very much for your careful review.

frsela:
In that case, I agree leaving the comments there.
Thank you for your feedback.

PR #27427, opened 2015-01-15
Bug 1118407 - [RTL][Gallery] Numeric value appears to the left of the word selected in gallery header r=wilsonpage

punamdahiya:


apps/gallery/style/gallery.css
```diff
@@ -257,6 +257,11 @@ body.pickView #thumbnails {
   outline-offset: -0.2rem;
 }
 
+/* RTL modifications */
+html[dir=rtl] #thumbnails-number-selected {
+  direction: rtl;
+}
+
```
anefzaoui: I just stumbled upon this, for consistency with the rest of Gaia css files and the way RTL is done, it's better to have everything-rtl at the bottom of the file. :)

wilsonpage: I disagree; I think it's better to be able to see all the different states a single element can be in in one place. Otherwise it's easy for someone new to not know that there are overrides in other parts of the code base.

PR #27440, opened 2015-01-16
Bug 1037351 - Double clicking to add an event in month view r=gaye

millermedeiros:


try-server-hook:
[millermedeiros Miller Medeiros (millermedeiros)](https://github.com/millermedeiros) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=3e4f19f60ba2)

apps/calendar/js/views/month.js
```diff
@@ -105,12 +108,25 @@ Month.prototype = {
...
         break;
       case 'monthChange':
         this.changeDate(e.data[0]);
         break;
     }
+    this._lastTarget = target;
+  },
+
+  _goToAddEvent: function(date) {
+    // slight delay to avoid tapping the elements inside the add event screen
```
gaye: If this is an issue, would it perhaps be better to have transitions between views that took long enough for things like this to not happen?

apps/calendar/js/views/month.js
```diff
@@ -105,12 +108,25 @@ Month.prototype = {
...
+        // make sure we discard double taps that started on a different day
+        if (this._lastTarget === target) {
+          this._goToAddEvent();
+        }
         break;
       case 'monthChange':
         this.changeDate(e.data[0]);
         break;
     }
+    this._lastTarget = target;
```
gaye: Don't we only want to set this in the `'click'` case?

millermedeiros: it doesn't make any difference, `target` only exists for delegated events.. which in this case are only `dbltap` and `click`

apps/calendar/js/views/month.js
millermedeiros: we would still need to block interaction during the transition. I would say we keep it like this for now and implement transitions later as a separate bug.

apps/calendar/js/views/month.js
```diff
@@ -105,12 +108,25 @@ Month.prototype = {
         this.controller.selectedDay = date;
         break;
       case 'dbltap':
-        this.app.go('/day/');
+        // make sure we discard double taps that started on a different day
+        if (this._lastTarget === target) {
+          this._goToAddEvent();
```
gaye: Shouldn't you reset `this._lastTarget`?

millermedeiros: also doesn't change anything since target is only checked inside `dbltap` and `click` always dispatches on the first tap (which will override the `_lastTarget`)

PR #27476, opened 2015-01-19
Bug 1119713 - [Stingray] KeyNavigationAdapter handles both keydown and k...

huchengtw-moz:
...eyup event


rexboy7:
Would it be better if we note somewhere (perhaps when adding event listener) that `enter-keyup` is used to prevent broken animation?

try-server-hook:
[huchengtw-moz John Hu (huchengtw-moz)](https://github.com/huchengtw-moz) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=e80287e15efc)

PR #27530, opened 2015-01-20
Bug 1091511 - [Message]The number/email or URL can't be tapped if there is full-width text in this message

steveck-chung:


try-server-hook:
[steveck-chung steveck-chung](https://github.com/steveck-chung) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=3628638fcdda)

apps/sms/js/link_helper.js
```diff
@@ -58,7 +58,7 @@ function checkDomain(domain) {
 }
 
 // defines things that can match right before to be a "safe" link
-var safeStart = /[\s,:;\(>]/;
+var safeStart = /[\s,:;\(>\u0080-\uFFFF]/;
```
azasypkin: I may ask stupid question now :) It's not related to this patch, but I tested patch with one of the built-in keyboards and I see there're two types of the following symbols ```()``` and ```（ ）```. I tried like ```(175)63759``` ((175)63759 recognized as phone number) and ```（175）63759``` (63759 recognized as phone number).

So, is only the first pair maybe used for phone numbers in Chinese and my second example isn't common\valid ?

steveck-chung: Well... full-width char might be used in Chinese/Japan/Korea IME, and it's always painful for developers ;) For this case you might be likely to see a phone number will all full-width symbol and number like `（６６６）６６６６`, and our logic will even not treat this string as valid phone number

azasypkin: Heh, nice :) Thanks for the explanation, I'm curious if iOS and Android support such numbers

steveck-chung: I think it depends on app, and you could see each app have their different logic for handling the link :)

PR #27566, opened 2015-01-21
Bug 1122820 - [Camera] While USB Storage is enabled, user can enter Camera and tap 'Shutter' before 'No Access' UI is present; can stall flash on

wilsonpage:


try-server-hook:
[wilsonpage Wilson Page (wilsonpage)](https://github.com/wilsonpage) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=1b5ae1a2813c)

apps/camera/js/app.js
```diff
@@ -274,6 +273,7 @@ App.prototype.loadLazyModules = function() {
 
   this.loadL10n(done());
   this.loadLazyControllers(done());
```
justindarc: Isn't this also still going to trigger `done()`?

wilsonpage: Maybe the naming isn't clear here. The `all-done` util returns a function that when called returns a new callback function. Each time the function is called it counts 1. The callback function passed on line `279` line be called until the three callback above have fired.

Soz finding it really hard to explain! Check out the new test I wrote as part of this patch in `app_test.js`.

PR #27576, opened 2015-01-21
Bug 1124081 - Warning about configuring email without SSL has not enough space for localization

jrburke:
Overrides some shared styles for confirm dialogs:
- Buttons are vertically stacked, so that "Learn More" is on the top, and buttons take full width of screen.
- The existing padding-bottom in the shared styles for the form is increased to account for the use of vertical space of the buttons.
- Paragraph border is only applied to the first paragraph.

The paragraph area is scrollable, so if the localization needs more space, the user will be able to scroll to see the rest of the text.

Picture of how it looks now on a flame device:

### DIFF
--- a/apps/email/style/setup_cards.css
+++ b/apps/email/style/setup_cards.css
@@ -455,3 +455,28 @@ cards-settings-debug button {
   margin: 1.5rem auto;
   width: calc(100% - 3rem);
 }
+
+/* Shared style overrides for the "are you sure" unencrypted connection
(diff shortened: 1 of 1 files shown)


try-server-hook:
[jrburke James Burke (jrburke)](https://github.com/jrburke) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=177cfd133211)

PR #27577, opened 2015-01-21
Bug 1115083 - Intermittent day_observer_test.js | day_observer cached #on: yesterday + persist + remove + visibility +autoland

millermedeiros:
this should avoid intermittent failures caused by `persist`.. it's an "ugly hack" but better than adding a _lock_ to avoid dispatching events while busytimes are being persisted, that way user can see busytimes _pop_ into the views while the calendar is still syncing.


try-server-hook:
[millermedeiros Miller Medeiros (millermedeiros)](https://github.com/millermedeiros) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=ed2f68ec8fc7)

PR #27590, opened 2015-01-22
Bug 1124265 - Wait for app to install before inspecting it's manifest +autoland

timdream:


try-server-hook:
[timdream Timothy Guan-tin Chien (timdream)](https://github.com/timdream) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=7781ca3a31e4)

apps/system/test/unit/input_app_list_test.js
```diff
@@ -282,6 +282,37 @@ suite('InputAppList', function() {
       });
     });
 
+    test('install app -- mark as downloading', function(done) {
+      var newApp = Object.create(INPUT_APP);
+      newApp.manifestURL = 'app://myfunnykeyboard.org/manifest.webapp';
```
mnjul: As discussed, please move the `manifestURL` assignment to the place where `downloadsuccess` is fired to better simulate the scenario that it is only available after the (mock) app has been downloaded.

Yes, the current test case code would fail without the patch on line 305. But it does not really test the offending behavior: without the patch, the `deepEqual()`s would actually pass as the `install` event would go through all the `_addInputApp()` process and this is the core behavior we want to guard.

try-server-hook:
[timdream Timothy Guan-tin Chien (timdream)](https://github.com/timdream) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=305e0a0590fc)

PR #27606, opened 2015-01-22
Ability to disable mutation observers and null-check noCompleteBug

msujaws:
This will help us with Loop development, which uses gaia-l10n on other browsers where mutation observers may not be available. See https://bugzilla.mozilla.org/show_bug.cgi?id=1097852 for more information.

### DIFF
--- a/shared/js/l10n.js
+++ b/shared/js/l10n.js
@@ -1476,7 +1476,8 @@
     characterData: false,
     childList: true,
     subtree: true,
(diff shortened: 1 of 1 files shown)


try-server-hook:
[msujaws Jared Wein (msujaws)](https://github.com/msujaws) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=d3ee1265c2c6)

msujaws:
I have a simpler patch that we can provide.

PR #27608, opened 2015-01-23
Bug 1124426 - [email/UI] The cookie cache may permanently screw up if we ever have more than 40 cookies worth of cache

jrburke:
Max segments now in a global HTML_COOKIE_CACHE_MAX_SEGMENTS, and html_cache.js does not write past that now for cookies, and html_cache_restore.js will not use values past that value.

Since html_cache.js has always kept that 40 segment area clean, this should be enough to ignore any cases where we get garbage above the 40 segment max.


try-server-hook:
[jrburke James Burke (jrburke)](https://github.com/jrburke) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=e69a03fe6799)

PR #27636, opened 2015-01-23
Bug 1106970 - Search box fixes

hfiguiere:


try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=e9782ac4d2c9)

apps/music/index.html
```diff
@@ -62,7 +62,8 @@ <h1 id="title-text" data-l10n-id="music">Music</h1>
         <form id="views-tiles-search" role="search" class="skin-dark">
           <button id="views-tiles-search-close" data-l10n-id="search-close" type="submit">close</button>
           <p>
-            <input id="views-tiles-search-input" type="text" dir="auto"
+            <input id="views-tiles-search-input" type="search" dir="auto"
+                   x-inputmode="verbatim"
```
dominickuo: Though I don't know what's `x-inputmode` is use for, but I look on the contacts app's html and it use this as well, so I guess this can make the music search has the same behaviour.

apps/music/js/ui/views/list_view.js
```diff
@@ -432,6 +433,14 @@ var ListView = {
 
         break;
 
+      case 'keypress':
+        if (target.id === 'views-list-search-input') {
+          if (evt.keyCode == 13) { // VK_Return
```
dominickuo: Probably we can use `KeyboardEvent.DOM_VK_RETURN` which is the same as the keyCode 13, and no need to add comment.

apps/music/index.html
hfiguiere: I lifted that from Contacts, and it is definitely to disable auto correct.u

apps/music/js/ui/views/list_view.js
hfiguiere: I lifted that from contacts as well :-/

hfiguiere: Will fix.

try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=f31ae4529959)

PR #27638, opened 2015-01-23
Bug 1122501 - [Messages][Drafts] Unsaved draft is silently discarded when user tries to forward message. r=schung

azasypkin:


try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=ff2e2527dc39)

apps/sms/js/activity_handler.js
```diff
@@ -241,7 +246,11 @@ var ActivityHandler = {
         return;
       }
 
-      Utils.confirm('discard-new-message').then(() => {
+      Utils.confirm(
+        'discard-new-message',
+        'unsent-message-title',
```
azasypkin: not sure what title will be better here (default or the same for all similar cases), what do you think?

apps/sms/js/activity_handler.js
```diff
@@ -250,54 +259,30 @@ var ActivityHandler = {
...
     if (Compose.isEmpty()) {
-      this.launchComposer(activity);
-    } else {
-      // ask user how should we do
-      ActivityHandler.displayUnsentConfirmation(activity);
+      Navigation.toPanel('composer', composerArgs);
+      return;
     }
+
+    Utils.confirm(
```
steveck-chung: Originally I was thinking that this should be triggered when in new activity, but the fact is opening new activity from other app will not trigger this(maybe it's because of new activity became inline, and it's reasonable from my POV because the message is saved as draft). After diving further I found it was mainly use for ActivityPicker.sendMessage. I'm not sure if it's suitable to reuse Activity.toView in that case especially it's not related to activity, but leaving this part in activity handler and no activity will actually uses it seems weird to me. Please correct me if I misunderstood.

So, here is some idea(but maybe not in this patch): Apply new confrim/toPanel logic in ActivityPicker.sendMessage instead of reuse ActivityHandler.toView(if didn't enter the thread directly, maybe we can refine it as well), and we can revisit if this part is necessary for ActivityHandler, wdyt?

apps/sms/js/activity_handler.js
steveck-chung: Maybe same for similar case is better? I don't see much benifit if we differentiate the discard confrim here.

apps/sms/js/activity_handler.js
julienw: Just a note of history: the idea of using ActivityHandler.toView in sendMessage was to behave just as if we were in an activity. But that's right: at that time we didn't have Navigation.toPanel yet, so maybe we can use that now.

azasypkin: > Originally I was thinking that this should be triggered when in new activity, but the fact is opening new activity from other app will not trigger this(maybe it's because of new activity became inline, and it's reasonable from my POV because the message is saved as draft). After diving further I found it was mainly use for ActivityPicker.sendMessage. 

Yeah, I also hard time trying to understand when it's called just via reading the code :)

> I'm not sure if it's suitable to reuse Activity.toView in that case especially it's not related to activity, but leaving this part in activity handler and no activity will actually uses it seems weird to me. Please correct me if I misunderstood.

Totally agree.

> So, here is some idea(but maybe not in this patch): Apply new confrim/toPanel logic in ActivityPicker.sendMessage instead of reuse ActivityHandler.toView(if didn't enter the thread directly, maybe we can refine it as well), and we can revisit if this part is necessary for ActivityHandler, wdyt?

Yeah, good idea, let me check if it doesn't require much changes and if so, I can do it in this patch!

azasypkin: The only thing is that we can someday extract number/email->thread_or_contact logic from ActivityHandler and use it here as well, so that when you'd like to send a message to number or email from message bubble and you already have appropriate thread - you'll navigate there instead of composer

julienw: oh yeah, good idea; will you file a bug for this ? :)

azasypkin: Sure!

try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=95c7d0adda15)

apps/sms/js/activity_picker.js
```diff
@@ -76,13 +76,6 @@ var ActivityPicker = {
       }
     }), onsuccess, onerror);
   },
-  sendMessage: function ap_sendMessage(number) {
```
azasypkin: Basically after I adressed your suggestion it's clear that ```sendMessage``` should not be in activity picker anymore :)

apps/sms/js/thread_ui.js
```diff
@@ -2799,10 +2816,14 @@ var ThreadUI = {
...
-            ActivityPicker.sendMessage(param);
+          method: () => {
+            this.navigateToComposer({
+              activity: { number: email }
+            });
           },
-          params: [email]
+          // As we change panel here, we don't want to call 'complete' that
+          // changes the panel as well
+          incomplete: true
```
azasypkin: That's the fix for a small group-email thread bug, currently it's impossible to send MMS from participants view if you're in group-email thread - if user tries to do that he's just returned to Thread panel

apps/sms/test/unit/thread_ui_test.js
```diff
@@ -4133,7 +4152,7 @@ suite('thread_ui.js >', function() {
             MockSettings.supportEmailRecipient = true;
 
             this.sinon.spy(ActivityPicker, 'email');
-            this.sinon.spy(ActivityPicker, 'sendMessage');
+            this.sinon.spy(ThreadUI, 'navigateToComposer');
```
azasypkin: I'm asserting this method instead of ```Navigation.toPanel``` below just to be sure that we support "discard" confirm dialog for those cases and not just navigation to Composer

try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=358067d146e7)

apps/sms/js/activity_picker.js
steveck-chung: Ya, it looks good :+1:

apps/sms/js/thread_ui.js
steveck-chung: Sorry I don't quite understand this part and I can't reproduce it...  Could you please explain more?

apps/sms/test/unit/activity_handler_test.js
```diff
@@ -981,14 +877,37 @@ suite('ActivityHandler', function() {
...
+
+      ActivityHandler.handleMessageNotification(message);
+
+      sinon.assert.notCalled(MessageManager.getMessage);
+      sinon.assert.notCalled(Utils.confirm);
+      sinon.assert.notCalled(Navigation.toPanel);
+
+      // It shouldn't matter if message input has any text or not since target
+      // thread is already opened
+      Compose.isEmpty.returns(false);
```
steveck-chung: Maybe we can move this part to the suite below?

apps/sms/js/thread_ui.js
azasypkin: Discussed offline :)

try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=a9cf0aafb72a)

apps/sms/test/unit/activity_handler_test.js
azasypkin: Done!

PR #27653, opened 2015-01-24
Bug 819744 - Delay the light sensor response in screen_manager.js r=timdream

gpgreen:
Updated the patch to break out auto light sensor response to a different file: "screen_auto_brightness.js". Added a unit test for this class. Removed light sensor response tests from screen_manager_test.js.

### DIFF
--- a/apps/system/js/screen_auto_brightness.js
+++ b/apps/system/js/screen_auto_brightness.js
@@ -1,0 +1,149 @@
+'use strict';
+
+(function(exports) {
+
+  /*
+   * constructor
+   */
+  var ScreenAutoBrightness = function ScreenAutoBrightness() {
+      // does nothing
+  };
+
+  /*
+   * callback function to set screen brightness.
+   */
+  ScreenAutoBrightness.prototype.onbrightnesschange = null;
+
+  /*
+   * The auto-brightness algorithm will never set the screen brightness
+   * to a value smaller than this. 0.1 seems like a good screen brightness
+   * in a completely dark room on a Unagi.
+   */
+  ScreenAutoBrightness.prototype.AUTO_BRIGHTNESS_MINIMUM = 0.1;
+
+  /*
+   * This constant is used in the auto brightness algorithm. We take
+   * the base 10 logarithm of the incoming lux value from the light
+   * sensor and multiplied it by this constant. That value is used to
+   * compute a weighted average with the current brightness and
+   * finally that average brightess is and then clamped to the range
+   * [AUTO_BRIGHTNESS_MINIMUM, 1.0].
+   *
+   * Making this value larger will increase the brightness for a given
+   * ambient light level. At a value of about .25, the screen will be
+   * at full brightness in sunlight or in a well-lighted work area.
+   * At a value of about .3, the screen will typically be at maximum
+   * brightness in outdoor daylight conditions, even when overcast.
+   */
+  ScreenAutoBrightness.prototype.AUTO_BRIGHTNESS_CONSTANT = 0.27;
+
+  /*
+   * We won't set a new brightness value if the difference between the old and
+   * new ambient light sensor values is lower than this constant.
+   */
+  ScreenAutoBrightness.prototype.AUTO_BRIGHTNESS_MIN_DELTA = 10;
+
+  /*
+   * This variable contains the latest ambient light sensor value that was used
+   * to set a brightness.
+   */
+  ScreenAutoBrightness.prototype._previousLux = undefined;
+
+  /*
+   * auto adjust screen delays, see autoAdjustBrightness function
+   * wait for _autoAdjustDelay milliseconds before adjusting brightness value
+   */
+  ScreenAutoBrightness.prototype._state = 0;
+  ScreenAutoBrightness.prototype._autoAdjustDelay = 750;
+
+  /*
+   * id of delay timeout
+   */
+  ScreenAutoBrightness.prototype._delayTimeout = null;
+
+  /*
+   * keep track of last lux value during autoAdjustBrightness delay's
+   */
+  ScreenAutoBrightness.prototype._autoDelayPrevLux = undefined;
+
+  // reinitialize to start state
+  ScreenAutoBrightness.prototype.reset = function() {
+    this._state = 0;
+    if (this._delayTimeout !== null) {
+        clearTimeout(this._delayTimeout);
+    }
+    this._delayTimeout = null;
+    this._previousLux = undefined;
+    this._autoDelayPrevLux = undefined;
+  };
+
+  //
+  // Automatically adjust the screen brightness based on the ambient
+  // light (in lux) measured by the device light sensor
+  // This function uses a state machine to delay before
+  // adjusting the brightness. If the user waves a finger over
+  // the sensor, the display brightness will not be changed.
+  // The state machine (recorded in _autoAdjustState):
+  // state 0: startup state, lux value is stored in _autoDelayPrevLux
+  //          transitions to state 1 if lux delta > AUTO_BRIGHTNESS_MIN_DELTA
+  // state 1: in delay timer, lux value is stored in _autoDelayPrevLux
+  //          transitions to state 2 once time delay has elapsed
+  // state 2: end of delay, use _autoDelayPrevLux value for screen brightness
+  //          transitions to state 0 automatically
+  //
+
+  ScreenAutoBrightness.prototype.autoAdjust = function(lux) {
+    if (lux < 1) { // Can't take the log of 0 or negative numbers
+      lux = 1;
+    }
+    if (typeof this._previousLux === 'undefined') {
+        this.setBrightnessFromLux(lux);
+        return;
+    }
+    // lux change during delay timer
+    if (this._state === 1) {
+      this._autoDelayPrevLux = lux;
+      return;
+    }
+    if (this._state === 0 && 
+        Math.abs(this._previousLux - lux) > this.AUTO_BRIGHTNESS_MIN_DELTA) {
+        // lux delta over the threshold, start a time delay, state 0 -> 1
+        this._state = 1;
+        this._autoDelayPrevLux = lux;
+        this._delayTimeout = setTimeout((function () {
+          // function called when delay has elapsed, state 1 -> 2
+          this._state = 2;
+          this.autoAdjust(lux); // adjust if required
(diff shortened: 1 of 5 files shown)


try-server-hook:
[gpgreen Greg Green (gpgreen)](https://github.com/gpgreen) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=e565a6593f60)

try-server-hook:
[gpgreen Greg Green (gpgreen)](https://github.com/gpgreen) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=9e32d32eeca3)

apps/system/js/screen_auto_brightness.js
```diff
@@ -0,0 +1,136 @@
...
+
+  /*
+   * constructor - argument is the function that will adjust
+   * the screen brightness. This function takes 2 arguments,
+   * the brightness and whether the change is immediate (true),
+   * or will be transitioned (false)
+   */
+  var ScreenAutoBrightness = function ScreenAutoBrightness(
+      setScreenBrightnessFn) {
+    this._setScreenBrightness = setScreenBrightnessFn;
```
timdream: It's not a convention (at least for me personally) to pass a callback function.

I would do

```javascript
ScreenAutoBrightness.prototype.onbrightnesschange = null;
```

and in the calling site, do

```javascript
if (this.onbrightnesschange === 'function') {
  this.onbrightnesschange(clampedBrightness);
} else {
  console.error('ScreenAutoBrightness: Brightness should change but no callback attached.');
}
```

and attach the callback in ScreenManager with

```javascript
this._screenAutoBrightness.onbrightnesschange = function(brightness) {
  this.setScreenBrightness(brightness, false);
}.bind(this);
```

apps/system/js/screen_auto_brightness.js
```diff
@@ -0,0 +1,136 @@
...
+  ScreenAutoBrightness.prototype._autoAdjustState = 0;
+  ScreenAutoBrightness.prototype._autoAdjustDelay = 750;
+
+  /*
+   * keep track of last lux value during autoAdjustBrightness delay's
+   */
+  ScreenAutoBrightness.prototype._autoDelayPrevLux = undefined;
+
+  // reinitialize to start state
+  ScreenAutoBrightness.prototype.init = function() {
```
timdream: This should be `reset()`. You could provide a `start` function that simply do `this.reset()` as well.

The convention is having `start()` and `stop()` being called *once* when the instance is created/destroyed. Since you intend to have ScreenManager to reuse the instance already created, let's not use `start()` and use `reset()`.

`init()` is everywhere but unfortunately it's not well defined in our old code pattern so let's not copy it.

apps/system/js/screen_auto_brightness.js
```diff
@@ -0,0 +1,136 @@
...
+  // state 0: startup state, lux value is stored in _autoDelayPrevLux
+  //          transitions to state 1 if lux delta > AUTO_BRIGHTNESS_MIN_DELTA
+  // state 1: in delay timer, lux value is stored in _autoDelayPrevLux
+  //          transitions to state 2 once time delay has elapsed
+  // state 2: end of delay, use _autoDelayPrevLux value for screen brightness
+  //          transitions to state 0 automatically
+  //
+
+  ScreenAutoBrightness.prototype.autoAdjust = function(lux) {
+    var self = this;
```
timdream: Use `bind()` instead of `self = this` to avoid identity crisis :-P.

apps/system/js/screen_auto_brightness.js
```diff
@@ -0,0 +1,136 @@
...
+    var brightnessDelta;
+    if (lux < 1) { // Can't take the log of 0 or negative numbers
+      lux = 1;
+    }
+    if (this._previousLux !== undefined) {
```
timdream: nit: Let's flip the logic here and early return, to avoid long `if` block?

```javascript
if (typeof this._previousLux === 'undefined') {
  ...
  return;
}
```

You can also create another `_setBrightnessFromLux()` function to avoid duplicate of code in this early return block and the bottom of the function.

apps/system/js/screen_auto_brightness.js
```diff
@@ -0,0 +1,136 @@
...
+      if (this._autoAdjustState === 0) {
```
timdream: ocd nit: I would define the state constants as `ScreenAutoBrightness.prototype.STATE_XXX` and compare/set the state with it, but that's really not important.

apps/system/js/screen_auto_brightness.js
```diff
@@ -0,0 +1,136 @@
...
+        brightnessDelta = Math.abs(this._previousLux - lux);
+        if (brightnessDelta <= this.AUTO_BRIGHTNESS_MIN_DELTA) {
+            return;
+        }
+        // lux delta over the threshold, start a time delay, state 0 -> 1
+        this._autoAdjustState = 1;
+        this._autoDelayPrevLux = lux;
+        setTimeout(function () {
```
timdream: What happen if we want to break the transition? Do we need to save the `timer` here and call `clearTimeout()` accordingly?

This is a really needed since we are moving the set-brightness action from sync to async, and we want to be able to cancel the pending async operation if it's no longer applicable.

apps/system/js/screen_auto_brightness.js
timdream: `switch ... case` maybe?

apps/system/js/screen_auto_brightness.js
```diff
@@ -0,0 +1,136 @@
...
+      }
+      // end of delay, adjust brightness, state 2 -> 0
+      else if (this._autoAdjustState === 2) {
+        this._autoAdjustState = 0;
+        lux = this._autoDelayPrevLux;
+        brightnessDelta = Math.abs(this._previousLux - lux);
+        if (brightnessDelta <= this.AUTO_BRIGHTNESS_MIN_DELTA) {
+          return;
+        }
+        // fall through...
```
timdream: This is the only other place where we need the `_setBrightnessFromLux()` function right?

apps/system/test/unit/screen_auto_brightness_test.js
```diff
@@ -0,0 +1,106 @@
+'use strict';
+
+/* global MocksHelper, MockScreenManager, ScreenAutoBrightness */
+
+requireApp('system/test/unit/mock_screen_manager.js');
+require('/js/screen_auto_brightness.js');
+
+var mocksForAutoBrightness = new MocksHelper([
+  'ScreenManager'
```
timdream: If you are using the `onbrightnesschange` pattern you would not need the mock :)

apps/system/js/screen_auto_brightness.js
timdream: Maybe do `clearTimeout` when `reset()` is called would be sufficient?

apps/system/js/screen_auto_brightness.js
```diff
@@ -0,0 +1,136 @@
...
+
+  /*
+   * keep track of last lux value during autoAdjustBrightness delay's
+   */
+  ScreenAutoBrightness.prototype._autoDelayPrevLux = undefined;
+
+  // reinitialize to start state
+  ScreenAutoBrightness.prototype.init = function() {
+    this._previousLux = undefined;
+    this._autoAdjustState = 0;
```
timdream: nit: this can simply be `state` or `_state`.

apps/system/js/screen_auto_brightness.js
gpgreen: Ok. Works for me.

apps/system/js/screen_auto_brightness.js
gpgreen: Ok

apps/system/js/screen_auto_brightness.js
gpgreen: Ok

apps/system/js/screen_auto_brightness.js
gpgreen: A good idea

apps/system/js/screen_auto_brightness.js
gpgreen: I redid the logic in this function to make it flow better. I didn't use the constants as they just add a lookup for every state retrieval.

apps/system/js/screen_auto_brightness.js
gpgreen: the timer gets canceled in reset() now. I cannot think of a case where it would need to be canceled elsewhere.

apps/system/js/screen_auto_brightness.js
gpgreen: Yes the only place, however they use different values of lux when it is changed, which means creating a new function with no argument doesn't really work. I did create a new function so autoAdjust flows better however.

try-server-hook:
[gpgreen Greg Green (gpgreen)](https://github.com/gpgreen) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=f024db8c9208)

try-server-hook:
[gpgreen Greg Green (gpgreen)](https://github.com/gpgreen) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=cf339229e27b)

apps/system/js/screen_auto_brightness.js
```diff
@@ -0,0 +1,150 @@
...
+      Math.log10(lux) * this.AUTO_BRIGHTNESS_CONSTANT;
+
+    var clampedBrightness = Math.max(this.AUTO_BRIGHTNESS_MINIMUM,
+                                     Math.min(1.0, computedBrightness));
+
+    if (typeof this.onbrightnesschange === 'function') {
+        this.onbrightnesschange(clampedBrightness);
+    } else {
+        console.log(
+'ScreenAutoBrightness: Brightness should change but no callback attached.'
```
timdream: nit: 

```javascript
      console.error('ScreenAutoBrightness: ' +
        'Brightness should change but no callback attached.');
```

try-server-hook:
[gpgreen Greg Green (gpgreen)](https://github.com/gpgreen) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=fbf969956442)

PR #27691, opened 2015-01-26
WIP  Bug-1037650 

kumarrishav:


autolander:

try-server-hook:
[kumarrishav Kumar Rishav (kumarrishav)](https://github.com/kumarrishav) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=219ced8ac274)

apps/sms/style/sms.css
```diff
@@ -37,6 +37,49 @@ gaia-header {
...
+  font-weight: 500;
+}
+
+/*
+ * Right to Left View
+ */
+
+html[dir="rtl"] section[role="toast"],
+html[dir="rtl"] section[role="toast"] p {
+  text-align: right;
```
julienw: you can use "text-align: start" to make it work automagically

apps/sms/style/sms.css
```diff
@@ -37,6 +37,49 @@ gaia-header {
   z-index: 10;
 }
 
+/*
+  toast
+*/
+section[role="toast"] {
```
julienw: "role" is actually an aria role; it needs to have specific values (see http://www.w3.org/TR/wai-aria/roles) and "toast" is not one of them. Moreover it was a mistake that we used these roles in CSS, we should not do that anymore, please use CSS classes instead.

apps/sms/style/sms.css
```diff
@@ -37,6 +37,49 @@ gaia-header {
...
+  color: #ffffff;
+  text-align: left;
+  font-style: italic;
+  font-family: sans-serif;
+}
+
+section[role="toast"] p strong {
+  font-weight: 500;
+}
+
```
julienw: I think we should have a separate CSS file toast.css or undo_toast.css if we decide to do our own css. But I think we should use the css for the existing role="notification" with specific classes to override what we need to override, it should make it easier to update it system-wide. For example the background color could be updated in one place.

apps/sms/js/thread_ui.js
```diff
@@ -2100,14 +2100,11 @@ var ThreadUI = {
           {
             l10nId: 'delete',
             method: function deleteMessage(messageId) {
-              Utils.confirm(
-                'deleteMessage-confirmation', null,
-                { text: 'delete', className: 'danger' }
-              ).then(() => {
```
julienw: actually I'm not sure we should remove the confirmation; I think we should wait for the undo mechanism to be implemented :)

apps/sms/js/thread_list_ui.js
```diff
@@ -364,8 +449,17 @@ var ThreadListUI = {
         MessageManager.markThreadRead(thread.id, isRead);
       }
     });
-
+
```
julienw: nit: space

apps/sms/js/thread_list_ui.js
```diff
@@ -282,6 +288,84 @@ var ThreadListUI = {
         }
 
         break;
+      case 'contextmenu':
```
julienw: tell me if I'm wrong, but I think this was removed from Omega's proposal ?

apps/sms/js/thread_list_ui.js
```diff
@@ -340,6 +424,7 @@ var ThreadListUI = {
   },
 
   markReadUnread: function thlui_markReadUnread() {
+    var l10nKey1;
```
julienw: nit: why "1" ? :)

apps/sms/js/thread_list_ui.js
```diff
@@ -364,8 +449,17 @@ var ThreadListUI = {
         MessageManager.markThreadRead(thread.id, isRead);
       }
     });
-
+    
     this.cancelEdit();
+
+    isRead ? l10nKey1 = 'marked-as-read' : l10nKey1 = 'marked-as-unread';
```
julienw: nit: better write it: `l10nKey = isRead ? 'marked-as-read' : 'marked-as-unread';`

apps/sms/js/thread_list_ui.js
```diff
@@ -408,89 +502,92 @@ var ThreadListUI = {
...
-      }
+    promise.then(function() {
+      ThreadListUI.onDeletionReadUnread({
+        l10nId: 'deleted',
+        l10nArgs: {
+          n: selection.selectedCount
+        }
+      });
+    });
+  },
```
julienw: I don't understabd why you use a promise here :)

julienw: ok, I think I understand what you want to do:
* you want to run performDeletion
* once it's finished, you want to run onDeletioNReadUnread

but this doesn't work.

What you really need is that performDeletion returns a promise :)

julienw: To be clearer: your current code does this:
* run performDeletion -- note that performDeletion is asynchronous, so it starts all getMessages and then returns
* when performDeletion returns, you call "resolve". That means the promise will be resolved.
* so when you run "promise.then()", this will be executed (asynchronously) just now.

And that's not really what we want :)

Remember we can't really undo deletion. So what we want to do is _delay_ deletion while the toast is displayed, but still make it appear as if it's really deleted. That's why I proposed a "UndoManager" when we discussed on IRC.

I think we need to do this UndoManager first, and only then apply it for delete/mark as read.

How about working on a undo for the draft saving operation? We already have a toast for this, and we can simply update it later. Then we can come back at this bug when we'll have a working UndoManager for draft saving.

What do you think?

apps/sms/js/thread_list_ui.js
julienw: ah sorry, I just saw it's part of the other bug ;)

PR #27704, opened 2015-01-27
Bug 1124758 - [Messages] We should not add the hidden (+0) text at the end of the header. r=julien

azasypkin:


try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=01591bbb6afa)

apps/sms/js/thread_ui.js
```diff
@@ -1387,26 +1389,38 @@ var ThreadUI = {
...
    */
-  setHeaderContent: function thui_setHeaderContent(content) {
-    if (typeof content === 'string') {
-      this.headerText.removeAttribute('data-l10n-id');
-      this.headerText.removeAttribute('data-l10n-args');
-
-      this.headerText.innerHTML = content;
-    } else {
+  setHeaderContent: function thui_setHeaderContent(contentL10n) {
+    if (typeof contentL10n === 'string') {
```
azasypkin: Previously l10n guys advised to use approach [1] for ```setHeaderContent```, but patch was landed earlier, so I use the chance to unify this now. Also we already use the same approach in ```dialog.js```

[1] https://developer.mozilla.org/en-US/Firefox_OS/Developing_Gaia/localization_code_best_practices#Writing_APIs_that_operate_on_L10nIDs

apps/sms/js/thread_ui.js
```diff
@@ -1387,26 +1389,38 @@ var ThreadUI = {
    * markup to support bidirectional content, but other panels still use it with
    * mozL10n.setAttributes as it would contain only localizable text. We should
    * get rid of this method once bug 961572 and bug 1011085 are landed.
-   * @param {string|{ id: string, args: Object }} content Should be either safe
-   * HTML string or object with l10nId and l10nArgs.
+   * @param {string|{ raw: string }|{ html: string }|
```
julienw: how about keeping "id" so that we always get an object as parameter? Or even "l10nId" so that it's really clear to read at the caller site

apps/sms/js/thread_ui.js
julienw: mmm ok; but as I see you could merge the "string parameter" and the `{id: }` parameter cases because they're essentially the same.

apps/sms/js/thread_ui.js
```diff
@@ -1366,14 +1366,16 @@ var ThreadUI = {
...
-        }));
+        if (thread.participants.length > 1) {
+          this.setHeaderContent({
+            html: this.tmpl.header.interpolate({
+              name: contactName,
+              participantCount: (thread.participants.length - 1).toString()
+            })
+          });
+        } else {
+          this.setHeaderContent({ raw: contactName });
```
julienw: note that the contactName can be the number; I think we need the `<bdi>` in that case, right ? Otherwise numbers starting with "+" will have the "+" badly positioned in RTL ?

The best way IMO is to use 2 different templates.

azasypkin: Ah, yeah, right, nice catch!

apps/sms/js/thread_ui.js
azasypkin: Sure, I can join these two cases.  I remember some time ago someone from l10n team said that its better to use ```setAttributes``` only when we have l10nArgs, but anyway I think like you here :)

apps/sms/js/thread_ui.js
julienw: mmm looks like it works now because header is always in LTR but it's not future proof :) what do you think?

azasypkin: I think you're right and having ```bdi``` will be better

try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=260a1e53b8d5)

try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=b58fcc27d402)

PR #27706, opened 2015-01-27
Bug 1115326 - Don't allow entering edit mode when launching a collection. r=kgrandon

Cwiiis:


try-server-hook:
[Cwiiis Chris Lord (Cwiiis)](https://github.com/Cwiiis) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=bf2230eda518)

shared/elements/gaia_grid/js/grid_dragdrop.js
```diff
@@ -666,6 +672,12 @@
               return;
             }
 
+            if (this.gridView._launchingApp) {
```
KevinGrandon: Is it possible to set some flag on touchend and rely on that instead? The current debounce mechanism we have is a giant hack and should probably go away. Would prefer to avoid hooking into it if possible.

Cwiiis: The problem is I don't think we have a reliable way to know when the collection is opened or closed... If someone told me that collection-launch and collection-close were 100% reliable, I expect we could use those, though it would still need a mechanism similar to this flag, as events can still go to the homescreen after the collection-launch signal (which is the essence of this bug).

I guess an alternative would be to just stick an event-blocking layer over the grid when apps are launching, but I have a strong feeling it'd give users the feeling that the device has crashed/locked up in the worst case scenarios... And the DOM is crowded enough as it is :/ That said, it could probably be done 100% in CSS, maybe that would be the way to go...

Cwiiis: Actually, looking again, maybe launch/close are 100% reliable. I think I should restructure this around that...

KevinGrandon: It should be reliable, but if it's easier to leave as-is, I'm fine with that. Thanks for looking.

try-server-hook:
[Cwiiis Chris Lord (Cwiiis)](https://github.com/Cwiiis) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=7525ed753a38)

PR #27474, opened 2015-01-18
Bug 1121766 - Intermittent failing test, TEST-UNEXPECTED-FAIL | apps/sms/test/marionette/composer_test.js | Messages Composer Messages Composer Test Suite Message char counter and MMS label. r=julien

azasypkin:

PR #27710, opened 2015-01-27
Bug 1104961 - Intermittent share_activity_test.js | Messages as share target Share via Messages Activity close button Should return to Thread panel if in Participants panel. Do not use elements that maybe inside activity that is being disposed. r=schung

azasypkin:


try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=58f5e9165e49)

try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=5fc60c72e643)

try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=e9c0902b8de8)

apps/sms/test/marionette/lib/messages.js
```diff
@@ -237,13 +237,21 @@
...
 
+          // Since recipient.js re-renders recipients all the time (when new
+          // recipient is added or old is removed) and it can happen several
+          // times (!!!) during single "add" or "remove" operation we should
+          // wait until Recipients View is in a final state. The problem here is
+          // that between "findElement" and "displayed" calls element can
+          // actually be removed from DOM and re-created again that will lead to
+          // "stale element" exception. Eventually real fix should be introduced
+          // in recipients.js to avoid wasteful rendering.
+          client.helper.wait(500);
```
evanxd: Intermittent failures might probably happen due to the 500 ms waiting. Something the test might need more than 500 ms to wait for running correctly.

Looks like we could use `searchTimeout`[1] to avoid this issue.

[1]: https://github.com/mozilla-b2g/gaia/blob/master/apps/calendar/test/marionette/lib/calendar.js#L16

azasypkin: > Intermittent failures might probably happen due to the 500 ms waiting. Something the test might need more than 500 ms to wait for running correctly.

Yeah, I'm looking into other solutions right now, the case we have is a bit complex to avoid though.

> Looks like we could use searchTimeout1 to avoid this issue.

Nope, the problem here is that we get one element with ```findElement```, but at the time when we retrieve its ```displayed``` property it's already removed from the DOM and replaced with exaclty the same but _new_ element. Such replacement leads to intermittent ```stale element``` exception.

try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=47412808b8ff)

apps/sms/test/marionette/lib/messages.js
```diff
@@ -237,16 +237,38 @@
...
+          var numberOfRecipientsChangeEvents = 5;
+
+          // Since recipient.js re-renders recipients all the time (when new
+          // recipient is added or old is removed) and it can happen several
+          // times during single "add" or "remove" operation we should
+          // wait until Recipients View is in a final state. The problem here is
+          // that between "findElement" and "displayed" calls element can
+          // actually be removed from DOM and re-created again that will lead to
+          // "stale element" exception.
+          client.executeScript(function(numberOfEvents) {
```
azasypkin: @steveck-chung I know that it looks quite scary :) But can't think of another hook we can use other than timeout. Maybe you have better ideas?

try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=aec5a5ea89dc)

shared/test/integration/tbpl-manifest.json
```diff
@@ -33,8 +33,6 @@
     "apps/settings/test/marionette/tests/app_permission_settings_test.js": "Bug 1004348 - Disable intermittent failing test, manipulate app permissions set geolocation of first app to Grant",
     "apps/settings/test/marionette/tests/hotspot_wifi_settings_test.js": "Bug 1006527 - Disable intermittent failing travis test, manipulate hotspot wifi settings resets settings for second load",
     "apps/settings/test/marionette/tests/message_settings_test.js": "",
-    "apps/sms/test/marionette/composer_test.js": "Bug 1121766 - TEST-UNEXPECTED-FAIL | Messages Composer Messages Composer Test Suite Message char counter and MMS label",
```
evanxd: Sorry, why do we enable `composer_test.js` here? Does the patch fix that issue? But I saw some failures of running `composer_test.js` in [1]. And it is also an intermittent failures.

[1]: https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=aec5a5ea89dc

apps/sms/test/marionette/lib/messages.js
evanxd: Got it.
Thanks for the explanation of root cause.

shared/test/integration/tbpl-manifest.json
azasypkin: Sorry for the confusion, it definitely won't go to the final patch, I just wanted to see if it helps for composer_test intermittent failures as well - yes it helps, but not for everything. So composer_test will be enabled in bug 1121766 with additional change (see https://github.com/mozilla-b2g/gaia/pull/27474 and https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=db4af079b192)

apps/sms/test/marionette/lib/messages.js
```diff
@@ -9,6 +9,26 @@
...
   };
 
+  function observeElementStability(el) {
+    delete el.dataset.__stable;
+
+    function markElementAsStable() {
+      return setTimeout(function() {
+        el.dataset.__stable = 'true';
+        observer.disconnect();
+      }, 1000);
```
steveck-chung: Just wondering if it's necessary to wait for 1sec timeout for stable?

azasypkin: mmm, there can be several changes (eg. recipients is re-rendered 4 times for a single "add" operation) in a row with unknown time between. Maybe I'm just missing your idea, could you please elaborate more?

try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=4e9974a45db3)

try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=0020c744495a)

apps/sms/test/marionette/lib/messages.js
steveck-chung: ya I know there might be several changes, but doesn't each change trigger the observer callback? And each callback will reset the timeout again. Or the observer only callback once even we add 4 recipients?

azasypkin: > ya I know there might be several changes, but doesn't each change trigger the observer callback? And each callback will reset the timeout again. 

Yeah that's right, so the idea is to "debounce" marking element as stable. That's how I think of it - If change is occured, let's cancel (clearTimeout) previous intention to mark element as stable and wait for 1 sec one more time, if change is occurred once again (ex. after 900 ms) - that means the element isn't stable enough, so there is a chance that it will be changed again, so we cancel old timeout (~100ms left to mark element as stable) and again wait for 1sec.

So depending on the environment "speed" and length of string that we send with ```sendKeys``` there can happen different number of mutations just for one "addRecipient". Also there's a chance that MutationObserver call won't fire at all, that's why I call ```markElementAsStable``` at the very beginning.

PR #27722, opened 2015-01-28
Bug 1068998 - Label video player buttons.

eeejay:


try-server-hook:
[eeejay Eitan Isaacson (eeejay)](https://github.com/eeejay) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=237f5f87687b)

shared/locales/media/media.en-US.properties
```diff
@@ -0,0 +1,3 @@
+playbackPlay.ariaLabel=Play
+playbackPause.ariaLabel=Pause
+playbackFullscreen=Fullscreen
```
davidflanagan: I think you're missing .ariaLabel for the fullscreen line, right?

shared/locales/media/media.en-US.properties
```diff
@@ -0,0 +1,3 @@
+playbackPlay.ariaLabel=Play
```
davidflanagan: I did not know that you could do this with localization. Nice feature!

Is it normal to leave 'playbackPlay' completely undefined in this case, or would it be better to explicitly set it to the empty string?  I assume that the way you have it is best, but just wanted to check since I have not seen this done before.

shared/js/media/video_player.js
```diff
@@ -36,9 +39,11 @@ function VideoPlayer(container) {
   var poster = newelt(container, 'img', 'videoPoster');
   var player = newelt(container, 'video', 'videoPlayer');
   var controls = newelt(container, 'div', 'videoPlayerControls');
-  var playbutton = newelt(controls, 'button', 'videoPlayerPlayButton');
+  var playbutton = newelt(controls, 'button', 'videoPlayerPlayButton',
+                    'playbackPlay');
   var footer = newelt(controls, 'div', 'videoPlayerFooter hidden');
-  var pausebutton = newelt(footer, 'button', 'videoPlayerPauseButton');
+  var pausebutton = newelt(footer, 'button', 'videoPlayerPauseButton',
+                     'playbackPause');
```
davidflanagan: Nit: please fix the indentation here and above so that the wrapped argument lines up with the space after the open paren.  There's the same issue below, and I'd appreciate it if you'd fix that indentation as well, even though that already existed in the code.

shared/js/media/video_player.js
```diff
@@ -48,7 +53,7 @@ function VideoPlayer(container) {
   var durationText = newelt(slider, 'span', 'videoPlayerDurationText');
   // expose fullscreen button, so that client can manipulate it directly
   var fullscreenButton = newelt(slider, 'button',
-                          'videoPlayerFullscreenButton');
+                          'videoPlayerFullscreenButton', 'playbackFullscreen');
```
davidflanagan: I've never even seen the fullscreen button displayed. I think that is only used on tablets. I suppose it is worth fixing, but your patch may be untestable unless you have a tablet.

apps/gallery/index.html
```diff
@@ -20,6 +20,7 @@
     <meta name="availableLanguages" content="ar, en-US, fr, zh-TW">
     <link rel="localization" href="locales/gallery.{locale}.properties">
     <link rel="localization" href="/shared/locales/date/date.{locale}.properties">
+    <link rel="localization" href="/shared/locales/media/media.{locale}.properties">
```
davidflanagan: It seems like a good idea to make these shared since we've got very similar audio controls in a bunch of places. And different video controls in the video app.  I wonder if calling it media/controls.{locale}.properties of playback.{locale}.properties would be better, though, so it could be focused just on the controls labels.

shared/locales/media/media.en-US.properties
eeejay: This is the standard way to use it. You could localize other attributes like title and alt in the same way.

shared/locales/media/media.en-US.properties
eeejay: Yes, good catch!

try-server-hook:
[eeejay Eitan Isaacson (eeejay)](https://github.com/eeejay) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=9cea88e13e4d)

try-server-hook:
[eeejay Eitan Isaacson (eeejay)](https://github.com/eeejay) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=f464a2a91f21)

PR #27729, opened 2015-01-28
Bug 1126566 - [email] Account settings layout regression

jrburke:
Three main changes:
- Reduced the .list-pack-switch and .list-pack-checkbox margin-end spacing to 7.5rem, it was set much larger than it needed to be. I believe this was just a holdout from the older non-flexbox design, and it wasn't really seen given that for shorter translations it did not make a difference.
- For .list-text inside a  .select-item, it had a bigger right padding to account for a chevron, but since it is in a .select-item, not an item-with-children (which gets the chevron), it did not need to reserve that space.
- The .item-with-children makes more sense to apply as an li style instead of a child of that li (the chevron logically belongs to the li) and by doing so allowed targeted style targets for .list-text children inside of .item-with-children. This helped fixed crazy line-heights for items with chevrons, there we large gaps between the text (only visible in non-en-US locales with longer text).

Tested with French and RTL locale to confirm changes look good. Here is a screenshot of the French one now:

a-top

a-bottom

### DIFF
--- a/apps/email/js/cards/settings_account.html
+++ b/apps/email/js/cards/settings_account.html
@@ -124,10 +124,10 @@
     </header>
     <ul data-prop="serversContainer" data-l10n-id="settings-server-group"
         class="tng-list tng-account-server-container">
-      <li>
+      <li class="item-with-children">
         <a href="#" data-prop="accountCredNode"
                     data-event="click:onClickCredentials"
-          class="tng-account-credentials list-text item-with-children"></a>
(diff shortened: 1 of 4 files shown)


try-server-hook:
[jrburke James Burke (jrburke)](https://github.com/jrburke) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=b3652f9f70ed)

PR #27754, opened 2015-01-28
Bug 1037495: Factor out common telemetry code to bring AppUsageMetrics URL/data up to snuff

marshall:
https://bugzilla.mozilla.org/show_bug.cgi?id=1037495

### DIFF
--- a/apps/sharedtest/test/unit/telemetry_test.js
+++ b/apps/sharedtest/test/unit/telemetry_test.js
@@ -1,0 +1,125 @@
+'use strict';
+
+/* global TelemetryRequest */
+
+require('/shared/js/telemetry.js');
+
+suite('Telemetry:', function() {
+  var xhr, XHR;
+
+  suiteSetup(function() {
+    XHR = sinon.useFakeXMLHttpRequest();
+    XHR.onCreate = function(instance) { xhr = instance; };
+  });
+
+  suiteTeardown(function() {
+    XHR.restore();
+  });
+
+  test('ctor required args', function() {
+    var x;
+    assert.throws(function() { x = new TelemetryRequest(); });
+    assert.throws(function() { x = new TelemetryRequest({ deviceID: 'x' }); });
+    assert.throws(function() { x = new TelemetryRequest({ reason: 'x' }); });
+
+    assert.doesNotThrow(function() {
+      x = new TelemetryRequest({ reason: 'x', deviceID: 'x' });
+    });
+  });
+
+  suite('request args', function() {
+    var req;
+    setup(function() {
+      req = new TelemetryRequest({
+        deviceID: 1,
+        reason: 'a reason',
+        ver: 3,
+        appName: '+name',
+        appUpdateChannel: 'channel',
+        appBuildID: 'buildID',
+        appVersion: 'version',
+        url: 'http://foobar'
+      });
+    });
+
+    test('sets packet version', function() {
+      assert.equal(req.packet.ver, 3);
+    });
+
+    test('fills info', function() {
+      assert.ok(req.info);
+      assert.equal(req.info.reason, 'a reason');
+      assert.equal(req.info.appName, '+name');
+      assert.equal(req.info.appUpdateChannel, 'channel');
+      assert.equal(req.info.appVersion, 'version');
+      assert.equal(req.info.appBuildID, 'buildID');
+    });
+
+    test('builds proper URL', function() {
+      assert.equal(req.url,
+        'http://foobar/1/a%20reason/%2Bname/version/channel/buildID');
+    });
+  });
+
+  test('copies info', function() {
+    var req = new TelemetryRequest({
+      deviceID: 1,
+      reason: 'reason'
+    }, { x: '1', y: '2' });
+
+    assert.deepEqual(req.info, req.packet.info);
+    assert.equal(req.info.x, '1');
+    assert.equal(req.info.y, '2');
+  });
+
+  suite('send', function() {
+    var req, clock;
+    setup(function() {
+      req = new TelemetryRequest({
+        deviceID: 1,
+        reason: 'reason',
+        url: 'http://foobar'
+      }, { x: '1', y: '2' });
+
+      clock = this.sinon.useFakeTimers();
+    });
+
+    test('sends proper JSON request', function() {
+      req.send();
+
+      assert.ok(xhr);
+      assert.equal(xhr.method, 'POST');
+
+      var url = new URL(xhr.url);
+      assert.equal(url.protocol, 'http:');
+      assert.equal(url.hostname, 'foobar');
+      assert.ok(url.pathname.indexOf('/1/reason') === 0);
+
+      var payload = JSON.parse(xhr.requestBody);
+      assert.ok(payload);
+      assert.deepEqual(payload, req.packet);
+    });
+
+    test('calls onload', function() {
+      var loaded = sinon.spy();
+      req.send({
+        onload: loaded
+      });
+
+      xhr.respond(200, 'OK');
+      assert.ok(loaded.called);
+    });
+
+    test('sets error callbacks', function() {
+      var xhr = req.send({
+        ontimeout: function(){},
+        onerror: function(){},
+        onabort: function(){}
+      });
+
+      assert.ok(xhr.ontimeout);
+      assert.ok(xhr.onerror);
+      assert.ok(xhr.onabort);
+    });
+  });
+});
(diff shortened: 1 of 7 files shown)


try-server-hook:
[marshall Marshall Culpepper (marshall)](https://github.com/marshall) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=d66aa82b11e0)

apps/system/js/app_usage_metrics.js
```diff
@@ -653,7 +665,15 @@
         oldMetrics.merge(self.metrics);
         self.metrics = oldMetrics;
         self.metrics.save(true);
-      };
+      }
+
+      request.sendRequest({
+        timeout: AUM.REPORT_TIMEOUT,
+        onload: onload,
+        onerror: retry,
```
mreid-moz: Note that on 4XX, you should not retry the same request - it will fail again :)  It looks like the retry function saves data locally to be rolled in to the next submission, so this may not be an issue.

apps/system/js/ftu_ping.js
```diff
@@ -25,7 +25,8 @@
   const DEFAULT_PING_TIMEOUT = 60 * 1000;
 
   // Used by the telemetry server to help identify the payload format
-  const PING_PAYLOAD_VERSION = 3;
+  const TELEMETRY_VERSION = 1;
```
mreid-moz: It looks like the common code is now inserting the "info" section as described in comment 1 of bug 1037495 (for both the app usage ping and the ftu ping).  Is that right?  If that's the case, then setting this version back to 1 is correct.  If the ftu pings will continue to look like they did in bug 987232, the version should stay at 3.

apps/sharedtest/test/unit/telemetry_test.js
```diff
@@ -0,0 +1,124 @@
...
+      clock = this.sinon.useFakeTimers();
+    });
+
+    test('sends proper JSON request', function() {
+      req.sendRequest();
+
+      assert.ok(xhr);
+      assert.equal(xhr.method, 'POST');
+
+      var path = xhr.url.substring('http://foobar'.length+1).split('/');
```
fabricedesre: nit: spaces around the '+' operator

apps/system/js/ftu_ping.js
marshall: Yes, that's correct!

apps/system/js/app_usage_metrics.js
```diff
@@ -301,8 +310,7 @@
         }
         else {
           // Our device id does not need to be unique, just probably unique.
-          // And it doesn't even need to be a real UUID
-          self.deviceID = Math.random().toString(36).substring(2, 10);
+          self.deviceID = uuid();
           asyncStorage.setItem(DEVICE_ID_KEY, self.deviceID);
```
davidflanagan: You need to update DEVICE_ID_KEY to a new string to convert v2.1 devices to the new uuid format. Otherwise old devices will continue to report with a non uuid. Or you need to check that the value looks like a uuid (probably by checking that it has hyphens in it).

I'm assuming that you want to force everyone to use a uuid, even at the cost of losing continuity with the old style ids.

Or do something fancy like send oldDeviceID for devices that were upgraded.

shared/js/telemetry.js
```diff
@@ -0,0 +1,106 @@
...
+      args.url || DEFAULT_URL,
+      encodeURIComponent(args.deviceID),
+      encodeURIComponent(this.info.reason),           // 'reason'
+      encodeURIComponent(this.info.appName),          // 'appName'
+      encodeURIComponent(this.info.appVersion),       // 'appVersion'
+      encodeURIComponent(this.info.appUpdateChannel), // 'appUpdateChannel'
+      encodeURIComponent(this.info.appBuildID)        // 'appBuildID'
+    ];
+
+    this.url = uriParts.join('/');
```
davidflanagan: This would be a good place for an ES6 template string if you want to give those a try.

shared/js/telemetry.js
```diff
@@ -0,0 +1,106 @@
...
+    this.info.appVersion = args.appVersion || UNKNOWN;
+
+    var uriParts = [
+      args.url || DEFAULT_URL,
+      encodeURIComponent(args.deviceID),
+      encodeURIComponent(this.info.reason),           // 'reason'
+      encodeURIComponent(this.info.appName),          // 'appName'
+      encodeURIComponent(this.info.appVersion),       // 'appVersion'
+      encodeURIComponent(this.info.appUpdateChannel), // 'appUpdateChannel'
+      encodeURIComponent(this.info.appBuildID)        // 'appBuildID'
```
davidflanagan: I don't get what the comments on these lines are telling me. The quote marks make it look like the reason field will contain the literal string 'reason'.

shared/js/telemetry.js
```diff
@@ -0,0 +1,106 @@
...
+    ];
+
+    this.url = uriParts.join('/');
+  }
+
+  const TR = TelemetryRequest;
+
+  // Set to true to enable debugging
+  TR.DEBUG = false;
```
davidflanagan: You could actually put the const and this DEBUG property up at the top of the file, where it would be easier to find and toggle for debugging.

shared/js/telemetry.js
```diff
@@ -0,0 +1,106 @@
...
+      args.unshift('[Telemetry]');
+      console.log.apply(console, args);
+    }
+  }
+
+  // Open and send the underlying XMLHttpRequest
+  // xhrAttrs: attributes to set on the XHR
+  //   - timeout: The timeout for the request
+  //   - onload / onerror / onabort / ontimeout: callbacks
+  TelemetryRequest.prototype.sendRequest = function(xhrAttrs) {
```
davidflanagan: I notice that when you use this method, your code looks like |request.sendRequest|.  Consider renaming the method to send() instead of the more verbose sendRequest. That would also more closely mirror XHR.

shared/js/telemetry.js
```diff
@@ -0,0 +1,106 @@
...
+    var xhr = new XMLHttpRequest({ mozSystem: true, mozAnon: true });
+
+    xhr.open('POST', this.url);
+    debug(this.url);
+
+    if (xhrAttrs && xhrAttrs.timeout) {
+      xhr.timeout = xhrAttrs.timeout;
+      delete xhrAttrs.timeout;
```
davidflanagan: You're deleting a property passed in by the client. The client app might have a static object that it passes to multiple calls to sendRequest(), and if it does that, only the first such call will have a timeout!

shared/js/telemetry.js
```diff
@@ -0,0 +1,106 @@
...
+
+    var data = JSON.stringify(this.packet);
+    xhr.send(data);
+    debug(data);
+
+    if (xhrAttrs) {
+      for (var key in xhrAttrs) {
+        xhr[key] = xhrAttrs[key];
+      }
+    }
```
davidflanagan: I see why you're deleting the timeout now. I'd say that you're getting in trouble by being too general here. The comment on the method mentions four possible properties. I'd just check for each of the event handlers individually and set them if defined.

autolander:

marshall:
@davidflanagan thanks for all the feedback, I'm using a different string for the DEVICE_ID_KEY to simply force a new UUID to be created for v2.1->v2.2 users, and no longer deleting the timeout key, amongst other suggestions. Should I mark you for review again on the bug, or is this sufficient to change the existing r- to r+?

try-server-hook:
[marshall Marshall Culpepper (marshall)](https://github.com/marshall) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=be559f342c61)

autolander:

try-server-hook:
[marshall Marshall Culpepper (marshall)](https://github.com/marshall) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=839dec3c57e2)

autolander:

try-server-hook:
[marshall Marshall Culpepper (marshall)](https://github.com/marshall) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=ec6c0c584128)

autolander:

try-server-hook:
[marshall Marshall Culpepper (marshall)](https://github.com/marshall) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=e97ea418a0e0)

autolander:

PR #27771, opened 2015-01-29
Bug 1115311 - [LockScreen] (State-Component) Land Clock widget in the current codebase

snowmantw:
- Clock now is a widget.
- Component is the common resource keeper, while State is the resource
  controller.
- No more time-varying if...else. Compose states & components instead.
- Process sequentialize the async & sync operations, while Stream add
  the ability to handle events & interrupts on it.
- New abstraction layer (Source) for a uniform way to collect events
  from timer, clock, mozSettings and DOM events.


try-server-hook:
[snowmantw Greg Weng (snowmantw)](https://github.com/snowmantw) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=e614f714d039)

PR #27785, opened 2015-01-29
Bug 1108708 - change padding for RTL in import contacts page

autra:


apps/ftu/style/style.css
```diff
@@ -1197,3 +1197,8 @@ html[dir="rtl"] #configure_network_params .pack-checkbox {
   left: unset;
   right: auto;
 }
+
+html[dir="rtl"] #import_contacts li button.icon {
+    -moz-padding-start: unset;
```
arcturus: Is necessary to unset it? (Just curious)

try-server-hook:
[autra Augustin Trancart (autra)](https://github.com/autra) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=b6be154da4a9)

apps/ftu/style/style.css
autra: It might not be useful if the line is not very long, but we don't have any -moz-padding-end in LTR. So I think it is better to not have any -moz-padding-start in RTL, for the sake of consistency at least. I reckoned that leaving the start padding would bring the possibility to have some RTL-specific bugs...

PR #27791, opened 2015-01-29
Bug 966228 - Email notification with no subject displays subject as 'nul...

rsajdok:
...l'

### DIFF
--- a/apps/email/js/sync.js
+++ b/apps/email/js/sync.js
@@ -75,6 +75,8 @@ define(function(require) {
           });
         }
 
+        title = title || mozL10n.get('notification-no-subject');
+
         var notification = new Notification(title, notificationOptions);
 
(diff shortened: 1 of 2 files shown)


apps/email/js/sync.js
```diff
@@ -75,6 +75,8 @@ define(function(require) {
           });
         }
 
+        title = title || mozL10n.get('new-message-no-subject');
```
jrburke: Lets call it notification-no-subject so that the key indicates where/how it is used, and update the LOCALIZATION NOTE comment to mention that this is used for notifications.

try-server-hook:
[rsajdok Robert Sajdok (rsajdok)](https://github.com/rsajdok) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=162c696be49f)

PR #27797, opened 2015-01-29
Bug 1068976 - Give videos and images basic descriptions in fullscreen gallery.

eeejay:


apps/gallery/js/frames.js
```diff
@@ -444,7 +444,8 @@ function setupFrameContent(n, frame) {
         frame.displayVideo(videofile, imagefile,
                            fileinfo.metadata.width,
                            fileinfo.metadata.height,
-                           fileinfo.metadata.rotation || 0);
+                           fileinfo.metadata.rotation || 0,
+                           fileinfo.date);
```
davidflanagan: You shouldn't have to pass date in with an extra parameter. videofile.lastModifiedDate should be the same thing. We get the date from the filesystem, not from file metadata, so just using the File property is probably the easiest way to go here and it avoids modifying the existing API.

apps/gallery/js/frames.js
```diff
@@ -455,7 +456,8 @@ function setupFrameContent(n, frame) {
         fileinfo.metadata.height,
         fileinfo.metadata.preview,
         fileinfo.metadata.rotation,
-        fileinfo.metadata.mirrored);
+        fileinfo.metadata.mirrored,
+        fileinfo.date);
```
davidflanagan: Same comment here about not needing a new argument.

shared/js/media/media_frame.js
```diff
@@ -107,7 +108,8 @@ MediaFrame.prototype.displayImage = function displayImage(blob,
                                                           height,
                                                           preview,
                                                           rotation,
-                                                          mirrored)
+                                                          mirrored,
+                                                          timestamp)
```
davidflanagan: If the blob is a file, you can get the timestamp from it. If the blob is not a file, then we won't have a meaningful timestamp anyway.

shared/js/media/media_frame.js
```diff
@@ -137,6 +139,26 @@ MediaFrame.prototype.displayImage = function displayImage(blob,
   // It would be nice if users of this module could override this
   // background color.
   this.image.style.backgroundColor = '#222';
+  this.image.setAttribute('role', 'img');
+
+  this._localize = function _localize() {
```
davidflanagan: I'd prefer to have this code directly in the localize() method. Maybe define this.orientation here in displayImage(), and then put all of this localization code in MediaFrame.prototype.localize(). If you get the date from this.imageblob.lastModifiedDate (or whatever the File property is) and have the orientation, then I think the rest of this code does not need to be here.

shared/js/media/media_frame.js
```diff
@@ -371,6 +393,11 @@ MediaFrame.prototype._displayImage = function(backgroundImage, width, height) {
   var temp = this.image.clientLeft; // jshint ignore:line
 };
 
+MediaFrame.prototype.localize = function localize() {
+  if (this._localize) {
```
davidflanagan: Please just put the accessibliity code directly here. You may need to use this.displayingImage or this.displayingVideo to distinguish the two cases.  And maybe add a comment that this localization code is really just localizing the accessability information.

shared/js/media/media_frame.js
```diff
@@ -137,6 +139,26 @@ MediaFrame.prototype.displayImage = function displayImage(blob,
...
+    if (timestamp) {
+      var locale_entry = navigator.mozL10n.get('imageDescription',
+        { orientation: navigator.mozL10n.get(orientation) });
+      label = MediaUtils.formatDate(timestamp, locale_entry);
+    } else {
+      label = navigator.mozL10n.get('imageDescriptionNoTimestamp',
+        { orientation: navigator.mozL10n.get(orientation) });
+    }
+    this.image.setAttribute('aria-label', label);
+  };
```
davidflanagan: This is shared code you're modifying here, also used by the Camera app, so you'll want to modify that app to be sure these new l10n strings are defined for that app as well.

Also, since it is shared, you should probably write this code to be robust in the case that the get() calls return null.  Line 151 should be protected by if (locale_entry) and line 156 should be protected by if (label), I think.

shared/js/media/media_frame.js
```diff
@@ -1,5 +1,5 @@
 'use strict';
-/* global Downsample, VideoPlayer */
+/* global Downsample, VideoPlayer, MediaUtils */
```
davidflanagan: Please ensure that this new dependency does not break the Camera app which also uses media_frame.js

shared/js/media/video_player.js
```diff
@@ -1,5 +1,5 @@
 'use strict';
-/* global Format */
+/* global Format, MediaUtils */
```
davidflanagan: The camera app also uses this, so make sure it doesn't break.

shared/js/media/video_player.js
```diff
@@ -36,9 +39,11 @@ function VideoPlayer(container) {
   var poster = newelt(container, 'img', 'videoPoster');
   var player = newelt(container, 'video', 'videoPlayer');
   var controls = newelt(container, 'div', 'videoPlayerControls');
-  var playbutton = newelt(controls, 'button', 'videoPlayerPlayButton');
+  var playbutton = newelt(controls, 'button', 'videoPlayerPlayButton',
+                    'playbackPlay');
```
davidflanagan: As a general thing, I'd prefer wrapped lines like this to be indented so that arguments align with the first argument. That's what my editor does by default, and it looks really weird to me when code does not do that.

shared/js/media/video_player.js
```diff
@@ -491,6 +502,25 @@ function VideoPlayer(container) {
...
+    if (videotimestamp) {
+      var locale_entry = navigator.mozL10n.get('videoDescription',
+        { orientation: navigator.mozL10n.get(orientation) });
+      label = MediaUtils.formatDate(videotimestamp, locale_entry);
+    } else {
+      label = navigator.mozL10n.get('videoDescriptionNoTimestamp',
+        { orientation: navigator.mozL10n.get(orientation) });
+    }
+    poster.setAttribute('aria-label', label);
+  }
```
davidflanagan: For backward compatibility, please write this code so that it works even if these strings are not defined in the locale files.

shared/locales/media/media.en-US.properties
```diff
@@ -0,0 +1,13 @@
...
+videoDescription = {{orientation}} video, %I:%M %p, %B %e, %Y
+videoDescriptionNoTimestamp = {{orientation}} video
+
+orientationPortrait = portrait
+orientationLandscape = landscape
+
+# Video controls
+playbackPlay.ariaLabel=Play
+playbackPause.ariaLabel=Pause
+playbackFullscreen=Fullscreen
```
davidflanagan: Shouldn't there be an ariaLabel on this line?

apps/gallery/index.html
```diff
@@ -20,6 +20,7 @@
     <meta name="availableLanguages" content="ar, en-US, fr, zh-TW">
     <link rel="localization" href="locales/gallery.{locale}.properties">
     <link rel="localization" href="/shared/locales/date/date.{locale}.properties">
+    <link rel="localization" href="/shared/locales/media/media.{locale}.properties">
```
davidflanagan: I think you'll need to do this for Camera as well, since it also uses media_frame and video_player.

shared/js/media/media_frame.js
davidflanagan: On second thought, MediaUtils.formatDate() is such a trivial function that maybe it would be better to just reimplement it here rather than adding  a new dependency to this module.

shared/js/media/media_frame.js
eeejay: I will initialize the label to an empty string, but really that isn't necessary either. If there is no locale entry get() returns an empty string, and so does DateTimeFormat.locateFormat().

shared/js/media/video_player.js
eeejay: Like I said above, I tested it and all l10n functions fail gracefully by returning empty strings.

shared/locales/media/media.en-US.properties
eeejay: Yes, oops. Looks like this patch got squashed with a different one. I'll rebase and upload a cleaner one.

apps/gallery/index.html
eeejay: Right. Will do that.

try-server-hook:
[eeejay Eitan Isaacson (eeejay)](https://github.com/eeejay) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=a5a4668ebc5f)

apps/camera/js/app.js
```diff
@@ -363,7 +363,9 @@ App.prototype.onBeforeUnload = function() {
  * @private
  */
 App.prototype.loadL10n = function(done) {
-  this.require(['l10n'], done);
+  this.require(['l10n'], function() {
+    this.require(['l10n_date'], done);
+  });
```
davidflanagan: I haven't worked with require.js based code much, so I don't know about this kind of nested require. Can you pass l10n and l10n_date in the same array?  I think you can probably add something to js/config/require.js to specify that l10n_date requires l10n so that it should be safe to load them in parallel.

My concern here is that nesting these two will have an impact on app startup time.

I think you'll need to ask Wilson Page or Diego Marcos to review these changes.

davidflanagan: Or actually, maybe in config/require.js you can just add l10n_date as a dependency of media_frame and maybe things will just work right. MediaFrame only gets loaded when the user previews photos, so l10n_date shouldn't have to be on the startup path as it presumably is here.

shared/js/media/media_frame.js
```diff
@@ -371,6 +377,40 @@ MediaFrame.prototype._displayImage = function(backgroundImage, width, height) {
   var temp = this.image.clientLeft; // jshint ignore:line
 };
 
+// This function adds a label for accessibility to the image frame.
+// Videos are localized within the video player, so this is only for images.
+MediaFrame.prototype.localize = function localize() {
+  if (!this.imageblob) {
```
davidflanagan: I think !this.displayingImage would be the better test here.  (If I'm remembering the property name correctly)

apps/camera/js/app.js
eeejay: It is lazily loaded, it isn't in the startup path. I'll check to see if this nested business is plain wrong.

try-server-hook:
[eeejay Eitan Isaacson (eeejay)](https://github.com/eeejay) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=3c6dfbc712a3)

shared/js/media/media_frame.js
```diff
@@ -51,6 +51,7 @@ function MediaFrame(container, includeVideo, maxImageSize) {
     this.video = new VideoPlayer(container);
     this.video.hide();
   }
+  navigator.mozL10n.once(this.localize.bind(this));
```
davidflanagan: I think that this is what I suggested you do, but I'm having second thoughts.  If the l10n data is ready, this may call localieImage before displayImage() has been called, there may not be anything to be localized.  And if the l10n data is not ready yet, then this plus the ready() call in frames.js may mean that localize gets called twice.

Given the call to localize() in  displayImage(), I think that this once() call isn't actually necessary.

shared/js/media/media_frame.js
```diff
@@ -144,6 +146,10 @@ MediaFrame.prototype.displayImage = function displayImage(blob,
   this.fullsizeWidth = this.fullSampleSize.scale(width);
   this.fullsizeHeight = this.fullSampleSize.scale(height);
 
+  if (navigator.mozL10n.readyState === 'complete') {
```
davidflanagan: You could add a comment here to say that if readyState is not 'complete', then localization will have to be handled externally by the same handler that handles language changes.

shared/js/media/video_player.js
```diff
@@ -526,6 +532,34 @@ function VideoPlayer(container) {
...
+      }
+      label = self.dtf.localeFormat(videotimestamp, locale_entry);
+    } else {
+      label = navigator.mozL10n.get(
+        'videoDescriptionNoTimestamp', { orientation: orientation });
+    }
+    poster.setAttribute('aria-label', label);
+  }
+
+  navigator.mozL10n.ready(localize);
```
davidflanagan: get rid of this line, please.

shared/js/media/media_frame.js
```diff
@@ -371,6 +377,40 @@ MediaFrame.prototype._displayImage = function(backgroundImage, width, height) {
   var temp = this.image.clientLeft; // jshint ignore:line
 };
 
+// This function adds a label for accessibility to the image frame.
+// Videos are localized within the video player, so this is only for images.
+MediaFrame.prototype.localize = function localize() {
+  if (!this.imageblob) {
+    return;
+  }
+
```
davidflanagan: I think you now need to add code to handle localizing the video player if there is one.  Maybe something like this?

```
if (this.displayingVideo) {
  this.videoplayer.localize();
}
if (this.displayingImage) {
 // all the code below would go here...
}
```

shared/js/media/media_frame.js
eeejay: Yeah, I had doubts about this too.

try-server-hook:
[eeejay Eitan Isaacson (eeejay)](https://github.com/eeejay) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=45dc3da25e1e)

try-server-hook:
[eeejay Eitan Isaacson (eeejay)](https://github.com/eeejay) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=93b74fe1f2b5)

shared/js/media/media_frame.js
```diff
@@ -51,8 +51,26 @@ function MediaFrame(container, includeVideo, maxImageSize) {
     this.video = new VideoPlayer(container);
     this.video.hide();
   }
+
+  // Add a class to the container so we could find it later and use it as
+  // a key in the instance weakmap.
+  this.container.classList.add('media-frame');
+  MediaFrame.instancesToLocalize.set(container, this);
```
davidflanagan: line 57 uses this.container, and line 58 uses container.  They have the same value, but the code is confusing this way. Change one line or the other to be consistent.

shared/js/media/media_frame.js
```diff
@@ -51,8 +51,26 @@ function MediaFrame(container, includeVideo, maxImageSize) {
...
 }
 
+// WeakMap with the container nodes as keys and MediaFrame instances as values.
+MediaFrame.instancesToLocalize = new WeakMap();
+
+navigator.mozL10n.ready(function() {
+  // Retrieve MediaFrame instances by searching for container nodes.
+  for (var container of document.querySelectorAll('.media-frame')) {
```
davidflanagan: This is the first time I've see for/of in Gaia.  I was confused at first, but I guess that is the price of progress :-)

shared/js/media/media_frame.js
```diff
@@ -157,6 +176,11 @@ MediaFrame.prototype.displayImage = function displayImage(blob,
   // Keep track of what kind of content we have
   this.displayingImage = true;
 
+  // If a locale is present and ready, go ahead and localize now.
+  if (navigator.mozL10n.readyState === 'complete') {
```
davidflanagan: You might add...

// Otherwise, localization will be handled by the ready() callback above

or something like that.

shared/js/media/video_player.js
```diff
@@ -537,3 +577,15 @@ VideoPlayer.prototype.show = function() {
...
+
+navigator.mozL10n.ready(function() {
+  // Retrieve VideoPlayer instances by searching for container nodes.
+  for (var container of document.querySelectorAll('.video-player-container')) {
+    var instance = VideoPlayer.instancesToLocalize.get(container);
+    if (instance) {
+      instance.localize();
+    }
+  }
+});
```
davidflanagan: Because of the way functions work in JS, forward references are possible. Since the VideoPlayer() constructor is so long, consider putting this code up above before the constructor if you think that would make things clearer.

Also, I think that camera and gallery are the only apps to use video player, and that they both use it inside of a MediaFrame, so it occurs to me now (kind of late) that it might be cleaner to only have the ready() callback and WeakMap thing in MediaFrame, and have the media frame localize() method call the localize method of its video player.  On the other hand, that makes the component less independent.  In either case, we're phasing it out in favor of the web component that Russ is working on.

eeejay: I did not know forward references are possible, that is strange and magical.

eeejay: I think I will leave it at the bottom. There is so much nice documentation on top before the function declaration, and I feel like the localization bits just make that confusing.

try-server-hook:
[eeejay Eitan Isaacson (eeejay)](https://github.com/eeejay) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=a9fd3f2517b6)

try-server-hook:
[eeejay Eitan Isaacson (eeejay)](https://github.com/eeejay) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=65c4c07ca04b)

PR #27815, opened 2015-01-30
Bug 1126691 - Remove STK_EVENT_TYPE_USER_ACTIVITY and STK_EVENT_TYPE_LANGUAGE_SELECTION handlers before registering new events.

weilonge:


try-server-hook:
[weilonge Sean Lee (weilonge)](https://github.com/weilonge) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=6345c9f56f0a)

apps/system/js/icc_events.js
```diff
@@ -175,28 +174,21 @@ var icc_events = {
...
-    };
     conn.removeEventListener('voicechange',
       this.register_icc_event_voicechange);
 
-    this.register_icc_event_idlescreen = function() {
-      icc_events.handleIdleScreenAvailableEvent(message);
-    };
+    if (this.stkUserActivity) {
+      navigator.removeIdleObserver(this.stkUserActivity);
+    }
```
frsela: Here we'll remove the idle observer only when a new user activity handler is requested.
AFAIK, we should remove it after inform the SIM card of the first user activity. Correct me if I'm wrong

weilonge: When STK_CMD_SET_UP_EVENT_LIST command is received, the whole eventList should be handled again.
If we understand every STK events as a switch, eventList is a list that describes how many events should be turned "ON". It should be recognized as "OFF" if the event is not in the list.
So, all event registrations (turn off all switches) should be removed before checking the events.

The above comment is based on [3GPP TS 11.14, 6.4.16 SET UP EVENT LIST] for your reference. You can see the comment 13 at bug 1088611 as well.
https://bugzilla.mozilla.org/show_bug.cgi?id=1088611#c13

If you think there is any wrong, please tell me. Thank you very much!

frsela: Thank you for your clarification. Now I've a clear picture !

frsela: r+, of course ;)

PR #27825, opened 2015-01-30
Bug 1126466 - [music] Use full-size album art in the player UI

jimporter:


try-server-hook:
[jimporter Jim Porter (jimporter)](https://github.com/jimporter) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=8cc52127bf7d)

apps/music/js/ui/views/player_view.js
```diff
@@ -244,22 +244,19 @@ var PlayerView = {
...
+        this.coverImage.style.backgroundImage = 'url(' + url + ')';
+        this.coverImage.classList.remove('fadeIn');
+
+        this.offscreenImage.src = url;
+        this.offscreenImage.addEventListener('load', function pv_loaded(event) {
+          /* jshint validthis:true */
+          event.target.removeEventListener('load', pv_loaded);
+          this.coverImage.classList.add('fadeIn');
+        });
+      }
```
davidflanagan: I think that using hte offscreenImage here is unnecessary and possibly inefficient. 

1)  offscreenImage has never been assigned a size, and the onload handler does not cause it to be rendered anywhere, so I think this means that it will be decoded at full size. Gecko will assume you're trying to do an image cache and will decode for you so it is ready when you insert the image into the document.  I'm not sure how you can test what size the image gets decoded at, however. The patch for this landed late last week and then was backed out because it was not working correctly for code that used background-image.

2) Gecko fires the load event well before the image is actually decoded and ready for display. The onload hander is apparently used to know when to trigger the fadeIn animation.  Since there is no network latency here, the decode time is probably longer than the load time.  I suspect that you could just get rid of the onload handler completely and maybe just add a short animation-delay to the fadeIn animation to compensate.

3) I like that you're're adding the test so that you don't change the image if not necessary.  But you could do the same check on coverImage.style.backgroundImage as long as you add the "url()" before comparing.

4) I don't get how the this inside the event handler could possibly be valid.  Did you mean to add a .bind(this)? Doesn't this refer to the offscreenImage at that point in the code?

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=250c3802f5dc6e3a3751e3c90e1a3ed89d4139e9

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9892a5649b82b9b137dc1b392b5c5fca73b99d6c

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0150be75d5643718cc7507864d149113dceb3fc8

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3ea09e590a822be88ad021ed611066febca91d24

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b43326b2e5576e5d38c2f6c6578b3f79b8aad131

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=59091c545709c2474714445a03ed3c62d8b35673

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7c98a824b24bfdee61ecf156fdaf00c7ea91ba42

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9ad71e61b7fbf3119fbec1e6d4672722472b4a44

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0ed36e287ed5ad37306bd54013fcb2369aa291a9

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3512e3c85ef4ed311f40444c8709c4c249d40c57

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=fc73017b482957be2627fc250f75d8879a5aefc4

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b3c411a26b5b4e97cfbbb0ef15943813f34ed96c

apps/music/js/metadata/album_art_cache.js
```diff
@@ -13,9 +13,67 @@ var AlbumArtCache = (function() {
...
+    // See if we've already made a URL for this album. If so, return it.
+    var cacheKey = makeCacheKey(fileinfo);
+    if (cacheKey && cacheKey in L1Cache.fullsize) {
+      return Promise.resolve(L1Cache.fullsize[cacheKey]);
+    }
+
+    // Otherwise, get the blob from the album art, create a URL for it, and
+    // return it in the Promise.
+    return getAlbumArtBlob(fileinfo).then(function(blob) {
+      return makeURL(cacheKey, blob, 'fullsize');
```
davidflanagan: makeAndCacheURL() might be a better name for this function.

davidflanagan: Also consider defining FULLSIZE AND THUMBNAIL constants so you don't have to use string literals throughout your code. Saves the risk of typos.

apps/music/js/metadata/album_art_cache.js
```diff
@@ -38,17 +96,22 @@ var AlbumArtCache = (function() {
 
     // See if we've already made a URL for this album. If so, return it.
     var cacheKey = makeCacheKey(fileinfo);
-    if (cacheKey && cacheKey in L1Cache) {
-      return Promise.resolve(L1Cache[cacheKey]);
+    if (cacheKey && cacheKey in L1Cache.thumbnail) {
```
davidflanagan: I just noticed that you're checking whether cacheKey exists here and skipping the L1 cache if so. But below you are apparently just assuming that cacheKey is defined.  So I think you can just remove the `cacheKey &&` here.

apps/music/js/metadata/album_art_cache.js
```diff
@@ -38,17 +96,22 @@ var AlbumArtCache = (function() {
...
     }
 
     // Otherwise, see if we've saved a blob in asyncStorage. If not, create a
     // thumbnail blob and store it in the cache. Finally, create a URL for the
     // blob and return it in the Promise.
+    //
+    // XXX: There's a (minor) race condition here. If two requests for the same
+    // album art get here, the slower one will overwrite the faster one's cached
+    // blob URL, meaning that not all instances of a piece of album art will
+    // have the same URL. This only really matters to tests, though.
```
davidflanagan: As soon as you notice that the L1 cache is empty you could store a promise in it. Then when you've got the thumbnail swap the promise out for the URL. And resolve the promise to the thumbnail as well.

This seems like it would add a lot of complexity, though.  The consequences for the race condition are minor so this seems acceptable to me as it is.  Unless there is some common case (displaying the tracks on an album) where you make many requests for the same thumbnail at the same time. Then you might want to think about it.

Actually, another easier thing you could do here is modify makeURL so that it checks again that the cache is still empty. If, by the time that makeURL is called there is already a url stored, then it could discard the new blob and return the cached URL.

That would solve it, wouldn't it? You don't have any cases where you purposely want to overwrite the cache, do you?

I think this kind of double-checking is actually a thing in concurrent programming and there might even be a computer-sciencey word for it.

apps/music/js/metadata/album_art_cache.js
```diff
@@ -164,12 +227,13 @@ var AlbumArtCache = (function() {
...
    * @param {Blob} blob The blob.
+   * @param {String} type Either 'thumbnail' or 'fullsize'.
    * @return {String} The blob: URL.
    */
-  function makeURL(cacheKey, blob) {
+  function makeURL(cacheKey, blob, type) {
     var url = URL.createObjectURL(blob);
     if (cacheKey) {
-      L1Cache[cacheKey] = url;
+      L1Cache[type][cacheKey] = url;
```
davidflanagan: Before creating this new URL you could make sure that the cache is still empty. If not, return the cached URL instead of creating the new one.

Or, if you're not sure that would be safe, you could at least emit a console.log here to let you know that your race condition had occurred. That would give you some data about how common it is in practice.

apps/music/js/metadata/album_art_cache.js
```diff
@@ -213,13 +277,15 @@ var AlbumArtCache = (function() {
...
         // need to grab it from deviceStorage. This could also be an unsynced
         // picture that came from a regular file.
-        var getreq = AlbumArt.pictureStorage.get(picture.filename);
-        getreq.onsuccess = function() {
-          resolve(this.result);
-        };
-        getreq.onerror = function() {
-          reject(this.error);
-        };
+        LazyLoader.load('/js/metadata/album_art.js', function() {
```
davidflanagan: Since you're adding a lazy loading step here, I went looking for where you removed it. I can't find anywhere, so I'm guessing that means that this was just a bug: you were relying on db.js to load it and assuming that we'd never not load it.

Is this an issue in the 2.2 branch? Is there a bug that needs to be fixed there?

apps/music/js/music.js
```diff
@@ -160,88 +160,116 @@ var App = (function() {
     }
   }
 
-  function showCurrentView(callback) {
-    // We need AlbumArtCache.getCoverURL() to display thumbnails; it might not
-    // have been loaded yet, so make sure we load it first. This should prevent
-    // us from having to worry about loading it anywhere else in the code, since
-    // showCurrentView is called pretty early in the startup process.
```
davidflanagan: It looks like you've changed your mind about that and are now using LazyLoader before every use...

apps/music/js/music.js
```diff
@@ -160,88 +160,116 @@ var App = (function() {
...
+      // ready to interact with. We won't send the final moz-app-loaded
+      // event until we're completely stable and have finished scanning.
+      //
+      // XXX: Maybe we could emit these events earlier, when we've just
+      // finished the "above the fold" content. That's hard to do on
+      // arbitrary screen resolutions, though.
+      window.performance.mark('visuallyLoaded');
+      window.dispatchEvent(new CustomEvent('moz-app-visually-complete'));
+      window.performance.mark('contentInteractive');
+      window.dispatchEvent(new CustomEvent('moz-content-interactive'));
```
davidflanagan: I've already reviewed this stuff in your other startup refactoring patch.  Looking at it now, I'll add that Eli is trying to get these dispatchEvent() calls out now.  You could help him out here if you take them out yourself, since his patch will have been completely broken by this.

apps/music/js/music.js
```diff
@@ -160,88 +160,116 @@ var App = (function() {
...
+    } else if (callback) {
+      callback();
+    }
+  }
+
+  // This tracks if we've automatically hidden the search box on startup yet (it
+  // should only be done once!)
+  var hidSearchBox = false;
+
+  function refreshViews(callback) {
```
davidflanagan: Actually, I have not reviewed this code... I just gave feedback on the 'enumerable' parts of it.  I'm assuming that most of this stuff is under review by Dominic and is here because the two patches are interdependent.  So I'm going to skip this stuff and just go looking for the album art related code.

apps/music/js/ui/title_bar.js
```diff
@@ -97,25 +97,28 @@ var TitleBar = {
...
+                      title: currentMetadata.title,
+                      artist: currentMetadata.artist,
+                      album: currentMetadata.album,
+                      picture: picture
+                    }
+                  });
+
+                  cleanupPick();
+                });
+            });
```
davidflanagan: Looks like all you're doing here is adding a LazyLoader.load() wrapper.  Too bad that function doesn't have a cool name like `using()` or something.

I'll just add that this seems like an obvious refactoring opportunity to (in some future bug) move pick-related functionality out of the title bar view!

apps/music/js/ui/views/list_view.js
```diff
@@ -60,8 +60,8 @@ var ListView = {
     this.lastFirstLetter = null;
     this.anchor.innerHTML = '';
     this.anchor.style.height = 0;
-    this.view.scrollTop = 0;
-    this.hideSearch();
+    this.searchInput.value = '';
+    this.view.scrollTop = this.anchor.offsetTop;
```
davidflanagan: I'm going to assume that the changes in this file are part of some other bug, and that you don't need me to review them.  I don't know what's going on here.

apps/music/js/ui/views/mode_manager.js
```diff
@@ -107,21 +107,27 @@ var ModeManager = {
...
           PlayerView.setOptions(App.playerSettings);
         }
-
         // Music only share the playing file when it's in player mode.
         this._enableNFCSharing(true);
 
         if (callback) {
           callback();
         }
+        document.getElementById('views-player').classList.remove('hidden');
```
davidflanagan: Good call to move this line inside the lazy load. Note that by putting it here after the callback you're assuming that the callback will never throw an exception.  A more defensive coding style would have you use a try/finally block.  Or just show the element right before calling the callback...

apps/music/js/ui/views/mode_manager.js
```diff
@@ -107,21 +107,27 @@ var ModeManager = {
...
       }.bind(this));
     } else {
+      // Disable the NFC sharing when it's in the other modes.
+      this._enableNFCSharing(false);
+
+      if (callback) {
+        callback();
+      }
```
davidflanagan: Same comment here. try/catch or try/finally?

apps/music/js/ui/views/player_view.js
```diff
@@ -37,6 +37,14 @@ var PlayerView = {
     return document.getElementById('player-audio');
   },
 
+  get foreCoverImage() {
+    return document.querySelector('.player-cover-image.visible');
+  },
+
+  get backCoverImage() {
+    return document.querySelector('.player-cover-image:not(.visible)');
+  },
```
davidflanagan: How about naming these visibleCoverImage and hiddenCoverImage instead?

apps/music/js/ui/views/player_view.js
```diff
@@ -246,22 +254,18 @@ var PlayerView = {
...
-      this.offscreenImage.src = url;
+    LazyLoader.load('js/metadata/album_art_cache.js', function() {
+      AlbumArtCache.getFullSizeURL(fileinfo).then(function(url) {
+        if (this.coverImageURL !== url) {
+          this.coverImageURL = url;
+
+          var back = this.backCoverImage, fore = this.foreCoverImage;
+          back.style.backgroundImage = 'url(' + url + ')';
+          back.classList.add('visible');
+          fore.classList.remove('visible');
```
davidflanagan: This looks like a sweet hack!

apps/music/js/ui/views/player_view.js
```diff
@@ -246,22 +254,18 @@ var PlayerView = {
...
 
   setCoverImage: function pv_setCoverImage(fileinfo) {
-    // Reset the image to be ready for fade-in
-    this.offscreenImage.src = '';
-    this.coverImage.classList.remove('fadeIn');
-
-    AlbumArtCache.getCoverURL(fileinfo).then(function(url) {
-      this.offscreenImage.addEventListener('load', pv_showImage.bind(this));
-      this.offscreenImage.src = url;
+    LazyLoader.load('js/metadata/album_art_cache.js', function() {
```
davidflanagan: While you're doing this refactoring, feel free to switch to arrow functions here. It isn't the saved typing that matters, but the fact that you can leave out the bind(this) if you do it.

```
LazyLoader.load('js/metadata/album_art_cache.js', () => {
});
```

apps/music/js/ui/views/player_view.js
```diff
@@ -826,103 +832,107 @@ var PlayerView = {
...
+                if (PlayerView.playStatus === PLAYSTATUS_PAUSED) {
+                  PlayerView.play();
+                }
               }
-            }
-          }());
-        }
+            }());
+          }
+        });
```
davidflanagan: So I'm assuming that none of this share code has changed, and that all you've done is insert the lazyloading step which changes the indentation of everything else.

apps/music/js/ui/views/tiles_view.js
```diff
@@ -42,44 +38,50 @@ var TilesView = {
...
+    if (this.handle) {
+      musicdb.cancelEnumeration(this.handle);
     }
 
+    this.handle = musicdb.enumerateAll('metadata.album', null, 'nextunique',
+                                       function(songs) {
+      TilesView.clean();
+      songs.forEach(function(song) {
+        TilesView._addSong(song);
+      });
```
davidflanagan: I think this is part of your startup improvements patch and not part of this album art patch, but I can't help commenting here that I still think using enumerateAll() is the wrong way to go here.

For the future, I think you need to make TilesView smart enough to be able to add new tiles without completely rebuilding itself each time you do.

Then, I think you should use a regular enumerate and batch the results into pages. When you have a page full of tiles, you add them as a batch to the tiles view and continue with the enumeration.  The gallery app does it that way and it allows us to get the first page displayed more quickly without waiting for a complete enumeration.

apps/music/js/ui/views/tiles_view.js
```diff
@@ -120,22 +122,25 @@ var TilesView = {
...
+      };
+
+      if (index <= NUM_INITIALLY_VISIBLE_TILES) {
+        // Load this tile's background now, because it's visible.
+        AlbumArtCache.getThumbnailURL(result).then(setTileBackgroundClosure);
+      } else {
+        // Defer loading hidden tiles until the visible ones are done.
+        setTimeout(function() {
+          AlbumArtCache.getThumbnailURL(result).then(setTileBackgroundClosure);
+        }, INITIALLY_HIDDEN_TILE_WAIT_TIME_MS);
```
davidflanagan: I had never seen this code before... it is probably a good optimization, but yuck!  Every tile after the first 8 is delayed by a second even after the first page is fully loaded.  This is a problem that might just go away if the albums were enumerated with enumerate() instead of enumerateAll()

apps/music/style/music.css
```diff
@@ -660,31 +659,18 @@ html[dir="rtl"] .tile-title-album {
...
-
-#player-cover-image {
+.player-cover-image {
   position: absolute;
   width: 100%;
   height: 100%;
   opacity: 0;
   background-position: center;
   background-size: cover;
+  transition: opacity 500ms ease-in-out 50ms;
```
davidflanagan: What's the 50ms for? Might be worth a comment since it isn't obvious.

apps/music/test/marionette/metadata_test.js
```diff
@@ -5,6 +5,11 @@ var assert = require('assert');
 var Music = require('./lib/music.js');
 var ListviewHelper = require('./lib/listviewhelper.js');
 
+function assertStartsWith(start, actual) {
+  assert.equal(actual.indexOf(start), 0,
```
davidflanagan: String.prototype.startsWith() is a thing now:

assert.ok(actual.startsWith(start), 'message')

apps/music/test/marionette/metadata_test.js
```diff
@@ -69,41 +73,41 @@ marionette('Music player metadata', function() {
...
 
-      bg = ListviewHelper.albumArtForListItem(music.listItems[1]);
-      assert.notEqual(bg.indexOf('url("app:'), -1, 'Not a placeholder');
+      bg = ListviewHelper.albumArtForListItem(client, music.listItems[1]);
+      assertStartsWith('url("app:', bg);
 
-      bg = ListviewHelper.albumArtForListItem(music.listItems[2]);
-      assert.notEqual(bg.indexOf('url("blob:'), -1, 'Not a metadata art');
+      bg = ListviewHelper.albumArtForListItem(client, music.listItems[2]);
+      assertStartsWith('url("blob:', bg);
```
davidflanagan: With these changes you're losing the original assertion messages. Maybe you could give assertStartsWith an optional message argument?

You might also just use assert.match() with a regexp that begins with ^. Though you might have to escape some of those URL syntax characters, which could be a pain.

apps/music/js/metadata/album_art_cache.js
jimporter: It technically works in 2.2, but it was very brittle.

apps/music/js/music.js
jimporter: Yeah, that was prone to bustage. I actually removed it because minor changes to the code caused everything to break, so I decided we should just lazy-load at the point of use.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ab6b795b08f0774c6e830bf9c88f0538afe5346f

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=87897b9ba877a5582128273ecaa96b7a0f8624eb

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d28de2a72edfa9071e2c55762e1a8c282cdc0f5f

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=fa555259113425edca4e4ec8b43c2d8153b6cda7

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a904d7dc1100bc6ed56aadc1995a2dda00f6b6ad

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9524d95a7ed048b73aa04bd1aa4f5ff4117717ad

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=880f90e92f62b836a6baa693ac80e1919b21291f

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=075ca0f91b3b8fa7d5f699939393d44616b4bfde

autolander:
http://docs.taskcluster.net/tools/task-graph-inspector/#UB3Du4bdSKa6GVqxKOXWuw

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a0fc274b051f28e8c85d4ea9ed4520e05cfc6eac

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=944df3b020a5f40716ded346b6afaf5b45e66a93

autolander:
http://docs.taskcluster.net/tools/task-graph-inspector/#ZhpN1XvVQnyvh01wiflXrw

autolander:
http://docs.taskcluster.net/tools/task-graph-inspector/#pJyxFA7eSJKofOncEnnQdQ

PR #27827, opened 2015-01-30
Bug 1068703 - improve message list accessibility in edit mode.

yzen:


try-server-hook:
[yzen Yura Zenevich (yzen)](https://github.com/yzen) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=3bdc91565d3b)

try-server-hook:
[yzen Yura Zenevich (yzen)](https://github.com/yzen) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=7c35e4d8746d)

try-server-hook:
[yzen Yura Zenevich (yzen)](https://github.com/yzen) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=1cd58e939a7b)

apps/email/js/cards/message_list.html
```diff
@@ -102,7 +102,7 @@ <h1 data-prop="headerNode" class="msg-listedit-header-label">
       </p>
     </form>
     <div data-prop="messagesContainer" class="msg-messages-container"
-         role="listbox">
+         role="listbox" aria-multiselectable="true">
```
jrburke: Apologies for my aria newness, but is this OK to always set on this div, even when the card is not in edit mode, and is not really multi-select, but more "tap on item and goes to a new card"?

yzen: Yeah, especially since we navigate away on click in regular mode.

apps/email/js/cards/message_list.js
```diff
@@ -443,10 +443,10 @@ return [
...
 
         this.selectedMessages = [];
-        var cbs = this.messagesContainer
-                  .querySelectorAll('input[type=checkbox]');
-        for (i = 0; i < cbs.length; i++) {
-          cbs[i].checked = false;
+        var msgNodes = this.messagesContainer.querySelectorAll(
+          '.msg-header-item');
+        for (i = 0; i < msgNodes.length; i++) {
+          this.setMessageChecked(msgNodes[i], false);
```
jrburke: I think we should be doing this reset of the message checked to be false for all nodes, regardless of `if (editMode) {`. The else case, when editMode is false, is doing some work with 'msg-header-item-selected' that I believe is the same attempt at a reset, but based on very old code. So we can just remove that ''msg-header-item-selected' work and do this setMessageChecked false work outside the if/else.

Otherwise, going back to non-edit mode by clicking x will leave some msgNodes with aria-selected still as true.

try-server-hook:
[yzen Yura Zenevich (yzen)](https://github.com/yzen) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=b37175e2714b)

PR #27852, opened 2015-02-02
Bug 874368 - Improving the perceived performance of notification swipes +autoland

etiennesegonzac:
- prevent the constant reflow while dragging
- take speed into account to detect swipes

### DIFF
--- a/apps/system/js/notifications.js
+++ b/apps/system/js/notifications.js
@@ -7,6 +7,8 @@
 var NotificationScreen = {
   TOASTER_TIMEOUT: 5000,
   TRANSITION_FRACTION: 0.30,
+  TRANSITION_DURATION: 300,
+  SWIPE_INERTIA: 100,
   TAP_THRESHOLD: 10,
   SCROLL_THRESHOLD: 10,
   CLEAR_DELAY: 80,
@@ -228,6 +230,7 @@ var NotificationScreen = {
     this._touchStartX = evt.touches[0].pageX;
     this._touchStartY = evt.touches[0].pageY;
     this._touchPosX = 0;
+    this._touchStartTime = evt.timeStamp;
     this._touching = true;
     this._isTap = true;
   },
@@ -259,6 +262,7 @@ var NotificationScreen = {
     this._touchPosX = evt.touches[0].pageX - this._touchStartX;
     if (Math.abs(this._touchPosX) >= this.TAP_THRESHOLD) {
       this._isTap = false;
+      this.notificationsContainer.style.overflow = 'hidden';
     }
     if (!this._isTap) {
       this._notification.style.transform =
@@ -274,6 +278,8 @@ var NotificationScreen = {
     evt.preventDefault();
     this._touching = false;
 
+    this.notificationsContainer.style.overflow = '';
+
     if (this._isTap) {
       var event = new CustomEvent('tap', {
         bubbles: true,
@@ -284,12 +290,21 @@ var NotificationScreen = {
       return;
     }
 
-    if (Math.abs(this._touchPosX) >
+    // Taking speed into account to detect swipes
+    var deltaT = evt.timeStamp - this._touchStartTime;
+    var speed = this._touchPosX / deltaT;
+    var inertia = speed * this.SWIPE_INERTIA;
+    var finalDelta = Math.abs(this._touchPosX + inertia);
+
+    if (finalDelta >
         this._containerWidth * this.TRANSITION_FRACTION) {
       if (this._touchPosX < 0) {
         this._notification.classList.add('left');
       }
-      this.swipeCloseNotification();
+
+      var durationLeft = (1 - (finalDelta / this._containerWidth)) *
+                         this.TRANSITION_DURATION;
+      this.swipeCloseNotification(durationLeft);
     } else {
       this.cancelSwipe();
     }
@@ -581,7 +596,7 @@ var NotificationScreen = {
       }));
   },
 
-  swipeCloseNotification: function ns_swipeCloseNotification() {
+  swipeCloseNotification: function ns_swipeCloseNotification(duration) {
     var notification = this._notification;
     this._notification = null;
 
@@ -611,6 +626,8 @@ var NotificationScreen = {
     });
 
     notification.classList.add('disappearing');
+    duration = duration || this.TRANSITION_DURATION;
+    notification.style.transitionDuration = duration + 'ms';
     notification.style.transform = '';
   },
 
--- a/apps/system/style/notifications/notifications.css
+++ b/apps/system/style/notifications/notifications.css
@@ -268,7 +268,7 @@
 #notifications-container .notification.disappearing,
 #notifications-container .notification.disappearing-via-clear-all,
 .notification-toaster.disappearing {
-  transition: transform 0.2s linear;
+  transition: transform 0.3s linear;
   transform: translateX(100%);
 }
 
--- a/apps/system/test/marionette/lib/utility_tray.js
+++ b/apps/system/test/marionette/lib/utility_tray.js
@@ -20,11 +20,16 @@
       element: '#utility-tray',
       screen: '#screen.utility-tray',
       notifContainer: '#desktop-notifications-container',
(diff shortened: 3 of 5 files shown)


try-server-hook:
[etiennesegonzac Etienne Segonzac (etiennesegonzac)](https://github.com/etiennesegonzac) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=640dc13596ed)

try-server-hook:
[etiennesegonzac Etienne Segonzac (etiennesegonzac)](https://github.com/etiennesegonzac) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=960124678ffb)

apps/system/js/notifications.js
```diff
@@ -228,6 +229,7 @@ var NotificationScreen = {
     this._touchStartX = evt.touches[0].pageX;
     this._touchStartY = evt.touches[0].pageY;
     this._touchPosX = 0;
+    this._touchStartDate = Date.now();
```
Cwiiis: I'm amazed to find out that touch events really don't have a time component... This seems like a major bit of missing functionality :/

etiennesegonzac: good to know!

try-server-hook:
[etiennesegonzac Etienne Segonzac (etiennesegonzac)](https://github.com/etiennesegonzac) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=db2d7faa3737)

apps/system/js/notifications.js
Cwiiis: Any reason not to use the event timestamp here? It may well make it safe over clock changes to do so (not sure about that, but using Date.now() definitely isn't)

apps/system/js/notifications.js
```diff
@@ -284,12 +289,20 @@ var NotificationScreen = {
...
+
+    if (finalDelta >
         this._containerWidth * this.TRANSITION_FRACTION) {
       if (this._touchPosX < 0) {
         this._notification.classList.add('left');
       }
-      this.swipeCloseNotification();
+
+      var durationLeft = (1 - (finalDelta / this._containerWidth)) * 300;
+      this.swipeCloseNotification(durationLeft);
```
Cwiiis: Nice :) Might be nice to factor out the 300 into a const so this reads better (and we can document it a little better)

apps/system/js/notifications.js
```diff
@@ -611,6 +624,9 @@ var NotificationScreen = {
     });
 
     notification.classList.add('disappearing');
+    if (duration) {
+      notification.style.transitionDuration = duration + 'ms';
+    }
```
Cwiiis: Do we want to do a transitionDuration = '' in an else block here, to make sure? (or if we add that const as mentioned above, transitionDuration = CONST_VALUE + 'ms'? (and then remove from CSS to avoid confusion too)

apps/system/style/notifications/notifications.css
```diff
@@ -268,7 +268,7 @@
 #notifications-container .notification.disappearing,
 #notifications-container .notification.disappearing-via-clear-all,
 .notification-toaster.disappearing {
-  transition: transform 0.2s linear;
+  transition: transform 0.3s linear;
```
Cwiiis: I take it this is just because 0.3 feels better than 0.2?

apps/system/test/marionette/lib/utility_tray.js
```diff
@@ -27,6 +27,11 @@
       return this.client.findElement(this.Selectors.screen);
     },
 
+    get firstNotification() {
+      return this.client.findElement(this.Selectors.notifContainer +
+                                     ' .notification');
```
Cwiiis: Maybe just add notification to Selectors?

apps/system/test/marionette/notification_actions_test.js
```diff
@@ -58,4 +68,25 @@ marionette('launch an app via notification click', function() {
     });
   });
 
+  test('swiping to dismiss', function() {
```
Cwiiis: While I don't really have much of an opinion either way, I thought we were trending towards one test per file for marionette tests?

apps/system/test/marionette/notification_actions_test.js
```diff
@@ -58,4 +68,25 @@ marionette('launch an app via notification click', function() {
...
+    utilityTray.waitForOpened();
+
+    // make sure we have our notification to click
+    notificationList.refresh();
+    var notifications = notificationList.getForApp(TARGET_APP_MANIFEST);
+    assert.equal(notifications.length, 1);
+
+    actions.flick(utilityTray.firstNotification, 45, 15, 90, 15, 30).perform();
+
+    // email will clear the notification from the tray, so wait
```
Cwiiis: Won't the notification be cleared by the swipe that was just performed, rather than by email?

apps/system/test/unit/notifications_test.js
```diff
@@ -646,6 +652,67 @@ suite('system/NotificationScreen >', function() {
...
+      NotificationScreen.touchstart(fakeEvt(1, 1));
+      this.sinon.clock.tick(10);
+      NotificationScreen.touchmove(fakeEvt(25, 1));
+      this.sinon.clock.tick(20);
+      NotificationScreen.touchmove(fakeEvt(45, 1));
+      this.sinon.clock.tick(30);
+      NotificationScreen.touchend(fakeEvt(45, 1));
+
+      sinon.assert.calledOnce(close);
+      sinon.assert.calledWith(close, 190);
```
Cwiiis: This is going to break if the animation speed changes I would suppose - could the unit test be altered to either read the animation time from the class, or to assert in a way that would allow small/medium changes to the animation time? (if it was a particularly fast swipe, I assume you could get the number pretty close to zero, then you could just test for < 50 or something like that?)

apps/system/js/notifications.js
etiennesegonzac: ha! I think I made the change then trashed it at some point while working on something else... sorry :)

apps/system/style/notifications/notifications.css
etiennesegonzac: Yes `0.2` was pretty short once you only do a fraction of it (with the `durationLeft` stuff)

apps/system/test/marionette/notification_actions_test.js
etiennesegonzac: yep, the comment got copy and pasted my bad

apps/system/test/unit/notifications_test.js
etiennesegonzac: yes this could be finicky, I'll just assert that it's less than the new `TRANSITION_DURATION` constant.

apps/system/js/notifications.js
```diff
@@ -228,6 +230,7 @@ var NotificationScreen = {
     this._touchStartX = evt.touches[0].pageX;
     this._touchStartY = evt.touches[0].pageY;
     this._touchPosX = 0;
+    this._touchStartDate = evt.timeStamp;
```
Cwiiis: nit, call this touchStartTime instead of touchStartDate

apps/system/test/marionette/lib/utility_tray.js
```diff
@@ -20,6 +20,7 @@
       element: '#utility-tray',
       screen: '#screen.utility-tray',
       notifContainer: '#desktop-notifications-container',
+      firstNotification: '#desktop-notifications-container .notification',
```
Cwiiis: nit: You decide - I think it's better to just call this 'notification', as you could do say this.client.findElements(Selectors.firstNotification).pop() to get the last element, and that reads weirdly. On the other hand, we have form for calling this 'firstSomething' in other files. I don't like it, but it's up to you :)

try-server-hook:
[etiennesegonzac Etienne Segonzac (etiennesegonzac)](https://github.com/etiennesegonzac) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=d7fc8179c75d)

PR #27854, opened 2015-02-02
Bug 1120696 - fixing the display problems +autolander

martapiekarska:


apps/privacy-panel/templates/rp/screenlock.html
```diff
@@ -16,7 +16,7 @@ <h1 data-l10n-id="screen-lock-header"></h1>
         </li>
         <li class="lockscreen-enabled">
           <label class="pack-switch">
-            <input type="checkbox" name="lockscreen.passcode-lock.enabled" class="passcode-enable uninit">
+            <input type="checkbox" name="lockscreen.passcode-lock.enabled" class="passcode-enable" class="uninit">
```
KevinGrandon: You have two class attributes here, I don't think that was intended?

PR #27880, opened 2015-02-03
Bug 1128493 - [CostControl] Widget broken after switching data sim

gitmai:


apps/costcontrol/js/widget.js
```diff
@@ -90,6 +90,42 @@ var Widget = (function() {
   var initialized, widget, leftPanel, rightPanel, fte, views = {};
   var balanceView;
 
+  function _openCCBalance() {
```
lodr: nit: ton increase readability, rename to `_openBalanceTab()`.

apps/costcontrol/js/widget.js
```diff
@@ -90,6 +90,42 @@ var Widget = (function() {
...
+    activity = new MozActivity({ name: 'costcontrol/balance' });
+  }
+
+  function _openCCTelephony() {
```
lodr: Same here.

apps/costcontrol/js/widget.js
```diff
@@ -90,6 +90,42 @@ var Widget = (function() {
...
+    activity = new MozActivity({ name: 'costcontrol/telephony' });
+  }
+
+  function _openCCDataUsage() {
```
lodr: And here.

apps/costcontrol/js/widget.js
```diff
@@ -535,7 +539,19 @@ var Widget = (function() {
         LazyLoader.load(SCRIPTS_NEEDED, initWidget);
       }
     },
-    showSimError: _showSimError
+    showSimError: _showSimError,
+    reset: function() {
```
lodr: nit: add a couple of breaklines in between the properties, please.

apps/costcontrol/test/unit/widget_startup_test.js
```diff
@@ -461,4 +453,82 @@ suite('Widget Startup Modes Test Suite >', function() {
     AirplaneModeHelper._status = 'disabled';
     Widget.init();
   });
+
+  test('DSDS normal start-up after sim switch from dataUseOnly to prepaid mode',
```
lodr: Please, reword this: `DSDS start-up after SIM switching: from data usage only SIM to prepaid SIM`.

apps/costcontrol/test/unit/widget_startup_test.js
```diff
@@ -461,4 +453,82 @@ suite('Widget Startup Modes Test Suite >', function() {
...
+        datausage: requestDataUsageResult
+      };
+      var costControlConfig  = newCostControlRequest(fakeResultDataUsage);
+      setupCardState(costControlConfig);
+      var ftePending = false;
+      setupConfig('DATA_USAGE_ONLY', ftePending);
+
+      // Waiting for the first startup (DATA_USE_ONLY mode)
+      window.addEventListener('hashchange', function _onHashChange(evt) {
+        if (window.location.hash.split('#')[1] === 'updateDone') {
```
lodr: Move this to a meaningful variable:
```js
var message = window.location.hash.split('#')[1];
```

apps/costcontrol/test/unit/widget_startup_test.js
```diff
@@ -461,4 +453,82 @@ suite('Widget Startup Modes Test Suite >', function() {
...
+          window.removeEventListener('hashchange', _onHashChange);
+          // right Panel is hidden when data_usage_only mode is loaded
+          assert.ok(rightPanel.hidden);
+
+          this.sinon.stub(window.CostControl, 'getInstance')
```
lodr: The assert above is enough to check the layout. There is no need for checking or mocking up results.

apps/costcontrol/test/unit/widget_startup_test.js
```diff
@@ -461,4 +453,82 @@ suite('Widget Startup Modes Test Suite >', function() {
...
+            .yields(newCostControlRequest(fakePrePaidResult));
+
+          this.sinon.stub(window.ConfigManager, 'getApplicationMode',
```
lodr: Same here. There is no need to fake results, just check the layout. If you need additional setup to avoid functions to throw exceptions, make all the setup in functions out of the test case.

lodr: Try to be meaningful with the names taking into accoount what are you testing. For instance. If you need to say something like: _mock data usage only SIM, then assert the layout is ok, then mock prepaid SIM, then assert the layout is ok_ use something like:

```js
setupDataUsageOnlySIM()
  .then(assertDataUsageOnlyLayout)
  .then(setupPrepaidSIM)
  .then(assertPrepaidLayout);
```

Use promises if you want or callbacks if not. Whatever that increases readability.

apps/costcontrol/test/unit/widget_startup_test.js
lodr: We could add a reference to the bug in a comment. What do you think?

try-server-hook:
[gitmai Marina (gitmai)](https://github.com/gitmai) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=2c79da43d715)

apps/costcontrol/test/unit/widget_startup_test.js
```diff
@@ -461,4 +453,94 @@ suite('Widget Startup Modes Test Suite >', function() {
     AirplaneModeHelper._status = 'disabled';
     Widget.init();
   });
+
+  test('DSDS start-up after SIM switching: from data usage only SIM to ' +
+         ' prepaid SIM',
```
lodr: Align the start of the text with the text above and don't break after for the function callback, please.

apps/costcontrol/test/unit/widget_startup_test.js
```diff
@@ -461,4 +453,94 @@ suite('Widget Startup Modes Test Suite >', function() {
...
+          // Simulating a data slot change
+          var serviceSlotChangeEvent = new CustomEvent('dataSlotChange');
+          window.dispatchEvent(serviceSlotChangeEvent);
+        });
+      }
+
+      setupDataUsageOnlySIM()
+        .then(assertDataUsageOnlyLayout)
+        .then(setupPrepaidSim)
+        .then(assertPrepaidDataLayout);
```
lodr: Add a `.catch()` asserting false to capture exceptions in the test.

apps/costcontrol/test/unit/widget_startup_test.js
```diff
@@ -461,4 +453,94 @@ suite('Widget Startup Modes Test Suite >', function() {
...
+          var balanceDataText = _('currency', {
+            value: fakeResult.balance.data.balance,
+            currency: fakeResult.balance.data.currency
+          });
+          assertNonDataUseOnlyLayout('', balanceDataText);
+        });
+      }
+
+      function setupDataUsageOnlySIM() {
+        return new Promise(function(resolve) {
```
lodr: If you are not performing an async process inside, remeber you can write it as:
```js
function setupDataUsageOnlySIM() {
  setupConfig('DATA_USAGE_ONLY', ftePending);
  return Promise.resolve();
}
```

apps/costcontrol/test/unit/widget_startup_test.js
```diff
@@ -461,4 +453,94 @@ suite('Widget Startup Modes Test Suite >', function() {
...
+          setupConfig('DATA_USAGE_ONLY', ftePending);
+          resolve();
+        });
+      }
+
+      function assertDataUsageOnlyLayout() {
+        return new Promise(function(resolve) {
+          window.addEventListener('hashchange', function _onHashChange(evt) {
```
lodr: I'm thinking this is still configuration and we could move to the previous step or make a dedicated function in such a way, the `assert` function only asserts.

apps/costcontrol/test/unit/widget_startup_test.js
```diff
@@ -461,4 +453,94 @@ suite('Widget Startup Modes Test Suite >', function() {
...
+                },
+                is_free: false,
+                is_roaming_free: false
+              };
+              setupConfig('PREPAID', false, fakeConfiguration);
+              return 'PREPAID';
+            }
+          );
+
+          // Waiting for the end of the second startup (with prepaid sim)
```
lodr: Just like you did here.

apps/costcontrol/test/unit/widget_startup_test.js
```diff
@@ -461,4 +453,94 @@ suite('Widget Startup Modes Test Suite >', function() {
...
+      var fakeResult = {
+        datausage: requestDataUsageResult,
+        balance: resultRequestPrepaid
+      };
+      this.sinon.stub(SimManager, 'isMultiSim').returns(true);
+      this.sinon.stub(window.CostControl, 'getInstance')
+            .yields(newCostControlRequest(fakeResult));
+
+      function assertPrepaidDataLayout() {
+        done(function() {
```
lodr: Please, return a promise and add the `done` callback explictely at the end `.then()` chains.

try-server-hook:
[gitmai Marina (gitmai)](https://github.com/gitmai) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=1159aaa0da12)

PR #27884, opened 2015-02-03
Bug 1118884 - Correct text alignment in canvas for rtl texts.

autra:


autolander:

PR #27896, opened 2015-02-04
Bug 1120541 - Do not maximize titlebar if it is fatal error +autoland

alivedise:


try-server-hook:
[alivedise Alive.Kuo (alivedise)](https://github.com/alivedise) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=6c3e1b7b96cc)

apps/system/js/app_window.js
```diff
@@ -818,7 +818,9 @@
           // Some signals that chrome needs to respond to can occur before
           // chrome has loaded - in those cases, manually call the handlers.
           if (that.inError) {
-            that.appChrome.handleEvent({type: 'mozbrowsererror'});
+            that.appChrome.handleEvent({type: 'mozbrowsererror', detail: {
+              type: 'non-fatal'
```
Cwiiis: I don't think we need to add this 'non-fatal' detail, as that type doesn't correspond to any kind of error message that the mozbrowsererror signal would produce. Either we should save the type when we receive the error or we should leave this empty like before, in my opinion.

alivedise: Fixing!

try-server-hook:
[alivedise Alive.Kuo (alivedise)](https://github.com/alivedise) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=94a2b58dc866)

try-server-hook:
[alivedise Alive.Kuo (alivedise)](https://github.com/alivedise) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=c21814760ed0)

PR #27925, opened 2015-02-04
Bug 1126279 - [Message][Refactoring] Refactor the Delete and markThreadRead methods in thread_list_ui.js

kumarrishav:


try-server-hook:
[kumarrishav Kumar Rishav (kumarrishav)](https://github.com/kumarrishav) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=4f9a860a2bec)

try-server-hook:
[kumarrishav Kumar Rishav (kumarrishav)](https://github.com/kumarrishav) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=69bf4cc278b4)

try-server-hook:
[kumarrishav Kumar Rishav (kumarrishav)](https://github.com/kumarrishav) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=282e768dda5a)

try-server-hook:
[kumarrishav Kumar Rishav (kumarrishav)](https://github.com/kumarrishav) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=a017b17bd228)

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7e0efadfc9dd35c9b20eb8025a7ed1defda8de93

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=bc62ebcf8b70e5c679a2768361117a8c48ecc2fa

apps/sms/js/thread_list_ui.js
```diff
@@ -402,10 +403,18 @@ var ThreadListUI = {
         thread.unreadCount = isRead ? 0 : 1;
         this.mark(thread.id, isRead ? 'read' : 'unread');
```
julienw: should this move to markReadUnread as well ?

julienw: both lines actually :)

apps/sms/js/thread_list_ui.js
```diff
@@ -393,7 +393,8 @@ var ThreadListUI = {
     }
   },
 
-  markReadUnread: function thlui_markReadUnread(selected, isRead) {
+  markReadUnreadUI: function thlui_markReadUnreadUI(selected, isRead) {
+    var threadToMark = [];
     selected.forEach((id) => {
```
julienw: if you use `filter` instead of `forEach`, you can make the code simpler. It takes a function: if the function returns "true", it keeps the element, otherwise it ignores it. Look at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter for more information.

apps/sms/js/thread_list_ui.js
```diff
@@ -402,10 +403,18 @@ var ThreadListUI = {
...
+		threadToMark.push(thread.id);
       }
     });
 
+    this.markReadUnread(threadToMark, isRead);
+  },
+
+  markReadUnread: function thlui_markReadUnread(threadToMark, isRead) {
+    threadToMark.forEach((id) => {
+        MessageManager.markThreadRead(id, isRead);
```
julienw: be careful to your indentation

apps/sms/js/thread_list_ui.js
kumarrishav: @julienw why this should move to markReadUnread as well?

julienw:
can you explain the idea of the patch? What is the purpose ?

apps/sms/js/thread_list_ui.js
julienw: because this is "thread.unreadCount = ..;" and "this.mark(..." are both part of the "markReadUnread" processus, right ?

kumarrishav:
@julienw  this patch has made, taking undo action in consideration, where we will update ui first and later gecko depends on user undo action (setimeout)..that's why i separate the ui updating and gecko modification part separately.

apps/sms/js/thread_list_ui.js
kumarrishav: It was part of it, initially to know which thread are to be work upon. But now, here i am storing all those thread id in threadToMark and passing that array with isRead to markReadUnread @julienw hope it clarify this

apps/sms/js/thread_list_ui.js
kumarrishav: is identation incorrect here? though i am taking care of it, as line is getting >80 most of time, so i have to break it into lines, while taking care of identation at same time.

apps/sms/js/thread_list_ui.js
kumarrishav: @julienw thanks for introducing to these awesome methods :) where were these methods hidden? ;)

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3235f68da7dfad8770bace2f4d726ed59c7c8924

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8b7afdb0c6f5cd2312c542fb2cf7ba48d96480a9

apps/sms/js/thread_list_ui.js
kumarrishav: Ah..got it, there i made mistake. But i don't know why this identation error wasn't coming in my editor.Anyway fix.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=57de5646d63b8d745c1931f475765bdb1e45cf0d

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b4f4e301b9f514e40204d82e785deebe29ef3928

apps/sms/js/thread_list_ui.js
```diff
@@ -393,17 +394,24 @@ var ThreadListUI = {
...
-        MessageManager.markThreadRead(thread.id, isRead);
+        ThreadListUI.mark(thread.id, isRead ? 'read' : 'unread');
       }
+      return markable;
+    }
+
+    ThreadListUI.markReadUnread(selected.filter(filterIds), isRead);
+  },
+
+  markReadUnread: function thlui_markReadUnread(threadToMark, isRead) {
```
julienw: I'd change how the functions are called:
* markReadUnreadUI => markReadUnread
* markReadUnread => performMarkReadUnread

"perform" gives the idea of really doing the action.

apps/sms/js/thread_list_ui.js
```diff
@@ -393,17 +394,24 @@ var ThreadListUI = {
     }
   },
 
-  markReadUnread: function thlui_markReadUnread(selected, isRead) {
-    selected.forEach((id) => {
+  markReadUnreadUI: function thlui_markReadUnreadUI(selected, isRead) {
+    function filterIds(id) {
```
julienw: The function name does not give a lot of information here; we already know it's a filter because we use it in the "filter" function. Moreover it does an action but we don't know it from the name.

Instead you can use something like: "markVisuallyAndFilterMarkable"

apps/sms/js/thread_list_ui.js
```diff
@@ -393,17 +394,24 @@ var ThreadListUI = {
...
       var thread  = Threads.get(id);
       var markable = thread && (!thread.hasDrafts || isRead);
```
julienw: does this mean we can't mark as "unread" a thread that has a draft ? I don't exactly remember why we did that. Because we thought that if there is a draft it can't be "unread" ?

apps/sms/js/thread_list_ui.js
```diff
@@ -393,17 +394,24 @@ var ThreadListUI = {
...
         thread.unreadCount = isRead ? 0 : 1;
-        this.mark(thread.id, isRead ? 'read' : 'unread');
-
-        MessageManager.markThreadRead(thread.id, isRead);
+        ThreadListUI.mark(thread.id, isRead ? 'read' : 'unread');
       }
+      return markable;
+    }
+
+    ThreadListUI.markReadUnread(selected.filter(filterIds), isRead);
```
julienw: please use `this` instead of `ThreadListUI`, it helps understanding that we call something on the same object.

apps/sms/js/thread_list_ui.js
```diff
@@ -393,17 +394,24 @@ var ThreadListUI = {
...
+    function filterIds(id) {
       var thread  = Threads.get(id);
       var markable = thread && (!thread.hasDrafts || isRead);
 
       if (markable) {
         thread.unreadCount = isRead ? 0 : 1;
-        this.mark(thread.id, isRead ? 'read' : 'unread');
-
-        MessageManager.markThreadRead(thread.id, isRead);
+        ThreadListUI.mark(thread.id, isRead ? 'read' : 'unread');
```
julienw: please use `this` instead of `ThreadListUI`. You can use an arrow function so that `this` is available here:

```javascript
var markVisuallyAndFilterMarkable = (id) => {
  ... // you can use "this"
};
```

apps/sms/js/thread_list_ui.js
```diff
@@ -393,17 +394,24 @@ var ThreadListUI = {
...
+      return markable;
+    }
+
+    ThreadListUI.markReadUnread(selected.filter(filterIds), isRead);
+  },
+
+  markReadUnread: function thlui_markReadUnread(threadToMark, isRead) {
+    threadToMark.forEach((id) => {
+        MessageManager.markThreadRead(id, isRead);
     });
```
julienw: Actually I think you can put everything on one line:
```javascript
threadToMark.forEach((id) => MessageManager.markThreadRead(id, isRead));
```

julienw: if it's more than 80 characters, you can do this:
```javascript
threadToMark.forEach(
  (id) => MessageManager.markThreadRead(id, isRead)
);
```

apps/sms/js/thread_list_ui.js
```diff
@@ -530,7 +490,64 @@ var ThreadListUI = {
...
+        filter: { threadId: threadId },
+        each: onThreadMessageRetrieved,
+        end: onAllThreadMessagesRetrieved
+      });
+    });
+  },
+
+  exitEditMode: function thlui_exitEditMode() {
+    ThreadListUI.cancelEdit();
+    WaitingScreen.hide();
```
julienw: actually, you can move `WaitingScreen.hide()` to `cancelEdit` and just use `cancelEdit` everywhere.

apps/sms/js/thread_list_ui.js
```diff
@@ -393,17 +394,24 @@ var ThreadListUI = {
...
+
+    ThreadListUI.markReadUnread(selected.filter(filterIds), isRead);
+  },
+
+  markReadUnread: function thlui_markReadUnread(threadToMark, isRead) {
+    threadToMark.forEach((id) => {
+        MessageManager.markThreadRead(id, isRead);
     });
 
     this.cancelEdit();
```
julienw: we should call cancelEdit after the UI change, not after the "real" change (especially considering your next patch).

apps/sms/js/thread_list_ui.js
```diff
@@ -449,75 +457,27 @@ var ThreadListUI = {
...
         if (isDraft) {
-          Drafts.delete(Drafts.get(threadId));
-          ThreadListUI.removeThread(threadId);
+          draftIds.push(threadId);
         } else {
           list.push(threadId);
         }
-
         return list;
       }, []);
```
julienw: tell me what you think, but I think this loop is not really useful now; maybe you should remove it and just have one single loop in `deleteThreadDraftUI` instead?

apps/sms/js/thread_list_ui.js
```diff
@@ -530,7 +490,64 @@ var ThreadListUI = {
         text: 'delete',
         className: 'danger'
       }
-    ).then(performDeletion.bind(this));
+    ).then(performDeletion.bind(this, selected));
```
julienw: we don't really need performDeletion if we remove the loop, we can move what's remaining of performDeletion to `deleteThreadDraftUI` directly and call it directly here. Wha

apps/sms/js/thread_list_ui.js
```diff
@@ -530,7 +490,64 @@ var ThreadListUI = {
...
+  },
+
+  deleteThreadDraftUI:
```
julienw: call this function `deleteThreadDraft` and the other function `performDeleteThreadDraft`

apps/sms/js/thread_list_ui.js
kumarrishav: Yeah..that's what we decided (steve pitched this..) that time.  Though we can do 'unread' even if thread has a draft.

apps/sms/js/thread_list_ui.js
kumarrishav: Yeah That's also fine. But then,  there is no need of delete function.. right? we will have 'deleteThreadDraftUI ' instead. 
I didn't remove the delete method (and that loop) because of one reason.. delete method is used at many places (and test too). So i didn't modify the name of 'delete' method name.

apps/sms/js/thread_list_ui.js
kumarrishav: same as above reason :)

apps/sms/js/thread_list_ui.js
kumarrishav: ok

apps/sms/js/thread_list_ui.js
kumarrishav: ok

apps/sms/js/thread_list_ui.js
```diff
@@ -530,7 +490,64 @@ var ThreadListUI = {
...
+      });
+    }
+    if (!threadIds.length) {
+      ThreadListUI.exitEditMode();
+    } else if(threadIds) {
+      threadIds.forEach(function(threadId) {
+        ThreadListUI.deleteThread(threadId);
+      });
+    }
+    ThreadListUI.deleteThreadDraft(selected, draftIds, threadIds);
```
julienw: try to use "this" instead of "ThreadListUI" everywhere you can

julienw: and when you can't, find a way to be able to :D

apps/sms/js/thread_list_ui.js
julienw: I think the "delete" function is still useful: call confirm, then perform deletion (and the "perform deletion" part is done by "deleteThreadDraftUI")

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b9accfede284e2f934d6326bc02be6503bd91f1f

apps/sms/js/thread_list_ui.js
```diff
@@ -394,19 +395,24 @@ var ThreadListUI = {
   },
 
   markReadUnread: function thlui_markReadUnread(selected, isRead) {
-    selected.forEach((id) => {
+    var filterIds = (id) => {
```
julienw: I said in my previous review to rename this to something like "markVisuallyAndFilterMarkable" to make it clear that it's marking visually messages as read as well as filtering.

apps/sms/js/thread_list_ui.js
julienw: For `this.mark` I think you're right, but I'm not sure for thread.unreadCount. OK maybe you're right for now but when we'll have a DB in Gaia maybe we'll need something more subtle.

apps/sms/js/thread_list_ui.js
```diff
@@ -394,19 +395,24 @@ var ThreadListUI = {
...
-    });
+      return markable;
+    };
 
     this.cancelEdit();
+    this.performMarkReadUnread(selected.filter(filterIds), isRead);
+  },
+
+  performMarkReadUnread:
+    function thlui_performMarkReadUnread(threadToMark, isRead) {
```
julienw: nit: put the "function" part on the same column than "performMarkReadUnread", I think this is a little more readable.

apps/sms/js/thread_list_ui.js
```diff
@@ -449,88 +455,80 @@ var ThreadListUI = {
...
-        args: { n: selected.length }
-      },
-      null,
-      {
-        text: 'delete',
-        className: 'danger'
-      }
-    ).then(performDeletion.bind(this));
+  exitEditMode: function thlui_exitEditMode() {
+    this.cancelEdit();
```
julienw: nit: then we don't need exitEditMode at all :) just use cancelEdit everywhere !

julienw: and actually that's what you do already.. so you can just remove this function :)

apps/sms/js/thread_list_ui.js
```diff
@@ -323,7 +323,8 @@ var ThreadListUI = {
           params.items.push(
             {
               l10nId: l10nKey,
-              method: this.markReadUnread.bind(this, [parentThreadId], isRead)
+              method:
+                this.markReadUnread.bind(this, [+parentThreadId], isRead)
```
julienw: nit: I think you can put this line on the previous line now :)

apps/sms/js/thread_list_ui.js
```diff
@@ -449,88 +455,80 @@ var ThreadListUI = {
...
-      }, []);
+  performDeleteThreadDraft:
+    function thlui_performDeleteThreadDraft(selected, draftIds, threadIds) {
+    var threadCountToDelete = 0, messageIdsToDelete = [];
 
-      // That means that we've just removed some drafts
-      if (threadIdsToDelete.length !== selected.length) {
-        Drafts.store();
-      }
+    if(draftIds) {
```
julienw: do you need the "if"? I think we always have an array as argument

apps/sms/js/thread_list_ui.js
```diff
@@ -449,88 +455,80 @@ var ThreadListUI = {
...
+      draftIds.forEach(function(threadId) {
+        Drafts.delete(Drafts.get(threadId));
+      });
+    }
+    if (threadIds.length !== selected.length) {
```
julienw: `if (draftsIds.length)` ?

apps/sms/js/thread_list_ui.js
```diff
@@ -449,88 +455,80 @@ var ThreadListUI = {
...
+      Drafts.store();
+    }
 
-      if (!threadIdsToDelete.length) {
-        exitEditMode();
-        return;
+    var onAllThreadMessagesRetrieved = () => {
+      if (!--threadCountToDelete) {
+        MessageManager.deleteMessages(messageIdsToDelete);
+        this.cancelEdit();
```
julienw: I think we never call cancelEdit if there is no thread to delete (means only drafts)

apps/sms/js/thread_list_ui.js
```diff
@@ -449,88 +455,80 @@ var ThreadListUI = {
...
+      var threadId = +id;
+      var isDraft = typeof Threads.get(threadId) === 'undefined';
+
+      if (isDraft) {
+        draftIds.push(threadId);
+        this.removeThread(threadId);
+      } else {
+        threadIds.push(threadId);
+        this.deleteThread(threadId);
       }
```
julienw: wondering if we can call deleteThread for both drafts and threads... (I don't know if bug 1084298 changes this BTW)

julienw:
you'll need a rebase, likely because of Steve's patch

apps/sms/js/thread_list_ui.js
```diff
@@ -449,88 +455,80 @@ var ThreadListUI = {
...
-      }
-
-      function onThreadMessageRetrieved(message) {
-        messageIdsToDelete.push(message.id);
-        return true;
+    selected.forEach((id) => {
+      // Coerce the threadId back to a number MobileMessageFilter and all
+      // other platform APIs expect this id to be a number.
+      var threadId = +id;
+      var isDraft = typeof Threads.get(threadId) === 'undefined';
```
julienw: this condition needs to be changed after Steve's patch (ad4b2eba8)

julienw: new condition is simpler: `Threads.get(threadId).isDraft`

apps/sms/js/thread_list_ui.js
julienw: in Steve's patch (ad4b2eb) we actually use deleteThread for drafts too...  but maybe delete is destructive so maybe we don't want to do this so early?

apps/sms/js/thread_list_ui.js
```diff
@@ -449,88 +455,80 @@ var ThreadListUI = {
...
-      if (threadIdsToDelete.length !== selected.length) {
-        Drafts.store();
-      }
+    if(draftIds) {
+      draftIds.forEach(function(threadId) {
+        Drafts.delete(Drafts.get(threadId));
+      });
+    }
+    if (threadIds.length !== selected.length) {
+      Drafts.store();
```
julienw: I think we don't need this after ad4b2eb now, it's done in Drafts.delete()

PR #27931, opened 2015-02-04
Bug 1102481 - Display "no results" when there is none.

hfiguiere:


apps/music/js/ui/views/search_view.js
```diff
@@ -30,6 +30,15 @@ var SearchView = {
...
 
+  showNoResult: function sv_showNoResult(show) {
+    var view = document.getElementById('views-search-no-result');
+    if (show) {
+      view.classList.remove('hidden');
+    } else {
+      view.classList.add('hidden');
+    }
+  },
+
```
dominickuo: Probably we can write `view.classList.toggle('hidden', !show)`?

try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=a19dd7240b97)

apps/music/js/ui/views/search_view.js
hfiguiere: Good call. Fixing.

PR #27932, opened 2015-02-04
Bug 1068999 - Make video player seek bar accessible.

eeejay:


shared/js/media/video_player.js
```diff
@@ -237,6 +243,9 @@ function VideoPlayer(container) {
   // Set the video duration when we get metadata
   player.onloadedmetadata = function() {
     durationText.textContent = formatTime(player.duration);
+    slider.setAttribute('aria-valuemax', player.duration);
+    navigator.mozL10n.setAttributes(slider, 'playbackSeekBar',
+                                    {'duration': formatTime(player.duration)});
```
davidflanagan: define a variable here for the formatted duration so you only have to call formatTime() once in this function

davidflanagan: Could you add a comment on the setAttributes() line to say that this is actually setting the aria-label attribute? Otherwise it is completely unclear what is going on here without referring to the locale file.  Is that actually what this line is doing?

shared/js/media/video_player.js
```diff
@@ -471,6 +483,20 @@ function VideoPlayer(container) {
     }
   });
 
+  slider.addEventListener('keypress', function(e) {
+    // Currently, we only allow screen reader users to adjust sliders with a
+    // constant step size (there is no page up/down equivalent). In the case
+    // of videos, we make sure that the maximum amount of steps for the entire
+    // duration is 20, or 2 second increments if the duration is less then 40
+    // seconds.
```
davidflanagan: Please expand this comment a bit to actually say that even though we're on a device without arrow keys, screen readers on the device will send up and down key events to control sliders like this.  Otherwise future maintainers of the app won't understand this.

shared/js/media/video_player.js
```diff
@@ -471,6 +483,20 @@ function VideoPlayer(container) {
...
+    var step = Math.max(player.duration/20, 2);
+    if (e.keyCode == e.DOM_VK_DOWN) {
+      player.currentTime -= step;
+    } else {
```
davidflanagan: Please explicitly check for VK_UP here instead of just using else.  In the past, hardware buttons on the device have sent key events and there is talk of exposing those to apps. So if something else starts sending HOME or ESCAPE events, for example, I don't want those trigger this code.

davidflanagan: Is this actually working and testable yet? I remember Punam mentioning these key events in some other patch and saying that she wanted to wait until it was actually testable before landing.

shared/js/media/video_player.js
eeejay: Done.

shared/js/media/video_player.js
eeejay: Added:
    // The standard accessible control for sliders is arrow up/down keys.
    // Our screenreader synthesizes those events on swipe up/down gestures.

shared/js/media/video_player.js
eeejay: The screen reader is currently broken in gecko-37 and m-c. A patch in [bug 1125422](https://bugzilla.mozilla.org/show_bug.cgi?id=1125422) is about to land in inbound once the tree opens (and should eventually be uplifted to 37).

But generally, swipe up/down is already implemented and works great. I tested this locally with the patches from the bug above applied.

eeejay: Good catch on VK_UP, added.

try-server-hook:
[eeejay Eitan Isaacson (eeejay)](https://github.com/eeejay) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=d66993c204f6)

PR #27939, opened 2015-02-04
Bug 1129563 - use window.location.hash instead of mozHasPendingMessages() r=pdahiya

davidflanagan:


autolander:

davidflanagan:
You're freaking kidding me, autolander!

KevinGrandon:
Haha yes.. Autolander is currently very picky about formats. I'm not sure if it's a good or bad thing. It is a pain, but it's also nice to have all of our commits in the same format. It might also complain when you go to land this - it does validation of the commit messages as well. You'd have to rebase and rename the commit to use a dash.

apps/gallery/js/gallery.js
```diff
@@ -179,42 +179,35 @@ function init() {
...
     setView(LAYOUT_MODE.list);
   }
 
-  // Register a handler for activities. This will take care of the rest
-  // of the initialization process.
+  // This function handles pick and browse activity requests
   navigator.mozSetMessageHandler('activity', function activityHandler(a) {
     var activityName = a.source.name;
     switch (activityName) {
     case 'browse':
```
punamdahiya: nit: case statements needs to be indented

davidflanagan: All my case labels come out this way, throughout the gallery app code base. That's the way emacs indents switch statements, and I've never tried to figure out how to change it. 

I think it may be an old C convention. The K&R book does it this way. And logically, the entire switch statement is a single block of code with entry points labelled by the cases. Because we can fall through from one block to the next, one could argue (though I'm not sure I would) that double indentation is incorrect because it implies that each case is an independent block.

In any case, my code is littered with this kind of indentation, and I don't think it makes sense to fix just one of them in his patch, especially since the relevant lines aren't actually modified by the patch.

PR #27943, opened 2015-02-04
Bug 1129568 - remove media queries for tablet support to improve startup time

davidflanagan:


apps/gallery/index.html
```diff
@@ -7,10 +7,7 @@
     <title>Gallery</title>
     <link rel="stylesheet" href="style/gallery.css" type="text/css" />
     <link rel="stylesheet" href="style/VideoPlayer.css" type="text/css" />
-    <link rel="stylesheet" href="style/VideoPlayer_tablet.css" type="text/css" media="(min-height: 768px) and (min-width: 768px)" />
```
jimporter: Is it possible to completely remove these files? If we ever need them back, we can always dig through the history later.

PR #27953, opened 2015-02-05
Bug 1117652 - Cursor should not show up in drop down menu

cctuan:
https://bugzilla.mozilla.org/show_bug.cgi?id=1117652

### DIFF
--- a/apps/clock/style/alarm.css
+++ b/apps/clock/style/alarm.css
@@ -75,6 +75,13 @@ ul.compact li label:active {
   display: none;
 }
 
+#time-select {
+  /* Bug 1117652: copypaste bubble would be triggered if tapping or longpressing
+                  on the input field, so we let the cursor out of the screen to
+                  prevent it. */
+  text-indent: 110%;
+}
+
 /**
  * 1. COMPLEX: Scoped style-sheets within <gaia-header>
  * trump all other rules. !important is required
--- a/apps/calendar/style/forms.css
+++ b/apps/calendar/style/forms.css
@@ -48,6 +48,10 @@ form li {
   opacity: 0;
   /* input should not be selectable otherwise it might affect the layout */
   -moz-user-select: none;
+  /* Bug 1117652: copypaste bubble would be triggered if tapping or longpressing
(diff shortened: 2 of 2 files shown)


try-server-hook:
[cctuan George (cctuan)](https://github.com/cctuan) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=f8b8c745080b)

apps/clock/style/alarm.css
```diff
@@ -75,6 +75,10 @@ ul.compact li label:active {
   display: none;
 }
 
+#time-select {
+  text-indent: 1000rem;
```
millermedeiros: are you sure that this is the best way to solve the issue? if so, please add a comment explaining why we need the text-indent (I know it's to move the caret away from the screen but we might forget it in the future and re-introduce the bug, specially since there is no way to write a test for it).

I assume the same problem is also present in all `<input type="time" />` (at least on the calendar app it does happen) so I would rather see this solved globally (eg. `shared/style/input_areas.css`)

millermedeiros: what about using a different unit instead of `rem`? that way the hack would work even if the field is wider than `1000rem` (bigger devices, 4K, etc..) - I did a test and `110%` seems to work

try-server-hook:
[cctuan George (cctuan)](https://github.com/cctuan) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=c837bf1a54b2)

apps/clock/style/alarm.css
```diff
@@ -75,6 +75,10 @@ ul.compact li label:active {
   display: none;
 }
 
+#time-select {
+  text-indent: 110%;
```
millermedeiros: would you mind adding a comment explaining why the `text-indent` is needed and also fixing the calendar app? thanks!

try-server-hook:
[cctuan George (cctuan)](https://github.com/cctuan) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=575534e6eeb7)

PR #27972, opened 2015-02-05
Bug 1124624 - Add transition to the group view panel

yvtheja:


try-server-hook:
[ythej Vishnu Teja (ythej)](https://github.com/ythej) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=5b5620cde945)

julienw:
You need to rebase

apps/sms/style/sms.css
```diff
@@ -302,21 +302,34 @@ form[role="dialog"][data-type="edit"] .bb-tablist > li {
...
+  transition: transform 0.4s ease;
 }
 
-.panel {
+#thread-messages {
   position: absolute;
   left: 100%;
   width: 100%;
   height: 100%;
+  float: left;
```
julienw: you don't need a float, I think ?

apps/sms/style/sms.css
```diff
@@ -302,21 +302,34 @@ form[role="dialog"][data-type="edit"] .bb-tablist > li {
   background: #fff;
 }
 
-.panel:first-child {
+#thread-list {
+  position: absolute;
   left: 0;
+  width: 100%;
+  height: 100%;
+  float: left;
```
julienw: same here, no float needed

julienw: or no position: absolute :) but you choose one or the other

apps/sms/style/sms.css
```diff
@@ -302,21 +302,34 @@ form[role="dialog"][data-type="edit"] .bb-tablist > li {
...
 
-#main-wrapper[data-position='left'] .panel {
-  transform: translateX(-100%);
+#main-wrapper[data-position="panel-position-1"] {
+  margin-left: 0;
+  float: left;
+}
+
+#main-wrapper[data-position="panel-position-2"] {
+  margin-left: -100%;
```
julienw: better use transformations: transform: translate(-100%) works fine !

julienw:
The issue you have is that the wrapper has a "width: 100%" rule, and "overflow: hidden".

If you keep "position: absolute", you can just remove "overflow: hidden" (and the width rule too, I don't think it's useful) and this should work.

If you use floats, you'll need to increase "width" to "300%" so that the floats are correctly located left to right.

apps/sms/style/sms.css
julienw: using transformations we'll have nice transitions that are handled by the GPU, which is really smooth on device. If you use margins, this won't be handled by the GPU and it will take CPU and be jerky on a device.

ythej:
@julienw , regarding your previous comment, why both `position: absolute` and `overflow: hidden` are included before ?

julienw:
Because before we always had this space that has a width of "100%", and we were transitioning inside this space.

With the new patch, we're transitioning the whole space, and the whole space has a width of 300% (because 3 panels).

with "overflow: hidden" we don't see the content that goes beyond the width.

Hope this is clear !

With your patch we'll also need to measure whether we have some overpainting.  I'm not sure how we can measure this though, I'll ask my colleagues tomorrow.

ythej:
@julienw , I am not yet trying to implement the third panel, I though of making the current transition work first. I tried removing `overflow: hidden` also but didn't work.

julienw:
If you tried something I'll need to see the full code (because CSS properties interact with each other :) )

ythej:
@julienw , I am creating new PR as it's little difficult to rebase on master this time.

julienw:
whatever is easier for you !
Please close this PR when you want then :)

But I'm not sure how this will fit in the new architecture.. maybe this will not be used in the end because the group view panel will be in its own separate file soon..

ythej:
I am really sucked with getting my passport, sorry that I took so long for this bug. But now I can see real form of corruption in India !! :(

I will be back on this bug as soon as I am clear with passport things.

Thank you.

PR #27976, opened 2015-02-05
Bug 1117444 - improve app->homescreen performance

etiennesegonzac:


try-server-hook:
[etiennesegonzac Etienne Segonzac (etiennesegonzac)](https://github.com/etiennesegonzac) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=8834a637c808)

apps/verticalhome/js/app.js
```diff
@@ -219,6 +221,12 @@
           window.removeEventListener('gaiagrid-layout-ready', this);
           break;
 
+        case 'visibilitychange':
+          setTimeout(function() {
+            document.body.style.overflow = document.hidden ? 'hidden' : '';
```
Cwiiis: heh, nice :) I've been looking at similar hacks to this elsewhere too... It's a shame, because it indicates that our progressive tiled rendering code might not be working (and I think the cancelling code definitely isn't), but even if it was, it will make Gecko more responsive.

apps/system/js/app_window.js
```diff
@@ -1852,7 +1852,7 @@
           return;
         }
         invoked = true;
-        setTimeout(callback);
+        callback();
```
Cwiiis: It makes sense to remove this one, however - tryWaitForFullRepaint will always call the function in a timeout already.

apps/system/js/app_window.js
```diff
@@ -1841,7 +1841,7 @@
...
```
Cwiiis: Looking at the implementation of waitForNextPaint, it's possible that the callback can be called immediately (happens if either the browser or browser.element are null) - This is ok if this can't happen in ready (which I guess it can't?), but otherwise there's the chance the callback will execute before the code below.

I'd say that perhaps it makes sense to remove this setTimeout, but to add one again in browser_mixin.js's implementation of waitForNextPaint.

apps/verticalhome/js/app.js
Cwiiis: p.s. this was something I meant to have a look at to do during rotation transitions too - we should totally do it there also...

try-server-hook:
[etiennesegonzac Etienne Segonzac (etiennesegonzac)](https://github.com/etiennesegonzac) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=0cf7727e9b01)

try-server-hook:
[etiennesegonzac Etienne Segonzac (etiennesegonzac)](https://github.com/etiennesegonzac) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=52f04eec744b)

try-server-hook:
[etiennesegonzac Etienne Segonzac (etiennesegonzac)](https://github.com/etiennesegonzac) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=c250d2ebdb54)

apps/verticalhome/js/app.js
Cwiiis: One last comment about this - we should add a comment as to what this is doing (hiding overflow to stop displayport rendering). Also, is it necessary for it to be in a timeout?

apps/system/js/app_window.js
```diff
@@ -1830,33 +1844,42 @@
     }
 
     this.debug('requesting to open');
+
+    // The homescreen was scheduled for a setVisible(false) but we need it
+    // right away.
+    if (this.isHomescreen && this._visibilityChangeTimeout) {
+      this.setVisible(true);
+      callback();
```
Cwiiis: Sometimes this callback is called in a setTimeout, other times it's called immediately - do we want/need to be consistent on this?

shared/elements/gaia_grid/js/grid_view.js
```diff
@@ -10,7 +10,7 @@
   /* The time for which we'll disable launching other icons after tapping on
    * an app icon. This will also disable edit mode.
    */
-  const APP_LAUNCH_TIMEOUT = 3000;
+  const APP_LAUNCH_TIMEOUT = 1000;
```
Cwiiis: This seems like an unrelated change(?)

apps/verticalhome/test/marionette/app_launch_visibility_test.js
```diff
@@ -0,0 +1,75 @@
...
+    appInstall.installPackage(server.packageManifestURL);
+
+    client.switchToFrame(system.getHomescreenIframe());
+  });
+
+  teardown(function(done) {
+    server.close(done);
+  });
+
+  test('Update overflow for fast resume once hidden', function() {
```
Cwiiis: My objection to this test is that it doesn't actually test that overflow is hidden during the animation, only that it is hidden when an application is launched and removed when the home button is pressed. I think such a test would probably be easier as a unit test, and no less valuable... As a sanity-check, I guess this works, but I question its usefulness.

apps/verticalhome/test/marionette/app_launch_visibility_test.js
```diff
@@ -0,0 +1,75 @@
...
+  setup(function() {
+    subject = client.loader.getAppClass('verticalhome');
+    system = client.loader.getAppClass('system');
+    appInstall = new AppInstall(client);
+
+    system.waitForStartup();
+    subject.waitForLaunch();
+
+    client.switchToFrame();
+    appInstall.installPackage(server.packageManifestURL);
```
Cwiiis: Is there a reason to do this and not just launch any arbitrary app (and thus cut the size of this test down)?

apps/verticalhome/js/app.js
etiennesegonzac: yep (for the timeout), will add the comment

shared/elements/gaia_grid/js/grid_view.js
etiennesegonzac: If you press home just after taping an icon on the homescreen (probably to tap another icon) the visibility of the homescreen stays the same. And without a visibility change the homescreen won't let you launch another app before `APP_LAUNCH_TIMEOUT`.
I'm open to suggestions if there's a better way to do this than lowering the timeout :)

apps/verticalhome/test/marionette/app_launch_visibility_test.js
etiennesegonzac: all the other homescreen tests are "independent" of gaia apps, thought it was on purpose :)

etiennesegonzac: I'll probably switch to a unit test anyway, it'll help covering the `setTimeout` too.
Thanks for the recommendation!

try-server-hook:
[etiennesegonzac Etienne Segonzac (etiennesegonzac)](https://github.com/etiennesegonzac) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=d4c55299fc3c)

shared/elements/gaia_grid/js/grid_view.js
Cwiiis: Could we just cancel the timeout on home button press? I'm not too concerned about this, so not a blocker.

etiennesegonzac: are we dispatching the home event to apps? 

(might go away anyway if I remove the setVisible delay)

Cwiiis: I'm not sure, I see code listening to the hashchange event, is that equivalent?

try-server-hook:
[etiennesegonzac Etienne Segonzac (etiennesegonzac)](https://github.com/etiennesegonzac) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=5656b5647ce1)

apps/system/js/app_window.js
```diff
@@ -214,14 +214,9 @@
   };
 
   /**
-   * Represent the current pagee visibility state,
-   * i.e. what is currently visible. Possible value:
-   * 'foreground': setVisible(true)
-   * 'background': setVisible(false)
-   *
-   * Default value is foreground.
+   * Represent the current page visibility state
```
Cwiiis: nit, s/Represent/Represents/

apps/system/js/app_window.js
```diff
@@ -265,6 +260,11 @@
    */
   AppWindow.prototype.setVisible =
     function aw_setVisible(visible) {
+      if (this._visible == visible) {
```
Cwiiis: nit, s/==/===/ (I guess? I suppose if you want to support calling it with non-boolean values, maybe == is better?)

etiennesegonzac: No you're right I don't think we want to support non-boolean values.

try-server-hook:
[etiennesegonzac Etienne Segonzac (etiennesegonzac)](https://github.com/etiennesegonzac) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=30d49a460194)

try-server-hook:
[etiennesegonzac Etienne Segonzac (etiennesegonzac)](https://github.com/etiennesegonzac) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=083dd97f4262)

PR #28005, opened 2015-02-05
Bug 1069615 - [Music] Make main music player views exclusively visible.

eeejay:


try-server-hook:
[eeejay Eitan Isaacson (eeejay)](https://github.com/eeejay) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=d0e74640e0ab)

try-server-hook:
[eeejay Eitan Isaacson (eeejay)](https://github.com/eeejay) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=f57ab44a1c96)

apps/music/style/music.css
```diff
@@ -193,6 +193,11 @@ body.tiles-mode #views {
   transform: translateX(0%);
 }
 
+body.tiles-mode #views-tiles,
+body.search-from-tiles-mode #views-tiles {
```
dominickuo: @eeejay When searching from tiles mode, I think `#views-tiles` should be hidden because we will only display the search results.

apps/music/style/music.css
```diff
@@ -206,9 +211,16 @@ body.search-from-list-mode #views {
   transform: translateX(-100%) translateY(-5rem);
 }
 
+body.list-mode #views-list,
+body.picker-mode #views-list,
+body.search-from-list-mode #views-list {
```
dominickuo: Same here, `#views-list` is also hidden when we entered the mode for searching from list.

apps/music/style/music.css
```diff
@@ -266,12 +290,15 @@ body.picker-mode #title-player {
...
-  transition: transform 300ms ease;
+  transition: transform 300ms ease, visibility 300ms;
+  visibility: hidden;
 }
 
 .view {
   width: 100%;
   height: auto;
+  visibility: hidden;
+  transition: visibility 300ms;
```
dominickuo: I checked the spec and it does said the `visibility` property supports transition, but do we actually see it when we apply it? maybe we just simply show/hide the views by changing the `visibility` style without transition, and possibly gain some fps.

apps/music/style/music.css
```diff
@@ -266,12 +290,15 @@ body.picker-mode #title-player {
   overflow-y: auto;
   background: -moz-linear-gradient(center top, #303030 0%, #000000 100%) repeat scroll 0 0 transparent;
   z-index: 10;
-  transition: transform 300ms ease;
+  transition: transform 300ms ease, visibility 300ms;
```
dominickuo: Same here.

apps/music/style/music.css
eeejay: `#views-tiles-search` lives inside `#views-tiles`, so that won't work. I think you are right that the actual tiles should be hidden, so I'll fix that with a new rule.

apps/music/style/music.css
eeejay: Same issue as above. `#views-list-anchor` needs to be hidden separately.

apps/music/style/music.css
eeejay: Not sure what the issue you see here is.

apps/music/style/music.css
eeejay: Not sure what you mean. You want to get rid of the animations? That would require some UX changes, no?

try-server-hook:
[eeejay Eitan Isaacson (eeejay)](https://github.com/eeejay) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=960079e2b263)

apps/music/style/music.css
dominickuo: @eeejay What I mean is, does the`visibility` transition actually changes the animations after it's combined with the other existing transitions? cause I try to check with my eye but didn't notice obvious changes on the animations.

dominickuo:
@eeejay After applied this patch and as you can see, the music open activity has hidden the wrong elements, it's a noticeable issue and you should be able to address it by adding the visibility style to `#views-player.open-mode` class in https://github.com/mozilla-b2g/gaia/blob/master/apps/music/style/music.css#L330

apps/music/style/music.css
eeejay: Since it is a binary value (`visible`/`hidden`) it is not actually animated itself. `visibility` is designed to work correctly in conjunction with other transitions. For example, a transition from `visible` to `hidden` with 300ms duration will have it actually hidden at the end of 300ms. A transition from `hidden` to `visible` will work by having it becoming visible immediately. In this way, if the `visibility` property i combined with, for example, `opacity`, you will get the correct dissolve effect since the element will remain visible for the duration of the transition.

Did that answer your question? :)

try-server-hook:
[eeejay Eitan Isaacson (eeejay)](https://github.com/eeejay) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=4b930948cea5)

apps/music/style/music.css
dominickuo: Okay, that sounds good to me, thanks for explaining it!

PR #28025, opened 2015-02-06
Bug 1127939 - Wait for provider ready before searching.

daleharvey:


try-server-hook:
[daleharvey Dale  Harvey (daleharvey)](https://github.com/daleharvey) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=8b22974ad5c3)

apps/search/js/providers/suggestions.js
```diff
@@ -37,13 +37,15 @@
...
     search: function(input) {
-      return new Promise((resolve, reject) => {
-        var url = encodeTerms(SearchProvider('suggestUrl'), input);
-        LazyLoader.getJSON(url, true).then(result => {
-          var results = result[1];
-          // We add an item to search the entered term as well
-          results.unshift(result[0]);
-          resolve(results);
+      return SearchProvider.ready().then(function() {
+        return new Promise((resolve, reject) => {
```
KevinGrandon: Why do we need this nested promise? Can't we do all of this logic within the ```.then()```?

daleharvey: We will need to resolve with the results?

KevinGrandon: I think you can use the ```resolve``` as the first argument from the ```.then()```? So basically standard promise chaining..

```
return SearchProvider.ready().then(resolve => {

});
```

daleharvey: ah nice I didnt realise that would work, will do

try-server-hook:
[daleharvey Dale  Harvey (daleharvey)](https://github.com/daleharvey) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=6092992300c8)

apps/search/js/providers/suggestions.js
daleharvey: Changed it

daleharvey: err no you cant do that, .then resolves with the value of the promise its called on

apps/search/js/providers/suggestions.js
```diff
@@ -37,13 +37,15 @@
...
 
     search: function(input) {
-      return new Promise((resolve, reject) => {
-        var url = encodeTerms(SearchProvider('suggestUrl'), input);
-        LazyLoader.getJSON(url, true).then(result => {
-          var results = result[1];
-          // We add an item to search the entered term as well
-          results.unshift(result[0]);
-          resolve(results);
+      return SearchProvider.ready().then(resolve => {
```
KevinGrandon: Ok, I think I mis-read the code before, sorry.  You're right, I think then() should just return a new promise as you're doing here.

This looks good, I think you should remove the resolve here then, that was wrong.

```js
return SearchProvider.ready().then(() => {
```

PR #28026, opened 2015-02-06
Bug 1130479 - [Calendar][RTL] swipe direction should be mirrored on month view for RTL locales r=jrburke

millermedeiros:


try-server-hook:
[millermedeiros Miller Medeiros (millermedeiros)](https://github.com/millermedeiros) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=08032cfa6b66)

apps/calendar/js/views/month.js
```diff
@@ -56,7 +56,8 @@ Month.prototype = {
     if (Math.abs(data.dy) > (Math.abs(data.dx) - XSWIPE_OFFSET)) {
       return;
     }
-    this._move(data.dx < 0);
+    var delta = document.documentElement.dir === 'rtl' ? -1 : 1;
```
jrburke: "delta" as a word to me implies a difference, as in the result of a subtraction or comparison diff, where the + or - seems more like a direction change, so a different word for the variable name might be better. "dir" seems like a possible choice but I can understand if you do not want it to conflict with the "dir" meaning the text direction. In any case, your call on the name. The code works as expected when tested on flame device.

millermedeiros: I agree about the naming! `delta` was the wrong name for it.

PR #28029, opened 2015-02-06
Bug 1098289 - Migrate Email to use navigator.sync (RequestSync) API (Phase1)

jrburke:
Move from mozAlarms to navigator.sync API. The bulk of the changes were just name swaps:
- mozSetMessageHandler: from 'alarm' to 'request-sync'.
- mozAlarm and its DOMRequest style API to navigator.sync's promise API.
- naming local variables with 'alarm' to 'registrations' or a 'sync task', or 'task' in the name.

See the [gelam pull request](https://github.com/mozilla-b2g/gaia-email-libs-and-more/pull/368) for the email/js/ext and upgrade notes. For files tracked only in gaia:
# js/cards/settings_debug.js

It now has an option to print the sync tasks in the logcat. There is also the [requestsyncmanager](https://github.com/jrburke/requestsyncmanager) app that can be used to look at all sync tasks on the device. Hopefully it will be possible to use it to adjust the minInterval from a lower value than 100 second minimum that is enforced by default. See bug referenced in that project's README. The "fastSync" options are now 100 and 200 seconds because of the minimum in the request-sync code. I expect over time the requestsyncmanager would be used to shift syncs down to faster times for dev debugging.

https://mxr.mozilla.org/mozilla-central/source/dom/requestsync/ has the backing code for navigator.sync.
## js/sync.js

The change here was mostly to print out a number that was more useful than the plog timing, which was not really insightful. Now it shows how long the sync took once we started up the app.
## test/marionette/notification_set_interval_test.js

I simplified the test to avoid needing to mock the navigator.sync API. I took a pass at it, but it started to feel complicated with the promises .then returning other promises that all needed **exposedProps** to work, and in the end, I think the most value in the test is confirming that a UI change for the interval is properly persisted.
## shared/test/integration/mock_navigator_moz_set_message_handler.js

Expanded to allow 'request-sync' to work with this mock. Preserves the default of 'alarm' so other people's code does not have to change, but now allows request-sync to also work.

### DIFF
--- a/apps/email/js/app_messages.js
+++ b/apps/email/js/app_messages.js
@@ -115,8 +115,8 @@ define(function(require) {
     evt.on(
         'notification', appMessages.onNotification.bind(appMessages));
 
-    // Do not listen for navigator.mozSetMessageHandler('alarm') type, that is
-    // only done in the back end's cronsync for now.
+    // Do not listen for navigator.mozSetMessageHandler('request-sync')
+    // type, that is only done in the back end's cronsync for now.
   } else {
     console.warn('Activity support disabled!');
   }
--- a/apps/email/js/cards/settings_debug.html
+++ b/apps/email/js/cards/settings_debug.html
@@ -63,5 +63,9 @@
     <button data-event="click:fastSync"
             href="#"
             class="tng-dbg-fastsync">Enable faster sync options</button>
+
+    <button data-event="click:showSyncs" href="#">Show sync tasks
+    in logcat</button>
+
   </section>
 </section>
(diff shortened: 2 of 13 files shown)


try-server-hook:
[jrburke James Burke (jrburke)](https://github.com/jrburke) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=989432c4d126)

try-server-hook:
[jrburke James Burke (jrburke)](https://github.com/jrburke) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=185ef78fbbed)

PR #28030, opened 2015-02-06
Bug 1129572 - make MediaDB send an enumerable event before it is fully ready

davidflanagan:


try-server-hook:
[davidflanagan David Flanagan (davidflanagan)](https://github.com/davidflanagan) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=218dce63956f)

apps/gallery/js/gallery.js
```diff
@@ -493,8 +501,13 @@ function initThumbnails() {
     PerformanceTestingHelper.dispatch('media-enumerated');
 
     // Now that we've enumerated all the photos and videos we already know
-    // about go start looking for new photos and videos.
-    photodb.scan();
+    // about it is time to go and scan the filesystem for new ones. If the
+    // MediaDB is fully ready, we can scan now.  Either way, we always want
+    // to scan every time we get a new 'ready' event.
+    photodb.addEventListener('ready', function() { photodb.scan(); });
```
jimporter: How does this new event listener change the behavior of the gallery app? I'm guessing it fires whenever you unplug USB. Did Gallery just not handle that case before?

davidflanagan: It did handle the case before, but it was ugly code.  Previously the ready event always called initThumbnails().  And initThumbnails() had code that checked whether the thumbnails had already been created. If so, it just called scan() and exited.

This new logic is much nicer!

PR #28052, opened 2015-02-09
Bug 1130317 - Pre-determine length of Uint8Array of Keyboard user dictionary dict blob

mnjul:
- The total byte length of serialized nodes can be inferred when filling each node's offset information, and is subsequently combined with header & character table size to initialize the Uint8Array total length.
- We thus avoid pushing to a JS array one-byte-by-one-byte and convert the array to Uint8Array, which should lead to better performance.

### DIFF
--- a/apps/keyboard/js/settings/word_list_converter.js
+++ b/apps/keyboard/js/settings/word_list_converter.js
@@ -318,17 +318,17 @@ TSTSerializer.prototype.serializeToNodes = function() {
 
 
 // JSConv:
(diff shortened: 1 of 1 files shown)


try-server-hook:
[mnjul Min-Zhong "John" Lu (mnjul)](https://github.com/mnjul) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=1c287073d57f)

PR #28069, opened 2015-02-09
Bug 1128516 - Show entered term as search suggestion

daleharvey:


try-server-hook:
[daleharvey Dale  Harvey (daleharvey)](https://github.com/daleharvey) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=32e29741786d)

apps/search/js/providers/suggestions.js
```diff
@@ -44,8 +52,6 @@
 
     name: 'Suggestions',
 
-    remote: true,
```
KevinGrandon: Why was this removed? I think this is definitely still needed for private browsing cases.

apps/search/js/providers/suggestions.js
```diff
@@ -32,6 +36,10 @@
     suggestionsSelect.appendChild(selectFragment);
   });
 
+  SettingsListener.observe(enabledKey, true, function(enabled) {
```
KevinGrandon: Should we just access the state from the global Search object? It looks like we're already observing that setting there..

https://github.com/mozilla-b2g/gaia/blob/master/apps/search/js/search.js#L84

try-server-hook:
[daleharvey Dale  Harvey (daleharvey)](https://github.com/daleharvey) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=083f7c6745cf)

try-server-hook:
[daleharvey Dale  Harvey (daleharvey)](https://github.com/daleharvey) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=70d49b8cdb48)

apps/search/js/search.js
```diff
@@ -138,38 +138,35 @@
         Object.keys(providers).forEach((providerKey) => {
           var provider = providers[providerKey];
 
-          // If suggestions are disabled, only use local providers
-          if (this.suggestionsEnabled || !provider.remote) {
+          var fetchRemote =
+            UrlHelper.isURL(input) ||
+            msg.data.isPrivateBrowser ||
+            !this.suggestionsEnabled;
```
benfrancis: fetchRemote is true when you shouldn't fetch remotely? What am I missing?

apps/search/js/search.js
```diff
@@ -138,38 +138,35 @@
...
+            if (toShow) {
+              this.suggestionNotice.hidden = false;
+            }
           }
+
+          provider.search(input, fetchRemote).then((results) => {
+            this.collect(provider, results);
+          }).catch((err) => {
+            if (provider.remote) {
+              this.loadingElement.classList.remove('loading');
```
benfrancis: Because provider.remote is false for the suggestions provider, won't it be the case that this line never gets reached and the loading animation will continue to show?

apps/search/js/search.js
daleharvey: In the refactor the meaning for this switched and forgot to switch the name

apps/search/js/search.js
daleharvey: The loading icon is completely broken at the moment, I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1131016 to fix it, but you are right, I think its best to just take out the `classList.add` for now, never show it and add it again properly (I already have the patch done, but its based on this one)

daleharvey: Actually will just merge these bugs, easier

try-server-hook:
[daleharvey Dale  Harvey (daleharvey)](https://github.com/daleharvey) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=2ad4cacac259)

apps/search/js/providers/suggestions.js
KevinGrandon: The pull request in the current state seems to work, but I don't really like the removal of this. It kind of loses its meaning if we're not using it as before. It's also less consistent if not all remote providers have this.

Additionally it seems we're falling for the infamous "boolean trap" by adding this boolean argument.  Can we instead just read from the global Search object, and access Search.suggestionsEnabled? We already rely on the Search object in this file, so I believe this is cleaner.

daleharvey: Whether we search is dependant on whether its a url, a private browsing instance and whether searches are enabled. I can / have a patch to set a Search.preventRemoteFetch, however that is far less readable than the `boolean trap`

`remote: true` is removed because when suggestions are disabled the provider still needs to be called in order to render a default

We can possibly go down the 'make the single link a different provider' path again but there are issues with the shared logic (how it clears the containers / hides the heading) and I think its ultimately a more confusing than the current way

KevinGrandon: Ok, sounds good. I definitely think that we should remove the ```remote``` attribute from other providers, but we can do that in a follow-up. Thanks for investigating the alternative approach of using the global object. I think we're on the same page now, and this can probably land whenever you're ready.

try-server-hook:
[daleharvey Dale  Harvey (daleharvey)](https://github.com/daleharvey) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=a8eb1b0521aa)

try-server-hook:
[daleharvey Dale  Harvey (daleharvey)](https://github.com/daleharvey) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=2a555fa1d2d3)

PR #28104, opened 2015-02-10
Bug 1129844 - Keyboard UserDictionary displays word list as sorted alphabetically

mnjul:
- Only the UserDictionary model does the sorting. It returns a full list on each list modification call .
- The view maintains an extra word-to-domElem mapping and can re-order the domElems on retrieving the full list from model.
- A new UDListView.rearrangeList function replaces .appendList function, which is used to re-order domElems, and generate new domEleme for words that haven't bee displayed (such as initialization, add, and replace).

### DIFF
--- a/apps/keyboard/js/settings/user_dictionary_list_panel.js
+++ b/apps/keyboard/js/settings/user_dictionary_list_panel.js
@@ -6,12 +6,23 @@
  * The panel for the list of user dictionary words. When there is no word,
  * display a placeholder.
  *
- * We use a ES6 WeakMap to conveniently map a DOM element to the word it
- * contains. The architecture design is you never need to query a DOM element
- * from a word (a little funtional-programming sense).
+ * We want to keep the list in alphabetical order. The view does not do the
+ * sorting job -- the model keeps the list sorted, while the view merely
+ * reflects the order of the list and displays as such.
+ *
+ * Thus, the model is expected to return a full, sorted word list for each
+ * modification to the list. This includes adding a word, removing a word, and
+ * replacing a word. (The list isn't really used for removing a word, though).
+ *
+ * We use two-way mapping:
+ * - From user interaction of a word DOM element, to a word string value:
+ *   This is kept with an ES6 WeakMap.
+ * - From a word string value to a word DOM element (used to re-order a
+ *   displayed list according to the order returned from model):
+ *   This is kept with an object.
  *
  * When we're talking about "a word's DOM element", we're always referring to
- * the <a> element, and never its parent, the <li> element.
+ * the <li> element, and never its inner child, the <a> element.
  */
 
 (function(exports) {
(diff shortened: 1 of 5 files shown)


try-server-hook:
[mnjul Min-Zhong "John" Lu (mnjul)](https://github.com/mnjul) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=d16eddaa2cf7)

apps/keyboard/js/settings/user_dictionary_list_panel.js
```diff
@@ -106,16 +121,32 @@ UserDictionaryListPanel.prototype.handleEvent = function(evt) {
...
+  wordList.forEach(word => {
+    word = word.trim();
+
+    var wordLIElem;
 
-  var li = document.createElement('li');
-  li.appendChild(wordElem);
-  this._listContainer.appendChild(li);
+    if (!(word in this._wordDomMap)) {
+      var wordElem = document.createElement('a');
```
timdream: Could you move L133-141 to a new method?

apps/keyboard/js/settings/user_dictionary_list_panel.js
```diff
@@ -106,16 +121,32 @@ UserDictionaryListPanel.prototype.handleEvent = function(evt) {
...
+      wordElem.textContent = word;
 
-  this._domWordMap.set(wordElem, word);
+      wordLIElem = document.createElement('li');
+      wordLIElem.appendChild(wordElem);
+
+      this._domWordMap.set(wordElem, word);
+      this._wordDomMap[word] = wordElem;
+    } else {
+      wordLIElem = this._wordDomMap[word].parentNode;
```
timdream: This `parentNode` look strange to me. After some digging, I realized you are save the reference to `<a>` and we need to append the parent node (`<li>`) here, and the reason you save `<a>` is because you are using that reference in `handleEvent`.

Let me proposed an alternative here:
1. If we could get rid of `<a>` without mess up with the style, please do.
2. If we couldn't, how about use `pointer-event: none;` to make sure we are always getting `evt.target` points to `<li>`? That way, we could limit the `<a>` as a detail to be disregarded in JS.

If either thing is too much to do in this patch, for get about it.

apps/keyboard/style/settings.css
```diff
@@ -241,6 +241,10 @@ input:focus {
   display: none;
 }
 
+#ud-wordlist-list a{
```
timdream: nit: Missing a space.

try-server-hook:
[mnjul Min-Zhong "John" Lu (mnjul)](https://github.com/mnjul) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=e92fb7ad6b5f)

PR #28126, opened 2015-02-10
Bug 1069629 - Label repeat button.

eeejay:


try-server-hook:
[eeejay Eitan Isaacson (eeejay)](https://github.com/eeejay) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=97f510106b23)

apps/music/js/ui/views/player_view.js
```diff
@@ -274,6 +274,7 @@ var PlayerView = {
 
   setRepeat: function pv_setRepeat(value) {
     var repeatClasses = ['repeat-off', 'repeat-list', 'repeat-song'];
+    var repeatLabels = ['repeat-off', 'repeat-list', 'repeat-song'];
```
jimporter: I don't think we need this at the moment. We could just reuse `repeatClasses` (until we need to update strings, which may be never).

eeejay: Sure. I was worried that this would break if someone changed the class names and didn't notice it was used for strings too.

jimporter: I'd just change the variable name to something like `repeatModes`, and maybe add a comment that they're used for class names and L10N IDs.

try-server-hook:
[eeejay Eitan Isaacson (eeejay)](https://github.com/eeejay) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=f0f500f620f1)

PR #28127, opened 2015-02-10
Bug 1069624 - Make music shuffle button an ARIA toggle button.

eeejay:


apps/music/js/ui/views/player_view.js
```diff
@@ -300,6 +300,7 @@ var PlayerView = {
     } else {
       this.shuffleButton.classList.remove('shuffle-on');
     }
+    this.shuffleButton.setAttribute('aria-pressed', this.shuffleOption);
```
dominickuo: @eeejay I don't know the `aria-pressed` attribute well, but do yo think we can merge it with the `shuffle-on` class in some way? so that we don't have to apply them speartely.

eeejay: As a rule, we try to separate ARIA from styling, and not use those attributes in CSS rules. This allows us the freedom to adjust the ARIA without risking changing layout.

dominickuo: Okay, then I think that make sense to me.

try-server-hook:
[eeejay Eitan Isaacson (eeejay)](https://github.com/eeejay) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=9b3e61609abd)

PR #28134, opened 2015-02-11
Bug 1121904 - [Bluetooth][Settings] Show connect/disconnect/unpair dialog for a paired device/headset. r=arthur

ian-liu:


apps/settings/js/modules/bluetooth/bluetooth_connection_manager.js
```diff
@@ -0,0 +1,408 @@
...
+     * The method will disconnect the connected device currently
+     * with the device's adapter.
+     *
+     * @access private
+     * @memberOf BluetoothConnectionManager
+     * @returns {Promise}
+     */
+    _disconnectCurrentConnectedDevice: 
+    function btcm__disconnectCurrentConnectedDevice() {
+      return new Promise((resolve, reject) => {
```
crh0716: It is not required to wrap the code in a promise. We could return the promise chain starting from `this._getConnectedDevices`.

apps/settings/js/modules/bluetooth/bluetooth_connection_manager.js
```diff
@@ -0,0 +1,408 @@
...
+        if (!this.connectedAddress) {
+          // There is no connected device now. Resolve directly.
+          Debug('_disconnectCurrentConnectedDevice(): No connected device now');
+          resolve();
```
crh0716: Per the previous comment, we can return `Promise.reslove()`.

apps/settings/js/modules/bluetooth/bluetooth_connection_manager.js
```diff
@@ -0,0 +1,408 @@
...
+     * @param {Object BluetoothDevice} device
+     * @returns {Promise}
+     */
+    disconnect: function btcm_disconnect(device) {
+      return new Promise((resolve, reject) => {
+        if (!this._defaultAdapter) {
+          reject('default adapter is not existed!!');
+        }
+
+        var req = this._defaultAdapter.disconnect(device);
```
crh0716: Now a DOM request is actually a promise. We can use `this._defaultAdapter.disconnect(device).then`.

apps/settings/js/modules/bluetooth/bluetooth_connection_manager.js
```diff
@@ -0,0 +1,408 @@
...
+    _connectedDeviceItemsMap: {},
+
+    /**
+     * Default adapter of Bluetooth.
+     *
+     * @access private
+     * @memberOf BluetoothConnectionManager
+     * @type {Object BluetoothAdapter}
+     */
+    _defaultAdapter: null,
```
crh0716: Actually you may not require to store a reference to the default adapter. You can always get it directly from `AdadpterManager`.

apps/settings/js/modules/bluetooth/bluetooth_connection_manager.js
```diff
@@ -0,0 +1,408 @@
...
+        return Promise.reject('default adapter is not existed!!');
+      }
+
+      // In code base v1, it always stopDiscovery before connect with a device.
+      // Shall we stopDiscovery here too?
+
+      // Disconnect current connected device first.
+      Debug('connect(): Want to connect device address = ' + device.address +
+            ', name = ' + device.name);
+      return this._disconnectCurrentConnectedDevice().then(() => {
```
crh0716: IMHO it is not required to have a function designed specifically for disconnecting from the current connected device. We can do the work here.

apps/settings/js/modules/bluetooth/bluetooth_connection_manager.js
```diff
@@ -0,0 +1,408 @@
...
+    /**
+     * The method will connect the input device with the device's adapter.
+     *
+     * @access private
+     * @memberOf BluetoothConnectionManager
+     * @param {Object BluetoothDevice} device
+     * @returns {Promise}
+     */
+    _connect: function btcm__connect(device) {
+      return new Promise((resolve, reject) => {
```
crh0716: Same here, no need to wrap the code with a promise.

apps/settings/js/modules/bluetooth/bluetooth_connection_manager.js
```diff
@@ -0,0 +1,408 @@
...
+     */
+    _getConnectedDevices: function btcm__getConnectedDevices() {
+      return new Promise((resolve, reject) => {
+        // get connected device via Profiles.HFP
+        this._getConnectedDevicesByProfile(
+          this.Profiles.HFP).then((connectedDevices) => {
+          Debug('_getConnectedDevices(): connectedDevices = ' + 
+                JSON.stringify(connectedDevices));
+          // Update each connected devices in map.
+          this._updateDeviceItemsMap(this.Profiles.HFP, connectedDevices);
```
crh0716: This is confusing to me. It seems you always get the connected devices from gecko and construct the map everytime. If you would like to use `_connectedDeviceItemsMap` as a cache, you should return the cache directly and update the cache when there are change events from gecko notifying the connection change (although I'm not sure if such event exists).

apps/settings/js/panels/bluetooth/bt_template_factory.js
```diff
@@ -57,6 +62,11 @@ define(function() {
       _updateItemState(li, descSmall, statePaired);
     });
 
+    // sjdfis
```
crh0716: ?

apps/settings/js/modules/bluetooth/bluetooth_device.js
```diff
@@ -30,11 +30,15 @@ define(function(require) {
...
   return function ctor_bluetooth_device(device) {
     var type = BtClassOfDeviceMapper.getDeviceType(device.cod);
+    var connected = device.connected || false;  // default: false
     var observableBluetoothDevice = Observable({
       name: device.name,
       paired: device.paired,
       address: device.address,
-      type: type
+      type: type,
+      connected: connected,
```
crh0716: From the code the property would be set as different types of values and which is not desired. I would suggest to rename the property to `connectionStatus`. Possible values of it are all strings: 'connecting', 'connected', and 'disconnected'.

apps/settings/js/panels/bluetooth/panel.js
```diff
@@ -347,15 +356,108 @@ define(function(require) {
...
+            // Pop confirmation dialog to make sure user's decision.
+            this._confirmUserWantToUnpairDevice(deviceItem);
+            break;
+        }
+      },
 
-        // https://bugzilla.mozilla.org/show_bug.cgi?id=1121904
-        // Since cannot pair with headset, we do unpair here directly.
-        // unpair with the remote device
+      _createActionMenuByConnectedStatus: function(deviceItem) {
```
crh0716: nit: Suggest to use `_showActionMenu`

try-server-hook:
[ian-liu Ian Liu (ian-liu)](https://github.com/ian-liu) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=e14647d985f6)

try-server-hook:
[ian-liu Ian Liu (ian-liu)](https://github.com/ian-liu) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=af3122b1d14a)

apps/settings/js/modules/bluetooth/bluetooth_connection_manager.js
crh0716: Please ignore this comment. We need `_defaultAdapter` for unregister the handlers from it.

apps/settings/js/modules/bluetooth/bluetooth_connection_manager.js
```diff
@@ -0,0 +1,674 @@
...
+     * @param {Object BluetoothDevice} device
+     * @returns {Promise}
+     */
+    disconnect: function btcm_disconnect(device) {
+      if (!this._defaultAdapter) {
+        return Promise.reject('default adapter is not existed!!');
+      }
+
+      return this._defaultAdapter.disconnect(device).then(() => {
+        // this.connectedAddress = null;
```
crh0716: It seems `connectedAddress` is updated in `_updateConnectionStatus` right? If so, please remove these lines.

apps/settings/js/modules/bluetooth/bluetooth_connection_manager.js
```diff
@@ -0,0 +1,674 @@
...
+        // this._updateConnectionStatus(options);
+        Debug('disconnect(): connectedAddress = null, resolve');
+      }, () => {
+        Debug('disconnect(): onerror(): reject with disconnect failed');
+        return Promise.reject('disconnect failed');
+      });
+    },
+
+    /**
+     * The method will disconnect the input device with the device's adapter.
```
crh0716: Please update the doc.

apps/settings/js/modules/bluetooth/bluetooth_connection_manager.js
```diff
@@ -0,0 +1,674 @@
...
+ *   logic.
+ *
+ * @module BluetoothConnectionManager
+ */
+define(function(require) {
+  'use strict';
+
+  var AdapterManager = require('modules/bluetooth/bluetooth_adapter_manager');
+  var Observable = require('modules/mvvm/observable');
+  require('shared/async_storage');
```
crh0716: Suggest to use `var AsyncStorage = require('shared/async_storage');` so we can avoid using `window.AsyncStorage`.

apps/settings/js/modules/bluetooth/bluetooth_connection_manager.js
```diff
@@ -0,0 +1,674 @@
...
+    function btcm__onDefaultAdapterChanged(newAdapter, oldAdapter) {
+      Debug('_onDefaultAdapterChanged(): newAdapter = ' + newAdapter);
+      Debug('_onDefaultAdapterChanged(): oldAdapter = ' + oldAdapter);
+
+      // save default adapter
+      this._defaultAdapter = newAdapter;
+
+      if (oldAdapter) {
+        // unwatch event since the old adapter is no longer usefull
+        this._unwatchOnhfpstatuschanged(oldAdapter);
```
crh0716: As the handlers for the status changes of the profiles are the same (note that the only difference is the name of the profile), I would suggest to iterate all the profiles defined in `Profiles` to register/unregister the handlers.  The same thing also applies to `_updateConnectionStatus` and all other parts related to profiles.
By doing this we can support newly added profiles just by adding new profile names in `Profiles` without changing other parts.

apps/settings/js/modules/bluetooth/bluetooth_device.js
```diff
@@ -30,14 +30,105 @@ define(function(require) {
...
     /**
+     * Provide a function to update connection info.
+     *
+     * @access public
+     * @memberOf BluetoothDevice
+     * @param {Object} options
+     * @param {Boolean} options.connectionStatus - is connected or not
+     * @param {Object} options.profiles - profiles of the connection type
+     */
+    observableBluetoothDevice.updateConnectionInfo =
```
crh0716: We should declare the function in the outer scope and simply use the reference of it so the same function won't be created every time we create a new bluetooth device.
```
function updateConnectionInfo() {}
return function ctor_bluetooth_device(device) {
  var observableBluetoothDevice = Observable({
    updateConnectionInfo: updateConnectionInfo
  });
};

File: apps/settings/js/modules/bluetooth/bluetooth_device.js [hunk 16]
```diff
@@ -30,14 +30,105 @@ define(function(require) {
...
     var observableBluetoothDevice = Observable({
       name: device.name,
       paired: device.paired,
       address: device.address,
-      type: type
+      type: type,
+      connectionStatus: connectionStatus,
+      profiles: null,
+      descriptionText: '',
+      data: device
```
crh0716: Note that we can also define a function in this object.

apps/settings/js/modules/bluetooth/bluetooth_device.js
```diff
@@ -30,14 +30,105 @@ define(function(require) {
...
+      Debug('updateConnectionInfo(): this.address = ' + this.address +
+            ', options = ' + JSON.stringify(options));
+      if (options.connectionStatus) {
+        this.connectionStatus = options.connectionStatus;
+      }
+
+      if (options.profiles) {
+        var profileHFP, profileA2DP;
+        if (this.profiles) {
+          profileHFP = (options.profiles.HFP) ?
```
crh0716: Could be simplified as:
```
profileHFP = options.profiles.HFP || this.profiles.HFP || false;
```

apps/settings/js/modules/bluetooth/bluetooth_device.js
```diff
@@ -30,14 +30,105 @@ define(function(require) {
...
+            (options.profiles.HFP || false) : (this.profiles.HFP || false);
+          profileA2DP = (options.profiles.A2DP) ?
+            (options.profiles.A2DP || false) : (this.profiles.A2DP || false);
+          this.profiles = {
+            'HFP': profileHFP,
+            'A2DP': profileA2DP
+          };
+        } else {
+          this.profiles = {
```
crh0716: The code under `if (options.profiles) {` could be simplified as:
```
this.profiles = {
  'HFP': options.profiles.HFP || this.profiles && this.profiles.HFP || false;
  'A2DP': options.profiles.A2DP || this.profiles && this.profiles.A2DP || false;
};
```

apps/settings/js/modules/bluetooth/bluetooth_connection_manager.js
```diff
@@ -0,0 +1,674 @@
...
+     * @returns {Promise} reject: reason
+     */
+    _getConnectedDevicesByProfile: 
+    function btcm__getConnectedDevicesByProfile(profileID) {
+      return new Promise((resolve, reject) => {
+        if (!this._defaultAdapter) {
+          reject('default adapter is not existed!!');
+        }
+
+        var req = this._defaultAdapter.getConnectedDevices(profileID);
```
crh0716: Because the cost of calling to `getConnectedDevices` is high, gaia should cache the connected devices of each profile. We can achieve this by calling to `getConnectedDevices` in `_initProperties` and update the cache using hfp and a2dp status change events.

apps/settings/js/modules/bluetooth/bluetooth_context.js
```diff
@@ -824,6 +844,221 @@ define(function(require) {
...
+     * @access private
+     * @memberOf BluetoothContext
+     */
+    _initConnectingDevices: function btc__initConnectingDevices() {
+      // Init the paired device connection status for connecting device.
+      if (!ConnectionManager.connectingAddress) {
+        return;
+      }
+
+      var index = this.getPairedDevices().array.findIndex(
```
crh0716: Let's just wrap the code in a single function that returns the device item of a specific address as you use it a lot in other functions. `_findDeviceByAddress` is a good name to me.

apps/settings/js/modules/bluetooth/bluetooth_connection_manager.js
```diff
@@ -0,0 +1,674 @@
...
+    connectingAddress: null,
+
+    /**
+     * The address of device that we are connected.
+     *
+     * @public
+     * @memberOf BluetoothConnectionManager
+     * @type {String}
+     */
+    connectedAddress: null,
```
crh0716: There can be multiple connected addresses, making this an `ObservableArray` makes more sense to me.

crh0716: The item stored in the array can contain the information of the connected profiles.

apps/settings/js/modules/bluetooth/bluetooth_device.js
ian-liu: The script is not cover the following case.

`options.profiles.HFP = false;`
`this.profiles.HFP = true`;

The result of `profileHFP` is true. `profileHFP` cannot be updated if we get the connection profile with 'false'.

apps/settings/js/modules/bluetooth/bluetooth_device.js
ian-liu: Same here. Cannot cover case `options.profiles.HFP = false;` if we got the profile result from platform side.

try-server-hook:
[ian-liu Ian Liu (ian-liu)](https://github.com/ian-liu) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=440977f7c1d7)

ian-liu:
The major change in latest version:
- Support cache connected device information. We only do adapter.getConnectedDevices() while init the cache. It improves the performance while the adapter is connecting/disconnecting with a device. It also improves refresh paired devices list. I feel more better than v1 code base. Thanks for your suggestion here:)
- ConnectionManager is not a observable object. We use event listener to notify outer modules.
- Update Bluetooth Context for some duplicated code. Such as: `_findDeviceByAddress'.
- Move updateConnectionInfo(), updateDescriptionText() to be inner function in BluetoothDevice module.

apps/settings/js/modules/bluetooth/bluetooth_context.js
```diff
@@ -749,6 +771,35 @@ define(function(require) {
...
+     * @access private
+     * @memberOf BluetoothContext
+     * @param {String} address
+
+     * @param {Object} options
+     * @param {String} options.paired - is paired or not
+     * @param {String} options.address - the address of the device
+     * @return {Object} device
+     */
+    _findDeviceByAddress:
```
crh0716: I think it is okay not to take `paired` as a parameter. Address is sufficient for finding the device as there are only two device lists to be searched. This also makes the function easier to use.

apps/settings/js/modules/bluetooth/bluetooth_context.js
```diff
@@ -824,6 +875,177 @@ define(function(require) {
...
+     * Device 'connectionStatus' property is changed from 
+     * ConnectionManager operation.
+     * Update device 'connectionStatus' property via reference address in event.
+     *
+     * @access private
+     * @memberOf BluetoothContext
+     * @param {Object} event
+     * @param {String} event.address - address of device
+     */
+    _updateDeviceConnectingStatus:
```
crh0716: Note that there are only slight difference among `_updateDeviceConnectingStatus`, `_updateDeviceConnectedStatus`, `_updateDeviceDisconnectedStatus`, and `_updateDeviceConnectionProfile`. To be more specific, the only difference is the options object to be passed to the update function.

We can combine these four functions into one. The key is that the event object should have a property called 'type' allowing you to differentiate different kinds of events.

apps/settings/js/modules/bluetooth/bluetooth_connection_manager.js
```diff
@@ -0,0 +1,868 @@
...
+      return this._connectedDevicesInfo;
+    },
+
+    /**
+     * These listeners would be called when device is connecting.
+     *
+     * @memberOf BluetoothConnectionManager
+     * @type {Array}
+     */
+    _connectingListeners: [],
```
crh0716: This could be implemented is a easier and more flexible way. We can have a object like:
```
_listeners: {
  'connecting': [],
  'connected': [],
  'disconnected': [],
  'profileChanged': []
}
```

With this object, `addEventListener`, `removeEventListener`, and the function for emitting the events can be implemented like:
```
addEventListener: function(eventName, callback) {
  // TODO: check eventName is valid
  // TODO: check callback is valid
  this._listeners[eventName].push(callback);
}
```

apps/settings/js/modules/bluetooth/bluetooth_connection_manager.js
```diff
@@ -0,0 +1,868 @@
...
+      return hasConnectedProfile;
+    },
+
+    /**
+     * When device is connecting, we will call all registered listeners.
+     *
+     * @memberOf BluetoothConnectionManager
+     * @param {Event} event
+     */
+    _deviceConnecting: function btcm__deviceConnected(event) {
```
crh0716: Following the previous comment, the four functions for emitting events could be combined. Note that the object passed to the listener should contain a `type` property so the callback can handle the event more easily. We can use the structure borrowed from html events like:
```
{
  type: 'connecting',
  detail: { // the object you want to pass to the listener }
}
```

apps/settings/js/modules/bluetooth/bluetooth_connection_manager.js
```diff
@@ -0,0 +1,868 @@
...
+     *
+     * @access private
+     * @memberOf BluetoothConnectionManager
+     * @param {Object} options
+     * @param {String} options.address - address of the device
+     * @param {Boolean} options.connected - is connected or not
+     * @param {Object} options.profileID - profile ID of the connection type
+     * @param {Object} options.device - connect device, Bluetooth Object
+     */
+    _updateConnectionDevices:
```
crh0716: nit: `_updateConnectedDevices`.

apps/settings/js/modules/bluetooth/bluetooth_connection_manager.js
```diff
@@ -0,0 +1,868 @@
...
+                                       ', profileID = ' + options.profileID);
+      this.connectingAddress = null;
+      
+      // Update the profile in the cache of connected device info.
+      var connectionDeviceInfo = {
+        address: options.address,
+        connected: options.connected,
+        profileID: options.profileID
+      };
+      this._updateConnectionDevices(connectionDeviceInfo);
```
crh0716: Why not just `this._updateConnectionDevices(connectionDeviceInfo)`?

apps/settings/js/modules/bluetooth/bluetooth_connection_manager.js
```diff
@@ -0,0 +1,868 @@
...
+
+    /**
+     * Return the cache of connected devices in ConnectionManager.
+     *
+     * @access public
+     * @memberOf BluetoothConnectionManager
+     * @return {Object}
+     */
+    getConnectedDevices: function btcm_getConnectedDevices() {
+      return this._connectedDevicesInfo;
```
crh0716: This function should return a promise. This differs from the case of getting an adapter because in this case gaia actively requests the latest connection status from gecko.

Consider this case, bt context calls to this function when `_initConnectedDevicesInfo` does not finish the initialization. Bt context does not get the connected device correctly until it calls to this function again.

To solve the issue, we should cache the promise returned from `_initConnectedDevicesInfo` and always wait for it resolved before returning the result.
And even in this module, we should not use `this._connectedDeviceInfo` directly but always use this function. This guarantees that the update to `this._connectedDevicesInfo` is always made after `this._connectedDevicesInfo` is initialized.

apps/settings/js/modules/bluetooth/bluetooth_context.js
ian-liu: I was wondering that the found out device in list is not our expected one. Such as: We want to find a device in paired devices list. But it's possible to find out it in remote devices list. If `ondevicefound`, `ondeviceunpaired`, `ondevicepaired` callback is coming (sequence/item)un-normally, I think we might update the device info in wrong list. That is why I have to design the function given remote/paired devices parameter. Then, try to find the device by address.

try-server-hook:
[ian-liu Ian Liu (ian-liu)](https://github.com/ian-liu) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=c8c4fc16e246)

apps/settings/js/modules/bluetooth/bluetooth_context.js
crh0716: I see, thanks for the explanation!

try-server-hook:
[ian-liu Ian Liu (ian-liu)](https://github.com/ian-liu) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=41689023bb19)

ian-liu:
Update:
* Rename `_getConnectedDevices` to be `_getConnectedDevicesWithHFPAndA2DP`.
* Fix connected device information error.
  ** Re-launch settings app while device is connected.
* Fix device description missed to update while ListView is re-render.

apps/settings/js/modules/bluetooth/bluetooth_connection_manager.js
```diff
@@ -0,0 +1,959 @@
...
+    /**
+     * The method will get all connected devices profiles 
+     * which we are interested in. Profile: HFP, A2DP.
+     *
+     * @access private
+     * @memberOf BluetoothConnectionManager
+     * @returns {Promise} resolve: connectedDevices
+     * @returns {Promise} reject: reason
+     */
+    _getConnectedDevicesWithHFPAndA2DP:
```
crh0716: How about `_getConnectedDevicesFromPlatform`? This avoids hard coded the profile name.

apps/settings/js/modules/bluetooth/bluetooth_connection_manager.js
```diff
@@ -0,0 +1,959 @@
...
+     * @access private
+     * @memberOf BluetoothConnectionManager
+     * @returns {Promise} resolve: connectedDevices
+     * @returns {Promise} reject: reason
+     */
+    _getConnectedDevicesWithHFPAndA2DP: 
+    function btcm__getConnectedDevicesWithHFPAndA2DP() {
+      // Get connected device via profiles HFP, A2DP
+      return Promise.all([
+        this._getConnectedDevicesByProfile(this.Profiles.hfp),
```
crh0716: I did not notice this before. Let's do this without use the profile explicitly:
```
return Promise.all(Object.keys(this.Profiles).map((profile) => {
  return this._getConnectedDevicesByProfile(profile);
})).then(
```

apps/settings/js/modules/bluetooth/bluetooth_connection_manager.js
```diff
@@ -0,0 +1,959 @@
...
+        this._getConnectedDevicesByProfile(this.Profiles.a2dp)
+      ]).then((connectedDevices) => {
+        // Update each connected devices in map.
+        var collectedConnectedDevicesByProfile = {
```
crh0716:
```
var collectedConnectedDevicesByProfile = {};
Object.keys(this.Profiles).forEach((profile, index) => {
  collectedConnectedDevicesByProfile[profile] = connectedDevices[index];
});
```
```

File: apps/settings/js/modules/bluetooth/bluetooth_connection_manager.js [hunk 32]
```diff
@@ -0,0 +1,959 @@
...
+     * The method will update each device item in maintaining map.
+     *
+     * @access private
+     * @memberOf BluetoothConnectionManager
+     * @param {Oject} connectedDevices
+     * @return {Promise}
+     */
+    _constructDeviceItemsMap:
+    function btcm__constructDeviceItemsMap(connectedDevices) {
+      Debug('_constructDeviceItemsMap(): connectedDevices = ' +
```
crh0716: This looks like a synchronous function, so it is not required to return a promise.

apps/settings/js/panels/bluetooth/panel.js
```diff
@@ -144,15 +153,20 @@ define(function(require) {
...
+          }, (reason) => {
+            Debug('onShow(): startDiscovery failed, ' +
+                  'reason = ' + reason);
+          });
         }
         this._leftApp = false;
       },
 
-      onBeforeHide: function() {
+      onBeforeHide : function() {
```
crh0716: Why to make this change? The space is not required.

apps/settings/js/panels/bluetooth/panel.js
```diff
@@ -162,12 +176,12 @@ define(function(require) {
         BtContext.unobserve('discovering');
       },
 
-      onHide: function() {
+      onHide : function() {
```
crh0716: same here

apps/settings/js/panels/bluetooth/panel.js
```diff
@@ -182,7 +196,7 @@ define(function(require) {
         }
       },
 
-      _onEnableCheckboxClick: function() {
+      _onEnableCheckboxClick : function() {
```
crh0716: same here.

try-server-hook:
[ian-liu Ian Liu (ian-liu)](https://github.com/ian-liu) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=474b18dfcd50)

ian-liu:
Update: Add unit test to cover `bluetooth_context`, `bluetooth_device`, `bluetooth_connection_manager`.

apps/settings/test/unit/modules/bluetooth/bluetooth_connection_manager_test.js
```diff
@@ -0,0 +1,552 @@
+'use strict';
+
+suite('BluetoothContext', function() {
+  var adapterManager;
+  var asyncStorage;
+  var connectionManager;
+
+  suiteSetup(function() {
```
crh0716: remove these functions if they are empty.

apps/settings/js/modules/async_storage.js
```diff
@@ -16,6 +16,11 @@ define(function(require) {
       return new Promise(function(resolve, reject) {
         AsyncStorage.setItem(key, value, resolve);
       });
+    },
+    removeItem: function as_removeItem(key) {
```
ian-liu: Note: I think I no need to do the change because I `require('shared/async_storage');`. Will remove it later.

apps/settings/js/modules/bluetooth/bluetooth_connection_manager.js
```diff
@@ -0,0 +1,955 @@
...
+ *   logic.
+ *
+ * @module BluetoothConnectionManager
+ */
+define(function(require) {
+  'use strict';
+
+  var AdapterManager = require('modules/bluetooth/bluetooth_adapter_manager');
+  var AsyncStorage = require('shared/async_storage');
+  require('shared/async_storage');
```
ian-liu: Note: should remove it.

apps/settings/test/unit/modules/bluetooth/bluetooth_connection_manager_test.js
```diff
@@ -0,0 +1,552 @@
...
+          }, () => {
+            //reject case
+          });
+        }, function() {
+          // reject case
+        }).then(done, done);
+      });
+    });
+
+    suite('has default adapter, init connected devices info ', function() {
```
crh0716: should be "has no default adapter"?

apps/settings/test/unit/modules/bluetooth/bluetooth_connection_manager_test.js
```diff
@@ -0,0 +1,552 @@
...
+        connectionManager._connectedDevicesInfo = mockConnectedDevicesInfo;
+        mockConnectedDevicesInfoPromise = Promise.resolve();
+        this.sinon.stub(connectionManager,
+          '_initConnectedDevicesInfo').returns(mockConnectedDevicesInfoPromise);
+      });
+
+      test('should return connected device cached while promise resolve ',
+           function(done) {
+        connectionManager.getConnectedDevices().then((connectedDevicesInfo) => {
+          assert.ok(connectionManager.getConnectedDevices);
```
crh0716: What is this check for?

apps/settings/test/unit/modules/bluetooth/bluetooth_connection_manager_test.js
```diff
@@ -0,0 +1,552 @@
...
+    var mockAdapter;
+    setup(function() {
+      mockAdapter = {
+        addEventListener: function() {}
+      };
+      this.sinon.stub(mockAdapter, 'addEventListener');
+      this.sinon.stub(connectionManager, '_onAdapterAttributeChanged');
+    });
+
+    test('attributechanged should be regedit callback function ', function() {
```
crh0716: typo: registered

apps/settings/test/unit/modules/bluetooth/bluetooth_connection_manager_test.js
```diff
@@ -0,0 +1,552 @@
...
+    var mockAdapter;
+    setup(function() {
+      mockAdapter = {
+        removeEventListener: function() {}
+      };
+      this.sinon.stub(mockAdapter, 'removeEventListener');
+      this.sinon.stub(connectionManager, '_onAdapterAttributeChanged');
+    });
+
+    test('onattributechanged should be unregedit ', function() {
```
crh0716: typo: unregistered

apps/settings/test/unit/modules/bluetooth/bluetooth_connection_manager_test.js
```diff
@@ -0,0 +1,552 @@
...
+          attrs: ['state']
+        };
+        this.sinon.stub(connectionManager, '_restoreConnection');
+      });
+
+      test('_restoreConnection() should be called ', function() {
+        connectionManager._onAdapterAttributeChanged(mockAdapter, evt);
+        assert.isTrue(connectionManager._restoreConnection.called);
+      });
+    });
```
crh0716: This function is important. We should ensure `_restoreConnection` is not called in other states.
```
['enabling', 'disabling', 'disabled'].forEach(function(state) {
  test('_restoreConnection() should not be called when ' + state, function() {
    mockAdapter.state = state;
    connectionManager._onAdapterAttributeChanged(mockAdapter, evt);
    assert.isTrue(connectionManager._restoreConnection.notCalled);
  });
});
```

apps/settings/test/unit/modules/bluetooth/bluetooth_connection_manager_test.js
```diff
@@ -0,0 +1,552 @@
...
+  });
+
+  suite('_watchProfilesStatuschanged > ', function() {
+    var mockAdapter;
+    setup(function() {
+      mockAdapter = {};
+    });
+
+    test('profile status callback of adapter should be regedit ', function() {
```
crh0716: typo: registered

apps/settings/test/unit/modules/bluetooth/bluetooth_device_test.js
```diff
@@ -45,11 +48,90 @@ suite('BluetoothDevice', function() {
...
+        mockOptions = {
+          connectionStatus: 'connected',
+          profiles: {
+            'hfp': true,
+            'a2dp': false
+          }
+        };
+      });
+
+      test('"connectionStatus" property should be "connecting"', function() {
```
crh0716: connected?

apps/settings/test/unit/modules/bluetooth/bluetooth_context_test.js
```diff
@@ -1141,6 +1194,86 @@ suite('BluetoothContext', function() {
...
+    });
+
+    suite('did not find out match device ', function() {
+      var inputOptions, mockPairedDvices, mockRemoteDevices;
+      setup(function() {
+        inputOptions = {
+          address: '00:11:22:AA:BB:04',
+          paired: false
+        };
+        mockPairedDvices = observableArray([
```
crh0716: The array seems the same as the ones above. We can init the array in one place, maybe in `suiteSetup` of the outer suite.

try-server-hook:
[ian-liu Ian Liu (ian-liu)](https://github.com/ian-liu) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=4058d266ea7e)

try-server-hook:
[ian-liu Ian Liu (ian-liu)](https://github.com/ian-liu) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=66a52ad48a22)

try-server-hook:
[ian-liu Ian Liu (ian-liu)](https://github.com/ian-liu) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=96e4c06ab9aa)

ian-liu:
Looks all tests are passed on try-server twice. We can merge the pr.

PR #28146, opened 2015-02-11
Bug 1131966 - Update suggestions when provider changes

daleharvey:


try-server-hook:
[daleharvey Dale  Harvey (daleharvey)](https://github.com/daleharvey) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=5d781ce5d633)

apps/search/js/providers/suggestions.js
```diff
@@ -59,6 +69,8 @@
     },
 
     search: function(input, preventRemote) {
+
+      this.currentSearch = input;
```
KevinGrandon: I don't really like the idea of keeping a current search here, so let's remove it if possible. The problem is that we interact with the search input state outside of the provier, so it's easy to get out of sync. E.g., search for something, clear rocketbar, change provider. With this patch we would then show suggestions with an empty search term.

If keeping this is the best option, perhaps a quick fix would be to check to see if there is content that exists within the suggestions already? Skipping the search if there is no content?

KevinGrandon: Or if this is only a problem with clearing, I guess we could implement that method, and null the currentSearch on clear.

daleharvey: I think the only way to handle this is to keep a reference to the currentSearch somewhere, however I did miss clearing the input, good catch

try-server-hook:
[daleharvey Dale  Harvey (daleharvey)](https://github.com/daleharvey) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=47887097ef7b)

try-server-hook:
[daleharvey Dale  Harvey (daleharvey)](https://github.com/daleharvey) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=12be1df49eeb)

try-server-hook:
[daleharvey Dale  Harvey (daleharvey)](https://github.com/daleharvey) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=f5b76a456fbc)

try-server-hook:
[daleharvey Dale  Harvey (daleharvey)](https://github.com/daleharvey) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=eacd3c96bf58)

try-server-hook:
[daleharvey Dale  Harvey (daleharvey)](https://github.com/daleharvey) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=106f455e3bbe)

PR #28155, opened 2015-02-11
Bug 1069617 - Put text inside anchor tags in music list view, and assign...

eeejay:
... aria roles.

### DIFF
--- a/apps/music/index.html
+++ b/apps/music/index.html
@@ -80,7 +80,7 @@
             <button id="views-list-search-clear" type="reset">Clear</button>
           </p>
         </form>
-        <div id="views-list-anchor"></div>
+        <div id="views-list-anchor" role="listbox"></div>
       </div>
       <div id="views-sublist" class="view hidden">
         <div id="views-sublist-header">
@@ -92,7 +92,7 @@
           </div>
         </div>
         <!-- Bug 968456. Scrollable area needs to be inside an opaque background div to help gecko -->
-        <div id="views-scrollable-sublist-anchor"><div id="views-sublist-anchor"></div></div>
+        <div id="views-scrollable-sublist-anchor"><div id="views-sublist-anchor" role="listbox"></div></div>
       </div>
       <div id="views-player" class="view hidden">
         <div id="player-cover">
@@ -142,21 +142,21 @@
           <span data-l10n-id="search-artists">ArtistS</span>
             (<span class="search-result-count"></span>)
         </gaia-subheader>
-        <div class="search-results"></div>
+        <div class="search-results" role="listbox"></div>
       </div>
       <div id="views-search-albums" class="search-category hidden">
         <gaia-subheader skin="dark">
           <span data-l10n-id="search-albums">AlbumS</span>
             (<span class="search-result-count"></span>)
         </gaia-subheader>
-        <div class="search-results"></div>
+        <div class="search-results" role="listbox"></div>
       </div>
       <div id="views-search-titles" class="search-category hidden">
         <gaia-subheader skin="dark">
           <span data-l10n-id="search-titles">TrackS</span>
             (<span class="search-result-count"></span>)
         </gaia-subheader>
-        <div class="search-results"></div>
+        <div class="search-results" role="listbox"></div>
       </div>
     </div>
     <!-- BB tabs -->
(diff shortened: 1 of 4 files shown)


try-server-hook:
[eeejay Eitan Isaacson (eeejay)](https://github.com/eeejay) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=41d006fa7d14)

apps/music/js/utils.js
```diff
@@ -75,13 +77,13 @@ function createListElement(option, data, index, highlight) {
       a.dataset.keyRange = 'all';
       a.dataset.option = data.option;
 
-      li.appendChild(titleSpan);
+      a.appendChild(titleSpan);
```
jimporter: I'm a little concerned about this change; why is it needed?

apps/music/style/music.css
```diff
@@ -1006,14 +1006,8 @@ html[dir="rtl"] li.list-item {
...
 li.list-item a {
   display: block;
-  color: #fff;
   text-decoration: none;
-  outline: none;
   margin: 0;
-  text-shadow: black 0.1rem 0.1rem;
-  white-space: nowrap;
-  overflow: hidden;
-  text-overflow: ellipsis;
```
jimporter: Why are these removed? Are they no longer needed?

apps/music/js/utils.js
eeejay: Currently, the label (`<span>`) and link (`<a>`) are siblings. The screen reader "sees" the labels, but they are inoperable because the listener expects the link to be the target. When the user swipes through the list items they hear "label foo", and then on the next item simply "link". This needs to be cleaned up and made logical. So in this patch we make the labels proper children of the link elements to give these lists a logical markup and allow the screen reader to navigate and operate it correctly.

apps/music/style/music.css
eeejay: I couldn't find anywhere in the application where there is a child `<a>` in a `.list-item`. So I re-appropriated that rule and cleaned it up for the case above.

dominickuo: @eeejay If I recalled correctly, we used to have `<a>` in a list element but we tried to reduce the children counts in the DOM tree, so `<a>` was removed in some commit, and this might be the styles we forget to clean.

apps/music/js/utils.js
dominickuo: Does the list elements also use `<a>` in the other apps? I remember why we use `<li>` was because the building blocks suggested the apps to use `<li>` for list elements.

dominickuo: Never mind, I think I misunderstood it :P

apps/music/js/ui/views/list_view.js
```diff
@@ -95,6 +95,8 @@ var ListView = {
       headerLi = document.createElement('li');
       headerLi.className = 'list-header';
       headerLi.textContent = this.lastFirstLetter || '?';
+      headerLi.setAttribute('role', 'heading');
+      headerLi.setAttribute('aria-level', '2');
```
dominickuo: @eeejay When music app is in searching modes, like searching in tiles or lists, the search results will have something like the list headers(Artists, Albums or Tracks), do we need to also set role and aria-level for them?

eeejay: Do you mean the gaia-subheader? That needs to be given a role as well, but i think it is outside the scope of this bug.

dominickuo: @eeejay Yes, just thought that's something related to the list elements but we can do it another bug.

apps/music/index.html
```diff
@@ -156,7 +156,7 @@ <h1 id="title-text" data-l10n-id="music"></h1>
           <span data-l10n-id="search-titles">TrackS</span>
             (<span class="search-result-count"></span>)
         </gaia-subheader>
-        <div class="search-results"></div>
+        <div class="search-results" role="listbox"></div>
```
dominickuo: @eeejay There are two more elements use `search-results` class and we should also add `role="listbox"` to them?

eeejay: Oops, good catch!

try-server-hook:
[eeejay Eitan Isaacson (eeejay)](https://github.com/eeejay) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=cebdd7199854)

PR #28162, opened 2015-02-12
Bug 1124398 Edge Swipe transitioning between apps will sometimes encounter a black screen when settings app is open r=etienne

songben423:
...ter a black screen when settings app is open r=etienne

[bug number  ]
[root cause  ]
[changes     ]
[side effects]
[self test   ]
[whether AOB ]
[reviewers   ]


autolander:

PR #28182, opened 2015-02-12
Bug 1132479 - Fix test_IMAP_email_notification failure

mwargers:


try-server-hook:
[mwargers Martijn (mwargers)](https://github.com/mwargers) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=760e84fbc102)

try-server-hook:
[mwargers Martijn (mwargers)](https://github.com/mwargers) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=5cd18afc1b4c)

tests/python/gaia-ui-tests/gaiatest/apps/email/regions/setup.py
```diff
@@ -182,7 +183,10 @@ def tap_next(self):
                 *self._account_prefs_next_locator))))
 
     def check_for_emails_interval(self, value):
```
JohanLorenzo: `value` is not used anymore in the function

tests/python/gaia-ui-tests/gaiatest/apps/email/regions/setup.py
```diff
@@ -114,6 +114,7 @@ def type_password(self, value):
     def select_account_type(self, value):
         account_type = self.marionette.find_element(*self._account_type_locator)
         account_type.click()
+        self.marionette.switch_to_frame()
```
JohanLorenzo: Why do we need this change here? It seems unrelated to the intermittent failure.

tests/python/gaia-ui-tests/gaiatest/apps/email/regions/setup.py
```diff
@@ -182,7 +183,10 @@ def tap_next(self):
                 *self._account_prefs_next_locator))))
 
     def check_for_emails_interval(self, value):
-        self.marionette.execute_script('document.querySelector("[data-l10n-id = settings-check-every-5min]").value = "%s";' % value)
+        self.marionette.execute_script("""
+            SpecialPowers.setIntPref('dom.requestSync.minInterval', 1);
+            document.querySelector("[data-l10n-id = settings-check-every-5min]").value = 20000;
```
JohanLorenzo: If we change the pref directly, this line is not useful anymore, is it?

tests/python/gaia-ui-tests/gaiatest/apps/email/regions/setup.py
mwargers: This is needed to get the test working on b2g desktop

tests/python/gaia-ui-tests/gaiatest/apps/email/regions/setup.py
mwargers: The pref is for the minimum interval time for syncing, which is normally 100s. We need to set it lower than 20s, to be able to sync to 20s.

tests/python/gaia-ui-tests/gaiatest/apps/email/regions/setup.py
mwargers: I'll put it back.

try-server-hook:
[mwargers Martijn (mwargers)](https://github.com/mwargers) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=436bcd8da28d)

tests/python/gaia-ui-tests/gaiatest/apps/email/regions/setup.py
```diff
@@ -182,7 +183,10 @@ def tap_next(self):
                 *self._account_prefs_next_locator))))
 
     def check_for_emails_interval(self, value):
-        self.marionette.execute_script('document.querySelector("[data-l10n-id = settings-check-every-5min]").value = "%s";' % value)
+        self.marionette.execute_script("""
+            SpecialPowers.setIntPref('dom.requestSync.minInterval', 1);
```
JohanLorenzo: A comment explaining why we change the data_layer in a function which manipulates the UI would be helpful for future readers.

try-server-hook:
[mwargers Martijn (mwargers)](https://github.com/mwargers) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=13f317958add)

try-server-hook:
[mwargers Martijn (mwargers)](https://github.com/mwargers) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=6faf507ef430)

tests/python/gaia-ui-tests/gaiatest/apps/email/regions/setup.py
```diff
@@ -182,7 +183,14 @@ def tap_next(self):
...
     def check_for_emails_interval(self, value):
-        self.marionette.execute_script('document.querySelector("[data-l10n-id = settings-check-every-5min]").value = "%s";' % value)
+        # The following pref change allows us to check the mail within 1 second or longer,
+        # rather than the default value of 100 seconds
+        # The UI data layer of the UI is changed, because the minimum check mail time value is 5 min,
+        # which is far too long to check for in a test. This allows us to check earlier
+        self.marionette.execute_script("""
+            SpecialPowers.setIntPref('dom.requestSync.minInterval', 1);
+            document.querySelector("[data-l10n-id = settings-check-every-5min]").value = '%s';
+        """ % value, special_powers=True)
```
geoelectric: Are you sure that 1 means 1 second? The value we set for the data layer is 20000 by default, so I assume that's ms.

Anyway, I'm fine with this, with one concern: any very fast value (probably even 20 seconds) is abusive to the remote IMAP server we check, to the point where they might actually blacklist us. 1 second would certainly be unacceptable.

I checked test/app code, and am troubled with how we currently do this. 

We set it to 20 seconds as soon as we init the Email object, then it stays that way for the whole test. Might stay that way for the whole test -run-, I'm not sure what does and doesn't survive if we do or don't restart. Certainly, if the test run aborts and leaves the phone on the email app, it will survive that and leave the phone fast-checking.

So while I think this change is basically OK, my specific concerns would be:

a) That we might want to set a higher minimum time as a bumper towards messing up too badly. We seem to only need 20 seconds, so maybe we should set it to 15 or 20, modulo comment below.

b) I think we should consider rewriting the test code more like:

      setup = SetupEmail()
      ...
      EmailUtil().send(...)
      setup.check_for_emails_interval(20000)
      try:
          ...
          # Wait for email notification
          system.wait_for_notification_toaster_displayed(timeout=60)
          system.wait_for_notification_toaster_not_displayed()
      finally:
           setup.check_for_emails_interval(300000)

...or something like that. That way it's only accelerated just long enough to get a fast sync for that one part of the test.

If we did that, I'd be up for dropping the sync rate (and minimum) way down to maybe 5 or 10 seconds, since it'd in theory only last one sync as long as the IMAP server processes the mail quickly. But either way, I think you really need to consider whether the current test leaves the phone hammering the server in the background.

mwargers: The dom.requestSync.minInterval pref is in seconds.
The change makes the check for the imap server to 20 seconds, it's basically fixing this test, which did the same.

try-server-hook:
[mwargers Martijn (mwargers)](https://github.com/mwargers) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=69b8e880ab9b)

geoelectric:
Looks OK, thanks for adding the cleanup step.

PR #28189, opened 2015-02-12
Bug 1131665 - [Contacts] Update Carets

pivanov:


shared/style/buttons.css
```diff
@@ -204,7 +204,8 @@ li .button.icon-view:active {
 li button.icon-view:before,
 li .bb-button.icon-view:before,
 li .button.icon-view:before {
-  background-image: url(buttons/images/next.png);
+  background-image: url(buttons/images/forward-light.svg);
```
wilsonpage: Are we safe to use SVG (performance wise) now?

pivanov: I think so ... we use a lot of SVGs in /shared/style/

PR #28194, opened 2015-02-12
Bug 1130013 - Add a link to the Marketplace App Stats page in Settings.

marshall:
https://bugzilla.mozilla.org/show_bug.cgi?id=1130013

### DIFF
--- a/apps/settings/elements/improve_browser_os.html
+++ b/apps/settings/elements/improve_browser_os.html
@@ -33,6 +33,11 @@
             <span data-l10n-id="sharePerformanceAndUsageData">Submit Data</span>
           </label>
         </li>
+        <li>
+          <a id="appUsageByRegion" class="menu-item">
+            <span data-l10n-id="appUsageByRegion"></span>
+          </a>
+        </li>
       </ul>
 
       <header>
@@ -69,5 +74,6 @@
       </ul>
     </div>
 
+    <panel data-path="panels/improveBrowserOS/panel"></panel>
   </template>
 </element>
--- a/apps/settings/js/panels/improveBrowserOS/panel.js
+++ b/apps/settings/js/panels/improveBrowserOS/panel.js
@@ -1,0 +1,26 @@
+define(function(require) {
+  'use strict';
+
+  var SettingsPanel = require('modules/settings_panel');
+
+  return function ctor_improveBrowserOS() {
+    return SettingsPanel({
+      onInit: function(panel) {
+        var appUsage = panel.querySelector('#appUsageByRegion');
+        appUsage.addEventListener('click', this._openAppUsage.bind(this));
+      },
+
+      _openAppUsage: function() {
+        var activity = new window.MozActivity({
+          name: 'marketplace-search',
+          data: {
+            type: 'firefox-os-app-stats'
+          }
+        });
+        activity.onerror = function() {
+          console.log('Error opening app stats page in Marketplace');
+        };
+      }
+    });
+  };
+});
--- a/apps/settings/locales/settings.en-US.properties
+++ b/apps/settings/locales/settings.en-US.properties
@@ -1302,6 +1302,7 @@ improveBrowserOSTitle=Improve {{brandShortName}}
 sendMozillaFeedback=Send Mozilla Feedback
 performanceAndUsageData=Performance and Usage Data
 sharePerformanceAndUsageData=Submit data
+appUsageByRegion=App usage by region
 crashReports=Crash Reports
 # LOCALIZATION NOTE (crashReports): if you leave the original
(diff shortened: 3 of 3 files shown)


try-server-hook:
[marshall Marshall Culpepper (marshall)](https://github.com/marshall) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=c990f0919a69)

apps/settings/elements/improve_browser_os.html
```diff
@@ -33,6 +33,13 @@ <h2 data-l10n-id="performanceAndUsageData">Performance And Usage Data</h2>
             <span data-l10n-id="sharePerformanceAndUsageData">Submit Data</span>
           </label>
         </li>
+        <li>
+          <a id="openAppStats" class="menu-item">
+            <span data-l10n-id="openAppStats">
+              Open App Statistics
```
crh0716: Based on the l10n guideline the text we should not leave text in the html file.

apps/settings/js/panels/improveBrowserOS/panel.js
```diff
@@ -0,0 +1,34 @@
...
+
+  return function ctor_improveBrowserOS() {
+    return SettingsPanel({
+      onInit: function(panel) {
+        this.openAppStats = panel.querySelector('#openAppStats');
+        this.openAppStats.addEventListener('click',
+                                          this._onOpenAppStats.bind(this));
+      },
+
+      onUninit: function() {},
```
crh0716: There are default functions in `SettingsPanel` so these three functions can be removed.

apps/settings/elements/improve_browser_os.html
crh0716: Can you remove all other texts while you are on it? Thanks.

apps/settings/js/panels/improveBrowserOS/panel.js
```diff
@@ -0,0 +1,34 @@
...
+      _onOpenAppStats: function() {
+        var activity = new window.MozActivity({
+          name: 'marketplace-search',
+          data: {
+            type: 'firefox-os-app-stats'
+          }
+        });
+
+        activity.onerror = function() {
+          window.open(
```
crh0716: This way when users cancel the activity request they are still prompted with the dialog. We should not do this when `activity.error.name === 'ActivityCanceled'`.

apps/settings/js/panels/improveBrowserOS/panel.js
```diff
@@ -0,0 +1,34 @@
...
+
+      _onOpenAppStats: function() {
+        var activity = new window.MozActivity({
+          name: 'marketplace-search',
+          data: {
+            type: 'firefox-os-app-stats'
+          }
+        });
+
+        activity.onerror = function() {
```
andymckay: Is this likely to happen? Under what scenarios will it happen? The URL should be "https://marketplace.firefox.com/usage" as per this commit I'm working on: https://github.com/mozilla/fireplace/commit/8ab4e7ee557a#diff-9d9d2e177b183f25634866826d4c75d4R58 - but I'm wondering if we need it. It will open in a browser and create a whole new flow to test. If this isn't needed, I'd rather remove.

apps/settings/locales/settings.en-US.properties
```diff
@@ -1301,6 +1301,7 @@ improveBrowserOSTitle=Improve {{brandShortName}}
 sendMozillaFeedback=Send Mozilla Feedback
 performanceAndUsageData=Performance and Usage Data
 sharePerformanceAndUsageData=Submit data
+openAppStats=Open App Statistics
```
andymckay: We should make title consistent, I can change mine to `App Statistics` but I'm not sure its the best title. `App Usage Statistics` maybe? We've got App Statistics on the Marketplace that are something else. Not sure at this point :)

marshall: @andymckay I liked the suggestion on the bug 'App usage by region', I have that in my latest that I'm about to push.. wdyt?

apps/settings/elements/improve_browser_os.html
marshall: I'll remove it for the new string, but would you mind filing a follow up for the others? I want to keep the impact of this patch low because it's coming in late

try-server-hook:
[marshall Marshall Culpepper (marshall)](https://github.com/marshall) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=f712e7efeb73)

apps/settings/elements/improve_browser_os.html
crh0716: okay, let's do it in the follow up.

PR #28197, opened 2015-02-12
Bug 1119865 - LanguageList.get should return a promise

MichaelKohler:


MichaelKohler:
there is still one test failing.. I'm working on that, sorry about it

try-server-hook:
[MichaelKohler Michael Kohler (MichaelKohler)](https://github.com/MichaelKohler) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=5ca5b0929268)

apps/ftu/test/unit/language_test.js
```diff
@@ -55,19 +55,14 @@ suite('languages >', function() {
     document.body.appendChild(section);
     var list = document.createElement('ul');
     section.appendChild(list);
-
-    // XXX in reality this method is async b/c it uses LanguageList.get;  here
-    // however it uses the mock which is sync.  Fix this in bug 1119865.
-    LanguageManager.buildLanguageList();
+    LanguageManager.get().then(function(langs) {
```
stasm: Does FTU's `LanguagesManager` have a `get` method?

apps/ftu/test/unit/language_test.js
```diff
@@ -55,19 +55,14 @@ suite('languages >', function() {
     document.body.appendChild(section);
     var list = document.createElement('ul');
     section.appendChild(list);
-
-    // XXX in reality this method is async b/c it uses LanguageList.get;  here
-    // however it uses the mock which is sync.  Fix this in bug 1119865.
-    LanguageManager.buildLanguageList();
```
stasm: Last time I looked at this code, I remember thinking that `buildLanguageList` should take a callback so that we can know in the tests when it finishes.  Right now we work around this by using sync mocks inside of it, but once we switch to a Promise-based API for `LanguageList`, we'll need a way to notify the test's setup that the FTU has finished building the list.

shared/test/unit/mocks/mock_language_list.js
```diff
@@ -11,8 +11,8 @@
         callback(langs);
       }
     },
-    get: function(callback) {
-      callback(langs, 'en-US');
+    get: function() {
+      Promise.resolve([langs, 'en-US']);
```
stasm: You should also be able to change `_languages` above to a real Promise with `Promise.resolve`.  Right now it pretends to be a `then`-able but is in fact synchronous for easier testing.

PR #28211, opened 2015-02-13
RenderThread call path refinement

steveck-chung:


autolander:

try-server-hook:
[steveck-chung steveck-chung](https://github.com/steveck-chung) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=d9c4ec5e90cc)

apps/sms/js/thread_list_ui.js
```diff
@@ -903,28 +889,30 @@ var ThreadListUI = {
...
+        targetIdx = i;
         break;
       }
     }
 
-    if (!threadFound) {
+    if (targetIdx === lastIdx) {
       threadsContainer.appendChild(node);
+    } else {
+      threadsContainer.insertBefore(node, threads[targetIdx + 1]);
```
julienw: I think you can just use inserBefore; if the second argument is null/undefined it's properly appended.

apps/sms/js/thread_list_ui.js
```diff
@@ -792,17 +775,20 @@ var ThreadListUI = {
...
         break;
       }
     }
-    if (!groupFound) {
-      ThreadListUI.container.appendChild(group);
+
+    if (targetIdx === lastIdx) {
+      this.container.appendChild(group);
+    } else {
+      this.container.insertBefore(group, headers[targetIdx + 1]);
```
julienw: Did you miss .parentNode ?

julienw:
I think we can have yet a better speedup if we use a javascript-based model instead of using the DOM. Can you maybe give it a try, using threads.js?

PR #28221, opened 2015-02-14
Bug 1069221 - Fixed visibility for screen readers while warnings are displayed

funktr0n:


try-server-hook:
[FunkTron FunkTron](https://github.com/FunkTron) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=7c391d65adf7)

apps/fm/js/fm.js
```diff
@@ -184,11 +184,24 @@ function updatePowerUI() {
 }
 
 function updateAntennaUI() {
-  $('antenna-warning').hidden = mozFMRadio.antennaAvailable;
```
yzen: I think we can do it a little bit simpler. Instead of tracking the state (based on which warning to display etc). Lets just do the following:
in updateAntennaUI and updateAirplaneModeUI we should just add 'warning' class to the document body if the warning is displayed as follows:
```
document.body.classList.toggle('warning', !mozFMRadio.antennaAvailable || airplaneModeEnabled);
```
Then in fm.css just have this styling:
```
.warning #container {
  visibility: hidden;
}
```

apps/fm/test/unit/fm_test.js
```diff
@@ -327,6 +332,15 @@ suite('FM', function() {
       test('#airplane-mode-warning is shown', function() {
         assert.equal(!!$('airplane-mode-warning').hidden, false);
       });
+
+      test('#container is hidden', function() {
+        assert.equal(!!('container').hidden, false);
```
yzen: Missing $

apps/fm/test/unit/fm_test.js
```diff
@@ -327,6 +332,15 @@ suite('FM', function() {
...
+      });
+
+      test('#antenna-warning is hidden', function() {
+        assert.equal(!!('antenna-warning').hidden, false);
```
yzen: Missing $

apps/fm/test/unit/fm_test.js
```diff
@@ -338,6 +352,20 @@ suite('FM', function() {
       test('#airplane-mode-warning is hidden', function() {
         assert.equal(!!$('airplane-mode-warning').hidden, true);
       });
+
+      test('#antenna-warning is on and container is hidden', function() {
```
yzen: Lets combine these 2 tests into 1. We also need a similar one for the airplane mode as well as the permutation of both: e.g. have antenna but in airplane mode, not in airplane mode but do not have an antenna.

apps/fm/js/fm.js
FunkTron: But there are cases where both warnings are on and visible at the same time, so shouldn't we only be showing one of those warnings?

apps/fm/test/unit/fm_test.js
FunkTron: Depending upon the response to the above comment, (i.e. if we decide not to show both warnings at the same time), aren't both permutations taken care of in the following sense...?

- If airplane mode is on, as it's the top-most layer in the z-index, we only really need to know that both the container and the antenna-warning are hidden.  We're not really concerned if the antenna-warning is on because it'll never stack higher than the airplane-mode-warning

- If airplane mode is off, now we need to check if the antenna-warning is on, so as to hide itself and the container appropriately when it's on/off.

apps/fm/js/fm.js
yzen: Good point. What about unifying both methods into one called something like updateWarningModeUI():
we would then do something like:
```
$('airplane-mode-warning').hidden = !airplaneModeEnabled;
$('antenna-warning').hidden = airplaneModeEnabled || mozFMRadio.antennaAvailable;
$('container').hidden = airplaneModeEnabled || !mozFMRadio.antennaAvailable;
```

apps/fm/test/unit/fm_test.js
yzen: Yeah, I should've checked that there's z-index priority.

try-server-hook:
[FunkTron FunkTron](https://github.com/FunkTron) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=96ee9411e87e)

try-server-hook:
[FunkTron FunkTron](https://github.com/FunkTron) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=46583a7e5a45)

KevinGrandon:
Last gaia-try run was good here, so let's land it.

PR #28275, opened 2015-02-17
Bug 1126524 - Collect search metrics in AppUsage data.

tamarahills:


try-server-hook:
[tamarahills tamarahills](https://github.com/tamarahills) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=2e99fcea82dd)

apps/system/js/app_usage_metrics.js
```diff
@@ -759,6 +767,25 @@
     return dayKey.replace(/-/g, '');
   };
 
+  UsageData.prototype.getSearchCounts = function(provider) {
+    var search = this.data.searches[provider];
+    var dayKey = this.getDayKey();
+    if(!search) {
+      // If no usage exists for this provider, create a new empty object for it.
+      this.data.searches[provider] = search = {};
+      debug('creating new object');
```
marshall: nit: maybe add something about which provider these debug statements are talking about

apps/system/js/rocketbar.js
```diff
@@ -567,6 +568,14 @@
       if (Service.currentApp.isPrivateBrowser()) {
         this.setInput('');
       }
+
+      // Only record the search if we have internet connectivity
+      if(navigator.onLine) {
+        var provider = window.SearchProvider('title');
```
marshall: does this function return a string normally?

apps/system/js/app_usage_metrics.js
```diff
@@ -785,6 +812,17 @@
     }
   };
 
+  UsageData.prototype.recordSearch = function(provider) {
+    debug('recordSearch', provider);
+    if(provider == null) {
```
marshall: nit: add a space between `if` and `(` (and a few other places in this patch)

apps/system/js/rocketbar.js
```diff
@@ -567,6 +568,14 @@
       if (Service.currentApp.isPrivateBrowser()) {
         this.setInput('');
       }
+
+      // Only record the search if we have internet connectivity
+      if(navigator.onLine) {
```
marshall: it may not matter, but should we move the online logic into AppUsageMetrics? is there a usecase in the future where we might want to know if a search occurred from the rocketbar, but the phone wasn't online?

apps/system/js/app_usage_metrics.js
tamarahills: fixed these.

apps/system/js/rocketbar.js
tamarahills: It returns the user's currently selected search provider: (e.g. 'google' 'yahoo')  If there is some error condition, it returns a false. I could check the return and not send the event.  I think we should probably log an error in this case because I don't think it is possible to have a setup where there is no selected search provider.

So, I'm glad you asked this because I did some more investigation on this and it looks like this function does not work after I do a |make reset-gaia|.  It seems to only work after I reboot the device after the make-reset gaia.  I'm going to look some more into this.

try-server-hook:
[tamarahills tamarahills](https://github.com/tamarahills) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=3bdab088d968)

apps/search/js/search.js
```diff
@@ -60,6 +61,20 @@
         });
       };
 
+      // Initialize the port for connecting to the system app to report metrics.
+      navigator.mozApps.getSelf().onsuccess = function() {
+        var app = this.result;
+        app.connect('app-metrics').then(function onConnAccepted(ports) {
+          ports.forEach(function(port) {
+            self._portMetrics = port;
```
marshall: seems like it would be more effecient to just set `self._portMetrics` directly, i.e

`self._portMetrics = ports[0]`

Is it possible for `ports` to be an empty array? (FWIW, this is my first time seeing app ports in use)

apps/search/js/search.js
```diff
@@ -246,6 +261,16 @@
     },
 
     /**
+     * Called to notify the AppUsageMetrics of a search
+     */
+    metricsNotify: function(provider) {
+      this._portMetrics.postMessage({
```
marshall: is it possible for `this._portMetrics` to be uninitialized when `metricsNotify` is called?

apps/search/test/unit/search_test.js
```diff
@@ -83,7 +90,7 @@ suite('search/search', function() {
       Search.init();
       MockNavigatormozApps.mTriggerLastRequestSuccess();
       assert.equal(MockNavigatormozApps.mLastConnectionKeyword,
-        'search-results');
+        'app-metrics');
```
marshall: instead of removing the check for `search-results`, should you check for both `search-results` and `app-metrics` ?

apps/system/js/app_usage_metrics.js
```diff
@@ -759,6 +767,25 @@
     return dayKey.replace(/-/g, '');
   };
 
+  UsageData.prototype.getSearchCounts = function(provider) {
+    var search = this.data.searches[provider];
+    var dayKey = this.getDayKey();
+    if(!search) {
```
marshall: nit: add a space between `if` and `(` (a few other places you did this below as well)

apps/system/js/app_usage_metrics.js
```diff
@@ -759,6 +767,25 @@
...
+      this.data.searches[provider] = search = {};
+      debug('creating new object for provider', provider);
+    }
+
+    var daySearch = search[dayKey];
+    if(!daySearch) {
+      daySearch = search[dayKey] = {
+        count: 0
+      };
+      debug('creating new object with count 0');
```
marshall: nit: this debug message doesn't give any context, is it necessary?

apps/system/js/app_usage_metrics.js
```diff
@@ -785,6 +812,24 @@
...
+      return;
+    }
+
+    // We don't want to report search metrics for local search and any other
+    // situation where we might be offline.  Check this here as this may change
+    // in the future.
+    if(navigator.onLine) {
+      var search = this.getSearchCounts(provider);
+      search.count++;
+      debug('Search Count : ', search.count);
```
marshall: nit: add the provider to this debug statement so we can monitor counts for each?

apps/system/js/app_usage_metrics.js
```diff
@@ -850,6 +895,24 @@
...
     }
+
+    // loop through all the search providers that we have data for
+    // and merge the new searches into the old searches.
+    for (var provider in newbatch.data.searches) {
+      var newsearch = newbatch.data.searches[provider];
+      var oldsearch = this.data.searches[provider];
+
+      for (var daykey in newsearch) {
+        var daySearch = oldsearch[daykey];
```
marshall: Verify that `oldsearch` actually exists here.. it's not guaranteed that a new batch of data will have the same set of search providers as an old set is it?  Here's what I'm thinking:

I submit a search using yahoo, then the batch of AppUsageMetrics is saved when I turn off my phone. I turn it on again, and search with google this time, but google doesn't exist in the first batch.. so `oldsearch` here would actually be undefined. does that sound plausible?

apps/system/test/unit/app_usage_metrics_test.js
```diff
@@ -787,6 +812,7 @@ suite('AppUsageMetrics:', function() {
         assert.equal(keys[0], dayKey);
       });
 
+      assert.deepEqual(payload.searches, metrics.data.searches);
```
marshall: can you add an assertion that your search for 'provider1' also exists in `payload.searches`?

shared/test/unit/mocks/mock_search_provider.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+/* exported MockSearchProvider */
+
+
+var MockSearchProvider = function(info) {
+  return 'testProvider';
+};
```
marshall: Do we really need 2 different implementations of MockSearchProvider?

marshall:
Looks like you'll need to rebase this as well :)

apps/search/js/search.js
tamarahills: My understanding is that the Rocketbar in the system app will load the Search app.  When the Rocketbar gets put into the active state, it loads the SearchApp at [1].  Then it waits for that to be loaded via a promise (I think this goes back to the initial problem we had with putting this in rocketbar on |make reset-gaia|.  I believe this then triggers the init function which initializes the ports at [2].

[1] https://github.com/tamarahills/gaia/blob/master/apps/system/js/rocketbar.js#L124
[2] https://github.com/tamarahills/gaia/blob/master/apps/search/js/search.js#L367

apps/search/js/search.js
tamarahills: Good question. This is also my first time working with ports and it doesn't seem to be that well documented.  A lot of the same code copy/pasted.  What I found out through testing is that ports is an empty array if there is no entry in manifest.webapp in any of the apps for app-metrics.  It also prepends an 'iac-' to everything, which threw me off as well.  We have one entry in the system manifest.webapp.  I changed this to take your suggestion.  I believe your suggestion is safe in our case since the manifest.webapp is fixed by us.  I'm not going to modify the self._port algorithm above (which I pretty much copy/pasted) as I don't know enough about how they are using it and if it makes sense for them to index this way.

apps/system/js/app_usage_metrics.js
tamarahills: removed this.  It was a debugging leftover.

apps/system/js/app_usage_metrics.js
tamarahills: changed this.

apps/system/js/app_usage_metrics.js
tamarahills: fixed

apps/search/test/unit/search_test.js
tamarahills: I have moved the check for the app-metrics into the separate file for maintaining the app-metrics port as per Dale's comment.

apps/system/test/unit/app_usage_metrics_test.js
tamarahills: Added this.

apps/system/js/app_usage_metrics.js
tamarahills: Yes, this is a good catch.  Thank you.  I've added an initialization if it's null.

shared/test/unit/mocks/mock_search_provider.js
tamarahills: I'm removing the one in system.

try-server-hook:
[tamarahills tamarahills](https://github.com/tamarahills) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=e168f6fb1449)

apps/system/js/app_usage_metrics.js
```diff
@@ -869,6 +941,11 @@
       var usage = new UsageData();
       if (data) {
         usage.data = data;
+        //Handle a scenario with old app data that does not have searches
+        if (typeof usage.data.searches === undefined) {
```
marshall: This condition should never return true, I think you want either:
`if (typeof usage.data.searches === 'undefined')`, or..
`if (usage.data.searches === undefined)`, or even briefer.. :)
`if (!usage.data.searches)`

apps/system/js/app_usage_metrics.js
```diff
@@ -544,6 +546,15 @@
               new Date(newStartTime).toString());
       }
       break;
+
+      case IACMETRICS:
```
marshall: nit: it looks like the indentation is off on the case block, I'm not sure if github is just playing tricks with the code though?

tamarahills: You are right.  I fixed this.

apps/system/js/app_usage_metrics.js
tamarahills: Thanks.  Changed this to the first one.

try-server-hook:
[tamarahills tamarahills](https://github.com/tamarahills) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=fa00f0803351)

PR #28276, opened 2015-02-17
Bug 1130547 - [email] marionettejs tests that use notification iconURL fail

jrburke:
apps/email/test/marionette/lib/notification.js has the core of the change, to use Notification.get() to get the notifications.

There were some other changes for allowing the files to be linted by jshint, and then the tests themselves to use the new methods in lib/notification.js. With this, all the marionette email tests pass locally for me.

email_evt.js was no longer needed because lib/notification.js can do the event emitting internally, and it makes more sense to do so internally to keep the knowledge of notification structures to that file.


try-server-hook:
[jrburke James Burke (jrburke)](https://github.com/jrburke) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=7b4d757347c7)

PR #28291, opened 2015-02-18
Bug 1134229 - Fix sort order for playlist in descending order.

hfiguiere:


try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=2af702b62092)

apps/music/js/ui/views/list_view.js
```diff
@@ -348,11 +348,12 @@ var ListView = {
     var data = this.dataSource[index];
     var keyRange = (target.dataset.keyRange != 'all') ?
       IDBKeyRange.only(target.dataset.keyRange) : null;
-    var title = data.metadata.title;
+    var title = data.metadata.title ? data.metadata.title :
+        data.metadata.l10nId;
```
dominickuo: @hfiguiere Do you think it will be more readable if we check `data.metadata.l10nId` directly? Because only the playlist items have `l10nId` field, so if `l10nId` exists, it should be playlists and we can assign the correct direction.

Probably the code will look like:
```JavaScript
var data.metadata.l10nId = l10nId;

var direction =
  (l10nId === 'playlists-most-played' ||
   l10nId === 'playlists-recently-added' ||
   l10nId === 'playlists-highest-rated') ?
   'prev' : 'next';
```

We used to use `title` to check them but now we use `l10nId`, that's what I didn't aware while refactoring, thanks for finding it!

hfiguiere: Minus the code error:
````
var l10nId = data.metadata.l10nId;
````

It sounds like a better idea. Will fix that.

Thanks

try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=b1ca0b03bca6)

PR #28324, opened 2015-02-19
Bug 1134436 - [Gallery][RTL] - Exposure slider is not mirrored r= djf

punamdahiya:


try-server-hook:
[punamdahiya punamdahiya](https://github.com/punamdahiya) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=9951f315d78f)

apps/gallery/style/gallery.css
```diff
@@ -920,3 +920,31 @@ a.filter-faded {
...
+  left: calc(3.9rem + (100% - 2 * 3.9rem) * 2 / 6);
+}
+
+html[dir="rtl"] #slider2 {
+  left: calc(3.9rem + (100% - 2 * 3.9rem) / 6);
+}
+
+html[dir="rtl"] #slider3 {
+  left: 3.9rem;
+}
```
davidflanagan: I wish there was a way to make this all less verbose. Would the RTL ones work if you set right instead of left? That would make them more parallel with the LTR styles because they'd have the same calculations.

Or, can you make it work by settting left and right to 0, and then setting margin-start to the calculated value? If that works, then you wouldn't need the special case for rtl.

This seems okay as it is, and I don't want to ask you to do a lot more work, so if margin-start can work, I'd suggest using that, but otherwise this is fine.

apps/gallery/js/ImageEditor.js
```diff
@@ -276,7 +285,12 @@ var exposureSlider = (function() {
     pixel -= thumbWidth / 2;
 
     // Move the thumb to that position
-    thumb.style.left = pixel + 'px';
+    if (navigator.mozL10n.language.direction === 'ltr') {
+      thumb.style.left = pixel + 'px';
+    }
+    else {
+      thumb.style.right = pixel + 'px';
+    }
```
davidflanagan: If you hard-coded left and right to 0, could you just set style.marginStart here to move the thumb?

apps/gallery/js/ImageEditor.js
```diff
@@ -124,6 +124,11 @@ function editPhoto(n) {
     $('edit-tool-apply-button').disabled = true;
 
     window.addEventListener('resize', resizeHandler);
+    // If the locale changes while user is using exposure edit option,
+    // then we need to re-position slider thumb for respective locale.
+    navigator.mozL10n.ready(function() {
+      exposureSlider.forceSetExposure(exposureSlider.getExposure());
+    });
```
davidflanagan: Thanks for remembering this. It seems like an incredibly rare occurrance to have the language direction changes while the exposure slider is displayed, but I guess it is the right thing to do to handle it.

The resizeHandler is added here in startEdit and removed in exitEdit(). With this change, you're calling ready() each time the user starts editing a new photo. I think if you add logging output at line 130, then edit photos a few times, and then switch the language, you'll find that the function is being executed once for each previous invocation of the editor.

Instead, you might need to add and remove event listeners for the 'localized' event

try-server-hook:
[punamdahiya punamdahiya](https://github.com/punamdahiya) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=b181c14f154d)

PR #28326, opened 2015-02-20
Bug 1134257 - test_ftu_skip_tour should reflect new Send Data initial st...

geoelectric:
...ate and separate SIM- and network-available conditions, r=jlorenzo


try-server-hook:
[geoelectric Geo Mealer (geoelectric)](https://github.com/geoelectric) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=c0fa2cbd91a5)

tests/python/gaia-ui-tests/gaiatest/tests/functional/ftu/test_ftu_skip_tour.py
```diff
@@ -41,18 +41,27 @@ def test_ftu_skip_tour(self):
         psk = self.testvars['wifi'].get('psk')
         keymanagement = self.testvars['wifi'].get('keyManagement')
 
+        # Assume a SIM is present if we assigned a phone number
+        # Don't trust internal API for this, since it's an external condition
+        has_a_sim = bool(self.testvars['local_phone_numbers'])
```
JohanLorenzo: :+1:

tests/python/gaia-ui-tests/gaiatest/tests/functional/ftu/test_ftu_skip_tour.py
```diff
@@ -101,9 +110,17 @@ def test_ftu_skip_tour(self):
...
-        self.ftu.tap_statistics_checkbox()
-        self.assertTrue(self.data_layer.get_setting('debug.performance_data.shared'))
+        # Verify the statistics checkbox reflects the initial internal state
+        Wait(self.marionette).until(
+            lambda m: self.ftu.is_statistics_enabled == share_data_default,
+            message="Send Data checkbox should match initial internal state")
+
+        # Tap the statistics checkbox and verify that it toggled internal state
+        self.ftu.toggle_statistics()
+        self.assertEqual(self.data_layer.get_setting(
```
JohanLorenzo: Like said in bug 1134445, this test showed that the default value has changed in the product, which is an actual bug. To answer on comment 8, I would leave the hardcoded value here.

tests/python/gaia-ui-tests/gaiatest/apps/ftu/app.py
```diff
@@ -334,9 +334,20 @@ def a11y_click_next_to_welcome_browser_section(self):
             Wait(self.marionette).until(expected.element_present(
                 *self._section_welcome_browser_locator))))
 
+    @property
+    def is_statistics_enabled(self):
+        return self.marionette.find_element(
+            *self._statistic_checkbox_locator).is_selected()
+
     def tap_statistics_checkbox(self):
```
JohanLorenzo: This function does the same thing as the one you created (without the check though). I would remove it and update the other place where this function is used (test_ftu_skip_tour_for_desktop).

tests/python/gaia-ui-tests/gaiatest/apps/ftu/app.py
```diff
@@ -334,9 +334,20 @@ def a11y_click_next_to_welcome_browser_section(self):
...
         self.marionette.find_element(*self._enable_statistic_checkbox_locator).tap()
 
+    def toggle_statistics(self):
```
JohanLorenzo: This implementation is way better than `tap_statistics_checkbox`.

One thing though. This is debatable, but I would prefer to have functions in the page class which don't depend on the state of the UI. 
The main reason is it makes the test more deterministic: If I call twice the same method, the same thing will happen on the UI. Also, it makes the expectation of the test writer clearer, like: "If I use `ftu.enable_send_statistics()`, I expect a active checkbox after this function call."

What do you think?

try-server-hook:
[geoelectric Geo Mealer (geoelectric)](https://github.com/geoelectric) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=467a76f90790)

tests/python/gaia-ui-tests/gaiatest/apps/ftu/app.py
geoelectric: I don't think it's either-or. Enable, Disable, and Toggle are all valid semantics. The difference is in what's guaranteed. Enable should enable no matter the previous setting, ditto Disable. Toggle means literally, whatever it is now, make it the other thing. Sometimes that's what you want.

In this case I wanted Toggle, because doing so let me use a simple control variable in the test instead of having to edit Enable/Disable if the default changes again. It's also consistent with handling in Settings (home of many checkboxes :) which is what I based many consistency decisions on.

I suppose I could have done Set_Shared_Data(value) to unconditionally set to a given value, but in this case I really did want to say "flip it and make sure it flips," not so much "set it here."

It's a fine point, and I'm hoping we can go with this (given new naming) for now. It's something to discuss again in terms of pattern best practices, for sure.

tests/python/gaia-ui-tests/gaiatest/apps/ftu/app.py
geoelectric: Posted in the bug, but I see a difference between a UI operation "tap" and a semantic operation "toggle the option however, with a check"

That said, my naming sucked in terms of making that differentation. I followed the lead in Settings, and shouldn't have.

I've changed the naming in the new PR to make the difference clear.

As for changing or removing the other function, unless you grepped every possible test base based on this one, including MTBF and so forth that's not in our repo, that's usually not a great idea. Once you publish an interface, it's generally better to keep it. But in this case I would have anyway.

geoelectric:
Differences from last:

* I changed the expected start state of Share Data to a hardcoded True.
* In order to make that a complete verification, I'm now verifying three things:

1) UI matches hardcoded expected start state
2) The UI state is consistent with back end
3) Toggling the UI state toggled the back end

Together that verifies everything Johan was concerned with, and also establishes a before state (necessary to know that your after state actually changed anything).

* I changed the naming of toggle_ and is_*_enabled to share_data instead of statistics. 

That clearly differentiates between functions dealing with nuts and bolts of the specific checkbox and functions dealing with the more general concept of the UI state, and should make more clear why these aren't redundant.

tests/python/gaia-ui-tests/gaiatest/tests/functional/ftu/test_ftu_skip_tour.py
```diff
@@ -40,19 +40,27 @@ def test_ftu_skip_tour(self):
         ssid = self.testvars['wifi']['ssid']
         psk = self.testvars['wifi'].get('psk')
         keymanagement = self.testvars['wifi'].get('keyManagement')
+        share_data_default = True
```
JohanLorenzo: Not sure we still need this variable now that it's hard coded.

geoelectric: Sorry to not have replied to you. It's a single point of change when they inevitably flip-flop on this decision, rather than having to change lines 115 and 126 separately.

Also, it's usually better to put a value in an explicit variable when it's meaningful, rather than using a magic True/False where otherwise not obvious--even when only using it once. The value is the variable name, for self-commenting.

JohanLorenzo: I would agree if we weren't in a test.

To me, assertions/waits should be done against not computed values, except if we're testing functions like `copy`. Having a single point of change seems less important in a test rather than in production code, simplicity would be the motto. 

I don't really think that hiding the value behind a variable makes the wait more self-documented (but the failure message does). That is to say, if you don't have enough context with this function, you could end up like:
``` Python
        Wait(self.marionette).until(
            lambda m: self.ftu.is_share_data_enabled == share_data_default, # What is the expected value already here?
 message="Send Data UI state should match initial internal state")

# Whereas:
        Wait(self.marionette).until(
            lambda m: self.ftu.is_share_data_enabled == True,
 message="Send Data UI state should match initial internal state") # Okay, the initial state is supposed to be true here.
```

[Here's a blog post](http://googletesting.blogspot.fr/2014/07/testing-on-toilet-dont-put-logic-in.html) from the Google Testing blog with one example of the drawback of having factorized variables. 

What do you think?

geoelectric: Yep, I've seen that post. Think there are a few differences:

* The variable they're creating is a partial "base" value with no real semantic importance, then they're computing in the test with concatenation. That causes the "two slashes" issue.
* The function is so small that there's minimal value in renaming the literal just to give it a name.
* The URL itself is self-documenting.
* The Google URL isn't likely to change.

OTOH, if at the top of file or in a commons file or whatever they'd had:

* google_plus_url = 'http://plus.google.com/'
* google_plus_photos_url = 'http://plus.google.com/u/0/photos'

...then I would think using those values in the test would be better. Note that there's no concatenation--string building of constants is usually bad for other reasons, especially localization. And that's one reason this is arguably better, those strings can be localized. 

But as far as what I did:

* The value stands alone with semantic meaning: the default is indeed True, period.
* The only computation is "not", which doesn't have the same issues as concatenation and is clear by itself.
* It's a large, sprawling test scenario function. That has its own possible issues, but I'm working in existing code.
* The sharing default is likely to change.

As for your added code comments, this is the most important part:

When I test the value, I really do mean to test that the value matches the system default, not that the value specifically matches True. 

That's very, very important in terms of communicating code meaning. Assigning a name means you're assigning a level of abstraction. 

And further, writing this from scratch, that default value would come from a data file of system defaults. They're external givens and not otherwise important to the intention of the test.

I suspect that's the difference here. You think the test should test for True, but I disagree. It should test that the initial UI value matches the default system value. 

And honestly, I disagree with that too, which is why I didn't hardcode the test the first time--the actual functionality of the UI is to come in with the *current* system value, not the default system value. As it stands, the test isn't idempotent without a reset, because the test itself changes that value so when it's rerun it will fail.

So I guess, if anything, what I would say is this test is less robust because we've mixed back-end testing concerns with UI testing concerns. But since that was at your request, I think we both need to roll with that one.

JohanLorenzo: Okay, I don't see any reason why this constant should be used, then. Thank you for the explanation.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ce8ac915646b7b8c217c0ab617bd7c76f8301c4b

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=05a7a5455609b9113b47cd6959dccac908de5493

geoelectric:
Hosed this pull request on the rebase. Closing out and will regenerate.

PR #28333, opened 2015-02-20
Bug 1134910 - dialer are always LTR even if current language is RTL. So ...

autra:
...for ellipsis, the end is always to the right, and the beginning always to the left


try-server-hook:
[autra Augustin Trancart (autra)](https://github.com/autra) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=2cc951cfd92d)

PR #28341, opened 2015-02-20
Bug 1064929 - Don't reset scroll position while indexing music. r=dkuo

Cwiiis:


try-server-hook:
[Cwiiis Chris Lord (Cwiiis)](https://github.com/Cwiiis) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=9209b3a100d5)

apps/music/js/ui/views/tiles_view.js
```diff
@@ -33,15 +33,17 @@ var TilesView = {
...
     this.view.addEventListener('click', this);
     this.view.addEventListener('input', this);
     this.view.addEventListener('touchend', this);
+    this.view.addEventListener('scroll', this.cancelHideSearch.bind(this));
     this.searchInput.addEventListener('focus', this);
     this.searchInput.addEventListener('keypress', this);
   },
 
-  clean: function tv_clean() {
+  clean: function tv_clean(ignoreScroll) {
```
dominickuo: TilesView.clean() is only called in music.js and do you think it's necessary to have `ignoreScroll` for options? because it should be always true in this patch.

PR #28342, opened 2015-02-20
Bug 1132191 - improved overall accessibility of the video app.

yzen:


try-server-hook:
[yzen Yura Zenevich (yzen)](https://github.com/yzen) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=686f27de6c4b)

apps/video/js/video.js
```diff
@@ -785,6 +790,17 @@ function showOverlay(id) {
   });
 }
 
+function setVideoAccessibility(asButton) {
```
russnicoletti: Please add a comment/function description for this function as its functionality is not intuitive

russnicoletti: Please add a comment/function description for this function as its functionality is not intuitive

apps/video/view.html
```diff
@@ -16,9 +16,11 @@
   <body role="application" class="skin-dark layout-fullscreen-player view-activity theme-media">
     <!-- display this at startup while we create thumbnails -->
     <div id="spinner-overlay"><div id="spinner"></div></div>
+    <div id="spinner-overlay" role="progress" data-l10n-id="spinner-overlay">
+      <div id="spinner"></div></div>
     <section role="region" id="player-view">
       <div id="video-container">
-        <video id="player" preload="metadata"></video>
+        <video id="player" preload="metadata" aria-hidden="true"></video>
```
russnicoletti: Recognizing I know little about ARIA, what I've read says the "aria-hidden" attribute should be set to "false" or removed when the element is presented. I don't see that happen for the elements for which 'aria-hidden="true"' has been added.

apps/video/view.html
```diff
@@ -27,26 +29,31 @@
...
-          <footer id="videoBar">
-            <div id="timeSlider">
-              <span id="elapsed-text"></span>
-              <div id="slider-wrapper">
-                <div id="elapsedTime" class="progress"></div>
-                <div id="bufferedTime" class="progress"></div>
-                <div id="timeBackground" class="progress"></div>
-                <button id="playHead"></button>
+          <div id="hide-controls" role="button"
+               data-l10n-id="hide-controls-button"></div>
```
russnicoletti: Nit: to be consistent, the ending 'div' tag should be on a separate line, lined up with the beginning 'div' tag.

apps/video/view.html
yzen: So the video element is used to display the video itself, but it makes little sense for the screen reader user, unless it exposes default browser controls. So in our case here, if we use aria-hidden="true" we hide if from the assistive technology only, but keep it visible for the sighted user.

apps/video/index.html
```diff
@@ -65,59 +65,73 @@ <h1 id="thumbnails-number-selected"></h1>
 
         <!-- This element gets positioned over different view elements -->
         <!-- It gets styled differently depending on what it is over -->
-        <ul id="thumbnails"></ul> <!-- Thumbnails inserted here -->
+        <ul role="presentation" id="thumbnails"></ul> <!-- Thumbnails inserted here -->
```
russnicoletti: I think it would be helpful to include a comment regarding the "presentation" role and why it's being used here.

apps/video/index.html
```diff
@@ -65,59 +65,73 @@ <h1 id="thumbnails-number-selected"></h1>
...
-                <span id="elapsed-text"></span>
-                <div id="slider-wrapper">
-                  <div id="elapsedTime" class="progress"></div>
-                  <div id="bufferedTime" class="progress"></div>
-                  <div id="timeBackground" class="progress"></div>
-                  <button id="playHead"></button>
+            <div id="hide-controls" role="button"
+                 data-l10n-id="hide-controls-button"></div>
+            <div role="toolbar">
+              <footer id="videoBar" role="presentation">
```
russnicoletti: I think it would be helpful to include a comment regarding the "presentation" role and why it's being used here.

apps/video/index.html
```diff
@@ -65,59 +65,73 @@ <h1 id="thumbnails-number-selected"></h1>
...
-              <div id="videoToolBar">
-                <button id="seek-backward" class="player-controls-button" data-icon="skip-back"></button>
-                <button id="play" class="player-controls-button" data-icon="pause"></button>
-                <button id="seek-forward" class="player-controls-button" data-icon="skip-forward"></button>
-              </div>
-            </footer>
+                <div id="fullscreen-button" role="button"
+                     data-l10n-id="fullscreen-button"></div>
+              </footer>
+              <footer id="videoControlBar" role="presentation">
```
russnicoletti: I think it would be helpful to include a comment regarding the "presentation" role and why it's being used here.

apps/video/index.html
```diff
@@ -152,18 +167,20 @@ <h1 id="video-title"></h1>
     <!-- This element is the template of each thumbnail item -->
     <div id="thumbnail-group-header" class="hidden">
       <!--
-        <li>
-          <div class="thumbnail-group-header"></div>
-          <ul class="thumbnail-group-container"></ul>
+        <li role="presentation">
```
russnicoletti: I think it would be helpful to include a comment regarding the "presentation" role and why it's being used here.

apps/video/index.html
```diff
@@ -152,18 +167,20 @@ <h1 id="video-title"></h1>
...
       -->
     </div>
     <div id="thumbnail-template" class="hidden">
       <!--
-        <li class="thumbnail">
+        <li class="thumbnail" role="option">
           <div class="inner">
             <div class="unwatched"></div>
-            <div class="img"></div>
+            <div class="img" role="presentation"></div>
```
russnicoletti: I think it would be helpful to include a comment regarding the "presentation" role and why it's being used here.

apps/video/view.html
```diff
@@ -27,26 +29,31 @@
...
+          <div role="toolbar">
+            <footer id="videoBar" role="presentation">
```
russnicoletti: I think it would be helpful to include a comment regarding the "presentation" role and why it's being used here.

apps/video/index.html
```diff
@@ -65,59 +65,73 @@ <h1 id="thumbnails-number-selected"></h1>
...
 
       <!-- (maximized) video player -->
       <section role="region" id="player-view">
         <!-- display this at startup while we create thumbnails -->
-        <div id="spinner-overlay" class="hidden"><div id="spinner"></div></div>
+        <div id="spinner-overlay" class="hidden" role="progress"
+             data-l10n-id="spinner-overlay"><div id="spinner"></div></div>
         <div id="video-container">
-          <video src="about:blank" id="player"></video>
+          <video src="about:blank" id="player" aria-hidden="true"></video>
```
russnicoletti: Recognizing I know little about ARIA, what I've read says the "aria-hidden" attribute should be set to "false" or removed when the element is presented. I don't see that happen for the elements for which 'aria-hidden="true"' has been added.

apps/video/index.html
```diff
@@ -65,59 +65,73 @@ <h1 id="thumbnails-number-selected"></h1>
...
             </gaia-header>
-            <footer id="videoBar">
-              <div id="timeSlider">
-                <span id="elapsed-text"></span>
-                <div id="slider-wrapper">
-                  <div id="elapsedTime" class="progress"></div>
-                  <div id="bufferedTime" class="progress"></div>
-                  <div id="timeBackground" class="progress"></div>
-                  <button id="playHead"></button>
+            <div id="hide-controls" role="button"
```
russnicoletti: Why introduce a new 'hide-controls' element? It is functioning the same as the 'video-container' in that it is toggling the video controls visibility. Also, it is the 'video-controls' element that is being used to hold the 'show video controls' aria label. I'm not seeing a need for an additional element here.

apps/video/view.html
russnicoletti: I think it would be helpful to include a comment regarding the "presentation" role and why it's being used here. The comment can be written such that it can be used to understand why the "presentation" role is being used on other elements in the file so that the same comment doesn't need to be included wherever the role is used.

apps/video/index.html
yzen: Unfortunately, screen reader users can not activate elements on screen that are not focusable (e.g. the so called empty space). Screen reader would need to first be focused on something to be able to activate it. Hence the area (outside of header/footer) needs to be something that the screen reader can treat as a button.

apps/video/js/video.js
```diff
@@ -116,6 +116,7 @@ document.addEventListener('visibilitychange', function visibilityChange() {
   else {
     if (playerShowing) {
       setControlsVisibility(true);
+      setVideoAccessibility(false);
```
russnicoletti: Since 'setVideoAccessibility' is being called everywhere where 'setControlsVisibility' is being called, I suggest moving the 'setVideoAccessibility' call into 'setControlsVisibility'

apps/video/js/video.js
```diff
@@ -1289,6 +1325,22 @@ function handleSliderTouchMove(event) {
...
+  // The standard accessible control for sliders is arrow up/down keys.
+  // Our screenreader synthesizes those events on swipe up/down gestures.
+  // Currently, we only allow screen reader users to adjust sliders with a
+  // constant step size (there is no page up/down equivalent). In the case
+  // of videos, we make sure that the maximum amount of steps for the entire
+  // duration is 20, or 2 second increments if the duration is less then 40
+  // seconds.
+  var step = Math.max(dom.player.duration / 20, 2);
+  if (event.keyCode === event.DOM_VK_DOWN) {
+    dom.player.currentTime -= step;
```
russnicoletti: As the video app is using the fastSeek API call for seeking within a video, I suggest it be used here also.

yzen: Currently the only mode of interaction with a slider (with screen reader) is by performing one finger swipes up or down. They are translated into keypresses. There's not concept (ATM) of a position of the tap, and instead we track the position of the screen reader focus. In this case the screen reader focuses on the whole markup block that has a role="slider" and treats it as an atomic unit.

yzen: This is the same approach we use here: https://github.com/mozilla-b2g/gaia/blob/master/shared/js/media/video_player.js#L488

apps/video/js/video.js
```diff
@@ -19,7 +19,7 @@ var ids = ['thumbnail-list-title', 'thumbnails', 'thumbnails-video-button',
            'spinner-overlay', 'thumbnails-single-delete-button',
            'thumbnails-single-share-button', 'thumbnails-single-info-button',
            'info-view', 'info-close-button', 'player', 'overlay',
-           'overlay-title', 'overlay-text', 'overlay-menu',
+           'overlay-title', 'overlay-text', 'overlay-menu', 'hide-controls',
```
russnicoletti: TBD: is 'hide-controls' necessary/useful?

apps/video/js/video.js
```diff
@@ -470,21 +474,33 @@ function showOptionsView() {
...
 }
 
 function hideOptionsView() {
   dom.optionsView.classList.add('hidden');
+  document.body.classList.remove('options-view');
+}
+
+function setDisabled(element, disabled) {
+  element.classList.toggle('disabled', disabled);
+  element.setAttribute('aria-disabled', disabled);
```
russnicoletti: Comment would be helpful here

apps/video/js/video.js
```diff
@@ -470,21 +474,33 @@ function showOptionsView() {
...
+}
+
+function setSelected(element, selected) {
+  element.classList.toggle('selected', selected);
+  element.setAttribute('aria-selected', selected);
```
russnicoletti: Comment about 'aria-selected' would be helpful here.

apps/video/js/view.js
```diff
@@ -126,7 +126,7 @@ navigator.mozSetMessageHandler('activity', function viewVideo(activity) {
     var ids = ['player', 'player-view', 'videoControls',
                'player-header', 'play', 'playHead', 'video-container',
                'elapsedTime', 'video-title', 'duration-text', 'elapsed-text',
-               'slider-wrapper', 'spinner-overlay',
+               'slider-wrapper', 'spinner-overlay', 'hide-controls',
```
russnicoletti: TBD whether 'hide-controls' is necessary/useful.

apps/video/js/view.js
```diff
@@ -188,6 +192,17 @@ navigator.mozSetMessageHandler('activity', function viewVideo(activity) {
     }
   }
 
+  function setVideoAccessibility(asButton) {
```
russnicoletti: Same issue as in video.js regarding whether videoControls element can be used to store the aria label.

apps/video/index.html
```diff
@@ -65,59 +65,73 @@ <h1 id="thumbnails-number-selected"></h1>
...
-                <button id="seek-backward" class="player-controls-button" data-icon="skip-back"></button>
-                <button id="play" class="player-controls-button" data-icon="pause"></button>
-                <button id="seek-forward" class="player-controls-button" data-icon="skip-forward"></button>
-              </div>
-            </footer>
+                <div id="fullscreen-button" role="button"
+                     data-l10n-id="fullscreen-button"></div>
+              </footer>
+              <footer id="videoControlBar" role="presentation">
+                <div id="videoToolBar">
```
russnicoletti: Should videoToolBar have a l10n-id -- 'video-toolbar'? That id exists in the locales properties file...

apps/video/locales/video.en-US.properties
```diff
@@ -79,3 +79,25 @@ date-group-header = %B %Y
...
+# screen.
+video-button.ariaLabel = Video
+select-button.ariaLabel = Select
+delete-button.ariaLabel = Delete
+share-button.ariaLabel = Share
+info-view.ariaLabel = Info
+options-view.ariaLabel = Options
+confirm-dialog.ariaLabel = Confirm
+videos-list.ariaLabel = Videos
+video-toolbar.ariaLabel = Controls
```
russnicoletti: I don't see where this is used. Should videoToolBar have a l10n-id of 'video-toolbar'?

apps/video/style/video.css
```diff
@@ -29,13 +29,12 @@ body > section {
...
-/*
- * while videoControls is hidden,
- * it should be stay on screen for mousedown event
- */
-#videoControls.hidden {
-  display: block;
-  visibility: visible;
+/* Stretch the hide controls button to cover all space outside of the header
+   and the control toolbars. */
+#hide-controls {
```
russnicoletti: TBD whether 'hide-controls' is necessary/useful.

apps/video/style/video.css
```diff
@@ -443,21 +442,21 @@ header .icon:after {
 /* video player controls */
 #videoControls {
   position: absolute;
+  display: block;
```
russnicoletti: TBD as to whether this change is necessary based on whether 'hide-controls' element is necessary.

apps/video/style/video.css
```diff
@@ -443,21 +442,21 @@ header .icon:after {
...
   top: 0;
   left: 0;
   width: 100%;
   height: 100%;
   z-index: 10;
   opacity: 1;
-  will-change: opacity;
-  -moz-transition: opacity 0.5s;
+  visibility: visible;
```
russnicoletti: TBD as to whether this change is necessary based on whether 'hide-controls' element is necessary.

apps/video/style/video.css
```diff
@@ -443,21 +442,21 @@ header .icon:after {
...
+  will-change: opacity, visibility;
+  transition: visibility 0s, opacity 0.5s;
 }
```
russnicoletti: TBD as to whether these changes are necessary based on whether 'hide-controls' element is necessary.

apps/video/style/video.css
```diff
@@ -443,21 +442,21 @@ header .icon:after {
...
 #videoControls.hidden {
   opacity: 0;
-}
-
-#videoControls.hidden > * {
-  pointer-events: none;
+  visibility: hidden;
+  transition-delay: 0.5s, 0s;
 }
```
russnicoletti: TBD as to whether these changes are necessary based on whether 'hide-controls' element is necessary.

apps/video/view.html
```diff
@@ -27,26 +29,31 @@
...
+                  <div id="elapsedTime" class="progress"></div>
+                  <div id="bufferedTime" class="progress"></div>
+                  <div id="timeBackground" class="progress"></div>
+                  <button id="playHead"></button>
+                </div>
+                <span id="duration-text" aria-hidden="true"></span>
+              </div>
+            </footer>
+            <footer id="videoControlBar" role="presentation">
+              <div id="videoToolBar">
```
russnicoletti: Should this have a 'video-toolbar' l10n-id?

try-server-hook:
[yzen Yura Zenevich (yzen)](https://github.com/yzen) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=0201430a368f)

apps/video/index.html
```diff
@@ -64,59 +64,79 @@ <h1 id="thumbnails-number-selected"></h1>
 
         <!-- This element gets positioned over different view elements -->
         <!-- It gets styled differently depending on what it is over -->
-        <ul id="thumbnails"></ul> <!-- Thumbnails inserted here -->
+        <!-- The real thumbnails list is actually inside this ul. We need
+             to remove the semantics of this list to be less verbose, in the
+             screen reader is on. -->
+        <ul role="presentation" id="thumbnails"></ul> <!-- Thumbnails inserted here -->
```
russnicoletti: I'm thinking the comment is supposed to say "The real thumbnails list is actually inside this ul. We need to remove the semantics of this list if the screen reader is on." Is that correct?

yzen: Yep, ill clean that up

apps/video/index.html
```diff
@@ -64,59 +64,79 @@ <h1 id="thumbnails-number-selected"></h1>
...
-              </div>
-              <div id="fullscreen-button"></div>
-            </footer>
-            <footer id="videoControlBar">
-              <div id="videoToolBar">
-                <button id="seek-backward" class="player-controls-button" data-icon="skip-back"></button>
-                <button id="play" class="player-controls-button" data-icon="pause"></button>
-                <button id="seek-forward" class="player-controls-button" data-icon="skip-forward"></button>
+        <!-- video controls footer toolbar -->
+        <div role="toolbar" class="video-controls video-controls-bottom">
```
russnicoletti: I understand the motivation in moving the 'gaia-header' element so that the order of the elements is consistent with the presentation. However, I'm not able to see the motivation for removing the 'videoControls' element and replacing it with a different div element. Why not add the 'toolbar' role and 'video-controls' class to the existing videoControls element? That seems like less of a change, i.e., simpler.

apps/video/index.html
```diff
@@ -64,59 +64,79 @@ <h1 id="thumbnails-number-selected"></h1>
...
+          <!-- It is simpler and less verbose (for the screen reader) to not
+               expose that this block is a footer and consider it part of
+               its toolbar container. -->
+          <footer id="videoBar" role="presentation">
```
russnicoletti: I assume "expose this block is a footer" should be "expose this block as a footer"

apps/video/index.html
```diff
@@ -151,18 +172,20 @@ <h1 id="video-title"></h1>
     <!-- This element is the template of each thumbnail item -->
     <div id="thumbnail-group-header" class="hidden">
       <!--
-        <li>
-          <div class="thumbnail-group-header"></div>
-          <ul class="thumbnail-group-container"></ul>
+        <li role="presentation">
+          <div class="thumbnail-group-header" role="heading"
+               aria-level="1"></div>
+          <ul class="thumbnail-group-container" role="listbox"
```
russnicoletti: I'm curious why 'aria-level' is used here in the thumbnail-group-header template but not in the thumbnail template below.

apps/video/js/video.js
```diff
@@ -241,11 +241,14 @@ function initPlayerControls() {
   dom.player.addEventListener('ended', playerEnded);
 
   // handle user tapping events
-  dom.videoControls.addEventListener('click', toggleVideoControls, true);
   dom.play.addEventListener('click', handlePlayButtonClick);
   dom.playerHeader.addEventListener('action', handleCloseButtonClick);
   dom.pickerDone.addEventListener('click', postPickResult);
   dom.options.addEventListener('click', showOptionsView);
+  dom.videoContainer.addEventListener('click', toggleVideoControls);
```
russnicoletti: What is the reason for removing the videoControls element and instead using the video-container element to toggle the video controls?

apps/video/js/video.js
```diff
@@ -470,21 +470,37 @@ function showOptionsView() {
...
 }
 
 function hideOptionsView() {
   dom.optionsView.classList.add('hidden');
+  document.body.classList.remove('options-view');
+}
+
+function setDisabled(element, disabled) {
+  element.classList.toggle('disabled', disabled);
+  // Set ARIA disabled attribute to maintain semantic meaning for the assistive
```
russnicoletti: Please add newline at the beginning of the comment to increase readability.

apps/video/js/video.js
```diff
@@ -470,21 +470,37 @@ function showOptionsView() {
...
+  // technologies like screen reader.
+  element.setAttribute('aria-disabled', disabled);
+}
+
+function setSelected(element, selected) {
+  element.classList.toggle('selected', selected);
+  // Set ARIA selected attribute to maintain semantic meaning for the assistive
```
russnicoletti: Please add newline at the beginning of the comment to increase readability.

apps/video/js/video.js
```diff
@@ -791,12 +796,19 @@ function setControlsVisibility(visible) {
   if (isPhone || isPortrait ||
       currentLayoutMode !== LAYOUT_MODE.list) {
 
-    dom.videoControls.classList[visible ? 'remove' : 'add']('hidden');
+    dom.playerView.classList[visible ? 'remove' : 'add'](
+      'video-controls-hidden');
     controlShowing = visible;
```
russnicoletti: Same comment regarding videoControls vs. video-container

apps/video/js/video.js
```diff
@@ -1029,11 +1041,11 @@ function showPlayer(video, autoPlay, enterFullscreen, keepControls) {
       switchLayout(LAYOUT_MODE.fullscreenPlayer);
     }
 
-    dom.durationText.textContent = MediaUtils.formatDuration(
-      dom.player.duration);
+    var formattedTime = MediaUtils.formatDuration(dom.player.duration);
+    dom.durationText.textContent = formattedTime;
```
russnicoletti: IMO, the code would be more readable if ''formattedTime' was 'formatedDuration', since the currentTime is also formatted below; 'formattedTime' is a bit vague...

apps/video/js/video.js
russnicoletti: This is one example of the problems of having two video player implementations (one in the video app and one in shared/js/media used by gallery and camera). We are working on fixing this by creating a video player web component. In any event, the video app code is more active and current. There was an explicit decision to use fastSeek as opposed to manipulating currentTime. That this change was only made in the video app and not in the video player code used by gallery and camera was an oversight that was not fixed due to our goal to have a single web component shared by all apps using a video player.

apps/video/js/view.js
```diff
@@ -518,13 +555,22 @@ navigator.mozSetMessageHandler('activity', function viewVideo(activity) {
     return filename.substring(filename.lastIndexOf('/') + 1);
   }
 
+  function setDisabled(element, disabled) {
+    element.classList.toggle('disabled', disabled);
+    // Set ARIA disabled attribute to maintain semantic meaning for the
```
russnicoletti: Please add whitespace before the comment.

apps/video/js/view.js
```diff
@@ -173,7 +173,10 @@ navigator.mozSetMessageHandler('activity', function viewVideo(activity) {
     dom.playerHeader.addEventListener('action', done);
     dom.save.addEventListener('click', save);
     // show/hide controls
-    dom.videoControls.addEventListener('click', toggleVideoControls, true);
+    dom.videoContainer.addEventListener('click', toggleVideoControls);
```
russnicoletti: TBD: videoControls or video-container?

apps/video/js/view.js
```diff
@@ -283,11 +286,11 @@ navigator.mozSetMessageHandler('activity', function viewVideo(activity) {
   // show video player
   function showPlayer(url, title) {
     function handleLoadedMetadata() {
-      dom.durationText.textContent = MediaUtils.formatDuration(
-        dom.player.duration);
+      var formattedTime = MediaUtils.formatDuration(dom.player.duration);
```
russnicoletti: A more precise name for this variable would make the code more readable.

apps/video/js/view.js
```diff
@@ -503,6 +524,22 @@ navigator.mozSetMessageHandler('activity', function viewVideo(activity) {
...
+    // Currently, we only allow screen reader users to adjust sliders with a
+    // constant step size (there is no page up/down equivalent). In the case
+    // of videos, we make sure that the maximum amount of steps for the entire
+    // duration is 20, or 2 second increments if the duration is less then 40
+    // seconds.
+    var step = Math.max(dom.player.duration / 20, 2);
+    if (event.keyCode === event.DOM_VK_DOWN) {
+      dom.player.currentTime -= step;
+    } else if (event.keyCode === event.DOM_VK_UP) {
+      dom.player.currentTime += step;
```
russnicoletti: fastSeek vs currentTime

apps/video/test/unit/video_test.js
```diff
@@ -1787,22 +1857,29 @@ suite('Video App Unit Tests', function() {
...
       thumbnail.htmlNode.classList.add('selected'); // selected
+      thumbnail.htmlNode.setAttribute('aria-selected', true);
       updateSelection(videodata);
       assert.isFalse(thumbnail.htmlNode.classList.contains('selected'),
                      'thumbnail should not contain \'selected\' class');
+      assert.equal(thumbnail.htmlNode.getAttribute('aria-selected'), 'false',
+                   'thumbnail aria-selected attribute should be set to false');
     });
 
     test('#updateSelection: update (add to) selected filenames', function() {
```
russnicoletti: Looking at these tests, I don't see the difference between this test and the first 'updateSelection' test. They both seem to be testing the same scenario; they assert different information for some reason. Do you see any difference in these tests. If not, can you combine this with the first test so that there is one test with a union of the asserts from the two tests? That would be great.

apps/video/test/unit/video_test.js
```diff
@@ -1815,6 +1892,7 @@ suite('Video App Unit Tests', function() {
     test('#updateSelection: update (remove) selected filenames', function() {
```
russnicoletti: Again, I don't see any difference between this test and the second updateSelection test that tests a thumbnail that was previously selected is deselected. If you also see no difference, it would be great to combine the two into one test.

apps/video/index.html
```diff
@@ -64,59 +64,79 @@ <h1 id="thumbnails-number-selected"></h1>
...
+        <div id="spinner-overlay" class="hidden" role="progress"
+             data-l10n-id="spinner-overlay"><div id="spinner"></div></div>
+        <!-- video controls header -->
+        <gaia-header id="player-header" action="back" class="video-controls">
+          <h1 id="video-title"></h1>
+          <button id="options" data-icon="more"
+                  data-l10n-id="options-more"></button>
+          <button id="picker-done" data-l10n-id="done">Done</button>
+        </gaia-header>
+        <div id="video-container" role="button">
```
russnicoletti: Perhaps a comment about how the video container is acting like a button.

apps/video/view.html
```diff
@@ -15,47 +15,56 @@
...
-              <button id="play" class="player-controls-button" data-icon="pause"></button>
-              <button id="seek-forward" class="player-controls-button" data-icon="skip-forward"></button>
+      <!-- video controls footer toolbar -->
+      <div role="toolbar" class="video-controls video-controls-bottom">
+        <!-- It is simpler and less verbose (for the screen reader) to not
+             expose that this block is a footer and consider it part of
+             its toolbar container. -->
+        <footer id="videoBar" role="presentation">
+          <div id="timeSlider" role="slider">
+            <span id="elapsed-text" aria-hidden="true"></span>
```
russnicoletti: Why use 'aria-hidden=true' here and on the other time slider children?

apps/video/index.html
```diff
@@ -64,59 +64,79 @@ <h1 id="thumbnails-number-selected"></h1>
...
-                <button id="seek-forward" class="player-controls-button" data-icon="skip-forward"></button>
+        <!-- video controls footer toolbar -->
+        <div role="toolbar" class="video-controls video-controls-bottom">
+          <!-- It is simpler and less verbose (for the screen reader) to not
+               expose that this block is a footer and consider it part of
+               its toolbar container. -->
+          <footer id="videoBar" role="presentation">
+            <div id="timeSlider" role="slider">
+              <span id="elapsed-text" aria-hidden="true"></span>
+              <div id="slider-wrapper" aria-hidden="true">
```
russnicoletti: Why use 'aria-hidden=true' here and on the other time slider children?

apps/video/js/video.js
```diff
@@ -299,14 +302,9 @@ function toggleVideoControls(e) {
...
-      // If control not shown, tap any place to show it.
-      setControlsVisibility(true);
-      e.cancelBubble = true;
-    } else if (e.originalTarget === dom.videoControls) {
-      // If control is shown, only tap the empty area should show it.
-      setControlsVisibility(false);
-    }
+    e.cancelBubble = !controlShowing;
+    // If control not shown, tap any place to show it.
+    setControlsVisibility(!controlShowing);
```
russnicoletti: Given the code change here, this comment is no longer relevant. My suggestion is either no comment or something mentioning toggling the visibility of the controls.

apps/video/js/view.js
```diff
@@ -210,14 +218,9 @@ navigator.mozSetMessageHandler('activity', function viewVideo(activity) {
...
-      // If control not shown, tap any place to show it.
-      setControlsVisibility(true);
-      e.cancelBubble = true;
-    } else if (e.originalTarget === dom.videoControls) {
-      // If control is shown, only tap the empty area should show it.
-      setControlsVisibility(false);
-    }
+    e.cancelBubble = !controlShowing;
+    // If control not shown, tap any place to show it.
+    setControlsVisibility(!controlShowing);
```
russnicoletti: This comment is no longer relevant given the code change. My suggestion is to remove it.

apps/video/index.html
yzen: There are several things here, I believe. video-controls class is also used by the player-header. I could remove it and then use set botton and top to 0 fore video-controls, but I would get into the same problem that existed before this pull request. I would overlay the video container with video controls, and unless there was something actionable in between (which was the previous approach of creating a pseudo-button over the video) the screen reader would never be able to toggle controls.

apps/video/index.html
yzen: So here the header is for the group and it makes sense to expose it semantically as such. The thumbnail is represented as an option (actionable list item within a listbox). Our screen reader utters options in one sentence and treats them as atomic elements so we opt out for being terse there.

try-server-hook:
[yzen Yura Zenevich (yzen)](https://github.com/yzen) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=12f182cc2c5c)

apps/video/style/video.css
```diff
@@ -441,23 +435,23 @@ header .icon:after {
...
-#videoControls.hidden {
+.video-controls-hidden .video-controls {
   opacity: 0;
+  visibility: hidden;
+  transition-delay: 0.5s, 0s;
 }
 
-#videoControls.hidden > * {
-  pointer-events: none;
+.video-controls-bottom {
```
russnicoletti: IMO, it is cleaner to use 'footer .video-controls' here as opposed to introducing another class - 'video-controls-bottom'.

apps/video/index.html
russnicoletti: Ok, I understand. I have one last comment on this subject in the video.css section below.

apps/video/view.html
```diff
@@ -15,47 +15,62 @@
...
-              <span id="duration-text"></span>
-            </div>
-          </footer>
-          <footer id="videoControlBar">
-            <div id="videoToolBar">
-              <button id="seek-backward" class="player-controls-button" data-icon="skip-back"></button>
-              <button id="play" class="player-controls-button" data-icon="pause"></button>
-              <button id="seek-forward" class="player-controls-button" data-icon="skip-forward"></button>
+      <!-- video controls footer toolbar -->
+      <div role="toolbar" class="video-controls video-controls-bottom">
```
russnicoletti: 'bottom' instead of 'video-controls-bottom'?

try-server-hook:
[yzen Yura Zenevich (yzen)](https://github.com/yzen) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=e9e33fa361e4)

PR #28350, opened 2015-02-21
Bug 1134097 - Implemented Accessibility Hints into Gaia

funktr0n:


try-server-hook:
[FunkTron FunkTron](https://github.com/FunkTron) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=fa9e867ce9ae)

apps/system/js/accessibility.js
```diff
@@ -309,7 +325,17 @@
           return;
       }
 
-      this.speak(aDetails.data, null, {
+      var accessibilityObj = this;
+      function hintCallback() {
+        if (options.hints) {
+          if (accessibilityObj.hintTimer) {
```
FunkTron: Do we really need to check for hintTimer here since we're already checking for it in Accessibility.handleEvent()?

apps/system/js/accessibility.js
```diff
@@ -284,6 +284,22 @@
     },
 
     /**
+     * Start a timeout that waits to display hints
+     * @memberof Accessibility.prototype
+     */
+     setHintsTimeout: function ar_setHintsTimeout(hints) {
```
yzen: nit: aHints (we use 'a' prefix for argument in this file)

apps/system/js/accessibility.js
```diff
@@ -284,6 +284,22 @@
...
+      var accessibilityObj = this;
```
yzen: Nit here and below: initial indentation is only 1 space instead of 2

yzen: No need to retaining this here. just use the format of:
```setTimeout(function() {...}.bind(this), ...); ```
.bind lets you set 'this' for the function scope.

File: apps/system/js/accessibility.js [hunk 4]
```diff
@@ -284,6 +284,22 @@
...
+      this.hintTimer = setTimeout(function() {
+        accessibilityObj.isSpeakingHint = true;
+        accessibilityObj.speak(hints, function(){
+          accessibilityObj.isSpeakingHint = false;
+        }, {
```
yzen: you would need bind here as well.

apps/system/js/accessibility.js
```diff
@@ -284,6 +284,22 @@
...
+          enqueue: true
+        });
+      }, 3000);
```
yzen: Can you take 3000 as a variable similar to REPEAT_BUTTON_PRESS.

apps/system/js/accessibility.js
```diff
@@ -309,7 +325,17 @@
           return;
       }
 
-      this.speak(aDetails.data, null, {
+      var accessibilityObj = this;
```
yzen: Same here, use bind for this and no need to the function variable.

apps/system/js/accessibility.js
```diff
@@ -284,6 +284,22 @@
     },
 
     /**
+     * Start a timeout that waits to display hints
+     * @memberof Accessibility.prototype
+     */
+     setHintsTimeout: function ar_setHintsTimeout(hints) {
+      var accessibilityObj = this;
+      this.hintTimer = setTimeout(function() {
```
yzen: Name this function (```function onHintTimeout() {...}```) . Things like this are useful for debugging because you would get a name of the function in the callstack if you get an exception, for example, instead of 'undefined'

apps/system/js/accessibility.js
```diff
@@ -284,6 +284,22 @@
...
+        accessibilityObj.isSpeakingHint = true;
+        accessibilityObj.speak(hints, function(){
```
yzen: Name a function here as well

apps/system/js/accessibility.js
```diff
@@ -363,6 +391,20 @@
     },
 
     /**
+     * Check for Hint speech/timer and clear.
+     * @memberof Accessibility.prototype
+     */
+    checkHintStatus: function ar_checkHintStatus() {
+      if(this.hintTimer){
```
yzen: No need for if, clearTimeout will not throw if the argument is undefined.

apps/system/js/accessibility.js
```diff
@@ -348,9 +374,11 @@
         case 'mozChromeEvent':
           switch (aEvent.detail.type) {
             case 'accessibility-output':
+              this.checkHintStatus();
```
yzen: I would move it into handleAccessFuOutput and handleAccessFuControl

apps/system/js/accessibility.js
```diff
@@ -363,6 +391,20 @@
...
     /**
+     * Check for Hint speech/timer and clear.
+     * @memberof Accessibility.prototype
+     */
+    checkHintStatus: function ar_checkHintStatus() {
+      if(this.hintTimer){
+        clearTimeout(this.hintTimer);
+      }
+      if(this.isSpeakingHint){
+        this.speechSynthesizer.cancel();
```
yzen: Could you create a wrapper Accessibility.cancelSpeech similar to speak (it should be more helpful when stubbing/moching things in tests).

try-server-hook:
[FunkTron FunkTron](https://github.com/FunkTron) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=525218c84a6f)

apps/system/js/accessibility.js
```diff
@@ -284,11 +290,27 @@
...
+     */
+    setHintsTimeout: function ar_setHintsTimeout(aHints) {
+      this.hintTimer = setTimeout(function onHintTimeout() {
+        this.isSpeakingHint = true;
+        this.speak(aHints, function onSpeakHintEnd(){
+          this.isSpeakingHint = false;
+        }.bind(this), {
+          enqueue: true
+        });
+      }.bind(this), 3000);
```
FunkTron: Just saw this, will change '3000' to HINT_TIMEOUT as id'd above

apps/system/locales/accessibility.en-US.properties
```diff
@@ -336,3 +336,20 @@ accessibility-stateUncheckedAbbr = ( )
...
+accessibility-menuitem-hint = Double tap to activate
+#accessibility-'check menu item' = Double tap to activate
+#accessibility-'radio menu item' = Double tap to activate
+accessibility-option-hint = Double tap to activate
+accessibility-radiobutton-hint = Double tap to activate
+accessibility-slider-hint = Swipe up or down to progress
+accessibility-spinbutton-hint = Swipe up or down to increase/decrease
+accessibility-pagetab-hint = Double tap to activate
+accessibility-entry-hint = Press and hold keys to begin typing
+#accessibility-outlineitem-hint
```
FunkTron: Not sure what this is.  Could you explain so I can write an appropriate hint?

apps/system/locales/accessibility.en-US.properties
```diff
@@ -336,3 +336,20 @@ accessibility-stateUncheckedAbbr = ( )
...
+
+# Hint dialogues for actionable events
+accessibility-pushbutton-hint = Double tap to activate
+accessibility-checkbutton-hint = Double tap to activate
+accessibility-combobox-hint = Press and hold options to focus
+accessibility-key-hint = Double tap to activate
+accessibility-link-hint = Double tap to activate
+accessibility-menuitem-hint = Double tap to activate
+#accessibility-'check menu item' = Double tap to activate
+#accessibility-'radio menu item' = Double tap to activate
```
FunkTron: Not sure how these roles are actually defined in the HTML.  Am I just replacing the spaces with '-' (dashes) here?

yzen: Yeah, this one needs to be fixed in gecko, we need to replace ' ' with '' in place where we generate the hint. I'll open a bug. Assume here that spaces are taken out so 'radiomenuitem' etc

apps/system/locales/accessibility.en-US.properties
yzen: It's a used when we describe a tree semantically. Not really used in Gaia atm (not really common on mobile).

apps/system/locales/accessibility.en-US.properties
```diff
@@ -336,3 +336,20 @@ accessibility-stateUncheckedAbbr = ( )
 accessibility-statePressedAbbr = (x)
 # LOCALIZATION NOTE: "( )" is a braille abbreviation for unpressed
 accessibility-stateUnpressedAbbr = ( )
+
+# Hint dialogues for actionable events
+accessibility-pushbutton-hint = Double tap to activate
+accessibility-checkbutton-hint = Double tap to activate
+accessibility-combobox-hint = Press and hold options to focus
```
yzen: I think this one is also double tap to activate

apps/system/locales/accessibility.en-US.properties
```diff
@@ -336,3 +336,20 @@ accessibility-stateUncheckedAbbr = ( )
...
+accessibility-key-hint = Double tap to activate
```
yzen: Press and hold to activate

apps/system/locales/accessibility.en-US.properties
```diff
@@ -336,3 +336,20 @@ accessibility-stateUncheckedAbbr = ( )
...
+accessibility-link-hint = Double tap to activate
+accessibility-menuitem-hint = Double tap to activate
+#accessibility-'check menu item' = Double tap to activate
+#accessibility-'radio menu item' = Double tap to activate
+accessibility-option-hint = Double tap to activate
+accessibility-radiobutton-hint = Double tap to activate
+accessibility-slider-hint = Swipe up or down to progress
```
yzen: I think maybe Swipe up to increase or down to decrease

apps/system/locales/accessibility.en-US.properties
```diff
@@ -336,3 +336,20 @@ accessibility-stateUncheckedAbbr = ( )
...
+accessibility-spinbutton-hint = Swipe up or down to increase/decrease
```
yzen: Swipe up to increase or down to decrease as well

apps/system/locales/accessibility.en-US.properties
```diff
@@ -336,3 +336,20 @@ accessibility-stateUncheckedAbbr = ( )
...
+accessibility-pagetab-hint = Double tap to activate
+accessibility-entry-hint = Press and hold keys to begin typing
```
yzen: Double tap to activate

apps/system/locales/accessibility.en-US.properties
yzen: Double tap to actiave

apps/system/js/accessibility.js
```diff
@@ -45,6 +45,12 @@
     CONTRAST_CAP: 0.6,
```
yzen: Overall in this file, lets use hints consistently (avoid using hint) in all names. We can have more than 1 hint per element so hints is more appropriate.

apps/system/js/accessibility.js
```diff
@@ -309,12 +331,19 @@
...
       }
 
-      this.speak(aDetails.data, null, {
+      var hintCallback = function ar_hintCallback() {
+        if (options.hints) {
+          this.setHintsTimeout(options.hints);
+        }
+      }.bind(this);
+
+      this.speak(aDetails.data, hintCallback, {
```
yzen: just define hintsCallback inline here.

apps/system/js/accessibility.js
```diff
@@ -309,12 +331,19 @@
           return;
       }
 
-      this.speak(aDetails.data, null, {
+      var hintCallback = function ar_hintCallback() {
+        if (options.hints) {
+          this.setHintsTimeout(options.hints);
```
yzen: So I think we want to combine aDetails.data with options.hints , so instead of 
'Some link' timeout then 'Double tap to activate' we hear:
'Some link' timeout then 'Some link double tap to activate'

apps/system/js/accessibility.js
```diff
@@ -284,11 +290,27 @@
     },
 
     /**
+     * Start a timeout that waits to display hints
+     * @memberof Accessibility.prototype
+     */
+    setHintsTimeout: function ar_setHintsTimeout(aHints) {
+      this.hintTimer = setTimeout(function onHintTimeout() {
```
yzen: Also I just realized that it's probably best to use setInterval, clearInterval instead of setTimeout/clearTimeout. We want to keep suggesting an interaction in case the user did no hear it the first time.

apps/system/js/accessibility.js
```diff
@@ -45,6 +45,12 @@
     CONTRAST_CAP: 0.6,
 
     /**
+     * Timeout (in milliseconds) between when a vc-change event fires
+     * and when a usability hint (if it contains one) is spoken
+     */
+    HINT_TIMEOUT: 3000,
```
yzen: 3s is too short, lets do 5s

apps/system/js/accessibility.js
```diff
@@ -45,6 +45,12 @@
     CONTRAST_CAP: 0.6,
 
     /**
+     * Timeout (in milliseconds) between when a vc-change event fires
+     * and when a usability hint (if it contains one) is spoken
```
yzen: Nit: usability hint -> interaction hints

apps/system/js/accessibility.js
yzen: So if we use interval, call this one this.hintsInterval

apps/system/locales/accessibility.en-US.properties
FunkTron: Are you sure about this one?  On my phone this hint only comes up when the keyboard is showing and the entry is already active.

try-server-hook:
[FunkTron FunkTron](https://github.com/FunkTron) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=d922b52762ce)

apps/system/locales/accessibility.en-US.properties
```diff
@@ -336,3 +336,20 @@ accessibility-stateUncheckedAbbr = ( )
...
+
+# Hint dialogues for actionable events
+accessibility-pushbutton-hint = Double tap to activate
+accessibility-checkbutton-hint = Double tap to activate
+accessibility-combobox-hint = Double tap to activate
+accessibility-key-hint = Prese and hold to activate
+accessibility-link-hint = Double tap to activate
+accessibility-menuitem-hint = Double tap to activate
+accessibility-checkmenuitem' = Double tap to activate
+accessibility-radiomenuitem' = Double tap to activate
```
FunkTron: Just saw this... Will fix the quote here and in 347

apps/system/js/accessibility.js
```diff
@@ -284,11 +290,33 @@
     },
 
     /**
+     * Start a timeout that waits to display hints
+     * @memberof Accessibility.prototype
+     */
+    setHintsTimeout: function ar_setHintsTimeout(aDetails, aHints) {
+        clearTimeout(this.hintsTimer);
```
FunkTron: Had to add this here to avoid overlapping Timeouts... I think the problem stems from the fact that when a speech action is canceled, its "end" events still fire anyway, so this.hintsTimer gets overwritten and you have two simultaneous loops.  Not sure if this is what you were looking for here.

Additionally, were you just wanting this loop to keep suggesting things forever?  Or did you want it to stop after a certain number of suggestions?

yzen: Yes, I think forever is fine, but we need to make sure that we are quiet when the screen is off :)

apps/system/js/accessibility.js
```diff
@@ -309,12 +337,17 @@
           return;
       }
 
-      this.speak(aDetails.data, null, {
+      this.speak(aDetails.data, function hintsCallback() {
+        if (options.hints) {
+          this.setHintsTimeout(aDetails, options.hints);
```
yzen: you can still pass 1 argument: aDetails.data.concat(options.hints)

apps/system/js/accessibility.js
```diff
@@ -284,11 +290,33 @@
...
 
     /**
+     * Start a timeout that waits to display hints
+     * @memberof Accessibility.prototype
+     */
+    setHintsTimeout: function ar_setHintsTimeout(aDetails, aHints) {
+        clearTimeout(this.hintsTimer);
+        this.hintsTimer = setTimeout(function onHintsTimeout() {
+          this.isSpeakingHints = true;
+          this.speak(aDetails.data, function onSpeakLabelEnd(){
```
yzen: With 1 argument you will only need 1 call to speak.

apps/system/js/accessibility.js
FunkTron: Had contemplated going this route (in fact it's what I started with), but yeah, it does seem to be a better solution than the chained speak calls.  Will do this instead.

try-server-hook:
[FunkTron FunkTron](https://github.com/FunkTron) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=d3fd9483ebea)

apps/system/test/unit/accessibility_test.js
```diff
@@ -25,6 +26,28 @@ suite('system/Accessibility', function() {
...
+    {string: 'link-hint'}
+  ];
+
+  // Can't get Chai or Sinon to match this function when compared
+  // to the argument list.  Will remove if unusable
+  /*var hintsCallback = function hintsCallback() {
+    if (options.hints) {
+      this.setHintsTimeout(aDetails.data.concat(options.hints));
+    }
+  }.bind(accessibility);*/
```
FunkTron: As you can see from this comment, I'm having trouble getting Chai or Sinon to equate this defined function (or even an inline replica) to match the inline function given as an argument in accessibility.js (see more below)

apps/system/test/unit/accessibility_test.js
```diff
@@ -247,8 +270,9 @@ suite('system/Accessibility', function() {
       accessibility.handleEvent(getAccessFuOutput(liveRegionDetails));
       assert.isFalse(stub_playSound.called);
       assert.isTrue(stubSpeak.called);
-      assert.isTrue(stubSpeak.calledWith(liveRegionDetails.data, null,
-        liveRegionDetails.options));
+      // This is where the calledWith comparison fails
+      /*assert.isTrue(stubSpeak.calledWith(liveRegionDetails.data, 
+          hintsCallback, liveRegionDetails.options));*/
```
FunkTron: I had to remove this assert statement in order to get all of the tests to pass because Sinon doesn't seem to like comparing the hintsCallback function against the inline argument given in the original file.  No matter what I do (even if I declare the argument as an inline anonymous blank function in both places), this test won't ever seem pass.  Should I just compare the other arguments using something like...?

assert.equal(liveRegionsDetails.data, stubSpeak.args[0][0]);
assert.equal(liveRegionsDetails.data, stubSpeak.args[0][2]);

Let me know what you think.

apps/system/test/unit/accessibility_test.js
```diff
@@ -289,4 +313,67 @@ suite('system/Accessibility', function() {
...
+    var realL10n = navigator.mozL10n;
+
+    setup(function() {
+      screenNode = document.createElement('div');
+      screenNode.id = 'screen';
+      document.body.appendChild(screenNode);
+
+      clock = this.sinon.useFakeTimers();
+
+      navigator.mozL10n = MockL10n; // jshint ignore:line
```
FunkTron: For whatever reason, jshint wouldn't let me commit the file because it was saying MockL10n was undefined.  I have no idea why it wasn't picking up the 'require' statement earlier in the doc, but if there's something you notice as to why that would be happening, I'd love to know the answer!

apps/system/test/unit/accessibility_test.js
```diff
@@ -289,4 +313,67 @@ suite('system/Accessibility', function() {
...
+      clock.restore();
+      navigator.mozL10n = realL10n;
+    });
+
+    test('handle vc change event with hints enabled', function() {
+      var stubSetHintsTimeout = this.sinon.stub(accessibility,
+        'setHintsTimeout');
+      // This is where MozL10n was necessary in place of the stubbed
+      // accessibility.speak method
+      accessibility.handleAccessFuOutput(vcChangeHintsEnabledDetails);
```
FunkTron: Just FYI -- The reason I included MockL10n in the first place was to be able to test whether setHintsTimeout would run after a vc-event fired with 'options.hints' populated.  I felt this was better than just seeing if accessibility.speak would be called with a known list of arguments.  Again, let me know if you'd prefer I did something else here.

apps/system/js/accessibility.js
```diff
@@ -284,11 +290,31 @@
     },
 
     /**
+     * Start a timeout that waits to display hints
+     * @memberof Accessibility.prototype
+     */
+    setHintsTimeout: function ar_setHintsTimeout(aDataPlusHints) {
+        clearTimeout(this.hintsTimer);
+        this.hintsTimer = setTimeout(function onHintsTimeout() {
+          if(!this.speechSynthesizer.screen.classList.contains('screenoff')){
```
yzen: I think a more stable way of doing this is listening to the 'screenchange' event that the screen_manager fires: https://github.com/mozilla-b2g/gaia/blob/master/apps/system/js/screen_manager.js#L556-L559 and checking for it screenEnabled flag in event detail. So you would need to add a new handler for that event in handleEvent. The handler will simply check if screenEnabled is false and if so you would call checkHintsStatus from there.

apps/system/js/accessibility.js
```diff
@@ -284,11 +290,31 @@
...
+            this.isSpeakingHints = true;
+            this.speak(aDataPlusHints, function onSpeakHintsEnd(){
```
yzen: Nit: space after function onSpeakHintsEnd() and before the last {

apps/system/js/accessibility.js
```diff
@@ -284,11 +290,31 @@
...
+              this.isSpeakingHints = false;
+              this.setHintsTimeout(aDataPlusHints);
+            }.bind(this), {
+            enqueue: true
```
yzen: Nit: indentation.

apps/system/js/accessibility.js
```diff
@@ -363,6 +394,18 @@
     },
 
     /**
+     * Check for Hints speech/timer and clear.
+     * @memberof Accessibility.prototype
+     */
+    checkHintsStatus: function ar_checkHintsStatus() {
```
yzen: I was just thinking that the name is slightly misleading. I think it should be clearHints instead of checkHintsStatus, since we mainly doing that.

apps/system/js/accessibility.js
```diff
@@ -388,6 +431,14 @@
     speak: function ar_speak(aData, aCallback, aOptions = {}) {
       this.speechSynthesizer.speak(aData, aOptions, this.rate, this.volume,
         aCallback);
+    },
+
+    /**
+     * Cancel any utterances currently being spoken by speechSynthesis.
+     * @memberof Accessibility.prototype
+     */
+    cancelSpeech: function ar_cancelSpeech() {
```
yzen: Could you replace all other instances where this.speechSynthesizer.cancel(); is used with the new method.

apps/system/js/accessibility.js
yzen: Call the handler handleScreenChange or something similar.

apps/system/test/unit/accessibility_test.js
yzen: see below

apps/system/test/unit/accessibility_test.js
yzen: Correct, this should be good
assert.deepEqual(liveRegionDetails.data, stubSpeak.args[0][0]);
assert.equal(typeof stubSpeak.args[0][1], 'function');
assert.deepEqual(liveRegionDetails.options, stubSpeak.args[0][2]);

apps/system/js/accessibility.js
```diff
@@ -284,11 +290,31 @@
     },
 
     /**
+     * Start a timeout that waits to display hints
+     * @memberof Accessibility.prototype
+     */
+    setHintsTimeout: function ar_setHintsTimeout(aDataPlusHints) {
```
yzen: Nit: use 2 spaces instead of 4

apps/system/test/unit/accessibility_test.js
yzen: You just need to add MockL10n to global comment on line 2

yzen: in fact i think you can mock l10n in suiteSetup for the whole suite.

apps/system/js/accessibility.js
yzen: This will also simplify your tests.

apps/system/test/unit/accessibility_test.js
```diff
@@ -289,4 +313,67 @@ suite('system/Accessibility', function() {
...
+      clock.tick(2000);
```
yzen: No need to tick here, we moch SpeechSynthesis to be synchronous. see https://github.com/mozilla-b2g/gaia/blob/master/apps/system/test/unit/mock_speech_synthesis.js

apps/system/test/unit/accessibility_test.js
```diff
@@ -289,4 +313,67 @@ suite('system/Accessibility', function() {
...
+    });
+
+    test('make sure hints are spoken', function() {
+      var stubSpeak = this.sinon.stub(accessibility, 'speak');
+      accessibility.setHintsTimeout(aDataPlusHints);
+      clock.tick(accessibility.HINTS_TIMEOUT+50);
+      assert.isTrue(stubSpeak.called);
+    });
+
+    test('make sure hints are not spoken with screen off', function() {
```
yzen: When we start handling screenchange event, we should be able to just test its handler.

yzen: e.g. that the timeout is cleared.

apps/system/test/unit/accessibility_test.js
```diff
@@ -289,4 +313,67 @@ suite('system/Accessibility', function() {
         accessibility.quicknav.element.classList.contains('visible'));
     });
   });
+
+  suite('interaction hints', function() {
+    var screenNode, clock;
+    var realL10n = navigator.mozL10n;
+
+    setup(function() {
+      screenNode = document.createElement('div');
```
yzen: You probably won't need this if we start handling it via screenchange event.

apps/system/test/unit/accessibility_test.js
```diff
@@ -289,4 +313,67 @@ suite('system/Accessibility', function() {
...
+      assert.isTrue(stubSetHintsTimeout.called);
+      assert.isTrue(stubSetHintsTimeout.calledWith(
+        vcChangeHintsEnabledDetails.data.concat(
+          vcChangeHintsEnabledDetails.options.hints)));
+    });
+
+    test('make sure hints are spoken', function() {
+      var stubSpeak = this.sinon.stub(accessibility, 'speak');
+      accessibility.setHintsTimeout(aDataPlusHints);
+      clock.tick(accessibility.HINTS_TIMEOUT+50);
```
yzen: nit spaces around +

yzen: I know it's a little more work, but could we wrap this test in it's own suite with clock setup and teardown to not affect the ones that will not need it?

try-server-hook:
[FunkTron FunkTron](https://github.com/FunkTron) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=98e9874a1f76)

apps/system/test/unit/accessibility_test.js
```diff
@@ -25,6 +26,26 @@ suite('system/Accessibility', function() {
...
+  };
+
+  var screenChangeScreenDisabledEvent = {
+    type: 'screenchange',
+    detail: {
+      screenEnabled: false
+    }
+  };
+
+  var aDataPlusHints = [
```
yzen: I think you don't need to use it, just pass vcChangeHintsEnabledDetails.options.hints on line 368.

apps/system/test/unit/accessibility_test.js
```diff
@@ -289,4 +321,55 @@ suite('system/Accessibility', function() {
...
+    test('make sure hints are canceled on screen off', function() {
+      var stubCancelHints = this.sinon.stub(accessibility, 'cancelHints');
+      accessibility.handleEvent(screenChangeScreenDisabledEvent);
+      assert.isTrue(stubCancelHints.called);
+    });
+
+    test('make sure hints are canceled with output/control event', function(){
+      // Output event
+      accessibility.isSpeakingHints = true;
+      speechSynthesizer.speak(fakeSentence, {});
```
yzen: Speak is synchronous in tests. I think this is not necessary. (same below)

apps/system/test/unit/accessibility_test.js
```diff
@@ -247,8 +275,12 @@ suite('system/Accessibility', function() {
       accessibility.handleEvent(getAccessFuOutput(liveRegionDetails));
       assert.isFalse(stub_playSound.called);
       assert.isTrue(stubSpeak.called);
-      assert.isTrue(stubSpeak.calledWith(liveRegionDetails.data, null,
-        liveRegionDetails.options));
+      // This is where the calledWith comparison fails
```
yzen: You should be able to remove the commented out part.

try-server-hook:
[FunkTron FunkTron](https://github.com/FunkTron) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=b6226e780c79)

try-server-hook:
[FunkTron FunkTron](https://github.com/FunkTron) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=7cd7345c9c77)

autolander:
http://docs.taskcluster.net/tools/task-graph-inspector/#2tUfC9D2RaCZ4BmfN9m7_A

PR #28358, opened 2015-02-23
Bug 1135614 - Use audio channel interruptions instead of a Settings flag...

mancas:
..., to turn off the radio when the user have an incoming call.

### DIFF
--- a/apps/fm/js/fm.js
+++ b/apps/fm/js/fm.js
@@ -1,5 +1,7 @@
 'use strict';
 
+/* globals AudioCompetingHelper */
+
 function $(id) {
   return document.getElementById(id);
 }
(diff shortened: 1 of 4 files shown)


apps/fm/js/fm.js
```diff
@@ -870,47 +872,29 @@ function init() {
...
-    function(event) {
-      // An attention screen is in the process of opening. Save the
-      // current state of the radio and disable.
-      if (event.settingValue) {
-        window._previousFMRadioState = mozFMRadio.enabled;
-        window._previousEnablingState = enabling;
-        window._previousSpeakerForcedState = speakerManager.speakerforced;
-        mozFMRadio.disable();
-      }
+  // Compete for the content audio channel
```
timdream: Please expand this comment and explain when the audio channel will be interrupted.

PR #28365, opened 2015-02-23
Bug 1135721 - Modify test_that_device_info_is_populated to support running on devices with no SIM cards

davehunt:


try-server-hook:
[davehunt Dave Hunt (davehunt)](https://github.com/davehunt) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=45a4bd50e0ba)

tests/python/gaia-ui-tests/gaiatest/tests/functional/settings/test_settings_device_info.py
```diff
@@ -13,10 +13,16 @@ def test_that_device_info_is_populated(self):
         settings.launch()
         device_info = settings.open_device_info_settings()
 
+        # devices without sim cards have no phone number
+        if 'local_phone_numbers' in self.testvars:
+            self.assertTrue(len(device_info.phone_number) > 0)
+            self.assertIsNotNone(device_info.phone_number)
+        else:
+            self.assertEqual(device_info.phone_number, '')
```
geoelectric: A concern here is the template for our testvars has a "local_phone_numbers": [""] as the default, so would foul this if someone assumed that was OK to leave. 

  if ('local_phone_numbers' in self.testvars and 
      bool(self.testvars.local_phone_numbers) and 
      self.testvars.local_phone_numbers[0] != ""):

...is probably about as robust as I can think of. Also now verbose enough that it should go in a module somewhere. :)

On another note, realized my own local_phone_numbers check will error if a number isn't defined, so I have to update the ftu PR similarly.

Also, is there a situation where len(x) > 0 and x is None? If not (as len(None) throws an exception) you can remove line 19.

geoelectric:
Aside from my concern regarding the testvars template defaults conflicting with this assumption, this is fine. r=me after you make a decision on that.

tests/python/gaia-ui-tests/gaiatest/tests/functional/settings/test_settings_device_info.py
davehunt: Right, I was forgetting about the template. I'll change this to use `if self.testvars.get('local_phone_numbers')` as this will return None if the key is not present, or an empty string/list, all of which are falsey. I'll edit the template to have an empty list for now - once the remote lab changes are in place the value will be either a string or list, and shouldn't be a list with only one item. We can then handle that in the tests.

davehunt: Sorry, I didn't notice your update suggesting essentially the same thing. :)

davehunt:
Merged in https://github.com/mozilla-b2g/gaia/commit/2722edf0f20c4abfe101923035143cfd00ade5de

PR #28401, opened 2015-02-24
Bug 1132397 - Only SIM 2 PIN is shown when switching airplane mode r=alive

albertopq:


try-server-hook:
[albertopq albertopq (albertopq)](https://github.com/albertopq) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=cdd691497d03)

try-server-hook:
[albertopq albertopq (albertopq)](https://github.com/albertopq) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=8166e03018be)

try-server-hook:
[albertopq albertopq (albertopq)](https://github.com/albertopq) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=5bb4e6444356)

apps/system/js/sim_lock_manager.js
```diff
@@ -159,6 +159,7 @@
 
     '_handle_airplanemode-enabled': function(evt) {
       this.simLockSystemDialog.close();
+      this._alreadyShown = false;
```
alivedise: Don't need to fix in this bug: I hope we could use this.simLockDialog.isVisible() instead of keeping and maintaining this internal value in the future. It might be async with the real state and cause confusion - unless there is an obvious reason to do so.

albertopq: Actually, this flag is saving a different status. This one is used for showing the first slot PIN dialog always before the second one. It was a quick fix to make it land to 2.1. I hope we can get rid of it soon as well..

PR #28412, opened 2015-02-24
Bug 1135860 - Change LogShake activity to share r=janx

lissyx:
To be able to handle sharing with a bugzilla client, we need to change
the activity triggered by LogShake, since name "new" and type "email" is
really intended to limit to the Email client. We make use of the "share"
activity.


try-server-hook:
[lissyx  (lissyx)](https://github.com/lissyx) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=26e6e6239106)

PR #28413, opened 2015-02-24
Bug 1122079 - Test the detection and skip of mp4 videos.

hfiguiere:


try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=0f2f57d54196)

apps/music/test/marionette/loading_test.js
```diff
@@ -0,0 +1,76 @@
...
+      music.launch();
+      music.waitForFirstTile();
+
+      music.switchToAlbumsView();
+
+      music.selectAlbum('We crash computers');
+
+      music.waitForSongs(function(songs) {
+        return songs.length >= 1;
+      });
```
davidflanagan: Technically, this isn't proving that the movie isn't loaded as a song. I think it is only proving that it isn't the first song.  You need some way to wait for the scan to complete and then verify that the movie was not loaded.

apps/music/test/marionette/loading_test.js
```diff
@@ -0,0 +1,76 @@
...
+      client.fileManager.removeAllFiles();
+      client.fileManager.add([
+        {
+          type: 'music',
+          filePath: 'apps/music/test-data/playlists/a.ogg'
+        },
+        {
+          type: 'music',
+          filePath: 'test_media/Movies/gizmo2.mp4'
+        }
```
davidflanagan: With this selection of files, the test would succeed even if the app didn't support any .mp4 formats.  Please include a .mp4 audio file and verify that it does get listed. I think you can just copy the existing .m4a file and rename it to .mp4.

Even better would be if you could find a test .mp4 audio file that was not from apple, so it had "ftypmp42" just like the movie file does. That would test even deeper in the metadata parsing code. But I'm not even sure that such things exist.

apps/music/test/marionette/loading_test.js
```diff
@@ -0,0 +1,76 @@
+/* global require, marionette, setup, suite, test, __dirname */
+'use strict';
+
+var assert = require('assert');
+var Music = require('./lib/music.js');
+var FakeRingtones = require('./lib/fakeringtones.js');
+var FakeControls = require('./lib/fakecontrols.js');
+
+marionette('Music loading', function() {
```
davidflanagan: This is a very generic name for this file and for this test. Are you planning to add other suites to this file that are not about distinguishing movies from songs?  If not, I'd change the names to be more specific.

apps/music/test/marionette/loading_test.js
```diff
@@ -0,0 +1,76 @@
...
+          filePath: 'test_media/Movies/gizmo2.mp4'
+        }
+      ]);
+    });
+
+    test('Check the mp4 isn\'t loaded', function() {
+
+      music.launch();
+      music.waitForFirstTile();
+
```
davidflanagan: If you just wait for the first tile, are you actually guaranteed that the scanning process has finished? Could the app still be scanning the movie file while the rest of the test runs?

apps/music/test/marionette/loading_test.js
davidflanagan: In general, this is a hard test to do exhaustively because mp4 is such a generic container and there are so many codecs that get included within it. Ideally, we'd have a comprehensive library of mp4 audio and video samples and could write a test to verify that the app reads all of the audio files and none of the video files.  If you know of any such mp4 sample library, feel free to import a copy into test-data/mp4/ or something and expand this test.  If not, then just make sure that the test is written to be extensible so that we can easily add other .mp4 files to it in the future.

Speaking of a test-data/mp4/ directory. Consider creating one, and copying the gizmo file and test-data/*.m4a file into it so that there is one central location for the test media used here.

apps/music/test/marionette/loading_test.js
```diff
@@ -0,0 +1,76 @@
...
+      music.switchToAlbumsView();
+
+      music.selectAlbum('We crash computers');
+
```
davidflanagan: If you select just one album, then of course the movie file won't show up there, since it is not in the same album as the song. I think you need to switch to show the list of all songs and then verify that there is only one song in that list, right?

apps/music/test/marionette/loading_test.js
hfiguiere: test-data is in general the various test files. It is not specific to any test.

apps/music/test/marionette/loading_test.js
hfiguiere: It is possible I add more test in the area of loading files.

apps/music/test/marionette/loading_test.js
hfiguiere: Duh. I feel dumb now that I left that here. It is actually more complex to do it properly and maybe I was meaning to do it and forgot. Good catch. Will definitely change the test.

apps/music/test/marionette/loading_test.js
hfiguiere: I'll be honest I'm not sure we have a way to know if we are finished loading. I need to figure that out.

try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=26baa35ec4d8)

try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=0b2ece5cf15b)

try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=54d8a276e5cd)

try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=e51e57858b98)

apps/music/test/marionette/loading_test.js
```diff
@@ -0,0 +1,78 @@
...
+          filePath: 'test_media/Movies/gizmo2.mp4'
+        }
+      ]);
+    });
+
+    test('Check the mp4 video isn\'t loaded', function() {
+
+      music.launch();
+      music.waitForFirstTile();
+      music.waitFinishedScanning();
```
davidflanagan: I'm not convinced by the implementation of waitFinishedScanning(). It will return if the musicdb is not scanning. So if scanning has not started yet, it will also return immediately.  Ideally, music.launch() would inject an event handler that listens for the scanend event, and that event handler would set some kind of _initialScanComplete flag on the musicdb.  Then, waitFinishedScanning() can wait for that flag to be set.

I'd even be okay adding this line to shared/js/mediadb.js in the endscan() method at 1696:

  media.initialScanComplete = true;

hfiguiere: If you are OK with doing this in mediadb.js then I will. It seems to be the simple and reliable solution to this. Will update the PR.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f87eb9f68c9da0f5301d04763e24f6845350d764

hfiguiere:
Updated PR now.

apps/music/test/marionette/loading_test.js
```diff
@@ -0,0 +1,78 @@
...
+
+      music.switchToSongsView();
+
+      music.waitForSongsInList(function(songs) {
+        return songs.length >= 2;
+      });
```
davidflanagan: I don't like this because you're waiting for there two be two songs, then testing that there are only two songs.  If a third song is added to the list later, then the test could pass when it should have failed.

Given the music app's current architecture, I don't see an easy way to know when the songs view is fully populated. Maybe you could modify switchToSongsView() so that it patches App.showCorrectOverlay() and use that to detect when the songs view is full loaded.

If you don't do that, I'd like you to at least add a comment here about the issue and wait for an extra second to make sure that no more songs are added.

Actually, here's a better way: when list_view.js enumerates the db, it sets this.handle to the return value of the enumerate() call.  this.handle is an object with a state property.  If you wait for `this.handle.state === 'complete'` then you'll know that the enumeration is complete.

apps/music/test/marionette/loading_test.js
```diff
@@ -0,0 +1,78 @@
...
+
+      var songs = music.songsInList;
+      assert.equal(songs.length, 2, 'Wrong number of songs');
+
+      var num = client.executeScript(function() {
+        var w = window.wrappedJSObject;
+        return w.ListView.dataSource.length;
+      });
+      assert.equal(num, 2, 'Data source have wrong number of songs');
```
davidflanagan: Just counting the songs here doesn't seem like enough. We could have accepted the movie and rejected the .ogg file, for example.  Why not have this executeScript() call return `w.ListView.dataSource.map((record) => record.name).sort()` so that you get a list of the song files, in alphabetical order. Then you can assert that the expected files were accepted and the movie was rejected.

apps/music/test/marionette/loading_test.js
hfiguiere: will update PR with that.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e791264b33c323923d65e4f15c13178add14677c

apps/music/test/marionette/loading_test.js
hfiguiere: Good idea. (but I can't use the arrow function there as it is parsed by NodeJS... - no biggie)

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a031a3239410d101d32d171604ffc03556b9f0eb

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d02771733876f32455e085819c5f99e2be698634

PR #28423, opened 2015-02-24
Bug 1135251 - [Gallery] - Remove mozL10n.ready call from ThumbnailList class r=djf

punamdahiya:


try-server-hook:
[punamdahiya punamdahiya](https://github.com/punamdahiya) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=d18e79bd0c02)

apps/gallery/js/thumbnail_list.js
```diff
@@ -35,9 +35,9 @@ function ThumbnailList(groupClass, container) {
   this.groupClass = groupClass;
   this.container = container;
 
-  // If the locale changes while we're running, then we need to
-  // re-localize the dates in the group headers
-  navigator.mozL10n.ready(this.localize.bind(this));
+  // Localize dates in group headers and thumbnail item
+  // description when l10n resources are loaded
+  navigator.mozL10n.once(this.localize.bind(this));
```
davidflanagan: If you do it like this, then we may get localized twice on startup once here, and once for the ready call in startup.js.

thumbnail_item and thumbnail_date_group already have code to localize themselves if readyState === complete when they are created. So we only need to handle the case where the thumbnail list is being created before the locale data is loaded.  Assuming you make the ready() call right after calling new ThumbnailList, I think that is all we need.

punamdahiya: You are right, using  once l10n handler inside thumbnailList constructor and waiting for all thumbnails to be created before setting ready call in startup.js is causing localized call twice on app start. 

I will move the call to thumbnails.js right after we are creating instance of thumbnailList.

try-server-hook:
[punamdahiya punamdahiya](https://github.com/punamdahiya) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=adfa6bf9d1e9)

PR #28429, opened 2015-02-24
Bug 1069623 - Make music seek bar accessible

eeejay:


apps/music/index.html
```diff
@@ -116,7 +117,7 @@ <h1 id="title-text" data-l10n-id="music"></h1>
             <p data-l10n-id="music-paused-notification"></p>
           </section>
         </div>
-        <div id="player-seek">
+        <div id="player-seek" role="slider" aria-valuemin="0">
```
dominickuo: @eeejay If the `aria-valuemin` here is to indicate the minimum value, should it be on the progress element? because we actually set the value on the progress element.

eeejay: Unfortunately we have to override the progress element here and use a role=slider instead. So the min/max/value attributes are copied to this element.

apps/music/js/ui/views/player_view.js
```diff
@@ -1039,6 +1053,21 @@ var PlayerView = {
           this.next();
         }
         break;
+      case 'keypress':
+        // The standard accessible control for sliders is arrow up/down keys.
+        // Our screenreader synthesizes those events on swipe up/down gestures.
+        // Currently, we only allow screen reader users to adjust sliders with a
+        // constant step size (there is no page up/down equivalent). In the case
+        // of videos, we make sure that the maximum amount of steps for the
```
davidflanagan: This should presumably say music here instead of videos.

apps/music/js/ui/views/player_view.js
```diff
@@ -1039,6 +1053,21 @@ var PlayerView = {
...
+        // entire duration is 20, or 2 second increments if the duration is less
+        // then 40 seconds.
+        var step = Math.max(this.audio.duration / 20, 2);
+        if (evt.keyCode == evt.DOM_VK_DOWN) {
+          this.seekAudio(this.audio.currentTime - step);
+        } else if (evt.keyCode == evt.DOM_VK_UP) {
+          this.seekAudio(this.audio.currentTime + step);
+        }
+        break;
```
dominickuo: @eeejay This looks exactly the same as those logic in video so should be fine, just curious why they are mapping to up/down instead of left/right keys, cause usually the fastforward/rewind icons goes left and right.

dominickuo:
@eeejay While reviewing this patch, it reminds me that modifying index.html also imply we should modify open.html as well, because open.html(open activity use it) is basically the player part of index.html, so after this one landed, we should file a new bug for open.html and make it accessible.

apps/music/js/ui/views/player_view.js
```diff
@@ -775,10 +777,22 @@ var PlayerView = {
   },
 
   setSeekBar: function pv_setSeekBar(startTime, endTime, currentTime) {
+    if (this.seekBar.max != endTime) {
+      // Duration changed, update accessibility label.
+      navigator.mozL10n.setAttributes(this.seekSlider,
+        'playbackSeekBar', {'duration': formatTime(endTime)});
+    }
+
```
dominickuo: @eeejay Music app has listened to `durationchange` event of the audio element, so if the duration attribute should be updated when the duration changes, it might make more sense to place it in the `durationchange` handler.

apps/music/js/ui/views/player_view.js
eeejay: Slider widgets use the up/down key to step down/up. So that is what we need to use here, since the screen reader synthesizes those keys.

apps/music/js/ui/views/player_view.js
eeejay: In the case above, we creating a human readable label for the seekbar. It changes when seekBar.max changes, which is in this function.

PR #28431, opened 2015-02-24
Bug 1136393 - Label player buttons

eeejay:


apps/music/index.html
```diff
@@ -110,7 +111,7 @@ <h1 id="title-text" data-l10n-id="music"></h1>
               <button class="rating-star" data-rating="4" role="radio" data-l10n-id="rating-star" data-l10n-args='{"n":"4"}'></button>
               <button class="rating-star" data-rating="5" role="radio" data-l10n-id="rating-star" data-l10n-args='{"n":"5"}'></button>
             </div>
-            <button id="player-album-shuffle" class="player-cover-button" data-icon="shuffle"></button>
+            <button id="player-album-shuffle" class="player-cover-button" data-icon="shuffle" aria-pressed="true"></button>
```
jimporter: I'm not sure what `aria-pressed` does here, but I'll assume you know what you're doing...

eeejay: In a previous patch, that attribute was erroneously on `#views-sublist-controls-shuffle`. I corrected it here. It has no real significance since it is set in JS anyway. This just keeps it sane and in one place.

PR #27225, opened 2015-01-08
Bug 1091441 - [Messages] the thread view is flashing while loading if there are MMS, r=azasypkin

steveck-chung:

PR #28458, opened 2015-02-25
Bug 1043903 - [Messages][Tests] Add integration tests for the thread panel with a large number of messages. r=julien

azasypkin:


try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=56a2973ae66b)

apps/sms/test/marionette/lib/messages.js
```diff
@@ -350,6 +361,22 @@
...
+              return [thread.id, thread];
+            }));
+
+            window.wrappedJSObject.TestStorages.setStorage('messagesDB', {
+              threads: threadMap,
+              recipientToThreadId: new Map(),
+              messageIdUniqueCounter: 100
+            });
+
+            window.wrappedJSObject.TestStorages.setStorageReady('messagesDB');
```
azasypkin: Can't really think of anything better than this, but the idea is the following: 

1. There is some fake storage that mocks use internally;
2. This storage should be filled from test code to simulate various scenarios;
3. Mocks are injected when app is started (via ```contentScript.inject``` plugin);
4. Here is the main problem that influenced solution - we have to fill storage with fake data before application requests it, to solve this I have the following flow - ```app.getData() -> mock.getData() -> TestStorages.getStorage('key') -> storage.getData()```. Specific test/suite decides when ```TestStorages.getStorage('key')``` resolves and application gets the data, so that test creates fake data, populates storage and then marks that storage as ```ready``` and app request passes through.

Alternative would be dedicated files with fake data for various scenarios that is injected before test/suite, but test won't have chance to work with that data, it can only choose what workload to load before app is run.

Do you have any better idea in mind?

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
```diff
@@ -3,6 +3,7 @@
 
 const Cu = Components.utils;
 Cu.import('resource://gre/modules/Services.jsm');
+Cu.import('resource://gre/modules/Promise.jsm');
```
azasypkin: Seems that plain ```Promise``` isn't avaialable here.

apps/sms/test/marionette/mocks/mock_test_storages.js
```diff
@@ -0,0 +1,57 @@
...
+
+    if (!deferred) {
+      deferred = {
+        data: null
+      };
+
+      deferred.promise = new Promise(function(resolve, reject) {
+        deferred.resolve = resolve;
+        deferred.reject = reject;
+      });
```
julienw: actually, Promise.jsm already support `Promise.defer()`, see https://dxr.mozilla.org/mozilla-central/source/toolkit/modules/Promise.jsm#58

azasypkin: Hmm, I didn't use it because mdn says it's obsoleted: https://developer.mozilla.org/en-US/docs/Mozilla/JavaScript_code_modules/Promise.jsm

azasypkin: But at the same time, I don't see any note about it in actual code

azasypkin: Ah, we probably should use ```PromiseUtils.jsm``` instead :)

julienw: ah yeah !

try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=4937738f48d2)

try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=e64abf9a069a)

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
```diff
@@ -161,11 +183,29 @@ Services.obs.addObserver(function(document) {
...
+
+            if (!iteratorResult.done) {
+              cursor.result = iteratorResult.value;
+            } else {
+              cursor.done = true;
+              cursor.continue = null;
+              cursor.result = null;
+            }
+
+            window.setTimeout(function nextTick() {
```
azasypkin: So as promised :) Here I tried to use ```Promise.resolve().then(cursor.onsuccess.call(cursor))``` instead, but on ~50 iterations it fails with "too much recursion" exception.

So we have the following flow here:
```
--- continue() (first time)
--- --- Promise.resolve().then(() => onsuccess())
--- --- --- continue()
--- --- --- --- Promise.resolve().then(() => onsuccess())
--- --- --- --- --- continue()
...........
```

Looks like promise somehow keeps stack of everything inside callback even if ```Promise.resolve``` not returned from ```then``` callback.

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
```diff
@@ -192,12 +232,15 @@ Services.obs.addObserver(function(document) {
             } else {
               cursor.result = thread.messages[currentIndex];
             }
-            cursor.onsuccess.call(cursor);
+
+            window.setTimeout(function nextTick() {
```
azasypkin: So as promised :) Here I tried to use ```Promise.resolve().then(cursor.onsuccess.call(cursor))``` instead, but on ~50 iterations it fails with "too much recursion" exception.

So we have the following flow here:
```
--- continue() (first time)
--- --- Promise.resolve().then(() => onsuccess())
--- --- --- continue()
--- --- --- --- Promise.resolve().then(() => onsuccess())
--- --- --- --- --- continue()
...........
```

Looks like promise somehow keeps stack of everything inside callback even if ```Promise.resolve``` not returned from ```then``` callback.

try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=ab5432226bbb)

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e286e505b2be2afbdddfc4d3ef1ab587a45a71cd

apps/sms/test/marionette/thread_test.js
```diff
@@ -25,9 +32,74 @@ marionette('Thread Panel Tests', function() {
...
+
+        // Composer panel should always be visible
+        assertIsDisplayed(messagesApp.Composer.sendButton);
+        assertIsDisplayed(messagesApp.Composer.messageInput);
+
+        thread.messages.forEach(function(message, index) {
+          var messageNode = messagesApp.Thread.findMessage(message.id);
+
+          // If current page is loaded, then message node should be visible
+          if (index < pageSize * loadedPage) {
```
julienw: I'm afraid this is fragile; especially except the first page we don't load a fixed number of element... we just display whatever is already in the page.

So most of the time your logic will work, but we might have an intermittent issue.

I wonder if we could have an event when all the messages are added to the DOM ? `qa-messages-rendered` ?

In the (not so distant) future (because gerard-majax bugs me with this)  I'd like to not add everything to the DOM, that's why I think an event is better than a class... but it's maybe more difficult to handle from the test?

apps/sms/test/marionette/mocks/mock_test_storages.js
```diff
@@ -0,0 +1,53 @@
...
+    setStorageReady: function(key) {
+      var storage = getStorage(key);
+
+      storage.deferred.resolve(Cu.waiveXrays(storage.data));
+    }
+  };
+
+  Object.defineProperty(window.wrappedJSObject, 'TestStorages', {
+    configurable: false,
+    writable: true,
```
julienw: do you want it writable ?

apps/sms/test/marionette/thread_test.js
azasypkin: >  especially except the first page we don't load a fixed number of element... we just display whatever is already in the page.

Mmm, but we've changed it in https://github.com/mozilla-b2g/gaia/pull/27225, no?

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
```diff
@@ -109,50 +132,49 @@ Services.obs.addObserver(function(document) {
...
 
       return request;
     },
 
     send: function(recipients, content, options) {
-      var requests = Cu.waiveXrays(Cu.cloneInto(recipients.map(() => ({
-        onsuccess: null,
-        onerror: null
-      })), window));
+      var requests = recipients.map(function() {
```
julienw: requests have a "then" method now, right ?

apps/sms/test/marionette/lib/messages.js
```diff
@@ -350,6 +361,22 @@
...
+        setStorage: function(threads) {
+          client.executeScript(function(threads) {
+            var threadMap = new Map(threads.map(function(thread) {
+              return [thread.id, thread];
+            }));
+
+            window.wrappedJSObject.TestStorages.setStorage('messagesDB', {
+              threads: threadMap,
+              recipientToThreadId: new Map(),
+              messageIdUniqueCounter: 100
```
julienw: I don't really get why it starts at 100 here?

apps/sms/test/marionette/thread_test.js
julienw: oh, ok, I missed we changed this.

apps/sms/test/marionette/lib/messages.js
azasypkin: It was a long ago, let me try to remember :) I agree 100 looks weird for me now

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
azasypkin: But we don't use it currently, I left it for the developer who will migrate to then in the code so that mock contains code that is actually used :) Do you think it's better to do it in this patch to support future migration to ```send().then```?

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
azasypkin: As I mentioned on IRC, it seems that it works correctly now even with 500~ iterations - do you prefer Promise syntax?

apps/sms/test/marionette/mocks/mock_test_storages.js
azasypkin: Mmm I think, probably copy-paste thing. Will fix

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
julienw: it would be nice :)

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
julienw: If you can do it it would be nice, so that it's as close as possible to the real API.

azasypkin: Sure

azasypkin: I found a better way! There are ```Services.DOMRequest.createRequest``` and ```Services.DOMRequest.createCursor``` available, so we don't need to invent the wheel :) And we don't need ```setTimeout/Promise.resolve().then``` workaround as well!

apps/sms/test/marionette/mocks/mock_test_storages.js
azasypkin: Now I see that I need it, as script injection happens in test ```setup``` method that redefines mock from the scratch for every test.

julienw: ok then add a comment about it :)

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
julienw: ah yeah, way better ! We live in chrome space :)

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b9b8a0dd564a893339734d8f9b4928aac4f3ec8e

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=dc01fcbed4f2e387a3518c0bb31fe72171efae32

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
julienw: are you sure? I know some jsm that use promises without importing this... (for example https://dxr.mozilla.org/mozilla-central/source/dom/settings/SettingsRequestManager.jsm)

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
```diff
@@ -27,215 +24,210 @@ Services.obs.addObserver(function(document) {
...
 
-  function getOrCreateThreadForRecipient(recipient) {
-    var threadId = recipientToThreadId.get(recipient);
+  var storagePromise = null;
+  function getStorage() {
+    if (!storagePromise) {
+      var appWindow = window.wrappedJSObject;
+
+      if (!appWindow.TestStorages) {
+        storagePromise = Promise.resolve();
```
julienw: should we resolve to something sensible ?

julienw: sorry, I see the fallback is there below. Forget my comment

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
julienw: another possibility is to use window.Promise which should be here.

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
```diff
@@ -27,215 +24,210 @@ Services.obs.addObserver(function(document) {
...
-                segments: segments,
-                charsAvailableInLastSegment: charsUsedInLastSegment ?
-                  segmentLength - charsUsedInLastSegment : 0
-              }
-            }
-          }, window));
-        }
-      }, delayMs);
+      var segmentLength = 160;
+      var charsUsedInLastSegment = (text.length % segmentLength);
```
julienw: maybe add a comment to explain this is not like the real implementation

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
```diff
@@ -27,215 +24,210 @@ Services.obs.addObserver(function(document) {
...
+        var request = Services.DOMRequest.createRequest(window);
+
+        var parameters = {
+          type: 'sms',
+          body: content,
+          receiver: receiver
+        };
+
+        createMessage(parameters, options).then(function(message) {
+          Services.DOMRequest.fireSuccessAsync(request, null);
```
julienw: please add a comment explaining we fire a success asap for the tests. In the future we might want to fail for some cases, right?

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
azasypkin: > are you sure?

Well, it definitely didn't work at the time of my comment :) But let me re-check.

> window.Promise 

IMO, it looks worse than ```Cu.import('resource://gre/modules/Promise.jsm');``` - but will check it as well :)

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
azasypkin: Yes, developer who will want to cover this case should be responsible for extending mock :)

This comment should probably be at the top of the file, does it work for you?

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
azasypkin: Ooookay :)

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
```diff
@@ -27,215 +24,210 @@ Services.obs.addObserver(function(document) {
...
-      }, delayMs);
+      };
+
+      var cursor = Services.DOMRequest.createCursor(window, handleCursor);
+
+      getStorage().then(function(storage) {
+        var thread = storage.threads.get(filter.threadId);
+
+        // Remove this once Array.prototype.values() is landed (see bug 875433).
+        messages = new Set(thread && thread.messages || []).values();
```
julienw: `(thread && thread.messages || [])[Symbol.iterator]()` ? Not sure this is really better :)

`(thread && thread.messages || []).entries()` is maybe better though. In that case you'll need `iteratorResult.value[1]`. Or: `var { done, value: [_, value] } = messages.next();` ;)

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
```diff
@@ -27,215 +24,210 @@ Services.obs.addObserver(function(document) {
...
+  }
+
+  function getOrCreateThreadForRecipient(storage, recipient) {
+    var threadId = storage.recipientToThreadId.get(recipient);
 
     if (!threadId) {
-      threadId = recipientToThreadId.size + 1;
-      threads.set(threadId, Cu.waiveXrays(Cu.cloneInto({
+      threadId = storage.recipientToThreadId.size + 1;
+      storage.threads.set(threadId, Cu.waiveXrays(Cu.cloneInto({
```
julienw: maybe a comment to explain why we need waiveXrays and cloneInto ?

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
```diff
@@ -27,215 +24,210 @@ Services.obs.addObserver(function(document) {
...
-      sentTimestamp: Date.now(),
-      read: true
-    }, window));
-
-    thread.messages.push(message);
-
-    return message;
+      );
+
+      var message = Cu.waiveXrays(Cu.cloneInto({
```
julienw: same here, a comment would be nice

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
```diff
@@ -27,215 +24,210 @@ Services.obs.addObserver(function(document) {
...
+      var threads = null;
+
+      var handleCursor = function() {
+        var iteratorResult = threads.next();
 
-      window.setTimeout(function() {
-        if (typeof cursor.onsuccess === 'function') {
-          cursor.onsuccess.call(cursor);
+        if (!iteratorResult.done) {
+          Services.DOMRequest.fireSuccessAsync(cursor, iteratorResult.value);
```
julienw: should we clone the value so that we can't change it from the client code ?

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
```diff
@@ -27,215 +24,210 @@ Services.obs.addObserver(function(document) {
...
-              cursor.result = thread.messages[currentIndex];
-            }
-            cursor.onsuccess.call(cursor);
-          };
+      var handleCursor = function() {
+        var iteratorResult = messages.next();
 
-          cursor.continue();
+        if (!iteratorResult.done) {
+          Services.DOMRequest.fireSuccessAsync(cursor, iteratorResult.value);
```
julienw: should we clone the value so that we can't change it from the client code ?

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
```diff
@@ -27,215 +24,210 @@ Services.obs.addObserver(function(document) {
...
+        } else {
+          Services.DOMRequest.fireDone(cursor);
         }
-      }, delayMs);
+      };
+
+      var cursor = Services.DOMRequest.createCursor(window, handleCursor);
+
+      getStorage().then(function(storage) {
+        var thread = storage.threads.get(filter.threadId);
```
julienw: maybe throw early if we don't have this filter? This would help finding issues in the test if we change the client code (for example if we ever paginate using the timestamps using the API -- I'm not sure we'll ever do that this way but..).

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
```diff
@@ -27,215 +24,210 @@ Services.obs.addObserver(function(document) {
...
+      getStorage().then(function(storage) {
+        var requestedMessage = null;
+        storage.threads.forEach(function(thread) {
+          thread.messages.forEach(function(message) {
+            if (message.id === id) {
+              requestedMessage = message;
+            }
           });
-          request.onsuccess.call(request);
+        });
```
julienw: do you think it would make sense to precalculate and store a "messages" map containing all messages when we set the storage ?

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
azasypkin: Well I tried all these options, and decided to use what is here currently to be as much similar to ```getThreads``` as possible (with the idea to have one cursor handling method for both eventually) :) So I'd prefer to keep it if you don't mind?

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
```diff
@@ -27,215 +24,210 @@ Services.obs.addObserver(function(document) {
...
+
+        if (requestedMessage) {
+          Services.DOMRequest.fireSuccessAsync(request, requestedMessage);
```
julienw: what happens if we find no message ?

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
julienw: not that I really care, to be honest :)

apps/sms/test/marionette/mocks/mock_test_storages.js
```diff
@@ -0,0 +1,55 @@
...
+    },
+
+    setStorage: function(key, value) {
+      getStorage(key).data = value;
+    },
+
+    setStorageReady: function(key) {
+      var storage = getStorage(key);
+
+      storage.deferred.resolve(Cu.waiveXrays(storage.data));
```
julienw: maybe a comment to explain waiveXrays ?

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
azasypkin: Honestly, it's too much optimization for test-only code :) So I really don't care, if you want we can handle it in follow-up

julienw: maybe the opposite makes more sense: the test would send a "messages" array and setStorage would organize this into threads. This would make a simpler test. But maybe it's more readable to concretely show the thread in the test. Tell me what you think :)

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
azasypkin: Looks like now it works without ```Cu.import('resource://gre/modules/Promise.jsm');```, heh, just 1.5 months and two issues have gone :D

apps/sms/test/marionette/thread_test.js
julienw: if we flick up really fast, this should still be possible that the messages are not loaded when we try to flick up? What do you think?

apps/sms/test/marionette/thread_test.js
```diff
@@ -25,9 +32,75 @@ marionette('Thread Panel Tests', function() {
...
+        assertIsDisplayed(messagesApp.Composer.sendButton);
+        assertIsDisplayed(messagesApp.Composer.messageInput);
+      });
+    });
+  });
+
   suite('Forward message', function() {
     setup(function() {
       messagesApp.launch();
+      messagesApp.setStorage();
```
julienw: maybe add a comment besides to say we set an empty db ?

apps/sms/test/marionette/thread_test.js
azasypkin: But we wait enough time with ```client.helper.waitForElement``` until they appear? Or what do you mean?

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
julienw: maybe in both places ?

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
julienw: yeah I don't really mind

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
julienw: I don't really mind either, it's more a suggestion, so if you feel strongly against such optimization I don't really mind. It's not only optimization but also how you write the code that's maybe painful. If that's the only place we need it then OK, if in the future we need to access messages directly elsewhere then we can revisit.

azasypkin: I agree that it makes sense, basically it was so initially, and then I just simplified it until the moment we really feel we need it :)

azasypkin: I'll try to recover what I had previously, maybe I can make it better than it was when I abandoned it :)

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
azasypkin: Test will definitely fail :D You'd like to fire error in this case?

apps/sms/test/marionette/mocks/mock_test_storages.js
azasypkin: I'll add brief comment and very nice link https://developer.mozilla.org/en-US/docs/Xray_vision :)

apps/sms/test/marionette/thread_test.js
azasypkin: Yep!

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
azasypkin: Okay, I remembered why I didn't do it this way :) Mock is becoming too "smart" with all that incoming/outgoing message-to-thread mapping, I'd rather leave at it as is for now and revisit it only when\if needed it as you suggest.

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
azasypkin: Hmm, we actually don't need any ```Cu.waiveXrays``` here, only ```cloneInto``` - will add comment with the link

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
azasypkin: Previously I store already cloned object, now I changed this and clone only when I return anything to content code.

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
azasypkin: Okay!

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
azasypkin: Using ```Services.DOMRequest.fireErrorAsync``` now.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6631f3aba262dfc29afe31d3ea67b91d7443930a

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
julienw: okay :)

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
julienw: do you know what the real API is doing?

apps/sms/test/marionette/mocks/mock_test_storages.js
julienw: I think I read this arleady :) but this is useful for anyone coming from normal web development and looking at this

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
azasypkin: Yep, rejects with ```NotFoundError````

apps/sms/test/marionette/thread_test.js
julienw: I think it should be good enough, we'll see if there are any intermittent over time. What I'm saying is very theorical.

when you scroll up, we display the 10 next elements that are in the DOM already; so it's theorically possible that the element we're looking for in not in the DOM yet.

But it's very theorical here, so let's forget it.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=52fac4225ab94467cefbd60d6cb37dab1de0318e

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
```diff
@@ -27,215 +32,240 @@ Services.obs.addObserver(function(document) {
...
-      }, delayMs);
+      var request = Services.DOMRequest.createRequest(window);
+
+      var segmentLength = 160;
+      var charsUsedInLastSegment = (text.length % segmentLength);
+
+      // Very simple example of message segmentation logic, in reality it's more
+      // complex, e.g. see real "calculateLength" method here:
+      // https://android.googlesource.com/platform/frameworks/opt/telephony
+      // /+/android-5.0.1_r1/src/java/android/telephony/SmsMessage.java
```
julienw: I was gonna say to point to B2G code instead (https://dxr.mozilla.org/mozilla-central/source/widget/android/AndroidBridge.cpp#1069) but it's really not readable...

Note that you can have the full URL on one line if there is no word before the URL, jshint should be happy with this.

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
```diff
@@ -27,215 +32,240 @@ Services.obs.addObserver(function(document) {
...
+      var result = {
+        segments: Math.ceil(text.length / segmentLength),
+        charsAvailableInLastSegment: charsUsedInLastSegment ?
+          segmentLength - charsUsedInLastSegment : 0
+      };
+
+      // See description at the top of the file about "cloneInto" necessity.
+      Services.DOMRequest.fireSuccessAsync(
+        request, Cu.cloneInto(result, window)
```
julienw: maybe we don't really care, I wonder if cloneInto is necessary here, as result is a new object with only scalar values.

azasypkin: I'll double check

azasypkin: Nope, there is a ```Permission denied``` exception if I remove ```cloneInto```, does it work for you without ```cloneInto```? I believe it doesn't work for object, since content can remove/change properties of source object.

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
azasypkin: > I was gonna say to point to B2G code instead (https://dxr.mozilla.org/mozilla-central/source/widget/android/AndroidBridge.cpp#1069) but it's really not readable...

Yeah, in the end I decided to point to Android source directly as it has much more comments :)

> Note that you can have the full URL on one line if there is no word before the URL, jshint should be happy with this.

Cool, didn't know about it!

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8aa0574c8e78b8cd122edde87c47fb7a46aa7dab

autolander:
http://docs.taskcluster.net/tools/task-graph-inspector/#ZQQIDltMR4Gattb5h-WgRg

PR #28460, opened 2015-02-25
Bug 1131917 - mozCamera may null when resume to camera app and touch scr...

jinchaowang:
...een rapidly to focus

### DIFF
--- a/apps/camera/js/lib/camera/camera.js
+++ b/apps/camera/js/lib/camera/camera.js
@@ -846,6 +846,7 @@ Camera.prototype.takePicture = function(options) {
 };
 
 Camera.prototype.updateFocusArea = function(rect, done) {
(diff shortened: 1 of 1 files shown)


apps/camera/js/lib/camera/camera.js
```diff
@@ -846,6 +846,10 @@ Camera.prototype.takePicture = function(options) {
 };
 
 Camera.prototype.updateFocusArea = function(rect, done) {
+  if (!this.mozCamera) {
+    debug('will focus but this.mozCamera is null, return to ignore focus');
```
wilsonpage: This log is confusing to me. I think the code is self explainatory. Please change to:

```js
if (!this.mozCamera) { return; }
this.focus.updateFocusArea(rect, focusDone);
...
```

PR #28466, opened 2015-02-25
Bug 908262 - SMS ordered by delivery time instead of sent time

lanahuong:


autolander:

PR #28474, opened 2015-02-25
Bug 1135873 - Test Case - Add imagecompare video playback test case 

nojunpark:
This test case takes snapshot of still frames and perform comparison


try-server-hook:
[npark-mozilla No-Jun Park (npark-mozilla)](https://github.com/npark-mozilla) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=eeb08290f7db)

tests/python/gaia-ui-tests/gaiatest/tests/graphics/test_play_3gp_video.py
```diff
@@ -0,0 +1,70 @@
...
+
+        video_player = VideoPlayer(self.marionette)
+        video_player.launch()
+        video_player.wait_for_thumbnails_to_load(1, 'Video files found on device: %s' %self.data_layer.video_files)
+
+        # Assert that there is at least one video available
+        self.assertGreater(video_player.total_video_count, 0)
+        first_video_name = video_player.first_video_name
+
+        self.assertEqual('none', self.data_layer.current_audio_channel)
```
davehunt: Why is this check necessary here?

tests/python/gaia-ui-tests/gaiatest/tests/graphics/test_play_3gp_video.py
```diff
@@ -0,0 +1,70 @@
...
+        self.push_resource('VID_IC_0001.3gp')
+
+    def test_play_3gp_video(self):
+        """https://moztrap.mozilla.org/manage/case/2478/"""
+
+        self.take_screenshot()
+
+        video_player = VideoPlayer(self.marionette)
+        video_player.launch()
+        video_player.wait_for_thumbnails_to_load(1, 'Video files found on device: %s' %self.data_layer.video_files)
```
davehunt: Run PEP8 or Flake8, I strongly suspect a missing whitespace issue here.

tests/python/gaia-ui-tests/gaiatest/tests/graphics/test_play_3gp_video.py
```diff
@@ -0,0 +1,70 @@
...
+        # See bug 1109203, current_audio_channel switches context to the system
+        # app, so we need to return it back to the displayed app
+        self.apps.switch_to_displayed_app()
+
+        # Click on the first video.
+        fullscreen_video = video_player.tap_first_video_item()
+
+        # Video will play automatically
+        # After playback, it rewinds to the beginning
+        time.sleep(15)
```
davehunt: Why 15 seconds?

tests/python/gaia-ui-tests/gaiatest/tests/graphics/test_play_3gp_video.py
```diff
@@ -0,0 +1,70 @@
...
+
+        # take snapshot of random frames
+        fullscreen_video.tap_forward()  # tapping forward or rewinds moves frames in 10 sec increment
+        self.take_screenshot()
+        fullscreen_video.tap_forward()  # go to the end of the frame
+        self.take_screenshot()
+
+        # go back to the beginning
+        fullscreen_video.tap_rewind()
+        fullscreen_video.tap_rewind()
```
davehunt: Why tap twice? Could we have an app object method for returning to the start?

davehunt: Ah I see from above regarding single/double tap. I think this would be clearer as an app object method, but a comment would also be fine.

tests/python/gaia-ui-tests/gaiatest/tests/graphics/test_play_3gp_video.py
npark-mozilla: I missed the warning here, thanks

tests/python/gaia-ui-tests/gaiatest/tests/graphics/test_play_3gp_video.py
davehunt: Well, you already have a comment, but change it to something like "tap rewind twice to go back to the beginning".

tests/python/gaia-ui-tests/gaiatest/tests/graphics/test_play_3gp_video.py
npark-mozilla: this was in the original script, and apparently the comment explains it but I should bundle it below the comments. thanks!

tests/python/gaia-ui-tests/gaiatest/tests/graphics/test_play_3gp_video.py
npark-mozilla: the video is 14 seconds, so I give enough time to play until the end and rewinds to the beginning, will add comment

tests/python/gaia-ui-tests/gaiatest/tests/graphics/test_play_3gp_video.py
npark-mozilla: comment added

try-server-hook:
[npark-mozilla No-Jun Park (npark-mozilla)](https://github.com/npark-mozilla) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=c5aed492d8fc)

tests/python/gaia-ui-tests/gaiatest/tests/graphics/test_play_3gp_video.py
```diff
@@ -0,0 +1,67 @@
...
+        # Click on the first video.
+        fullscreen_video = video_player.tap_first_video_item()
+
+        # Video will play automatically
+        # After playback, it rewinds to the beginning.  The video file is 14 seconds long
+        time.sleep(15)
+
+        begin_time = time.strptime('00:00', '%M:%S')
+        # display the data overlay to show the elapsed time
+        fullscreen_video.show_controls()
```
mwargers: show_controls is a programmatic way of showing the controls, perhaps you could use tap_video_control() to mimick what a user would be doing to show the controls here?

npark-mozilla: I thought so, then i checked the code for show_controls, and all it did was tapping the screen! not sure why the original comment said it is a programmatic way though.

tests/python/gaia-ui-tests/gaiatest/tests/graphics/test_play_3gp_video.py
```diff
@@ -0,0 +1,67 @@
...
+        # app, so we need to return it back to the displayed app
+        self.assertEqual('none', self.data_layer.current_audio_channel)
+        self.apps.switch_to_displayed_app()
+
+        # Click on the first video.
+        fullscreen_video = video_player.tap_first_video_item()
+
+        # Video will play automatically
+        # After playback, it rewinds to the beginning.  The video file is 14 seconds long
+        time.sleep(15)
```
mwargers: Ugh, this can't be sped up by just fast-forwarding to the end of the video (mimicking a tap just before the end of the seeker), then wait a second or so?

npark-mozilla: I thought about that, then i realized some bugs actually happen when you let it play until the end (i.e., time mismatching, screen turning black) and does not happen if you only do seek operation, so I decided to let it play, and make sure nothing out of ordinary is happening.

geoelectric:
Unsolicited review:

I think it's OK to have a separate test, since they have different purposes, even if a lot of flow is shared.

And since we have a version that already makes sure it does everything user-like, I actually think it'd be fine to shortcut non-essential UI manipulation with programmatic calls instead to make the test robust.

However, because our logging only reports filename, class name, and test function name, those need to be made distinct. 

Right now, this is the same as the functional/videoplayer/test_play_3gp_video.py file in those respects and would be confusing.

tests/python/gaia-ui-tests/gaiatest/tests/graphics/test_play_3gp_video.py
mwargers: Oops, sorry, you're right! I looked at the wrong show_controls function (I looked at the html5_player one).

tests/python/gaia-ui-tests/gaiatest/apps/videoplayer/regions/fullscreen_video.py
```diff
@@ -26,6 +33,24 @@ def show_controls(self):
         self.marionette.find_element(*self._video_player_locator).tap()
         self.wait_for_element_displayed(*self._video_controls_locator)
 
+    # move the slider
+    def move_seek_slider(self, dir_x):
```
davehunt: Nit: I would rename `dir_x` to `offset`.

tests/python/gaia-ui-tests/gaiatest/tests/graphics/test_play_3gp_video.py
davehunt: My concern is that any lag in Marionette or Gaia of 1 second or more will mean we haven't reached the end of the video. Is this a case you'd want a failure? If not, perhaps it may be an idea to pad it a little more. It also might make sense to set a constant for the video length, and just add a couple of seconds here, so if it changes it's easier to locate.

tests/python/gaia-ui-tests/gaiatest/tests/graphics/test_play_3gp_video.py
```diff
@@ -0,0 +1,67 @@
...
+
+        # Assert that there is at least one video available
+        self.assertGreater(video_player.total_video_count, 0)
+        first_video_name = video_player.first_video_name
+        self.take_screenshot()
+
+        # See bug 1109203, current_audio_channel switches context to the system
+        # app, so we need to return it back to the displayed app
+        self.assertEqual('none', self.data_layer.current_audio_channel)
+        self.apps.switch_to_displayed_app()
```
davehunt: This check might be appropriate for the functional test, but I'm, not sure it benefits the graphics one. It adds complexity, and is already tested elsewhere.

davehunt:
I agree with Geo that these tests can take a few liberties. It's valuable for them to check things they rely on like videos being present/displayed - although the following steps would likely fail anyway if they didn't - but if there are shortcuts we can take it will help to reduce maintenance.

tests/python/gaia-ui-tests/gaiatest/tests/graphics/test_play_3gp_video.py
npark-mozilla: good idea, fixed

tests/python/gaia-ui-tests/gaiatest/apps/videoplayer/regions/fullscreen_video.py
npark-mozilla: done

tests/python/gaia-ui-tests/gaiatest/tests/graphics/test_play_3gp_video.py
npark-mozilla: aha, thought I needed  self.apps.switch_to_displayed_app() for the script to work, but guess that is only for the audio. removed.

try-server-hook:
[npark-mozilla No-Jun Park (npark-mozilla)](https://github.com/npark-mozilla) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=8ee7a30d3eca)

npark-mozilla:
squashed

PR #28487, opened 2015-02-26
Bug 962409 - (conninfo-widget) [LockScreen] Implement ConnInfo widget

snowmantw:
1. Fixed exposed defects of process/stream/state/component
2. Extend the above components according to the use cases
3. Make sure all basic cases are implemented

### DIFF
--- a/apps/system/lockscreen/js/lockscreen.js
+++ b/apps/system/lockscreen/js/lockscreen.js
@@ -24,7 +24,7 @@
     /*
     * Lockscreen connection information manager
     */
-    _lockscreenConnInfoManager: null,
+    _lockscreenConnectionStatesWidget: null,
 
     /*
     * Boolean return true when initialized.
@@ -177,6 +177,12 @@
           this.chargingStatus.stop();
         } else {
           this._passCodeTimeoutCheck = this.checkPassCodeTimeout();
+          // XXX: [lock] -> screen enabled, since when it's 'locked' by button,
+          //                the screen is still off, so no need to draw widgets
+          //                and make them functional
+          //      [unlock] -> slide to unlock, no problem
+          //
+          // This remains until the top-most panels get implemented.
           if (!this.lockScreenClockWidget) {
             this.createClockWidget();
           }
@@ -491,10 +497,8 @@
     // mobile connection state on lock screen.
     // It needs L10n too. But it's not a re-entrable function,
     // so we need to check if it's already initialized.
-    if (window.navigator.mozMobileConnections &&
-        !this._lockscreenConnInfoManager) {
-      this._lockscreenConnInfoManager =
-        new window.LockScreenConnInfoManager(this.connStates);
+    if (window.navigator.mozMobileConnections) {
+      this.createConnectionStatesWidget();
     }
   };
 
@@ -662,8 +666,22 @@
       return;
     }
 
-    this.lockScreenClockWidget.stop().destroy();
-    delete this.lockScreenClockWidget;
+    this.lockScreenClockWidget.stop()
+      .next(() => {
+        this.lockScreenClockWidget.destroy().next(() => {
+          delete this.lockScreenClockWidget;
+        });
+      });
+
+
+    if (window.navigator.mozMobileConnections) {
+      this._lockscreenConnectionStatesWidget.stop()
+        .next(() => {
+          this._lockscreenConnectionStatesWidget.destroy().next(() => {
+          delete this._lockscreenConnectionStatesWidget;
+        });
+      });
+    }
 
     if (this.unlockSoundEnabled) {
       var unlockAudio = new Audio('/resources/sounds/unlock.opus');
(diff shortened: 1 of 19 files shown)


try-server-hook:
[snowmantw Greg Weng (snowmantw)](https://github.com/snowmantw) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=0ef30c6fbe42)

try-server-hook:
[snowmantw Greg Weng (snowmantw)](https://github.com/snowmantw) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=edc34de61628)

try-server-hook:
[snowmantw Greg Weng (snowmantw)](https://github.com/snowmantw) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=35501c788521)

try-server-hook:
[snowmantw Greg Weng (snowmantw)](https://github.com/snowmantw) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=3d69c3f960ae)

apps/system/lockscreen/js/lockscreen.js
```diff
@@ -662,8 +666,22 @@
...
-    delete this.lockScreenClockWidget;
+    this.lockScreenClockWidget.stop()
+      .next(() => {
+        this.lockScreenClockWidget.destroy().next(() => {
+          delete this.lockScreenClockWidget;
+        });
+      });
+
+
+    if (window.navigator.mozMobileConnections) {
```
timdream: `if (this._lockscreenConnectionStatesWidget)` ?

apps/system/lockscreen/js/state/lockscreen_basic_state.js
```diff
@@ -57,6 +61,17 @@
    * Or the component cannot track the last active state.
    */
   LockScreenBasicState.prototype.transferTo = function() {
+    if (this._transferred) {
+      this.logger.debug('Prevent transferring racing');
+      var nullifized = new Stream();
+      // This would return a process could be concated but would do nothing.
+      // It's better to formally provide a API from Process, like
+      // Process.maybe() or Process#nullize(), but this is a simplier solution.
+      return nullifized.start().next(() => nullifized.stop());
```
timdream: Isn't this simply equal to `Promise.resolve()`?

apps/system/lockscreen/js/state/lockscreen_basic_state.js
```diff
@@ -57,6 +61,17 @@
    * Or the component cannot track the last active state.
    */
   LockScreenBasicState.prototype.transferTo = function() {
+    if (this._transferred) {
```
timdream: I am not convinced we need this flag. It sounds like a workaround to a deeper issue.

apps/system/lockscreen/js/state/lockscreen_basic_state.js
snowmantw: This function would return a Process/Stream could be concated with 'next' since the nature of the `stop` and `start` method. So if I return a Promise this would probably break unknown user code, and add the extra burden of detecting this un-uniform type on each caller.

apps/system/lockscreen/js/state/lockscreen_basic_state.js
snowmantw: This is a simple deduction progress:

1. Just focus on this method: whether calling a State to transfer itself twice is reasonable or not. 
2. Obviously it's not reasonable. So, how to make sure this wouldn't happen?
3.1. We could ask every caller to detect whether the state is transferred.
3.2. Or the method could 'nullify' it's second call
4. However, to adopt 3.1 would, as I mentioned, to ask every state to do the duplicated thing to detect the similar condition using the similar code
5. So here I adopted 3.2, to solve this case while prevent from scattering duplicated code about everywhere

apps/system/lockscreen/js/lockscreen.js
snowmantw: Yes I would change it.

apps/system/lockscreen/js/state/lockscreen_basic_state.js
timdream: My question is simply that we shouldn't be transferring twice anyway. If so that means something is wrong with our concurrent wiring... you could certainly workaround and nullify the second call but I recommend we find the real cause.

apps/system/lockscreen/js/widgets/connection_states/lockscreen_connection_states_sim_widget_setup.js
```diff
@@ -0,0 +1,220 @@
...
+   * A necessary forwarding hack since the original 'voicechange'
+   * event bring nothing about which SIM slot is the event source.
+   */
+  LockScreenConnectionStatesSIMWidgetSetup.prototype.startForwardVoicechange =
+  function() {
+    this.component.resources.simslot.conn.addEventListener('voicechange',
+      this.stream.configs.sources[0].onchange);
+  };
+
+  LockScreenConnectionStatesSIMWidgetSetup.prototype.stopForwardVoicechange=
```
timdream: ` =`

timdream:
General comment: I am not seeing any old code being purged.

PR #28491, opened 2015-02-26
Bug 1136633 - [Messages] Blank page when launching the SMS application in some conditions. r=julien

azasypkin:


try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=c8bfc4645042)

apps/sms/js/activity_handler.js
```diff
@@ -511,7 +511,11 @@ var ActivityHandler = {
     //
     var params = Utils.params(message.imageURL);
 
+    // When notification is removed from notification tray, notification system
+    // message will still be fired, but "clicked" property will be equal to
+    // false. In this case we should open default ThreadList panel.
     if (!message.clicked) {
+      Navigation.toPanel('thread-list');
```
julienw: Proposal:

`if (document.hidden) { window.close() } else { Navigation.toPanel('thread-list'); }`

julienw: rationale: I don't want to take 100% CPU for some seconds just because we closed a notification

julienw: the "if" is to overcome race conditions like the user launches the app just as soon as he closed the notification.

azasypkin: I'm fine with proposal for master, but not sure that we should do it for v2.2 or you think it's safe enough?

julienw: Not sure... I wonder what happens if we're doing something else in the same time (like receiving a SMS/MMS) but are sitll not displayed...

julienw: this is where a task/queue based logic helps :D

julienw: Can you check what happens in v2.1, and stick to this behavior ?

azasypkin: Yep, checking...

azasypkin: v2.1 runs and keeps Messages app when we remove notification as well

try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=17a391f50eae)

PR #28502, opened 2015-02-26
Bug 1122105 - Test that we display the album art.

hfiguiere:


try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=3e94721030be)

try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=36eae73824c5)

apps/music/test/marionette/lib/listviewhelper.js
```diff
@@ -0,0 +1,18 @@
+/* global module, require */
+'use strict';
+
+var assert = require('assert');
+
+var ListviewHelper = {
+
+  backgroundImageForListItem: function(list_items, index) {
```
davidflanagan: I'd change the function name. If the implementation changes to use an  tag instead of background-image, then this name wouldn't make sense. albumArtForListItem() would make sense.  (Of course, then there is the issue of whether you should remove the "url()" from the returned string or not...)

The other helper classes I've seen take a single element as their argument. Here you're using an array and index which seems non-parallel. I'd modify to just pass the element and do the indexing in the code that calls this function.

apps/music/test/marionette/metadata_test.js
```diff
@@ -0,0 +1,73 @@
+/* global require, marionette, setup, suite, test, __dirname */
+'use strict';
+
+var assert = require('assert');
+var Music = require('./lib/music.js');
+var FakeRingtones = require('./lib/fakeringtones.js');
```
davidflanagan: I've got no idea what the fakeringtones is for, but have you checked whether it is really necessary here?
Since it seems so unrelated to the point of this test, if you really do need it, maybe add a comment just saying that it is necessary but irrelevant.

apps/music/test/marionette/metadata_test.js
```diff
@@ -0,0 +1,73 @@
...
+      'device.storage.prompt.testing': true
+    },
+
+    settings: {
+      'lockscreen.enabled': false,
+      'ftu.manifestURL': null
+    },
+
+    apps: apps
+  });
```
davidflanagan: Ah. I think I see. the fakecontrols and fakeringtones are fake apps that are needed by some of the test.  Most of the stuff from this line above seems like boilerplate marionette setup. Could you refactor this stuff into lib/marionette_setup.js and then replace all this with:

   var client = setup_marionette();

Or something like that?

This is just an idea. I don't feel strongly about making that change.

apps/music/test/marionette/metadata_test.js
```diff
@@ -0,0 +1,73 @@
...
+
+      bg = ListviewHelper.backgroundImageForListItem(music.listItems, 0);
+      assert.ok(bg.indexOf('url("blob:') != -1);
+
+      music.switchToSongsView();
+
+      bg = ListviewHelper.backgroundImageForListItem(music.listItems, 0);
+      assert.ok(bg.indexOf('url("blob:') != -1);
+    });
+
```
davidflanagan: This would be a more convincing test if you had two songs with different album art, and you verified that each song had the same blob url in each of the views and that the two songs always had different blob urls.  Or at least have one song with album art and one without and verify that they display different urls.

Of course, the ideal thing would be to verify that the album art was actually correct. Like make on song have an image that is a solid black square and the other have a solid white square and then somehow check that the right colors are displayed. I don't know how you could actually do that, though, so I don't want to require it here.

apps/music/test/marionette/lib/listviewhelper.js
hfiguiere: Totally make sense. Done.

apps/music/test/marionette/metadata_test.js
hfiguiere: Probably cruft from cut and pasting another test. Removed. FakeControls too.

apps/music/test/marionette/metadata_test.js
hfiguiere: I can file a bug for that.

apps/music/test/marionette/metadata_test.js
hfiguiere: Currently the art in the sample file is ````apps/music/test-data/album-art.jpg````. So we could perform an image comparison.
Indeed it make sense to check with a second file that has different art. I am not sure if the blob URL would be different for the same image or not. If there is no art we don't have a blob URL but rather a file URL.

I'll rework the test to improve this the way you suggested.

hfiguiere: We already unit test the extraction of the album art from the metadata. So I think comparing that they differ between the two ought to be enough.

try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=1ca6904b8cfe)

try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=93fd7c7372be)

apps/music/test/marionette/metadata_test.js
```diff
@@ -0,0 +1,109 @@
...
+        },
+        {
+          // red art
+          type: 'music',
+          filePath: 'apps/music/test-data/playlists/01.ogg'
+        },
+        {
+          // no art
+          type: 'music',
+          filePath: 'test_media/samples/Music/b2g.ogg'
```
davidflanagan: I think it would be better to create a new file somewhere under in apps/music/test-data/  rather than relying on this external file. Someone might add art to it or delete it.

apps/music/test/marionette/metadata_test.js
```diff
@@ -0,0 +1,109 @@
...
+        },
+      ]);
```
davidflanagan: There are a bunch of undocumented assumptions in this fileManager.add() call. In order for the test to work, I think that the album, artist and song names for all three files have to be in a certain alphabetical order. Since these test files are shared with other tests, their metadata might change. So I think it would be good to add a comment documenting the requirements for these files, and maybe also specifying the title, album and artist for each of the files.

apps/music/test/marionette/metadata_test.js
hfiguiere: In both case they'd break the test. So it would get caught. I can find/use a different file.

apps/music/test/marionette/metadata_test.js
hfiguiere: 1. As I said the tests would break
2. they are ours so the amount of people changing them is lower (I'll take comment above into account)

try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=3d37ada45b9c)

try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=e8e439a9bc30)

PR #28503, opened 2015-02-26
Bug 1136633 - [Messages] Blank page when launching the SMS application in some conditions. r=julien

azasypkin:


try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=2aa25308fc82)

apps/sms/js/navigation.js
```diff
@@ -116,6 +116,17 @@ var Navigation = {
   },
 
   /**
+   * Ensures that current panel is correctly set, if no current panel set and
+   * we're not in the process of transitioning then set default panel as current
+   * one.
+   */
+  ensureCurrentPanel: function() {
```
azasypkin: I'm not sure about the correct name for this method, but the idea is to fix app navigation state when it's not valid and there is a chance that app instance will be used by user (I'll add unit tests once we figure out best approach to deal with the issue). As you found on Friday, my solution doesn't work correctly when notification click is handled by system message + it wouldn't work correctly when you have already opened app (navigated to Composer for example) and then you remove notification (that wasn't created by current app instance, so click will be propagated through system message) - user will be navigated to thread list :)

Initially I put this code into ActivityHandler, but it uses only Navigation internal data, so I moved it to Navigation.js.

So what are your thoughts about that?

try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=f3f6f0d3e982)

apps/sms/js/navigation.js
```diff
@@ -72,6 +72,8 @@ var Navigation = {
     this.mainWrapper = document.getElementById('main-wrapper');
     this.transitioning = false;
 
+    currentPanel = null;
```
azasypkin: Can't think of any better way to nullify this in unit tests :(

apps/sms/js/navigation.js
azasypkin: Also was thinking to return a promise here, smth like this:
```
if (this.transitioning) {
  return xPromise;
}

if (!currentPanel) {
  return this.toPanel(this.defaultPanel);
}

return Promise.resolve();
```

But don't have anything for ```xPromise```, that's how the following question appeared: do you know why can't we use smth like ```Navigation.transitionPromise (that equals to local promise from toPanel)``` instead of boolean ```Navigation.transitioning```?

It's not important as there is no need in this promise currently :)

julienw: > But don't have anything for xPromise, that's how the following question appeared: do you know why can't we use smth like Navigation.transitionPromise (that equals to local promise from toPanel) instead of boolean Navigation.transitioning?

no reason, I think we could :) "transitioning" is currently a private var anyway.

julienw: > Initially I put this code into ActivityHandler, but it uses only Navigation internal data, so I moved it to Navigation.js.

I think it's good in Navigation.

For the name, I don't really have a strong opinion.

In my mind, long-term, I think Navigation could handle the initial loading: given a hash, it would lazy load the right JS file (which in turn would lazy load all dependencies), and then call toPanel after the right JS file has returned something (or not) in a promise. So... this name is as good as any other name.

apps/sms/js/navigation.js
julienw: it's ok

apps/sms/js/activity_handler.js
```diff
@@ -558,7 +558,15 @@ var ActivityHandler = {
     //
     var params = Utils.params(message.imageURL);
 
+    // When notification is removed from notification tray, notification system
+    // message will still be fired, but "clicked" property will be equal to
+    // false. When user clicks on notification we'll get two system messages,
+    // first to notify app that notification is clicked and the second one to
+    // indicate that notification is removed from notification tray.
```
julienw: maybe go to a new line before "When", and explain we get 2 system messages because _we_ remove the notification from SMS?

Also add a reference to the bug # where we'll remove this close event ?

apps/sms/test/unit/navigation_test.js
```diff
@@ -521,6 +521,66 @@ suite('navigation >', function() {
...
+
+      Navigation.init().catch((e) => {
+        Navigation.toPanel.reset();
+
+        // It's expected error for the unknown panel
+        if (e.message === 'Panel x-unknown-panel is unknown.') {
+          return;
+        }
+
+        return Promise.reject(e);
```
julienw: nit: maybe "throw e" is just as good.

apps/sms/test/unit/navigation_test.js
```diff
@@ -521,6 +521,66 @@ suite('navigation >', function() {
     });
   });
 
+  suite('ensureCurrentPanel >', function() {
+    setup(function(done) {
+      this.sinon.stub(Navigation, 'slide').returns(Promise.resolve());
+      this.sinon.spy(Navigation, 'toPanel');
+
+      // Set custom hash to have non-initialized current panel
+      window.location.hash = '#x-unknown-panel';
```
julienw: nit: use the hash we'll get from the notification ? `#notification` ?

apps/sms/test/unit/navigation_test.js
```diff
@@ -521,6 +521,66 @@ suite('navigation >', function() {
...
+    });
+
+    test('navigates to default panel when current panel is not set',
+    function(done) {
+      // We don't have any correct current panel, so navigation to default
+      // panel should be triggered.
+      Navigation.ensureCurrentPanel();
+
+      sinon.assert.calledOnce(Navigation.toPanel);
+      sinon.assert.calledWith(Navigation.toPanel, 'panel1');
```
julienw: instead of looking whether internal methods are called, maybe we can return a promise from ensureCurrentPanel like you suggested, and then check whether the current panel is 'panel1' using the public method isCurrentPanel ?

apps/sms/test/unit/navigation_test.js
```diff
@@ -521,6 +521,66 @@ suite('navigation >', function() {
...
+      }).then(done, done);
+    });
+
+    test('does not navigate to default panel when transitioning is happening',
+    function() {
+      // We don't have any correct current panel, but we're in transitioning
+      // state currently that means that if nothing goes wrong we'll have
+      // initialized current panel soon and no need to force navigation to
+      // default panel.
+      Navigation.transitioning = true;
```
julienw: do you think it would work to call toPanel and then ensureCurrentPanel, instead of fiddling with internal vars ?

apps/sms/test/unit/navigation_test.js
```diff
@@ -521,6 +521,66 @@ suite('navigation >', function() {
...
+      Navigation.ensureCurrentPanel();
+
+      sinon.assert.notCalled(Navigation.toPanel);
+
+      // When we're not in the transitioning state and current panel is not
+      // available then we should navigate to default panel;
+      Navigation.transitioning = false;
+      Navigation.ensureCurrentPanel();
```
julienw: not sure this case is worth it; we already handle this in the previous test, right ?

apps/sms/js/activity_handler.js
azasypkin: > maybe go to a new line before "When", and explain we get 2 system messages because we remove the notification from SMS?

Okay!

> Also add a reference to the bug # where we'll remove this close event ?

Mmm, not sure that we have one, afaik Michael was talking about some non-existent-yet bug where notification guys are going to make Gecko Notification API conformant with the latest spec (including removing of "show" and "close" events). But I can file one for "onclose" only if you prefer.

apps/sms/js/navigation.js
azasypkin: Okay, will go with ``` ensureCurrentPanel``` for now. I also have some things in mind similar to what  you've described, so we're good :)

apps/sms/test/unit/navigation_test.js
azasypkin: mm, why not :) Will change

apps/sms/test/unit/navigation_test.js
azasypkin: yeah, just habbit to preserve original stack trace, but it's just test, so can re-throw if you prefer :)

apps/sms/test/unit/navigation_test.js
azasypkin: Yeah, with promise-based method it will look better, will try smth like ```transitioningPromise``` and see if it's small enough change for v2.2 (I believe it's!) ;)

apps/sms/test/unit/navigation_test.js
azasypkin: Yep, it should + we wouldn't do it with ```transitioningPromise```

apps/sms/test/unit/navigation_test.js
azasypkin: Mm, it's just for the sake of being more explicit in the test, will see how it will look with ```boolean ----> promise``` change for ```Navigation.transitioning```

apps/sms/test/unit/navigation_test.js
julienw: isn't the stack trace set once and for all when the error is created ?

apps/sms/js/activity_handler.js
julienw: yeah, file a bug and tell it in bug 1068564

apps/sms/test/unit/navigation_test.js
azasypkin: Ha-ha, you're right! As always I want JS to be like C#, but here it's more like Java that preserves original stack trace when rethrown. My bad :)

apps/sms/js/activity_handler.js
azasypkin: Done https://bugzilla.mozilla.org/show_bug.cgi?id=1139363

try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=411dcae37ceb)

try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=9b74d7ccc9dc)

apps/sms/js/navigation.js
```diff
@@ -268,16 +285,16 @@ var Navigation = {
...
       }.bind(this)
     );
 
-    promise.
+    this.transitionPromise.
       then(() => this.emit('navigated', { panel: panel, args: args })).
       then(nextQueuedPanel, nextQueuedPanel);
 
-    return promise;
+    return this.transitionPromise;
```
julienw: could lead to less change if you kept the local variable `promise` and finish with `return this.transitionPromise = promise`.

This would work the same and could make the code slightly more readable.

apps/sms/js/navigation.js
```diff
@@ -152,7 +171,7 @@ var Navigation = {
     //   - Panel is still transitioning
     //   - trying to navigate when the app is not loaded completely
     var notReadyNavigate = (panel !== this.defaultPanel && !this.isReady);
-    if (this.transitioning || notReadyNavigate) {
+    if (this.transitionPromise || notReadyNavigate) {
       queuedPanel = {
         panel: panel,
         args: args
```
julienw: I've been thinking that we could likely not nullify `this.transitionPromise` and encapsulate everything here with this.transitionPromise.then(...). But this is more than we'd want to do here obviously :)

azasypkin: Yeah, maybe later we can refine this :)

apps/sms/js/navigation.js
azasypkin: Yeah, makes sense

try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=bc8b707bdcfc)

try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=33a9189bcbce)

try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=9438cea24f39)

PR #28509, opened 2015-02-26
Bug 1128739 - [email] Refactor to streamline startup

jrburke:
A refactor of the email startup that uses localStorage and a flatter startup structure once the start card is known, where html_cache_restore is the decision maker. There are some notes about how the code is structured in docs/startup.md in the pull request.

That doc is meant to be a long term statement of the startup structure, below are some notes for those familiar with how the code works now and the diffs from that state:
### No more "set up account?" dialog

I removed the confirm dialog when the user does not have an account but calls email with an activity. This simplified the flow and I think is a better user experience. Now setup_account_info is shown with an error message in the card indicating an account is needed. This part needs ux-review, and here is a capture of how it looks:

new_setup_account_info
### Removed some async setup

Both `cards` and `model` now depend on a `cards_init` and `model_init` respectively that allow it to do more synchronous configuration. The contents or mapping of the init modules can be changed via module config for different environments. This helped reduce the number of async hops at runtime between files for setting up those modules.

So the change in model_acctsslice_startup_test.js is mainly about that change: model setup does not emit an "api" notification, so the `api` module is just directly referenced as a dependency.
### No fragment ID dispatch

The ticket talked about doing fragment ID dispatch to any arbitrary card, but these changes alone were big enough that I felt just getting this stabilized was the first step. So I will file a follow-on bug for that once we have sat with this for a bit. It could be that service worker use comes along fast enough that it would just fall out of that work too.
### Startup performance

I seemed to get about a 40-50ms averaged hit by going with localStorage when tested locally. So instead of it being 400ms for visually complete, it was more like 450ms instead. That is still the lowest startup for visually complete for an app, and I think worth the cost given the move away from horrible cookie APIs and the ability to cache more cards.

In addition, some later events, moz-content-interactive and moz-chrome-interactive seemed to slightly improve, but the final app-loaded metric did not, it seemed to reflect the startup cost delay. So those intermittent metrics probably just got pushed forward ahead of some work that needs to complete anyway. My bet is that some of the card pieces that are less dependent on model being completely ready before executing can now run sooner.
### request-sync testing shows an IDB error

When the app closes itself after syncing, I will see messages like this:

```
W/E-Mail  (10208): [JavaScript Warning: "An IndexedDB transaction that was not yet complete has been aborted due to page navigation." {file: "app://email.gaiamobile.org/js/ext/main-frame-setup.js" line: 5186}]
```

I can get these on master too, so not caused by this change, and they do not seem to be fatal, but want to call out that I have seen them.
### document.hidden on startup

html_cache_restore uses a setTimeout to test document.hidden when notifications start up the app, because document.hidden was false at that point in startup without the setTimeout. I would like to think the app is so fast now we are testing things before they are ready, but I think it is more likely just a bug in the platform, and will be filing a bug for that issue. However I want to work up a simpler test case for it first.
### Beware zombie syncs!

If testing periodic sync, due to bug xxx, canceled sync operations never die! A restart does not kill them. A reset-gaia though seems to do the trick. On the plus side, the email app does not freak out when getting these older syncs, just lets them pass through.


try-server-hook:
[jrburke James Burke (jrburke)](https://github.com/jrburke) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=93e7490d0edd)

apps/email/js/activity_composer_data.js
```diff
@@ -0,0 +1,62 @@
...
+      attachData = {
+        body: url
+      };
+    } else {
+      attachData = queryUri(url);
+      attachData.attachmentBlobs = data.blobs || [];
+      attachData.attachmentNames = data.filenames || [];
+
+      attachmentName.ensureNameList(attachData.attachmentBlobs,
+                                 attachData.attachmentNames);
```
asutherland: nit: identation

apps/email/js/config.js
```diff
@@ -0,0 +1,72 @@
...
+
+// Set up loading of scripts, but only if not in tests, which set up
+// their own config.
+if (typeof TestUrlResolver === 'undefined') {
+  requirejs.config({
+    // waitSeconds is set to the default here; the build step rewrites
+    // it to 0 in build/email.build.js so that we never timeout waiting
+    // for modules in production. This is important when the device is
+    // under super-low-memory stress, as it may take a while for the
+    // device to get around to loading things email for background tasks
```
asutherland: not: comment phrasing hiccup.  maybe wants to be: device to get around to loading things email NEEDS for background tasks

apps/email/js/html_cache.js
```diff
@@ -37,56 +39,66 @@ exports.cloneAsInertNodeAvoidingCustomElementHorrors = function(node) {
...
-  html = encodeURIComponent(HTML_COOKIE_CACHE_VERSION +
+  html = HTML_CACHE_VERSION +
                             (langDir ? ',' + langDir : '') +
-                            ':' + html);
+                            ':' + html;
 
-  // Set to 20 years from now.
-  var expiry = Date.now() + (20 * 365 * 24 * 60 * 60 * 1000);
-  expiry = (new Date(expiry)).toUTCString();
+  // console.log('SAVING: ' + 'html_cache_' + id + ': ' + html);
```
asutherland: nit: this commented out line seems potentially mooted by the active console.log below.

apps/email/js/html_cache_restore.js
```diff
@@ -21,23 +66,143 @@ var startupCacheEventsSent = false;
...
+      // done on non-trivial entry points, like manual_config.
+      startupData = {}; //parseFragmentId();
+
+  /**
+   * startupOnModelLoaded is set to this function if the
+   * data_has_account localStorage value is not known. Called eventually by
+   * config.js once the model module has been loaded.
+   */
+  function localOnModelLoaded(model, callback) {
+    model.once('acctsSlice', function(acctsSlice) {
```
asutherland: Would it be safer for this to be a call to latestOnce()?  If not or undesired, a comment seems appropriate.

apps/email/js/html_cache_restore.js
```diff
@@ -21,23 +66,143 @@ var startupCacheEventsSent = false;
...
+ * happen, see localOnModelLoaded for more details. config.js looks for this
+ * callback.
  */
-var HTML_COOKIE_CACHE_MAX_SEGMENTS = 40;
-
-// Use a global to work around issue with
-// navigator.mozHasPendingMessage only returning
-// true to the first call made to it.
-window.htmlCacheRestorePendingMessage = [];
+var startupOnModelLoaded;
```
asutherland: I think it would be helpful for this to be a bit more blatant about the fact that it's a global.  I tend to like explicitly doing "window.strtupOnModelLoaded" (probably because it's both obvious and shuts up js2-mode's syntax highlighting since it expects window to be a global), but other things could work.  At the very least, I think the doc-block should say the word global.

apps/email/js/html_cache_restore.js
```diff
@@ -21,23 +66,143 @@ var startupCacheEventsSent = false;
...
+
+  // Set up the default view, but if that is not possible to know yet, since
+  // the status of hasAccount is unknown, wait for the callback to set it up.
+  if (!startupOnModelLoaded) {
+    setDefaultView();
+  }
+
+  startupData.entry = 'default';
+
+  /*
```
asutherland: Please tag this with a comment like: "imminent future work, bug NNNNNN"

apps/email/js/html_cache_restore.js
```diff
@@ -21,23 +66,143 @@ var startupCacheEventsSent = false;
...
-// navigator.mozHasPendingMessage only returning
-// true to the first call made to it.
-window.htmlCacheRestorePendingMessage = [];
+var startupOnModelLoaded;
 
 (function() {
-  var selfNode = document.querySelector('[data-loadsrc]'),
+  var evt,
+      handlers = {},
+      pendingMessage = null,
```
asutherland: Please add a comment on the intent and use of pendingMessage.  Especially with the way ignoreMessage manipulates this, it's not super obvious and the manipulations are somewhat scattered.

apps/email/js/html_cache_restore.js
```diff
@@ -21,23 +66,143 @@ var startupCacheEventsSent = false;
...
-// Use a global to work around issue with
-// navigator.mozHasPendingMessage only returning
-// true to the first call made to it.
-window.htmlCacheRestorePendingMessage = [];
+var startupOnModelLoaded;
 
 (function() {
-  var selfNode = document.querySelector('[data-loadsrc]'),
+  var evt,
+      handlers = {},
```
asutherland: The markdown has some treatment of this I think, but a brief comment indicating the types of the keys and values I think is helpful.

apps/email/js/html_cache_restore.js
```diff
@@ -21,23 +66,143 @@ var startupCacheEventsSent = false;
...
-
-// Use a global to work around issue with
-// navigator.mozHasPendingMessage only returning
-// true to the first call made to it.
-window.htmlCacheRestorePendingMessage = [];
+var startupOnModelLoaded;
 
 (function() {
-  var selfNode = document.querySelector('[data-loadsrc]'),
+  var evt,
```
asutherland: doc nit: brief comment on evt about it being dynamically loaded on an as-needed basis please.

apps/email/js/html_cache_restore.js
```diff
@@ -21,23 +66,143 @@ var startupCacheEventsSent = false;
...
+      handlers = {},
+      pendingMessage = null,
+      startingUp = true,
+      lastEntryTime = 0,
```
asutherland: Maybe just move lastEntryTime down to just above the one function that uses it?

apps/email/js/html_cache_restore.js
```diff
@@ -21,23 +66,143 @@ var startupCacheEventsSent = false;
...
+      handlerQueues = {
```
asutherland: Docs for handlerQueues too, although it seems like if this is moved closer to the 2 funcs that use it, less comments are required.

apps/email/docs/startup.md
```diff
@@ -0,0 +1,63 @@
+## Email startup flow
```
asutherland: This doc is an awesome addition!  A partner engineer in particular was looking for some docs about the email app startup the other day, and I think they would have found this invaluable.  (And appreciated the cleaned-up flow.)  The actual request for this is partially why I'm requesting the extra comments in the implementation.

One small improvement here I think would be to briefly mention that `model` is responsible for starting up the back-end (and is synonymous with the back-end really) in the "Email startup flow" section.

apps/email/js/html_cache_restore.js
```diff
@@ -102,103 +250,236 @@ window.htmlCacheRestorePendingMessage = [];
...
-    if (navigator.mozHasPendingMessage('activity')) {
-      window.htmlCacheRestorePendingMessage.push('activity');
+  /**
+   * Perform requested activity.
+   *
+   * @param {MozActivityRequestHandler} req activity invocation.
+   */
+  function onActivityRequest(req) {
+    console.log('mozSetMessageHandler: received an activity');
+    if (messageTooFast('activity') || ignoreMessage('activity')) {
```
asutherland: ignoreMessage should come befor messageTooFast, but its lack of caring about its type/pendingMessage is a little concerning.  Specifically, based on the code as written, if we have both 'notification' and 'activity' pending, 'activity' will win.  But onNotification will fire first and its invocation of messageTooFast() will return false, then ignoreMessage will kill it.  Then onActivityRequest will fire and messageTooFast will return true for it and then things maybe get broken for a while until another activity eventually gets past the messageTooFast gauntlet so ignoreMessage() can have its pendingMessage-clearing side-effect.

apps/email/js/html_cache_restore.js
```diff
@@ -102,103 +250,236 @@ window.htmlCacheRestorePendingMessage = [];
...
-    'moz-app-loaded'
-  ].forEach(function(eventName) {
-      window.addEventListener(eventName, function mozPerfLoadHandler() {
-        dispatch(eventName);
-      }, false);
-    }));
+  // mozHasPendingMessage seems like it can only be called once per message
+  // type, so only asking once. These are ordered by least important to most
+  // important. Similar to gateEntry, we only want to process one event on
+  // startup, ignore the others.
```
asutherland: This rationale seems to contradict the rationale in messageTooFast where you suggest 'request-sync' is fine and effectively orthogonal to other-things.  Certainly it does seem undesirable that if the user triggers an activity around the time request-sync wakes us up that we'll ignore the user's actviity.

Specifically, I think:

- It makes sense that only one of notification or activity can win since they both are effectively trying to decide on the state of the UI.  (Arguably we could show the message/inbox under the compose card.  I agree we don't want to do that, though.)

- request-sync should be orthogonal.  In the game of prioritization, I think an argument can be made that request-sync is the least important to deal with promptly, but it should be dealt with.  Given the current pendingMessage flow here, I think the right thing is probably to only set pendingMessage to "request-sync" if none of the prior cases fired.  If they did fire, instead have the triggering of the actual request-sync logic then() on a promise that gets resolved once the UI has reached the user-interactive state.

apps/email/js/html_cache_restore.js
```diff
@@ -102,103 +250,236 @@ window.htmlCacheRestorePendingMessage = [];
...
-    startApp();
-  } else {
-    var parsedResults = retrieve();
+  function onRequestSync(e) {
+    console.log('mozSetMessageHandler: received a request-sync');
+    if (ignoreMessage('request-sync')) {
+      return;
+    }
+
+    dispatch('request-sync', [e]);
```
asutherland: This indirection is a problem at startup because you *must* acquire your wake-locks before you return from the method being called by mozSetMessageHandler.  You can probably just move the wake-lock acquisition in here and then pass it through.

asutherland: (And that's because although the system acquired the wake-locks on our behalf to get us far enough to get to mozSetMessageHandler being invoked for the given message type, it gets rid of them once our method returns.)

apps/email/js/html_cache_restore.js
```diff
@@ -21,23 +66,143 @@ var startupCacheEventsSent = false;
...
+
+    lastEntryTime = entryTime;
+    return false;
+  }
+
+  /**
+   * If waiting for the first startup message, only accept the message if its
+   * type matches the pending expected type.
+   */
+  function ignoreMessage(type) {
```
asutherland: So, it's not the right name, but I think something like "acquireSystemMessageSemaphore" is more appropriate.  Maybe something involving a "gate"?  Basically, I understand this function for being responsible for:

- Telling the other less important message types to get lost because they're being ignored by returning false.
- Letting the message type we decided we do want to process through, and taking down the "get lost" sign on our way in.

It might make sense to introduce a function shouldProcessMessage() that internally uses messageTooFast and ignoreMessage.  As noted elsewhere, it's currently possible to cause problems by mis-ordering ignoreMessage and messageTooFast, so this makes an arguably more clear contract.

apps/email/js/html_cache_restore.js
```diff
@@ -102,103 +250,236 @@ window.htmlCacheRestorePendingMessage = [];
...
+  }
+
+  function onNotification(msg) {
+    console.log('mozSetMessageHandler: received a notification');
+    // Skip notification events that are not from a notification "click". The
+    // system app will also notify this method of any close events for
+    // notifications, which are not at all interesting.
+    if (!msg.clicked) {
+      if (startingUp) {
+        window.close();
```
asutherland: I think this is too aggressive.  The intent here is "if the only reason I woke up was for this stupid close notification, then close the window", but the implementation will close us in other situations too.  Because of the ordering of the calls to mozSetMessageHandler and the fact that I think multiple messages could have been queued even for 'notification', you need to do something like check if after the call to mozSetMessageHandler if pendingMessage === 'notification' and only then trigger the close.

asutherland: (That is, infer from the fact that pendingMessage did not get nulled out that we didn't find a useful 'notification' activity and therefore the 'notification' event was a lie, and since we clearly had nothing better to do, we should now shutdown.  I guess based on my request-sync improvement, though, you'd need to check that flag too.  You could also have a boolean like "doingSomething" and just set that to true anyplace we start something.)

apps/email/js/html_cache_restore.js
```diff
@@ -102,103 +250,236 @@ window.htmlCacheRestorePendingMessage = [];
...
+
+    // The notification infrastructure does not automatically bring the app to
+    // the foreground, so if still hidden, show it now. Ideally this would not
+    // use a setTimeout, but it was getting incorrect document.hidden values on
+    // startup, where just a bit later the value does seem to be set correctly.
+    // The other option was to do this work inside the notification handler in
+    // mail_app, but the delay is long enough waiting for that point that the
+    // user might be concerned they did not correctly tap the notification.
+    setTimeout(function() {
+      if (document.hidden && navigator.mozApps) {
```
asutherland: Given the number of activity-related bug reports we get, I think this would be a good place for a console.log that indicates the decision we made.  Mainly whether we were document.hidden and whether we tried to foreground ourselves.  Especially since it's conceivable that there could be future platform regressions about us foregrounding ourselves without a user-initiated event handler/mozSetMessageHandler call directly on the stack.  (I think various guards may have been lessened and there may be some attribution that follows setTimeout calls and such, but conceivably dynamic script loading might lose the attribution and thereby cause troubles, etc. etc..

try-server-hook:
[jrburke James Burke (jrburke)](https://github.com/jrburke) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=018e4f04dbdf)

apps/email/js/activity_composer_data.js
jrburke: Fixed in comment commit: 7d8f41f19303f833552bc490fe6bf87e1349526a

apps/email/js/config.js
jrburke: Fixed in comment commit: 7d8f41f19303f833552bc490fe6bf87e1349526a. I also spread the comments out to reach the full 80 character width.

apps/email/js/html_cache.js
jrburke: Yeah, a debug one I used earlier. Removed it. Fixed in comment commit: 7d8f41f19303f833552bc490fe6bf87e1349526a

apps/email/js/html_cache_restore.js
jrburke: Good catch. Fixed in comment commit: 7d8f41f19303f833552bc490fe6bf87e1349526a

apps/email/js/html_cache_restore.js
jrburke: I went ahead and converted all the global vars in this file to be referenced via `window.`. Fixed in comment commit: 7d8f41f19303f833552bc490fe6bf87e1349526a

apps/email/js/html_cache_restore.js
jrburke: I just removed the code. It was not that complicated, so not worth taking up space. And not sure yet how the fragment ID work will shake out yet. I did file [bug 1137510](https://bugzilla.mozilla.org/show_bug.cgi?id=1137510) to track the fragment ID work. Removed in commit 7d8f41f19303f833552bc490fe6bf87e1349526a.

apps/email/js/html_cache_restore.js
jrburke: Moved top variables to more local locations and added docs for each one. Fixed in comment commit: 7d8f41f19303f833552bc490fe6bf87e1349526a

apps/email/js/html_cache_restore.js
jrburke: Moved top variables to more local locations and added docs for each one. Fixed in comment commit: 7d8f41f19303f833552bc490fe6bf87e1349526a

apps/email/js/html_cache_restore.js
jrburke: Moved top variables to more local locations and added docs for each one. Fixed in comment commit: 7d8f41f19303f833552bc490fe6bf87e1349526a

apps/email/js/html_cache_restore.js
jrburke: Fixed in comment commit: 7d8f41f19303f833552bc490fe6bf87e1349526a

apps/email/js/html_cache_restore.js
jrburke: Fixed in comment commit: 7d8f41f19303f833552bc490fe6bf87e1349526a

apps/email/docs/startup.md
jrburke: Fixed in comment commit: 7d8f41f19303f833552bc490fe6bf87e1349526a

try-server-hook:
[jrburke James Burke (jrburke)](https://github.com/jrburke) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=4ab5e314fd59)

apps/email/js/html_cache_restore.js
jrburke: Log statement added in d2e9acac7ff25d0c078c024641580dedbdc10fbb

apps/email/js/html_cache_restore.js
jrburke: In d2e9acac7ff25d0c078c024641580dedbdc10fbb, this has been improved with the appDispatchedMessage that indicates if anything has been dispatched, and comments to what might be missed near this code.

jrburke:
Commit d2e9acac7ff25d0c078c024641580dedbdc10fbb should address the other feedback so far:

* the ignoreMessage and entryTime checks could mess up, were badly ordered, and would have been more developer-proof if had a combo method.
* the wake locks that need to be acquired when the mozSetMessageHandler listener is notified.

The main complication was trying to handle mozSetMessageHandler request-syncs in html_cache_restore. This has been moved back to cronsync-main (and the gelam pull request updated to match), since dealing with the wake locks is complicated.

Plus, the policing of pendingMessage in html_cache_restore was really around policing messages that dealt with UI insertions. So I reworked the naming around that, and now that request-sync is not a message handler in html_cache_restore, it did mean that the checking for the pendingMessage and entryTime could be just in one method, named `isUiMessageTypeAllowedEntry`.

I'm open to other names though (and really any feedback). I am thinking of renaming html_cache_restore to something different. Currently favoring gate_lord, but open to other suggestions. Also OK if we keep it the same name now. Other names though just for fun: mr_gates, gatwick, pearly_gates, gandalf (you shall not pass), cache_gate (like watergate!). Maybe startup_cache if you want to be more literal.

asutherland:
Ha! Destiny reference!  And anytime something asynchronous happens, you could have a comment like "please wait here and fight off monsters while I investigate this "l10n" machine left from the golden age."

startup_cache sounds nice and obvious.  You could always put a "// codename: gate_lord" in there.  bikeshed-wise, I really do like the symmetry of html_cache.js/html_cache_restore.js, so if renaming, maybe startup_cache_save.js/startup_cache.js would be a nice pairing.  Or save_startup_cache.js/startup_cache.js.

jrburke:
When I wrote that last night, I had forgotten about html_cache. Thought about it more overnight, and since html_cache is still there, I'll just keep the names the same for now. It is a bit of an uneven pairing now though as html_cache just worries about the cache, where html_cache_restore is doing cache work and some message capturing. That's OK though, with the removal of the request-sync capturing the names are probably fine as-is. So unless you indicate you prefer a renaming I'll just leave it as-is.

try-server-hook:
[jrburke James Burke (jrburke)](https://github.com/jrburke) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=2b1c3d4946a0)

try-server-hook:
[jrburke James Burke (jrburke)](https://github.com/jrburke) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=d65166c6b06e)

try-server-hook:
[jrburke James Burke (jrburke)](https://github.com/jrburke) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=fb8023d70f33)

PR #28531, opened 2015-02-28
Bug 825318 - Handle downloads that have already succeeded at add-time. r=jrburke, r=crdlc

asutherland:
This patch consists of a few parts:
- Changes to download_notification.js (r=crdlc)

The new mozDownloadManager adoptDownload mechanism adds downloads with their
state already set to succeeded.  In that case we want to avoid the "started"
notification being displayed and instead jump directly towards displaying the
"succeeded"/completed notification.  Additionally, we want to process the
completion so the download is added to the data store.
- Changes to the email app (r=jrburke) (Now with test fixes!)

Improvements to MIME-type derivation logic.

Register all downloads with the download manager.  This requires the
"downloads" permission and a version of mozDownloadManager that implements
adoptDownload.

Display an error message if the user attempts to view/open a downloaded
attachment and there is no eligible activity.

NOTE: the gaia-email-libs-and-more parts of this patch are being layered on
top in a separate commit for ease of reviewing, etc.
- Make the PDF viewer not break itself it loses a race. (r=jrburke)

It's not okay to access the DOM before it has been populated, so we use defer.
And we use defer on all script nodes for consistency and to avoid any
unpleasant surprises from the webapp optimizer logic, should it start running
on the file.  (Currently pdfjs is blacklisted at an l10n level which also
defeats the other general _optimize steps.)

See https://bugzilla.mozilla.org/show_bug.cgi?id=1071449#c18 for more info.

### DIFF
--- a/apps/email/js/cards/compose.js
+++ b/apps/email/js/cards/compose.js
@@ -1013,7 +1013,8 @@ return [
         var activity = new MozActivity({
           name: 'pick',
           data: {
-            type: ['image/*', 'video/*', 'audio/*', 'application/*'],
+            type: ['image/*', 'video/*', 'audio/*', 'application/*',
+                   'text/vcard'],
             nocrop: true
           }
         });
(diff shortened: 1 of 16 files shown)


try-server-hook:
[asutherland Andrew Sutherland (asutherland)](https://github.com/asutherland) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=f1c724d31374)

PR #28537, opened 2015-02-28
Bug 1068224 - Limit the amount of calls shown in the call info view

yvtheja:
...ug Sherk

### DIFF
--- a/apps/communications/dialer/elements/call-info-view.html
+++ b/apps/communications/dialer/elements/call-info-view.html
@@ -14,6 +14,8 @@
 
       <div id="call-durations">
       </div>
+
+      <a id="show-more" data-l10n-id="showMore"></a>
     </div>
 
     <div class="ci__contact-wrapper">
--- a/apps/communications/dialer/js/call_info.js
+++ b/apps/communications/dialer/js/call_info.js
@@ -67,8 +67,7 @@
   }
 
   function updateCallDurations(group) {
-    var countToBeDisplayed;
-    var callInfoCount = 0;
+    var groupCalls;
     var callDurationsElt = document.getElementById('call-durations');
     callDurationsElt.innerHTML = '';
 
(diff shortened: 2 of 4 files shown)


try-server-hook:
[ythej Vishnu Teja (ythej)](https://github.com/ythej) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=b4de71fff69f)

try-server-hook:
[ythej Vishnu Teja (ythej)](https://github.com/ythej) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=9984fad08ab0)

apps/communications/dialer/js/call_info.js
```diff
@@ -68,34 +77,44 @@
       return;
     }
 
+    if(_showMoreButtonClicked == false) {
+      countToBeDisplayed = 9;
+    } else if (_showMoreButtonClicked == true ) {
+      // Hide the showMoreButton once it is clicked.
+      showMoreButton.classList.add('hide-show-more');
+      countToBeDisplayed = 100;
```
DouglasSherk: This is an unnecessary restriction. We should just unlimit the count displayed.

apps/communications/dialer/js/call_info.js
```diff
@@ -68,34 +77,44 @@
       return;
     }
 
+    if(_showMoreButtonClicked == false) {
```
DouglasSherk:
```js
if (!_showMoreButtonClicked) {
  // ...
}
```

apps/communications/dialer/js/call_info.js
```diff
@@ -68,34 +77,44 @@
...
+      countToBeDisplayed = 9;
+    } else if (_showMoreButtonClicked == true ) {
+      // Hide the showMoreButton once it is clicked.
+      showMoreButton.classList.add('hide-show-more');
```
DouglasSherk: You don't need this class. We could just use something like this instead:
```css
/* Only select the .show-more-button if it has more than 11 sibling <li> elements */
.show-more-button ~ li:nth-child(11) ~ .show-more-button {
  display: none;
}
```

apps/communications/dialer/js/call_info.js
```diff
@@ -68,34 +77,44 @@
...
-      startTime.dataset.date = call.date;
-      startTime.textContent = Utils.prettyDate(call.date);
-
-      var durationElt = document.createElement('p');
-      durationElt.classList.add('cd__duration');
-      navigator.mozL10n.once(function() {
-        if (call.duration === 0) {
-          if (group.type === 'incoming') {
-            durationElt.setAttribute('data-l10n-id', 'info-missed');
+      if(callInfoCount < countToBeDisplayed) {
```
DouglasSherk: Use ```slice()``` instead to only select the part of ```group.calls``` that we want.

apps/communications/dialer/js/call_info.js
ythej: Those are not list though. They are div's. Doing this thing with div's will be hacky I think ?

apps/communications/dialer/js/call_info.js
ythej: yeah, I agree, Little confused seeing your comment on the bug :)
Thank you.

apps/communications/dialer/js/call_info.js
DouglasSherk: I don't think so, if it's scoped to something concrete like ```.show-more-button```.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0b8b6f721da234ccd5c0354dc93f5909d21d2b3f

apps/communications/dialer/elements/call-info-view.html
```diff
@@ -14,6 +14,8 @@ <h1 id="call-info-title"></h1>
 
       <div id="call-durations">
       </div>
+
+      <a id="show-more" data-l10n-id="showMore"></a>
```
DouglasSherk: You will have to move this into ```<div id="call-durations"></div>```. Then you can use this selector to match the case I was talking about before:
```css
#show-more ~ li:nth-child(11) ~ #show-more {
  display: none;
}
```

ythej: @DouglasSherk , We will not have to display the "show more" button if there are less than 9 rows and if we apply the above method, I think it's not possible ?

PR #28557, opened 2015-03-02
Bug 1138343 - [UserDictionary, KBSettingsApp] Instantiate and store an ArrayBuffer instead of Uint8Array during dictionary blob generation

mnjul:
Unit test also explicitly checks for Converter.toBlob()'s return object's type.

### DIFF
--- a/apps/keyboard/js/settings/word_list_converter.js
+++ b/apps/keyboard/js/settings/word_list_converter.js
@@ -405,7 +405,7 @@ TSTBlobBuilder.prototype.toBlobArray = function() {
     this._emitNode(node);
   }, this);
 
-  return this._output;
+  return this._output.buffer;
 };
 
 
--- a/apps/keyboard/test/unit/settings/word_list_converter_test.js
+++ b/apps/keyboard/test/unit/settings/word_list_converter_test.js
@@ -40,9 +40,13 @@ suite('WordListConverter', function() {
     assert.equal(actual.length, expected.length, 'length of two blobs differs');
   };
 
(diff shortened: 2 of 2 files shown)


try-server-hook:
[mnjul Min-Zhong "John" Lu (mnjul)](https://github.com/mnjul) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=fcbdedbecea2)

apps/keyboard/js/settings/word_list_converter.js
```diff
@@ -405,7 +407,7 @@ TSTBlobBuilder.prototype.toBlobArray = function() {
     this._emitNode(node);
   }, this);
 
-  return this._output;
+  return this._outputBuffer;
```
timdream: I wonder if `this._output.buffer` is a better pointer to the buffer instance instead of yet another private property? But that's not really important.

try-server-hook:
[mnjul Min-Zhong "John" Lu (mnjul)](https://github.com/mnjul) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=8c2dcfd48f3d)

apps/keyboard/js/settings/word_list_converter.js
mnjul: Actually...I can still just instantiate a Uint8Array without explicitly instantiating an ArrayBuffer and just return ``this._output.buffer`` in ``toBlobArray()``. This approach is *much* cleaner and eliminates most code changes.

PR #28568, opened 2015-03-02
Bug 1136791 - test_marketplace_packaged_app times out after opening Mark...

mwargers:
...etplace, unable to search


try-server-hook:
[mwargers Martijn (mwargers)](https://github.com/mwargers) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=b1c13ddbf526)

npark-mozilla:
Ah, I think this is caused by the phone performance, I did notice it takes time to load the search prompt. (and generally laggy when using search) What if you increase the timeout value of the wait() call that fails?  we could increase it by 10 seconds like: 
         search_box = Wait(self.marionette,timeout = 20).until(
            expected.element_present(*self._search_locator))

npark-mozilla:
since the wait() doesn't work, sleep() should do it for now.

PR #28609, opened 2015-03-03
Bug 1139072 - [System] Some notifications tests fail when the test-agent...

julienw:
... runs in a backround tab r=mhenretty

### DIFF
--- a/apps/system/test/unit/notifications_test.js
+++ b/apps/system/test/unit/notifications_test.js
@@ -50,6 +50,7 @@ suite('system/NotificationScreen >', function() {
     fakeNotifContainer;
   var fakePriorityNotifContainer, fakeOtherNotifContainer;
   var realVersionHelper, realMozL10n;
+  var isDocumentHidden;
 
   function sendChromeNotificationEvent(detail) {
     var event = new CustomEvent('mozChromeNotificationEvent', {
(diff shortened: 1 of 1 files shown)


try-server-hook:
[julienw Julien Wajsberg (julienw)](https://github.com/julienw) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=9530ae2ce9e0)

apps/system/test/unit/notifications_test.js
```diff
@@ -144,6 +145,12 @@ suite('system/NotificationScreen >', function() {
     };
     navigator.mozL10n = MockL10n;
 
+    isDocumentHidden = false;
+    Object.defineProperty(document, 'hidden', {
+      configurable: true,
+      get: () => isDocumentHidden
```
mikehenrty: Should we do `set: (val) => isDocumentHidden = val` too for future proofing and consistency?

julienw: actually I made this on purpose: the real one does not change anything on set, and really we should not let anybody to change it anywhere, it would be a bug. So I think it's better to not have a set.

Tell me what you think :)

julienw: ping @mikehenrty

mikehenrty: Hmmm, then why even have a isDocumentHidden variable?

mikehenrty: In any case, I won't block the review on this :) Merging...

julienw: > Hmmm, then why even have a isDocumentHidden variable?

You can change the variable from a test (so that you can test other branches in your app code), but not from the app code.

If we have a setter, the app code could change it.

That's 2 very different ways to change the value.

PR #28626, opened 2015-03-04
Bug 1139336 - Implement EventEmitter r=eragonj

crh0716:
https://bugzilla.mozilla.org/show_bug.cgi?id=1139336

### DIFF
--- a/apps/settings/js/modules/base/event_emitter.js
+++ b/apps/settings/js/modules/base/event_emitter.js
@@ -1,0 +1,137 @@
+define(function(require) {
+  'use strict';
+
+  var Module = require('modules/base/module');
+
+  /**
+   * EventEmitter implements basic event emitting logics. It also provides
+   * function that augments the ability to other objects.
+   *
+   * @example
+   *   var eventEmitter = EventEmitter(['event1', 'event2']);
+   *   var eventHandler = function(event) {
+   *     console.log(event.detail);
+   *   };
+   *   var obj = {
+   *     handleEvent: function(event) {
+   *       console.log(event.detail);
+   *     }
+   *   };
+   *
+   *   eventEmitter.addEventListener('event1', eventHandler);
+   *   eventEmitter.addEventListener('event2', obj);
+   *
+   *   eventEmitter._emitEvent('event1', 'value1'); // value1
+   *   eventEmitter._emitEvent('event2', 'value2'); // value2
+   *
+   * Extending from EventEmitter:
+   * @example
+   *   var NewModule = Module.create(function() {
+   *     this.super(EventEmitter).call(this, ['event1', 'event2']);
+   *   }).extend(EventEmitter);
+   *
+   *   var module = NewModule();
+   *
+   * @class EventEmitter
+   * @returns {EventEmitter}
+   */
+  var EventEmitter = Module.create(function EventEmitter(eventNames) {
+    if (eventNames && eventNames.length) {
+      this._eventListeners = eventNames.reduce((result, eventName) => {
+        result[eventName] = [];
+        return result;
+      }, {});
+    } else {
+      this.throw('no valid registered events');
+    }
+  });
+
+  /**
+   * Emit an event of a specific type and a given value.
+   *
+   * @access private
+   * @memberOf EventEmitter.prototype
+   * @param {String} eventName
+   * @param {Object} value
+   */
+  EventEmitter.prototype._emitEvent = function(eventName, value) {
+    var listeners = this._eventListeners[eventName];
+    if (!listeners) {
+      this.throw('invalid event name: ' + eventName);
+    }
+
+    this.debug('_emitEvent:' + eventName + ' ' + value);
+    var eventObj = {
+      type: eventName,
+      detail: value
+    };
+    listeners.forEach((listener) => {
+      listener.call(this, eventObj);
+    });
+  };
+
+  /**
+   * Add an event listener on an event. If listener is an object,
+   * object.handleEvent is called when the event emits.
+   *
+   * @access public
+   * @memberOf EventEmitter.prototype
+   * @param {String} eventName
+   * @param {(Function|Object)} listener
+   */
+  EventEmitter.prototype.addEventListener = function(eventName, listener) {
+    var listeners = this._eventListeners[eventName];
+    if (listener && listeners) {
+      switch (typeof listener) {
+        case 'function':
+          break;
+        case 'object':
+          if (typeof listener.handleEvent === 'function') {
+            listener = listener.handleEvent;
+          } else {
+            listener = null;
+          }
+          break;
+        default:
+          listener = null;
+          break;
+      }
+      if (listener && listeners.indexOf(listener) < 0) {
+        listeners.push(listener);
+      } 
+    } else {
+      this.error('addEventListener: invalid listener for ' + eventName);
+    }
+  };
+
+  /**
+   * Remove the event listener from an event.
+   *
+   * @access public
+   * @memberOf EventEmitter.prototype
+   * @param {String} eventName
+   * @param {(Function|Object)} listener
+   */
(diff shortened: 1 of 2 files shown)


try-server-hook:
[crh0716 Arthur Chen (crh0716)](https://github.com/crh0716) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=ea9eac40eca4)

apps/settings/js/modules/base/event_emitter.js
```diff
@@ -0,0 +1,121 @@
...
+   */
+  EventEmitter.prototype.addEventListener = function(eventName, listener) {
+    var listeners = this._eventListeners[eventName];
+    if (listener && listeners && listeners.indexOf(listener) < 0) {
+      var type = typeof listener;
+      switch(type) {
+        case 'function':
+          listeners.push(listener);
+          break;
+        case 'object':
```
EragonJ: The `Object` case is for the case that we want to addEventListener on `this` object right ?

crh0716: yes

apps/settings/js/modules/base/event_emitter.js
```diff
@@ -0,0 +1,121 @@
...
+   * @memberOf EventEmitter.prototype
+   * @param {String} eventName
+   * @param {(Function|Object)} listener
+   */
+  EventEmitter.prototype.removeEventListener = function(eventName, listener) {
+    var listeners = this._eventListeners[eventName];
+    if (listener && listeners) {
+      var index = -1;
+      var type = typeof listener;
+      switch(type) {
```
EragonJ: one more space

apps/settings/js/modules/base/event_emitter.js
```diff
@@ -0,0 +1,121 @@
...
+   * @access public
+   * @memberOf EventEmitter.prototype
+   * @param {String} eventName
+   * @param {(Function|Object)} listener
+   */
+  EventEmitter.prototype.addEventListener = function(eventName, listener) {
+    var listeners = this._eventListeners[eventName];
+    if (listener && listeners && listeners.indexOf(listener) < 0) {
+      var type = typeof listener;
+      switch(type) {
```
EragonJ: one more space

apps/settings/js/modules/base/event_emitter.js
```diff
@@ -0,0 +1,121 @@
+/**
+ * EventEmitter implements basic event emitting logics. It also provides
+ * function that augments the ability to other objects. For example:
+ *  
+ *   function Module() {
+ *     this.EventEmitter.call(this, [eventName1, eventName2]);
```
EragonJ: This line makes me confused. Why we have to predefined `eventName1` & `eventName2` before using ? Without this line, `this._eventListeners` will never be `{}` and would break other functionalities, so I can understand that you need this line to initialize it, but I think we should directly make `this._eventListeners = {}` when this `Object` is augmented.

apps/settings/js/modules/base/event_emitter.js
```diff
@@ -0,0 +1,121 @@
...
+define(function() {
+  'use strict';
+
+  /**
+   * @class EventEmitter
+   * @returns {EventEmitter}
+   */
+  function EventEmitter(eventNames) {
+    if (eventNames && eventNames.length) {
+      this._eventListeners = {};
```
EragonJ: Here

apps/settings/js/modules/base/event_emitter.js
```diff
@@ -0,0 +1,121 @@
...
+
+  /**
+   * Emit a event of a specific type and a given value.
+   *
+   * @access private
+   * @memberOf EventEmitter.prototype
+   * @param {String} eventName
+   * @param {Object} value
+   */
+  EventEmitter.prototype._emitEvent = function(eventName, value) {
```
EragonJ: Why this method is `private` while the others are `public` ? If we got an object which is augmented by `EventListener`, for the object, `emitEvent`, `addEventListener` and `removeEventListener` should have the same privileges no matter `public` or `private`. For my personal opinion, they should all be `public`. WDYT ?

apps/settings/js/modules/base/event_emitter.js
crh0716: This line is considered as calling to the constructor of `EventEmitter`. If the constructor is not called correctly, the emitter won't work. Just like prototypical inheritance.

apps/settings/js/modules/base/event_emitter.js
crh0716: That's why we throw an error when the constructor is not called correctly.

apps/settings/js/modules/base/event_emitter.js
crh0716: The module augmented by `EventEmitter` is considered as `EventEmitter` itself. So if `_emitEvent` event is not exposed, it is private.

try-server-hook:
[crh0716 Arthur Chen (crh0716)](https://github.com/crh0716) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=e72c20df481f)

apps/settings/js/modules/base/event_emitter.js
```diff
@@ -0,0 +1,127 @@
...
+   * object.handleEvent is called when the event emits.
+   *
+   * @access public
+   * @memberOf EventEmitter.prototype
+   * @param {String} eventName
+   * @param {(Function|Object)} listener
+   */
+  EventEmitter.prototype.addEventListener = function(eventName, listener) {
+    var listeners = this._eventListeners[eventName];
+    if (listener && listeners && listeners.indexOf(listener) < 0) {
```
EragonJ: If we are going to mimic the natural EventEmitter, we should not check `listeners.indexOf(listener) < 0` here.

apps/settings/test/unit/modules/base/event_emitter_test.js
```diff
@@ -0,0 +1,84 @@
...
+        var eventHandler = {
+          handleEvent: sinon.spy()
+        };
+        this.emitter.addEventListener('event1', eventHandler);
+        this.emitter._emitEvent('event1', 'value1');
+        sinon.assert.calledWith(eventHandler.handleEvent, sinon.match({ 
+          type: 'event1',
+          detail: 'value1'
+        }));
+      });
```
EragonJ: Can we add one more test case if we try to addEventListener twice with the same functions ?

apps/settings/js/modules/base/event_emitter.js
```diff
@@ -0,0 +1,127 @@
...
+   *     console.log(event.detail);
+   *   };
+   *   var obj = {
+   *     handleEvent: function(event) {
+   *       console.log(event.detail);
+   *     }
+   *   };
+   *
+   *   module.addEventListener('event1', eventHandler);
+   *   module.addEventListener('event2', handleEvent);
```
EragonJ: this should be `module.addEventListener('event2', obj);` instead.

apps/settings/js/modules/base/event_emitter.js
crh0716: Per the offline discussion the event types should be predefined instead of run-time determined, so we still need the check.

apps/settings/test/unit/modules/base/event_emitter_test.js
crh0716: Indeed, we should prevent a callback from being added multiple times.

apps/settings/js/modules/base/event_emitter.js
crh0716: good catch!

apps/settings/test/unit/modules/base/event_emitter_test.js
EragonJ: :+1:

PR #28628, opened 2015-03-04
Bug 1139239 - Fix yahoo search url

daleharvey:


apps/search/test/marionette/lib/search.js
```diff
@@ -124,6 +126,12 @@ Search.prototype = {
     this.client.helper.waitForElement('body');
   },
 
+  searchDataVersion: function() {
+    var file = path.resolve(__dirname +
```
KevinGrandon: Since we already have two places we have to bump, might as well hardcode the version here so we can add a comment?

daleharvey: Heh I was trying to fix this another way first which is why I ended up parsing json, but thats far simpler, thanks

try-server-hook:
[daleharvey Dale  Harvey (daleharvey)](https://github.com/daleharvey) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=cc6644e7e036)

PR #28629, opened 2015-03-04
Bug 1139338 - Refactor Observable and implement read-only and dependency property r=eragonj

crh0716:
https://bugzilla.mozilla.org/show_bug.cgi?id=1139338

### DIFF
--- a/apps/settings/js/modules/base/dependency_graph.js
+++ b/apps/settings/js/modules/base/dependency_graph.js
@@ -1,0 +1,134 @@
+define(function(require) {
+  'use strict';
+
+  var Module = require('modules/base/module');
+
+  /**
+   * DependencyGraph is a basic data structure that helps decide the evaluation
+   * order of a certain node.
+   *
+   * @class DependencyGraph
+   * @requires module:modules/base/module
+   * @params {DependencyGraph} dpGraph
+   *                           The newly created graph will be initialized using
+   *                           dpGraph.
+   * @returns {DependencyGraph}
+   */
+  var DependencyGraph = Module.create(function DependencyGraph(dpGraph) {
+    this._nodes = dpGraph ? JSON.parse(JSON.stringify(dpGraph._nodes)) : {};
+  });
+
+  /**
+   * Print the content of the nodes for debugging.
+   *
+   * @access private
+   * @memberOf DependencyGraph.prototype
+   */
+  DependencyGraph.prototype._printNodes = function dg_printNodes() {
+    Object.keys(this._nodes).forEach((name) => {
+      var node = this._nodes[name];
+      console.log('name: ' + name);
+      console.log('children: ' +
+      node.children.reduce((result, name) => {
+        return result + ' ' + name;
+      }, ''));
+      console.log('dependent nodes: ' +
+      node.dependentNodes.reduce((result, name) => {
+        return result + ' ' + name;
+      }, ''));
+      console.log('======================');
+    });
+  };
+
+  /**
+   * Traverse the tree and add the dependent nodes to all nodes in the path.
+   *
+   * @access private
+   * @memberOf DependencyGraph.prototype
+   * @param {Object} node
+   *                 The node that we have new dependent nodes to add to.
+   * @param {Array} dependentNodes
+   *                New dependent nodes to add.
+   */
+  DependencyGraph.prototype._addDependentNodes =
+    function dg_addDependentNodes(node, dependentNodes, traversedNodes) {
+      traversedNodes = traversedNodes || [];
+      if (traversedNodes.indexOf(node.name) >= 0) {
+        traversedNodes.push(node.name);
+        var error = traversedNodes.join(' -> ');
+        this.throw('circular dependency detected! ' + error);
+      }
+
+      if (dependentNodes) {
+        // add all nodes in "dependendNodes" to the current node
+        dependentNodes.forEach((name) => {
+          if (node.dependentNodes.indexOf(name) < 0) {
+            node.dependentNodes.push(name);
+          }
+        });
+      } else {
+        // Nothing to add, prepare the nodes to be added to the children.
+        dependentNodes = Array.prototype.slice.call(node.dependentNodes);
+      }
+
+      dependentNodes.push(node.name);
+      traversedNodes.push(node.name);
+      node.children.forEach((name) => { 
+        this._addDependentNodes(this._nodes[name], dependentNodes,
+          traversedNodes);
+      });
+      dependentNodes.pop();
+      traversedNodes.pop();
+  };
+
+  /**
+   * Add dependency. Calling to this function adds a dependency from the first
+   * parameter towards the second one.
+   *
+   * @access public
+   * @memberOf DependencyGraph.prototype
+   * @param {String} name1
+   *                 The name that depends on the other.
+   * @param {String} name2
+   *                 The name that is being depended on.
+   */
+  DependencyGraph.prototype.addDependency = function dg_addDep(name1, name2) {
+    var node1 = this._nodes[name1];
+    var node2 = this._nodes[name2];
+
+    if (!node1) {
+      node1 = this._nodes[name1] = {
+        name: name1,
+        children: [],
+        dependentNodes: []
+      };
+    }
+
+    if (!node2) {
+      node2 = this._nodes[name2] = {
+        name: name2,
+        children: [],
+        dependentNodes: []
+      };
+    }
+
+    if (node1.children.indexOf(name2) < 0) {
+      node1.children.push(name2);
+      this._addDependentNodes(node1);
+    }
+  };
+
+  /**
+   * Get all dependent names of a name.
+   *
+   * @access public
+   * @memberOf DependencyGraph.prototype
+   * @param {String} name
+   */
+  DependencyGraph.prototype.getAllDependent = function dg_getAllDep(name) {
+    var node = this._nodes[name];
+    return node && node.dependentNodes;
+  };
+
+  return DependencyGraph;
+});
--- a/apps/settings/js/modules/base/module.js
+++ b/apps/settings/js/modules/base/module.js
@@ -1,0 +1,152 @@
+/**
+ * Module provides methods to create a new module. Any module created here has
+ * the ability of being extended by other modules. Existing properties are
+ * overridden.
+ *
+ * Creating a module
+ * @example
+ *   var NewModule = Module.create(function() {
+ *     this.value = 100;
+ *   });
+ *   NewModule.prototype.print = function() { console.log(this.value); };
+ *
+ *   var instance = NewModule();
+ *   instance.print(); // 100
+ *
+ * Extending a module
+ * @example
+ *   var AnotherModule = Module.create();
+ *   AnotherModule.prototype.inc = function() { this.value++; };
+ *   var ExtendingModule = NewModule.extend(AnotherModule);
+ *
+ *   var instance = ExtendingModule();
+ *   instance.inc();
+ *   instance.print(); // 101
+ *
+ * @module modules/base/module
+ */
+define(function() {
+  'use strict';
+
+  const LOG_LEVEL = {
+    NONE: 0,
+    DEBUG: 1,
+    INFO: 2,
+    WARN: 3,
+    ERROR: 4,
+    ALL: 5
+  };
+
+  var _constructorMap = (function() {
+    var _ctorMap = new Map();
+    var _getConstructor = function(Module) {
+      return _ctorMap.get(Module) || function() {};
+    };
+    var _registerConstructor = function(Module, constructor) {
+      _ctorMap.set(Module, constructor || function() {});
+    };
+    return {
+      getConstructor: _getConstructor,
+      registerConstructor: _registerConstructor
+    };
+  })();
+
+  var _emptyFunction = function() {};
+  var _createLogger = function(name) {
+    switch (name) {
+      case 'DEBUG':
+        return function(msg) {
+          console.log(this._msgPrefix + msg);
+        };
+      case 'INFO':
+        return function(msg) {
+          console.info(this._msgPrefix + msg);
+        };
+      case 'WARN':
+        return function(msg) {
+          console.warn(this._msgPrefix + msg);
+        };
+      case 'ERROR':
+        return function(msg) {
+          console.error(this._msgPrefix + msg);
+        };
+    }
+  };
+
+  var ModulePrototype = {
+    get _msgPrefix() {
+      return '[' + this.$name + ']: ';
+    },
+    debug: _emptyFunction,
+    info: _emptyFunction,
+    warn: _emptyFunction,
+    error: _emptyFunction,
+    throw: function(msg) {
+      throw new Error(this._msgPrefix + msg);
+    },
+    set _logLevel(value) {
+      Object.keys(LOG_LEVEL).forEach((name) => {
+        var level = LOG_LEVEL[name];
+        if (value >= level && value > 0) {
+          this[name.toLowerCase()] = _createLogger(name);
+        }
+      });
+    },
+    super: _constructorMap.getConstructor
+  };
+
+  var _extend = function() {
+    switch (arguments.length) {
+      case 0:
+        break;
+      case 1:
+        var Module = arguments[0];
+        for (var prop in Module.prototype) {
+          if (prop === '$name') {
+            continue;
+          }
+          var pd = Object.getOwnPropertyDescriptor(Module.prototype, prop);
+          if (pd) {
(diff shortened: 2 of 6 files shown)


try-server-hook:
[crh0716 Arthur Chen (crh0716)](https://github.com/crh0716) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=5f4bd8ede1f1)

apps/settings/js/modules/mvvm/observable.js
```diff
@@ -1,102 +1,219 @@
...
+      } else {
+        _defineObservableProperty(this, name, {
+          value: obj[name]
+        });
+      }
+    }
+  };
+
+  Observable.prototype._notify = function o__notify(name, newValue, oldValue) {
+    var observers = !!this._observers && this._observers[name];
```
EragonJ: why we have to use `!!` here ? can't understand.

apps/settings/js/modules/mvvm/observable.js
```diff
@@ -1,102 +1,219 @@
...
+        observer(newValue, oldValue);
+      });
+    }
+  };
+
+  Observable.prototype._removeObserver =
+    function o__removeObserver(observer, name) {
+      // arguments in reverse order to support .bind(observer) for the
+      // unbind from all case
+      var observers = !!this._observers && this._observers[name];
```
EragonJ: same here.

apps/settings/js/modules/mvvm/observable.js
```diff
@@ -1,102 +1,219 @@
...
-        handlers.splice(index, 1);
+        observers.splice(index, 1);
       }
+  };
+
+  Observable.prototype.observe = function o_observe(name, observer) {
+    var propertyInfo =
+      this._observablePropertiesInfo && this._observablePropertiesInfo[name];
+    if (!propertyInfo) {
+      throw new Error('Observable: property ' + name + ' does not exist');
```
EragonJ: I think we can directly `console.log` the error here instead of throwing them away directly because by doing so, we can still make our code workable even if one day someone accidentally remove the `observablePropertiesInfor[name]`

apps/settings/js/modules/mvvm/observable.js
```diff
@@ -1,102 +1,219 @@
...
+    }
+
+    if (typeof observer !== 'function') {
+      return;
+    }
```
EragonJ: I think this one can be moved to the top because without `observer`, we don't need to do all the stuffs below.

apps/settings/js/modules/mvvm/observable.js
```diff
@@ -1,102 +1,219 @@
...
+
+    if (!this._observers) {
+      this._observers = {};
+    }
+    var observers = this._observers[name];
+    if (typeof observers === 'undefined') {
+      observers = this._observers[name] = [];
     }
+    observers.push(observer);
```
EragonJ: I think we can just use `this._observers[name]` directly instead of creating another `observers` variable.

apps/settings/js/modules/mvvm/observable.js
```diff
@@ -1,102 +1,219 @@
...
-      // (prop, handler) -- remove handler from the specific prop
-      removeHandler(handler, prop);
-    } else if (prop in _eventHandlers) {
-      // (prop) -- otherwise remove all handlers for property
-      _eventHandlers[prop] = [];
+  Observable.prototype.unobserve = function o_unobserve(name, observer) {
+    if (typeof name === 'function') {
+      // (observer) -- remove from every key in _observers
+      Object.keys(this._observers).forEach(
+        this._removeObserver.bind(this, name));
```
EragonJ: I am kinda curious the real use case for this. Does this interface make sense ?

apps/settings/js/modules/mvvm/observable.js
```diff
@@ -1,102 +1,219 @@
...
+    } else {
+      var propertyInfo =
+      this._observablePropertiesInfo && this._observablePropertiesInfo[name];
+      if (!propertyInfo) {
+        throw new Error('Observable: property ' + name + ' does not exist');
```
EragonJ: Same question here, maybe we can just console and return directly ? This may happen sometimes if developers don't notice this and I think this is not a big deal and can still make the program work.

try-server-hook:
[crh0716 Arthur Chen (crh0716)](https://github.com/crh0716) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=bfef2a15c85c)

try-server-hook:
[crh0716 Arthur Chen (crh0716)](https://github.com/crh0716) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=9dd590e871b8)

apps/settings/js/modules/mvvm/observable.js
crh0716: This is useful when destroying an object. (we want to remove every dependency to the object so it can be recycled correctly.)

try-server-hook:
[crh0716 Arthur Chen (crh0716)](https://github.com/crh0716) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=fd88811c9442)

apps/settings/js/modules/base/module.js
```diff
@@ -0,0 +1,150 @@
...
+    return {
+      getConstructor: _getConstructor,
+      registerConstructor: _registerConstructor
+    };
+  })();
+
+  var _emptyFunction = function() {};
+  var _createLogger = function(name) {
+    switch (name) {
+      case 'TRACE':
```
EragonJ: This may be wrong, please check : https://developer.mozilla.org/en-US/docs/Web/API/console#Stack_traces

apps/settings/js/modules/base/module.js
```diff
@@ -0,0 +1,150 @@
...
+      case 'DEBUG':
+      case 'INFO':
```
EragonJ: We have console.info by the way https://developer.mozilla.org/en-US/docs/Web/API/Console/info

apps/settings/js/modules/base/module.js
```diff
@@ -0,0 +1,150 @@
...
+ *   var instance = ExtendingModule();
+ *   instance.inc();
+ *   instance.print(); // 101
+ *
+ * @module modules/base/module
+ */
+define(function() {
+  'use strict';
+
+  var LOG_LEVEL = {
```
EragonJ: `const LOG_LEVEL` instead.

apps/settings/js/modules/base/module.js
```diff
@@ -0,0 +1,150 @@
...
+    throw: function(msg) {
+      throw new Error(this._msgPrefix + msg);
+    },
+    set _logLevel(value) {
+      Object.keys(LOG_LEVEL).forEach((name) => {
+        var level = LOG_LEVEL[name];
+        if (value >= level && value > 0) {
+          this[name.toLowerCase()] = _createLogger(name);
+        } else {
+          this[name.toLowerCase()] = _emptyFunction;
```
EragonJ: Because we already set all these helper function to "_emptyFunction" at first, I don't think it is necessary to iterate all of them and set the others to "_emptyFunctions" again. Because normally, this is debug used code, we won't dynamically change the log level at runtime.

apps/settings/js/modules/base/module.js
```diff
@@ -0,0 +1,150 @@
...
+    super: _constructorMap.getConstructor
+  };
+
+  var _extend = function() {
+    switch (arguments.length) {
+      case 0:
+        break;
+      case 1:
+        var Module = arguments[0];
+        for (var i in Module.prototype) {
```
EragonJ: `i` is too vague here for people to understand what is going to be iterated. Can we use something straightforward ? `eachProperty`, `prop`, ... for example.

apps/settings/js/modules/base/module.js
```diff
@@ -0,0 +1,150 @@
...
+  function _create(constructor) {
+    var ModuleFunc = function() {};
+    ModuleFunc.prototype = Object.create(ModulePrototype);
+    ModuleFunc.prototype.$name = constructor && constructor.name || '';
+
+    var Module = function() {
+      var instance = new ModuleFunc();
+      if (typeof constructor === 'function') {
+        constructor.apply(instance, arguments);
+      }
```
EragonJ: if constructor is not a `function`, should we throw out some error message for developers here ?

apps/settings/js/modules/base/module.js
```diff
@@ -0,0 +1,150 @@
...
+    return this;
+  };
+
+  // Returns a module function. The function returns an instance of the
+  // module when it is called. The passed constructor is called using the
+  // instance as "this".
+  function _create(constructor) {
+    var ModuleFunc = function() {};
+    ModuleFunc.prototype = Object.create(ModulePrototype);
+    ModuleFunc.prototype.$name = constructor && constructor.name || '';
```
EragonJ: This means that developers have to explicitly define the constructor function with a name, otherwise its name would always be `""`. They can override `xxx.$name` later of course when designing module, but I am not sure whether the `$` would be easy to understand or not.

apps/settings/js/modules/mvvm/observable.js
```diff
@@ -1,102 +1,353 @@
...
-        if (oldValue !== value) {
-          _observable['_' + p] = value;
-          var handlers = _eventHandlers[p];
-          handlers.forEach(function(handler) {
-            handler(value, oldValue);
+        // caches the old values of all dependent
+        if (dependentList && dependentList.length) {
+          oldValues = {};
+          dependentList.forEach((name) => {
+            oldValues[name] = this[name];
```
EragonJ: I think this block is quite isolated enough that can be extracted as another function.

apps/settings/js/modules/base/dependency_graph.js
```diff
@@ -0,0 +1,134 @@
...
+   * @param {Array} dependentNodes
+   *                New dependent nodes to add.
+   */
+  DependencyGraph.prototype._addDependentNodes =
+    function dg_addDependentNodes(node, dependentNodes, traversedNodes) {
+      traversedNodes = traversedNodes || [];
+      if (traversedNodes.indexOf(node.name) >= 0) {
+        traversedNodes.push(node.name);
+        var error = traversedNodes.join(' -> ');
+        this.Error('circular dependency detected! ' + error);
```
EragonJ: we should return early here.

apps/settings/js/modules/mvvm/observable.js
```diff
@@ -1,102 +1,353 @@
...
+   * @access public
+   * @memberOf Observable.prototype
+   * @param {String} name
+   * @param {Function} observer
+   */
+  Observable.prototype.observe = function o_observe(name, observer) {
+    if (typeof observer !== 'function') {
+      return;
+    }
+    (this._observers[name] = this._observers[name] || []).push(observer);
```
EragonJ: this code is cool

apps/settings/test/unit/modules/base/module_test.js
```diff
@@ -0,0 +1,216 @@
...
+      var constructor = function FakeModule() {};
+      var Module = this.Module.create(constructor);
+      assert.equal(Module().$name, constructor.name);
+    });
+  });
+
+  suite('Extendibility', function() {
+    suite('extend', function() {
+      var A, B, C;
+      var instanceOfA, instanceOfB, instanceOfC;
```
EragonJ: break them into several lines.

apps/settings/js/modules/base/module.js
crh0716: cool, let's use it.

apps/settings/js/modules/base/module.js
crh0716: Actually the meaning of trace here is a finer grain logging than debugging messages instead of the stack.

crh0716: Let's remove this level as there might be confusion between these two concepts.

apps/settings/js/modules/base/module.js
crh0716: :100:

apps/settings/js/modules/base/module.js
crh0716: Good point, but we still need to set the logger under the specified logging level

apps/settings/js/modules/base/module.js
crh0716: `prop` sounds good!

apps/settings/js/modules/base/module.js
crh0716: `constructor` could be `undefined`. We can throw in other cases.

apps/settings/js/modules/base/module.js
crh0716: `$name` is only for internal use and `$` is for avoiding naming conflicts.

apps/settings/js/modules/mvvm/observable.js
crh0716: Move to a separate function called `_getAllDependentValues`.

apps/settings/js/modules/base/dependency_graph.js
crh0716: Ah, we should actually throw here.

try-server-hook:
[crh0716 Arthur Chen (crh0716)](https://github.com/crh0716) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=ae8d94369734)

apps/settings/js/modules/base/module.js
EragonJ: Yes, let's remove it !

apps/settings/js/modules/base/module.js
EragonJ: ++

apps/settings/js/modules/base/module.js
EragonJ: okay !

apps/settings/js/modules/base/module.js
EragonJ: good idea !

apps/settings/js/modules/base/module.js
EragonJ: okay, never mind, because I only see dollar sign when using jQuery xD

apps/settings/js/modules/mvvm/observable.js
EragonJ: ++

apps/settings/js/modules/base/dependency_graph.js
EragonJ: Because in your previous revision, `this.Error` refers to `console.log()`, I think we should return after this line to make sure we won't do following works. But now, you would throw error in this case, I think your approach do fix the problem I said.

apps/settings/test/unit/modules/base/module_test.js
```diff
@@ -0,0 +1,206 @@
...
+        this.sinon.stub(console, 'error');
+        var Module = this.Module.create(function A() {});
+        module = Module();
+      });
+
+      test('NONE', function() {
+        module._logLevel = this.Module.LOG_LEVEL.NONE;
+        doLog(module);
+
+        assert.ok(console.log.notCalled);
```
EragonJ: `console.info` is missing.

apps/settings/test/unit/modules/mvvm/observable_test.js
```diff
@@ -1,110 +1,358 @@
...
+  };
+
+  function executeTest() {
+    test('shape: Observable({}).observe is a function', function() {
+      var observable = this.Observable({});
+      assert.equal(typeof observable.observe, 'function');
+    });
+    test('shape: Observable({}).unobserve is a function', function() {
+      var observable = this.Observable({});
+      assert.equal(typeof observable.observe, 'function');
```
EragonJ: this should be `observable.unobserve` instead.

apps/settings/test/unit/modules/base/module_test.js
crh0716: good catch!

apps/settings/test/unit/modules/mvvm/observable_test.js
crh0716: :100:

try-server-hook:
[crh0716 Arthur Chen (crh0716)](https://github.com/crh0716) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=2431ea6fb225)

try-server-hook:
[crh0716 Arthur Chen (crh0716)](https://github.com/crh0716) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=60f4d219a618)

PR #28644, opened 2015-03-04
Bug 1139443 - [Calendar] simplify utils/overlap

millermedeiros:
I ended up moving the `ConflictSpan` logic into the same file as `utils/overlap` since it's only used there and it's an _internal_ module..

tried to add a lot of comments to explain things, maybe this will make it easier to maintain it in the future.

please note that the order of the elements changed, now elements that starts earlier have a higher chance of being on the left side of the screen on LTR and on the right side of the screen on RTL (which in my opinion is a good improvement).

I also tried to keep as much from the old unit tests as possible just to make sure we are covering all the previous cases. (had to remove/replace any test that used `remove()` tho, since we don't need this feature anymore)

let me know what you think


try-server-hook:
[millermedeiros Miller Medeiros (millermedeiros)](https://github.com/millermedeiros) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=4e594496d776)

apps/calendar/js/utils/overlap.js
```diff
@@ -1,135 +1,167 @@
...
-   * @param {Object|String} busytime id or busytime object.
-   * @return {HTMLElement} associated dom element.
+   * gets ConflictSpan that can fit the item or create a new ConflictSpan
    */
-  getElement: function(busytime) {
-    var id = this._busytimeId(busytime);
-    return this.elements[id];
+  _getConflict: function(item) {
+    var conf, i = -1;
+    while ((conf = this._conflicts[++i])) {
```
gaye: Why double parens?

millermedeiros: because jshint doesn't allow assignments inside `while()` unless you set `boss:true` to avoid errors like `while (foo = bar)` when you really mean `while (foo == bar)`.

gaye: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

millermedeiros: ohh nice, forgot that `Array#find` was added to the spec.. been [using it for a very long time](http://moutjs.com/docs/latest/array.html#find)

PR #28648, opened 2015-03-05
Bug 1102835 - Keyboard UserDictionary: Formally enable prediction engine to predict from user dictionary

mnjul:
- We use only one worker which instantiates two Predictions object and use Promise.all() to receive the asynchronous results and merge the results.
- A setUserDictionary() function has been added to worker since user dictionary may change even when layout language isn't changed. Necessary wiring from latin.js is added.
- An empty user dictionary is stored as |undefined| in IndexedDB, and worker uses a strategy- & proxy-derived pattern to decide whether user dictionary prediction is invoked.
- Predictions() are no longer IIFE since we need more than one instances now.
- Worker.js is now handled by jshint, removed from jshint xfail list.
- The suggestions passed from worker to latin.js annotate whether a suggestion is from user dictionary; latin.js will always consider to show at least one suggestion from user dictionary if its weight is high enough.
- Enable user dictionary in keyboard settings by default.


try-server-hook:
[mnjul Min-Zhong "John" Lu (mnjul)](https://github.com/mnjul) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=4e239aa3f483)

try-server-hook:
[mnjul Min-Zhong "John" Lu (mnjul)](https://github.com/mnjul) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=b1be8a0d642b)

Makefile
```diff
@@ -507,7 +507,6 @@ define BUILD_CONFIG
   "LOCALES_FILE" : "$(subst \,\\,$(LOCALES_FILE))", \
   "GAIA_KEYBOARD_LAYOUTS" : "$(GAIA_KEYBOARD_LAYOUTS)", \
   "GAIA_KEYBOARD_PRELOAD_DICT_LAYOUTS" : "$(GAIA_KEYBOARD_PRELOAD_DICT_LAYOUTS)", \
-  "GAIA_KEYBOARD_ENABLE_USER_DICT" : "$(GAIA_KEYBOARD_ENABLE_USER_DICT)", \
```
timdream: We still want to keep the switch here.

timdream: Please add a `GAIA_KEYBOARD_ENABLE_USER_DICT?=1` in `Makefile`.

apps/keyboard/js/imes/latin/predictions.js
```diff
@@ -969,4 +987,4 @@ var Predictions = function() {
     setNearbyKeys: setNearbyKeys,
     predict: predict
   };
-}();
```
timdream: Something wrong here?

try-server-hook:
[mnjul Min-Zhong "John" Lu (mnjul)](https://github.com/mnjul) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=49e40fec885a)

apps/keyboard/js/imes/latin/predictions.js
mnjul: I changed that from `}();` to `};` such that it's no longer IFFE'ed; we need multiple instances of `Predictions` now.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ff2b012f205ce926a6f0a78148409da23af4a8d9

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0ae1825329dc84400a38fa4e48a4a28897101af5

apps/keyboard/js/imes/latin/latin.js
```diff
@@ -235,6 +250,28 @@
...
+    // Get user dictionary blob.
+    getUserDictionaryBlobPromise =
+      dbStore.getItem('dictblob').then(function(blob) {
+      // If worker is there, tell it we have a (new) user dictionary blob,
+      // since setLanguage() may be bypassed due to language not changing.
+      // (User dictionary is langauge-neutral.)
+      // Do pass the argument if blob is undefined (no user dictionary words),
+      // Since the worker needs to know if user has deleted all words.
+      // Worker will bypass UserDictionary prediction as needed.
+      // |blob| may be |undefined| so we don't Transfer it yet.
```
timdream: How about we do `typeof blob` check here to decide if we want to use Transfer? I don't know if it's worth it though.

apps/keyboard/js/imes/latin/worker.js
```diff
@@ -28,111 +31,282 @@
...
+      } else {
+        activePredictor = userDictPredictor;
+        userDictPredictor.setDictionary(buffer);
+      }
+    },
+    setNearbyKeys: function(data) {
+      userDictPredictor.setNearbyKeys(data);
+    },
+    // Directly pass arguments to wrapped predictor
+    predict: function(...args) {
```
timdream: This will break demo on iOS :'(

apps/keyboard/js/imes/latin/worker.js
```diff
@@ -28,111 +31,282 @@
...
+    // [4] latin.js will always retain at least one user dict prediction (unless
+    //     it decides to drop the prediction because it equals user input)
+
+    var BUILT_IN_CANDIDATES_LOW = 24;
+    var BUILT_IN_CANDIDATES_HIGH = 50;
+    var BUILT_IN_CANDIDATES_EXTRA = 60;
+    var USER_DICT_CANDIDATES = 40;
+    var BUILT_IN_TOO_LOW_THRESHOLD = 4;
+
+    var builtInPromise = new Promise(function(resolve, reject){
```
timdream: Interesting... The `ProxyUserDictPredictor.predict` and `Predictions.predict` are both sync, and we are in a worker, so it make less sense to wrap them inside of a Promise.

Wrapping in a Promise also cause to create `pending*Prediction` to make sure we don't run two predictions at the same time, although it's almost impossible.

How about flatting the code flow and make it run sync again? it make more sense to me.

Unfortunately there the callback function passed in `new Promise()` comes with different complex logic to post-process the result according to user/build-in types... we would have to organize it in some way inside the `worker.predict` function.

apps/keyboard/js/imes/latin/worker.js
mnjul: I can definitely ES5-fy it, but I guess those arrow functions need ES5-fication too?

apps/keyboard/js/imes/latin/worker.js
mnjul: Predictions.predict is sync, but the prediction process isn't. Wrapping both predictions in Promise means I can better synchronize things. We should discuss about this.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ffc236b2293caacb6d161e7399df52d389a8ecfb

PR #28650, opened 2015-03-05
Bug 1139238 - [FFOS2.0][Woodduck][Airplane mode]The airplane mode is invalid.

mpizza:
--  disable buletooth click handler when 'this.airplaneMode.dataset.enabling' is true.

### DIFF
--- a/apps/system/js/quick_settings.js
+++ b/apps/system/js/quick_settings.js
@@ -236,7 +236,8 @@ var QuickSettings = {
 
           case this.bluetooth:
             // do nothing if bluetooth isn't ready
-            if (this.bluetooth.dataset.initializing)
+            if (this.bluetooth.dataset.initializing 
+                || this.airplaneMode.dataset.enabling)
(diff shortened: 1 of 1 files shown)


apps/system/js/quick_settings.js
```diff
@@ -236,7 +236,8 @@ var QuickSettings = {
 
           case this.bluetooth:
             // do nothing if bluetooth isn't ready
-            if (this.bluetooth.dataset.initializing)
+            if (this.bluetooth.dataset.initializing 
+                || this.airplaneMode.dataset.enabling)
```
crh0716: Let's also early return when `this.airplaneMode.dataset.disabling`. And the problem should also exist for wifi, so I would suggest to add the same check for it.

crh0716: Access to dataset is costly. I would suggest to keep the current airplane mode status on `QuickSettings`.

PR #28651, opened 2015-03-05
Bug 1094759 - new Bootstrap

alivedise:
Conflicts:
    apps/system/index.html
    apps/system/js/app_transition_controller.js
    apps/system/js/app_window_manager.js
    apps/system/js/bootstrap.js
    apps/system/js/browser.js
    apps/system/js/browser_context_menu.js
    apps/system/js/input_window_manager.js
    apps/system/js/software_button_manager.js
    apps/system/js/task_manager.js
    apps/system/test/unit/bootstrap_test.js
    apps/system/test/unit/icc_test.js
    apps/system/test/unit/sound_manager_test.js
    apps/system/test/unit/task_manager_test.js


try-server-hook:
[alivedise Alive.Kuo (alivedise)](https://github.com/alivedise) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=1dd96e6228fd)

try-server-hook:
[alivedise Alive.Kuo (alivedise)](https://github.com/alivedise) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=112edf2d47cb)

try-server-hook:
[alivedise Alive.Kuo (alivedise)](https://github.com/alivedise) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=bb046f477d04)

apps/system/index.html
```diff
@@ -1193,5 +905,22 @@ <h1 data-l10n-id="deviceMenu"></h1>
       <div id="bottom-panel" class="gesture-panel" data-z-index-level="gesture-panel"></div>
 
     </div> <!-- end of #screen -->
+
+    <script defer src="shared/js/settings_listener.js"></script>
```
KevinGrandon: What's the reasoning for placing these scripts at the bottom? Since we have the ```defer``` attribute, I think it's cleaner for them to be defined in the ```<head>```. I think it's cleaner to read that way.

shared/test/unit/mocks/mock_service.js
```diff
@@ -28,8 +48,52 @@ var MockService = {
   },
   query: function(state) {
     switch (state) {
+      case 'isUpgrade':
```
KevinGrandon: The number of things under Service is growing at quite a fast speed. Is this the long term plan and is this scalable? It seems like just another namespace besides window to hold global soup.

apps/system/index.html
```diff
@@ -3,84 +3,33 @@
...
-    <script defer src="js/base_ui.js"></script>
-    <script defer src="js/places.js"></script>
 
-    <script defer src="js/orientation_manager.js"></script>
-    <!--
-    <script defer src="shared/js/input_mgmt/input_app_list.js"></script>
-    <script defer src="shared/js/keyboard_helper.js"></script>
-    <script defer src="shared/js/template.js"></script>
-    -->
+    <link rel="resource" type="image/png" href="shared/resources/branding/browser_84.png"/>
```
KevinGrandon: These two resources currently seem duplicated on line 117?

apps/system/js/bluetooth_core.js
```diff
@@ -16,12 +20,22 @@
   BaseModule.create(BluetoothCore, {
     name: 'BluetoothCore',
 
-    start: function() {
+    _start: function() {
+      console.log('1');
```
KevinGrandon: nit: just a reminder, I assume we want to remove these before landing?

apps/system/js/homescreen_launcher.js
```diff
@@ -216,7 +160,9 @@
         return null;
       }
       if (typeof this._instance == 'undefined') {
-        this._instance = new HomescreenWindow(this._currentManifestURL);
+        console.trace();
+        this._instance = new window.HomescreenWindow(this._currentManifestURL);
```
KevinGrandon: nit: we already define ```HomescreenWindow``` as a global, so maybe we should just access it directly instead of through ```window```?

apps/system/js/shortcuts.js
```diff
@@ -0,0 +1,60 @@
...
+  BaseModule.create(Shortcuts, {
+    name: 'Shortcuts',
+
+    _start: function() {
+      window.addEventListener('touchstart', this, true);
+      window.addEventListener('touchend', this, true);
+      window.addEventListener('mousedown', this, true);
+      window.addEventListener('mouseup', this, true);
+    },
+    /* === XXX Bug 900512 === */
```
KevinGrandon: I think all of this weird touch handler code was removed recently by @etiennesegonzac ?

try-server-hook:
[alivedise Alive.Kuo (alivedise)](https://github.com/alivedise) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=ce59e07be720)

try-server-hook:
[alivedise Alive.Kuo (alivedise)](https://github.com/alivedise) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=e63a94e0453b)

try-server-hook:
[alivedise Alive.Kuo (alivedise)](https://github.com/alivedise) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=90db670d006d)

apps/system/js/shortcuts.js
etiennesegonzac: yes we did ! (it's a perf issue for APZC in the main process and we don't have the issue on the flame)

etiennesegonzac:
I could see all the `onidle` waits becoming a pain for integration tests. Good thing is: we already have a `waitForStartup` method in `apps/system/test/marionette/lib/system.js`!
I think it should wait for the system to be fully loaded, maybe we'll need an event.

shared/test/unit/mocks/mock_service.js
```diff
@@ -28,8 +48,52 @@ var MockService = {
...
+        return this.mLayoutManager_height -
+          (arguments[2] ? 0 : this.mKeyboardHeight);
+      case 'SoftwareButtonMnager.height':
+        return this.mSoftwareButtonManager_height;
+      case 'SoftwareButtonMnager.width':
+        return this.mSoftwareButtonManager_width;
+      case 'InputWindowManager.height':
+        return this.mInputWindowManager_getHeight;
+      case 'getFtuOrigin':
+        return 'app://ftu.gaiamobile.org';
```
etiennesegonzac: This doesn't look healthy :)

For example, the tricky EdgeSwipeDetector unit test failure just comes from the fact that `SoftwareButtonManager` is misspelled as `SoftwareButtonMnager` in this switch case.

I think the `MockService` should be a really shallow mock with barely no logic and we should let each test do things like:

`this.sinon.stub(MockService, 'query').withArgs('SoftwareButtonManager.height').returns(42)`

or, even better the we could declare a little helper in `mock_service.js` to let testers do 

`mockQueryWith('SoftwareButtonManager.height', 42)`

Some tests will have a fair amount of those but it makes it very explicit in code and in tests when a modules has tones for `Service` dependencies which is a good thing IMO since we don't want it to go out of hands.

apps/system/test/unit/edge_swipe_detector_test.js
```diff
@@ -577,7 +565,7 @@ suite('system/EdgeSwipeDetector >', function() {
         var recvEvents = verticalSwipe();
 
         var call = fwSpy.secondCall;
-        assert.equal(call.args[0], recvEvents[7]);
+        assert.equal(call.args[0], recvEvents[8]);
```
etiennesegonzac: I'm pretty sure we shouldn't change this. I agree that this is hard to read,
But I'm pretty sure this test will never pass since it asserts that the second and third call to forward got the same argument (`recvEvents[8]` instead of `recvEvents[7]` then `recvEvents[8]`)

apps/system/js/base_module.js
```diff
@@ -110,7 +102,83 @@
    */
   BaseModule.STATES = [];
 
+  BaseModule.SIDE_MODULES = [];
```
etiennesegonzac: Not sure about the "side modules" name, but not sure I fully understand the concept either.
Are those "lazy modules"? If yes I think `LAZY_MODULES` and `LayModuleMixin` would be more explicit (and a comment block would probably help too).

apps/system/js/base_module.js
```diff
@@ -110,7 +102,83 @@
...
+        unloaded.forEach(function(module) {
+          var moduleName = BaseModule.lowerCapital(module);
+          if (!this[moduleName]) {
+            this._initialSideModule(moduleName, module);
+          }
+        }, this);
+        this.__side_module_loaded && this.__side_module_loaded();
+      }.bind(this)).catch(function(err) {
+        console.error(err);
+      });
```
etiennesegonzac: this code is pretty hard to read and _very_ similar to the sub module code, we should factor it :)

apps/system/js/base_module.js
```diff
@@ -121,33 +189,50 @@
...
+          if (!this[name]) {
+            this._initialSubModule(name, submodule);
+          }
+        } else {
+          unloaded.push(submodule);
+        }
+      }, this);
+
+      if (unloaded.length === 0) {
+        this.__sub_module_loaded && this.__sub_module_loaded();
```
etiennesegonzac: Is the `__` prefix supposed to be for "stuff the child class should/can override"?
I think we need to be more consistent with those (we also have `_start`, `_stop` etc...).
We even have a comment mentioning `onSubModuleInited` ??

How about:
```
baseStart()
baseStop()
baseSubModuleLoaded()
baseSideModuleLoaded()
```
etc...

apps/system/js/base_module.js
etiennesegonzac: Also I'm not sure that we need a top level `SIDE_MODULES` concept if each module is responsible to trigger the `_startSideModules()`.

Either the logic of "when should we load side modules" is something common, then we should have it in the "declarative part" of the `BaseModule` (ie. `BaseModule.SIDE_MODULES`) and the default implementation should take care of loading them.
Or the load logic really depends on each module and we should just add a `BaseModule.lazyLoad('ModuleName')` helper method and let each module explicitly load stuff lazily when it fits.

apps/system/js/base_module.js
```diff
@@ -110,7 +102,83 @@
...
+        if (this[moduleName]) {
+          this.debug('Stopping submodule: ' + moduleName);
+          this[moduleName].stop && this[moduleName].stop();
+        }
+      }, this);
+    }
+  };
+
   var SubmoduleMixin = {
+    START_SUB_MODULES_ON_START: true,
```
etiennesegonzac: I'm not completely opposed to it but we should always think twice before adding a special case to what is supposed to be the "common part" of the system modules.
Looks like we could remove this special case and let `AppCore` do a few (tested) `this.lazyLoad('ModuleName')` calls once it gets the `applicationready`.

apps/system/js/edge_swipe_detector.js
```diff
@@ -370,14 +369,16 @@
       // but we still want to redispatch touch events to the "overlayed"
       // software home button
       var softwareButtonOverlayed =
-        Service.currentApp &&
-        Service.currentApp.isFullScreenLayout();
+        Service.query('getTopMostWindow') &&
+        Service.query('getTopMostWindow').isFullScreenLayout();
       if (softwareButtonOverlayed) {
-        return x > (layoutManager.width - softwareButtonManager.width) ||
-            y > (layoutManager.height - softwareButtonManager.height);
```
etiennesegonzac: please extract the `Service.query` calls inside variables

apps/system/js/launcher.js
```diff
@@ -0,0 +1,61 @@
...
+  var Launcher = function() {};
+  Launcher.SETTINGS = [
+    'ftu.manifestURL',
+    'homescreen.manifestURL',
+    'deviceinfo.previous_os',
+    'deviceinfo.os',
+    'lockscreen.enabled'
+  ];
+  Launcher.STATES = [
+    'isUpgrade'
```
etiennesegonzac: nit: `justUpgraded`

try-server-hook:
[alivedise Alive.Kuo (alivedise)](https://github.com/alivedise) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=8c9e28dc471e)

apps/system/js/core.js
```diff
@@ -9,18 +9,55 @@
...
+    'SourceView',
+    'TtlView',
+    'MediaRecording',
+    'QuickSettings',
+    'Shortcuts',
+    'UsbStorage',
+    'MobileIdManager', // nonblocking
+    'FindmydeviceLauncher', // nonblocking
+    'FxAccountsManager', // nonblocking
+    'FxAccountsUI', // nonblocking
```
mikehenrty: What does it mean to be non-blocking? Are they loaded different in the code?

apps/system/js/wallpaper_manager.js
```diff
@@ -349,6 +349,10 @@
       // Create a new blob:// url for this blob
       this._blobURL = URL.createObjectURL(blob);
 
+      document.getElementById('screen').style.backgroundImage =
+        'linear-gradient(rgba(0, 0, 0, 0.1), rgba(0, 0, 0, 0.1)),' +
+        'url(' + this._blobURL + ')';
+
```
mikehenrty: What does this have to do with your patch?

apps/system/js/core.js
```diff
@@ -58,6 +98,46 @@
...
+            'js/devtools/logshake.js',
+            'js/devtools/remote_debugger.js',
+            'shared/js/date_time_helper.js'
+          ]).then(function() {
+            self.remoteDebugger = new RemoteDebugger();
+            // self.remoteDebugger.start();
+          });
+        }
+      };
+      navigator.addIdleObserver(idleObserver);
```
mikehenrty: Right now it looks like we have 3 places that use this 10s idle timer to do things like loading imports and starting side modules. Perhaps we should abstract this into BaseModule such that you could have things like:
```
Core.LOAD_ON_IDLE = [
  'js/download/download_manager.js',
  'js/payment.js',
  ...
];
Core.START_MODULES_ON_IDLE = [
  'NotificationScreen',
  'AirplaneMode',
  ...
];
```
and even perhaps have a function`__on_idle` that gets run after the IDLE modules have been loaded/started. In any case, it seems a little unfortunate that these modules have to repeat the idle observer logic, and yet all do very similar things.

apps/system/js/shortcuts.js
alivedise: good news!

apps/system/js/homescreen_launcher.js
alivedise: right, fixing

shared/test/unit/mocks/mock_service.js
alivedise: I have no long term plan... but I am opened to other solution to avoid put instances under global object.

shared/test/unit/mocks/mock_service.js
alivedise: I am sure this(mock things in MockService instead of mocking it in tests) was requested by you in certain review :)

apps/system/test/unit/edge_swipe_detector_test.js
alivedise: Yes, I am just blindly trying what's that...

apps/system/js/wallpaper_manager.js
alivedise: This one is moved from the original bootstrap.js to assign the blob into the wallpaper background image.

apps/system/js/core.js
alivedise: right

try-server-hook:
[alivedise Alive.Kuo (alivedise)](https://github.com/alivedise) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=7c3080836a80)

try-server-hook:
[alivedise Alive.Kuo (alivedise)](https://github.com/alivedise) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=983e125b5a5f)

try-server-hook:
[alivedise Alive.Kuo (alivedise)](https://github.com/alivedise) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=b459b3ab65a2)

apps/system/js/launcher.js
```diff
@@ -0,0 +1,61 @@
...
+    name: 'Launcher',
+    ready: function() {
+      this._version = {};
+      return Promise.all([
+        this.readFtu(),
+        this.readSetting('homescreen.manifestURL'),
+        this.readSetting('ftu.manifestURL'),
+        this.readSetting('deviceinfo.os'),
+        this.readSetting('deviceinfo.previous_os')
+      ]).then(function(results) {
```
julienw: just a suggestion, but maybe you can use destructing assignment to have more meaningful variable names instead of results[0,1,2,3,4]

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ee2e6d945cf3c99cb0913217b224f80e905f0db3

apps/system/js/base_module.js
```diff
@@ -111,6 +103,24 @@
...
   var SubmoduleMixin = {
+    // We could dynamically add and load new modules on demand.
+    loadWhenIdle: function(modules, time) {
+      return new Promise(function(resolve) {
+        this.constructor.SUB_MODULES =
+          this.constructor.SUB_MODULES.concat(modules);
+        var self = this;
+        var idleObserver = {
+          time: time || 10,
+          onidle: function() {
```
timdream: `onidle` will invoke instantly if the phone has already idle for more than 10 sec when `addIdleObserver` is called. Are you sure this is what's intended? If so we can use IdleObserver directly.

timdream: Otherwise we should be using idleTimer in shared.

alivedise: Yes, it's intended. And even 10secs is still slow..I am thinking of a way to improve it if necessary.

apps/system/js/launcher.js
alivedise: Good suggestion! Thanks!

apps/system/js/launcher.js
alivedise: done

apps/system/js/edge_swipe_detector.js
alivedise: done

apps/system/js/base_module.js
alivedise: As you wish.

PR #28663, opened 2015-03-05
Bug 1133506 2.2 - [NFC/Secure Element] Update wallet demo code and move ...

tauzen:
...it to UI Test, r=timdream


autolander:

PR #28704, opened 2015-03-09
Bug 1139031 - update l10n file to fix () problem r=Stas

EragonJ:


try-server-hook:
[EragonJ EragonJ (E.J.) (EragonJ)](https://github.com/EragonJ) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=4579bc41a13a)

apps/settings/locales/settings.en-US.properties
```diff
@@ -290,7 +290,13 @@ message-delivery-reports = Delivery Reports
...
 message-read-reports = Read Reports
 message-read-reports-details = Request read report for each message sent
-mmsTitle = Multimedia Messaging (MMS)
+# [For localizers]
+# We can't render these strings correctly for RTL locales, so we have to put LRE (&#x202a;)
+# and PDF (&#x202c;) entities to these translations to force the correct direction.
+#
+# RTL locales should have a translation similar to this:
+# mmsTitle.innerHTML= &#x202a;Multimedia Messaging (MMS)&#x202c;
+mmsTitle.innerHTML = Multimedia Messaging (MMS)
```
stasm: I think you should also include LRE and PDF in the English copy here.  If the translation is not available, the build sytsem will copy the English value and without the Unicode marks the parentheses will look broken.

stasm: The comment is too long, too, and doesn't follow the commenting style of .properties files.  I suggest to change it as follows:

```
# LOCALIZATION NOTE (mmsTitle): LRE (&#x202a;) and PDF (&#x202c;)
# force the proper directionality of parentheses at the end of the string
# regardless of the direction of the surrounding layout. Remove them only
# if the translation puts them between other text.
mmsTitle.innerHTML= &#x202a;Multimedia Messaging (MMS)&#x202c;
```

EragonJ: thanks for the hint @stasm ! I'll update it based on your suggestion.

EragonJ: Okay, I didn't notice the fallback mechanism here, so we should also add these special characters here +++

try-server-hook:
[EragonJ EragonJ (E.J.) (EragonJ)](https://github.com/EragonJ) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=919e08f41d34)

apps/settings/locales/settings.en-US.properties
```diff
@@ -290,7 +290,12 @@ message-delivery-reports = Delivery Reports
 message-delivery-reports-details = Request notification of delivery for each message sent
 message-read-reports = Read Reports
 message-read-reports-details = Request read report for each message sent
-mmsTitle = Multimedia Messaging (MMS)
+# LOCALIZATION NOTE (mmsTitle): LRE (&#x202a;) and PDF (&#x202c;)
+# force the proper directionality of parentheses at the end of the string
+# regardless of the direction of the surrounding layout. Remove them only
+# if the translation puts them between other text.
+# mmsTitle.innerHTML= &#x202a;Multimedia Messaging (MMS)&#x202c;
```
stasm: You can probably remove line 297 here as it is the same as the original string one line below.

EragonJ: okay !! Thanks Stas

PR #28777, opened 2015-03-11
Bug 1032128 - Add "always allow" for cert-based authentication. r=janx

jryans:
In addition to the "always allow" option, this reworks the "selectone" modal style to allow more customization and localization.

### DIFF
--- a/apps/system/js/app_chrome.js
+++ b/apps/system/js/app_chrome.js
@@ -11,7 +11,6 @@
 
 (function(exports) {
   var _id = 0;
-  var _ = navigator.mozL10n.get;
 
   var newTabManifestURL = null;
   SettingsListener.observe('rocketbar.newTabAppURL', '',
@@ -820,19 +819,27 @@
       }
     }
 
-    var data = {
-      title: _('add-to-home-screen'),
-      options: []
-    };
+    var title = 'add-to-home-screen';
+    var options = [];
 
     if (this.isSearch()) {
       var dataset = this.app.config;
-      data.options.push({ id: 'search', text: dataset.searchName });
+      options.push({
+        id: 'search',
+        text: {
+          raw: dataset.searchName
+        }
+      });
     } else {
-      data.options.push({ id: 'origin', text: this.title.textContent });
+      options.push({
+        id: 'origin',
+        text: {
+          raw: this.title.textContent
+        }
+      });
     }
 
-    ModalDialog.selectOne(data, selected);
+    ModalDialog.selectOne(title, options, selected);
   };
 
   AppChrome.prototype.showWindows = function ac_showWindows() {
--- a/apps/system/js/devtools/remote_debugger.js
+++ b/apps/system/js/devtools/remote_debugger.js
@@ -4,6 +4,41 @@
 (function(exports) {
 
   /**
+   * Note: Values copied from Gecko:
+   *       toolkit/devtools/security/auth.js
+   */
+  var AuthenticationResult = {
+
+    /**
+     * Close all listening sockets, and disable them from opening again.
+     */
+    DISABLE_ALL: 'DISABLE_ALL',
+
+    /**
+     * Deny the current connection.
+     */
+    DENY: 'DENY',
+
+    /**
+     * Additional data needs to be exchanged before a result can be determined.
+     */
+    PENDING: 'PENDING',
+
+    /**
+     * Allow the current connection.
+     */
+    ALLOW: 'ALLOW',
+
+    /**
+     * Allow the current connection, and persist this choice for future
+     * connections from the same client.  This requires a trustable mechanism to
+     * identify the client in the future, such as the cert used during OOB_CERT.
+     */
+    ALLOW_PERSIST: 'ALLOW_PERSIST'
+
+  };
+
+  /**
    * RemoteDebugger displays a prompt asking the user if they want to enable
    * remote debugging on their device. This is generally called when the user
    * attempts to access the device from the App Manager.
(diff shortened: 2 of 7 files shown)


try-server-hook:
[jryans J. Ryan Stinnett (jryans)](https://github.com/jryans) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=39c036fc2fe8)

try-server-hook:
[jryans J. Ryan Stinnett (jryans)](https://github.com/jryans) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=31f9761d77f5)

apps/system/js/modal_dialog.js
```diff
@@ -362,23 +362,24 @@ var ModalDialog = {
 
   buildSelectOneDialog: function md_buildSelectOneDialog(data) {
     var elements = this.elements;
-    elements.selectOneTitle.textContent = data.title;
-    elements.selectOneMenu.innerHTML = '';
+    this._localizeElement(elements.selectOneTitle, data.title);
```
jankeromnes: With your refactor, you're changing:

    elements.selectOneTitle.textContent = data.title

to

    elements.selectOneTitle.setAttribute('data-l10n-id', data.title);

I'm not sure what this does, was it on purpose or did you intend to write the following instead?

    this._localizeElement(elements.selectOneTitle, {raw: data.title});

apps/system/js/devtools/remote_debugger.js
```diff
@@ -4,6 +4,58 @@
...
+  /**
+   * Note: Values originally defined in Gecko:
+   *       toolkit/devtools/security/auth.js
+   *
+   * |allowConnection| implementations can return various values as their
+   * |result| field to indicate what action to take.  By specifying these, we
+   * can centralize the common actions available, while still allowing embedders
+   * to present their UI in whatever way they choose.
+   */
+  var AuthenticationResult = createEnum({
```
jankeromnes: I'm not fond of additional bootstrapping code that makes things slightly harder to read.

Would you mind removing `createEnum` and create the `AuthenticationResult` enum in a "lamer" way like `ALLOW: "ALLOW"`, `DENY: "DENY"`, etc?

apps/system/js/modal_dialog.js
jankeromnes: I don't understand why all other modals use `evt.detail.title` but "selectone" uses `evt.detail.message.title`. If there isn't a good reason, could you please fix it? Also might be worth to make `setTitle` use `_localizeElement` and use it here (or use it in the `case 'selectone'` line 239 to do like the other modals).

apps/system/js/devtools/remote_debugger.js
jankeromnes: Oh, I see this code is copy/pated. Instead of a plain copy of the code and comments, I'd prefer it just as:

    /** Note about original file */
    var AuthenticationResult = {
      DISABLE_ALL: "DISABLE_ALL",
      DENY: "DENY", // ...
    };

jankeromnes:
devtools-prompt-new

The new prompt looks a little weird on my device. Please address the nits I'll post on Bugzilla.

apps/system/js/modal_dialog.js
jryans: I did not mean to write `{ raw: data.title }`, as I want to look up the properties by it's ID, which is the "default" behavior of passing just a string.

This is an API change for other users of `selectOne`. I believe I've now fixed the only other user of it

Also, I think I've cleaned up the events in the way you describe.

apps/system/js/devtools/remote_debugger.js
jryans: Alright, I removed `createEnum` here as you suggest, even though I feel it's clearer by reducing the possibility for errors. However, I won't be removing it from Gecko. :wink:

I also removed the comment about `allowConnection` which did not make sense here. However, I left the comments on the values, as I think they are generally helpful and unlikely to change over time.

try-server-hook:
[jryans J. Ryan Stinnett (jryans)](https://github.com/jryans) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=25a412e3b20a)

apps/system/locales/system.en-US.properties
```diff
@@ -477,7 +477,10 @@ payment-flow=Purchase
 
 # Remote Debugger Connection Dialog
 remoteDebuggerPromptUSB=Allow USB debugging connection?
-remoteDebuggerPromptTCP=Allow remote debugging connection from {{host}}:{{port}}?
+remoteDebuggerPromptTCP.innerHTML=Allow remote debugging connection from {{host}}:{{port}}?<br/><br/>Client ID: {{cert}}
```
jryans: I'll probably change `Client ID:` to `Their Cert:`. On @past's advice from Gecko changes, we've avoided saying the words "client" or "server" to the user so far.

apps/system/js/app_chrome.js
```diff
@@ -827,9 +827,19 @@
...
-      data.options.push({ id: 'origin', text: this.title.textContent });
+      data.options.push({
+        id: 'origin',
+        text: {
+          raw: this.title.textContent
+        }
+      });
     }
 
     ModalDialog.selectOne(data, selected);
```
jankeromnes: Didn't you refactor this method to take an additional non-optional `title` parameter?

apps/system/locales/system.en-US.properties
jankeromnes: Please use a more precise term like `SHA-256 certificate:`, add a line break after, and surround host and cert with `<code></code>`, like so:

    Allow remote debugging connection from <code>{{host}}:{{port}}</code>?<br/><br/>SHA-256 certificate:<br/><code>{{cert}}</code>

apps/system/style/modal_dialog/modal_dialog.css
```diff
@@ -85,6 +85,7 @@
 
 #modal-dialog-select-one h3 {
   font-size: 1.8rem;
+  line-height: 2rem;
 }
 
 #modal-dialog-select-one ul {
```
jankeromnes: The button text is badly placed. Please add the following rules to `#modal-dialog-select-one ul > li > button`:

    padding-left: 1.5rem;
    line-height: 2rem;

apps/system/locales/system.en-US.properties
jankeromnes: Note: Please also use "SHA-256 certificate:" or "My SHA-256 certificate:" in WebIDE, instead of "My Cert:".

jankeromnes:
In the test file `apps/system/test/unit/modal_dialog_test.js`, please replace:

    ModalDialog.buildSelectOneDialog({
      title: 'testTitle',
      options: testOptions
    });

with:

    ModalDialog.buildSelectOneDialog(testOptions);

in the test named `'call buildSelectOneDialog >'`.

apps/system/js/devtools/remote_debugger.js
```diff
@@ -35,35 +70,61 @@
...
+        title = 'remoteDebuggerPromptUSB';
       } else {
-        text = {
+        title = {
           id: 'remoteDebuggerPromptTCP',
-          args: session.client
+          args: {
+            host: session.client.host,
+            port: session.client.port,
+            cert: session.client.cert.sha256.replace(/:/g, ' ')
```
jankeromnes: Please don't replace `:` with spaces. Instead, split the SHA-256 with something like:

    // Split the SHA256 cert into blocks of 24 chars.
    var sha = session.client.cert.sha256;
    var block = 24;
    var splitsha = [];
    for (var i = 0; i < sha.length; i += block) {
      splitsha.push(sha.slice(i, i + block));
    }

and then assign `title.args.cert` like so:

    cert: splitsha.join('\u200b')

This will add zero-width spaces at strategic places in the SHA-256 so that it can break and be displayed nicely.

Note: Please also display the SHA-256 in WebIDE using the same uppercase, `:`-separated format.

apps/system/js/app_chrome.js
jryans: Hmm, I suppose I did... Not sure why I skipped this step, I thought I did it!

jryans:
I made the test change you suggested, as well as further cleanups of other test bits to align with the new API.

apps/system/style/modal_dialog/modal_dialog.css
jryans: It's not obvious to me that this is an improvement.

If you hover on the `h3` in the inspector, the vertical guide shows that the first pixel of button text is aligned with that of the `h3`.

For the `line-height`, what's your reasoning that these particular buttons needs a different line height than the default from the `[role="dialog"] button` block?

try-server-hook:
[jryans J. Ryan Stinnett (jryans)](https://github.com/jryans) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=7f4f7bdb1af6)

PR #28798, opened 2015-03-11
Bug 1140321 - [RTL][Download] Fail to download the file, then the error ...

gmarty:
...icon is left-aligned and overlap with refresh icon in download list

### DIFF
--- a/apps/settings/style/downloads.css
+++ b/apps/settings/style/downloads.css
@@ -192,8 +192,7 @@
 #downloadList ul > li[data-state="failed"] .download-status {
   width: 3rem;
   height: 3rem;
-  background: url('images/download_exclamation.png');
-  background-repeat: no-repeat;
+  background: url('images/download_exclamation.png') no-repeat center center;
   left: 0.2rem;
 }
 
@@ -235,8 +234,14 @@
 }
 
 /* RTL rules */
+html:-moz-dir(rtl) #downloadList aside.download-status {
+  margin: 1.75rem 0 0 0.5rem;
+  right: 0.2rem;
+  left: unset;
+}
+
 html:-moz-dir(rtl) #downloadList aside.pack-end {
-  right: auto;
+  right: unset;
   left: 0;
 }
 


try-server-hook:
[gmarty Guillaume C. Marty (gmarty)](https://github.com/gmarty) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=02dfb7a53042)

apps/settings/style/downloads.css
```diff
@@ -235,6 +235,11 @@
 }
 
 /* RTL rules */
+html:-moz-dir(rtl) #downloadList aside.download-status {
+  right: 0.2rem;
```
crh0716: As the icon is smaller than 30x30, it is aligned to the right. I would suggest to add `background-center` as you did in the previous patch. 
And `#downloadList aside` set `margin-right` to 0.5rem, so the value of right should be -0.3rem in rtl layout.

apps/settings/style/downloads.css
```diff
@@ -235,6 +235,11 @@
...
+  left: auto;
```
crh0716: nit: Suggest to use `unset` if you want to override it as a default value.

apps/settings/style/downloads.css
gmarty: That's overridden in line 192 `#downloadList ul > li[data-state="failed"] .download-status`. So I let it to 0.2rem for now.

try-server-hook:
[gmarty Guillaume C. Marty (gmarty)](https://github.com/gmarty) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=ac9f11319092)

apps/settings/style/downloads.css
crh0716: It's fine to set `right` to 0.2rem, but we should also flip the margin of `#downloadList aside` in RTL mode (`margin: 1.75rem 0 0 0.5rem`).

try-server-hook:
[gmarty Guillaume C. Marty (gmarty)](https://github.com/gmarty) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=633dcc85381a)

PR #28799, opened 2015-03-11
Bug 1139207 - [RTL] "Call Log" truncates long contacts name at the beginning of the name. Use bdi to ellipsis content. r=drs

azasypkin:


try-server-hook:
[azasypkin Aleh Zasypkin (azasypkin)](https://github.com/azasypkin) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=07ee074d0772)

apps/communications/dialer/js/call_log.js
```diff
@@ -524,17 +524,15 @@ var CallLog = {
       }
     }
     primInfoMain.appendChild(bdi);
-
-    var retryCount = document.createElement('span');
-    retryCount.className = 'retry-count';
+    primInfo.appendChild(primInfoMain);
```
azasypkin: Haven't noticed where we may need this empty element, am I missing something? I changed it because it occupies space even if not visible (because of ```-moz-margin-start: 0.5rem```). If it's needed I can use css-only solution like ```.retry-count:empty { display: none; }```

apps/communications/dialer/style/call_log.css
```diff
@@ -163,7 +163,7 @@ ol, ul {
 }
 
 .log-item .primary-info {
-  text-overflow: ellipsis;
+  display: flex;
```
azasypkin: With this we always display ```retry-count``` even if contact name is long.

apps/communications/dialer/style/call_log.css
```diff
@@ -175,7 +175,13 @@ ol, ul {
 }
 
 .log-item .primary-info-main {
+  min-width: 0;
```
azasypkin: We need this because of flex parent, see https://bugzilla.mozilla.org/show_bug.cgi?id=1043520 for deta for more details.

apps/communications/dialer/style/call_log.css
DouglasSherk: Yikes, I'm surprised that we've never noticed this before it was indeed a problem. Thanks for this fix.

PR #28820, opened 2015-03-12
Bug 1122097 - Test that a file with &, >, <, ~ and some Unicode is properly displayed.

hfiguiere:


try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=3a253fbb8c56)

try-server-hook:
[hfiguiere Hubert Figuière (hfiguiere)](https://github.com/hfiguiere) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=e48b4127048a)

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=03ea3ddb64e0a15835fe2c5913a90107cd2d2354

apps/music/test/marionette/lib/playlisthelper.js
```diff
@@ -4,6 +4,10 @@
 
 var PlaylistHelper = {
 
+  mainTitle: function(element) {
+    return element.findElement('.list-main-title').text();
```
davidflanagan: I assume that the marionette text() call returns the textContent property of the element, or that it recursively traverses the element tree and returns the concatenation of the textContent properties of all descendants? Assuming that's true, then this seems fine. If it returns innerHTML, then it isn't what we need for this test, of course.

apps/music/test/marionette/playlist_test.js
```diff
@@ -30,6 +30,30 @@ marionette('Music player playlist', function() {
...
+    });
+
+    test('Check name with >, <, ~, & and some Unicode', function() {
+      music.launch();
+      music.waitForFirstTile();
+      music.switchToSongsView();
+
+      // this will wait on the first song as well.
+      var song = music.firstSong;
+      assert.equal(PlaylistHelper.mainTitle(song), 'dump 2>&1 < ~/®');
```
davidflanagan: This song title goes well with the other ones on the fake album. 
But it isn't sufficient for this test because the angle brackets and the & are not close enough to real HTML.
If I understand it correctly, the point of this test is to make sure that we are not setting song titles on innerHTML (or are properly escaping the HTML if we do).  But the HTML parser is smart enough to figure out that this song title is not HTML. So, consider this code:

```
document.body.innerHTML = 'dump 2>&1 < ~/®';
assert.equal(document.body.textContent, 'dump 2>&1 < ~/®')
```

That assertion passes even though the innerHTML contained unescaped special characters.

For a valid test, I think you need a song title like 

```
<b.&amp; Injection Vulnerabliity</b>
```

apps/music/test/marionette/lib/playlisthelper.js
hfiguiere: http://mozilla-b2g.github.io/marionette-js-client/api-docs/files/lib_marionette_element.js.html#l168

That's what it returns. We use it elsewhere too.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c117decf6743bf310d83188c58b475e735351728

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5fb7c5ab648f08db9006811f0c0545d5c6ecac47

PR #28842, opened 2015-03-12
Bug 1141500 - Avoid lengthy work before utility tray animation. r=mhenretty

Cwiiis:


try-server-hook:
[Cwiiis Chris Lord (Cwiiis)](https://github.com/Cwiiis) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=4601a8a9c1ec)

try-server-hook:
[Cwiiis Chris Lord (Cwiiis)](https://github.com/Cwiiis) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=9e27ad4ceb4d)

try-server-hook:
[Cwiiis Chris Lord (Cwiiis)](https://github.com/Cwiiis) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=9ac7475b914d)

apps/system/js/utility_tray.js
```diff
@@ -97,13 +100,33 @@ window.UtilityTray = {
     Service.register('makeAmbientIndicatorInactive', this);
   },
 
+  /*
+   * The time after which we shouldn't use motion events to determine the
+   * speed of the utility tray opening/closing, in milliseconds.
+   */
+  VELOCITY_THRESHOLD: 50,
```
mikehenrty: nit: name is a little misleading, since this value is in ms an not px/ms. What about something like `SWIPE_TIME_THRESHOLD`?

apps/system/js/utility_tray.js
```diff
@@ -236,7 +259,15 @@ window.UtilityTray = {
           evt.preventDefault();
         }
 
+        var oldDelta = this.lastDelta;
+
         this.onTouchMove(evt.touches[0]);
+
+        var move = this.lastDelta - oldDelta;
+        if (Math.abs(move) > 0) {
```
mikehenrty: What about in the case where we are touch moving with a constant velocity? As unlikely as that seems, if we had two identical delta Ys in a row we wouldn't count that as a move, and I'm not sure that is what we want here?

apps/system/js/utility_tray.js
```diff
@@ -236,7 +259,15 @@ window.UtilityTray = {
...
+          this.lastTime = evt.timeStamp;
+          this.lastMove = move;
+        }
```
mikehenrty: A couple of notes here. First, this logic would make more sense to me if it were inside the `onTouchMove` function. Relying on `onTouchMove' to modify `this.lastDelta` seems a little funky and would be clearer if it was inline in the function. I guess we would need to pass the event timeStamp into the `onTouchMove`.

Second, `this.lastMove` is the change in Y deltas between confirmed touched moves, which means it's giving us an acceleration/deceleration if I'm understanding it correctly. I guess the onTouchMove might not be called with constant dt, but assuming performance is relatively consistent, I think `this.lastMove` is more or less telling us the change in swipe velocity. Assuming that is the case, how does change in velocity vs. whether or not the utility tray is shown [1] tell us whether or not to use velocity based animation? Shouldn't we be more concerned about velocity of the swipe (ie whether that is negative or positive) rather than acceleration of the swipe?

1.) https://github.com/mozilla-b2g/gaia/pull/28842/files#diff-690ac4f812d69b66c808d3e430042345R432

apps/system/js/utility_tray.js
```diff
@@ -251,6 +282,7 @@ window.UtilityTray = {
           return;
         }
 
+        this.lastTime = evt.timeStamp - this.lastTime;
```
mikehenrty: nit: it seems weird that `this.lastTime` starts life as a timestamp, but then becomes a time delta. I think better would be to call it `this.touchStartTime` or something above, and then in the `onTouchEnd` function itself we could calculate the delta time as a local var. Like the above comment about passing in the event timeStamp to `onTouchStart` we would have to do the same thing here with the event timeStamp to `onTouchEnd`.

apps/system/js/utility_tray.js
```diff
@@ -396,6 +428,17 @@ window.UtilityTray = {
       var significant = (Math.abs(this.lastDelta) > (this.screenHeight / 5));
       var shouldOpen = significant ? !this.shown : this.shown;
 
+      if (significant && this.lastTime <= this.VELOCITY_THRESHOLD &&
```
mikehenrty: I'm not sure I understand what the velocity threshold does. It seems like the spirit is to prevent a really slow drag from becoming a really slow animation? In that case, would it be better to just calculate the final velocity, and use a velocity threshold based on px/ms rather than just ms? Right now, with how `this.lastTime` gets updated (which appears to be only when the swipe is accelerating/decelerating) it's unclear what this does.

apps/system/js/utility_tray.js
```diff
@@ -434,64 +477,98 @@ window.UtilityTray = {
...
+    if (instant || style.transform === '') {
+      this.afterHide();
+    } else if (this.hideStartCallback === null) {
+      // We want to remove the utility-tray class from the screen at the start
+      // of the animation, but if we do it outside of a timeout, the work will
+      // align with the start of the animation and cause a noticeable delay.
+      this.hideStartCallback = setTimeout(() => {
+        this.hideStartCallback = null;
+        this.screen.classList.remove('utility-tray');
+      }, 20);
```
mikehenrty: Neat trick!

apps/system/js/utility_tray.js
mikehenrty: Ahhh, sorry I had misunderstood what `this.lastDelta` was. It's not the delta from the last touch move, but the delta from touch start. So you can ignore that second part from the comment above.

apps/system/js/utility_tray.js
mikehenrty: Ignore this comment, I misunderstood what `this.lastDelta` is.

try-server-hook:
[Cwiiis Chris Lord (Cwiiis)](https://github.com/Cwiiis) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=58ae7ca9c084)

PR #28846, opened 2015-03-12
Bug 1141559 - update time while seeking

russnicoletti:


autolander:

try-server-hook:
[russnicoletti Russ Nicoletti (russnicoletti)](https://github.com/russnicoletti) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=7cae5609bcc3)

apps/video/js/video.js
```diff
@@ -827,13 +827,10 @@ function movePlayHead(percent) {
...
-  // Don't do updates while we're seeking because the position we fastSeek()
-  // to probably isn't exactly where we requested and we don't want jerky
-  // updates
-  if (dom.player.seeking) {
-    return;
-  }
+  // Bug 1141559 -- update the slider when we get a 'seeked' event so that
+  // the elapsed time and play head move in real time while the slider is
+  // being dragged. Note that this will result in some "jeryky" updates for
+  // videos that lack keyframes for significant sections.
```
davidflanagan: I've got issues with this comment.
1) Including the bug number here isn't very helpful since you're only deleting code. If you were marking code that you had added and weren't sure about whether it should remain, then adding the bug number here might be helpful. Or if the comment included instructions about what code to add back in, maybe it would make sense. But anyway, the bug number doesn't seem helpful here.

2) misspelling of jerky

3) This function gets called for both timeupdate and seeked events, so the comment is kind of misleading since it only talks about seeked events.  Also, this function does not move the play head while dragging because that is done elsewhere, so that part of the comment is wrong, too.  And, I don't think "real time" is the correct way to describe what is happening here.  "real time" would be if you updated the time in the code that handles the drags and moves the slider while dragging. But in this case, you're only doing occasional updates to the time when a fastSeek finishes

apps/video/js/video.js
```diff
@@ -827,13 +827,10 @@ function movePlayHead(percent) {
 }
 
 function updateVideoControlSlider() {
-  // We update the slider when we get a 'seeked' event.
-  // Don't do updates while we're seeking because the position we fastSeek()
-  // to probably isn't exactly where we requested and we don't want jerky
-  // updates
-  if (dom.player.seeking) {
-    return;
-  }
```
davidflanagan: I wonder if the real issue here was that by the time we got one seeked event, we had already called fastSeek() again, so we weren't updating the slider even when we did get the seeked events.

try-server-hook:
[russnicoletti Russ Nicoletti (russnicoletti)](https://github.com/russnicoletti) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=1ed98a5e9262)

PR #28854, opened 2015-03-12
Bug 1039449 - Show and edit DevTools device name. r=janx

jryans:


apps/settings/locales/settings.en-US.properties
```diff
@@ -1130,8 +1130,9 @@ draw-tile-borders=Draw tile borders
 draw-layer-borders=Draw layer borders
 log-animations=Log slow animations
 remote-debugging-usb=Debugging via USB
-dev-tools-wifi=DevTools via Wi-Fi
-dev-tools-qps=Pseudo-localization
+devtools-wifi=DevTools via Wi-Fi
+devtools-device-name=Device Name
```
jankeromnes: Nit: We only capitalize the first letter and acronyms, so `Name` should be `name`.

apps/settings/locales/settings.en-US.properties
```diff
@@ -1130,8 +1130,9 @@ draw-tile-borders=Draw tile borders
 draw-layer-borders=Draw layer borders
 log-animations=Log slow animations
 remote-debugging-usb=Debugging via USB
-dev-tools-wifi=DevTools via Wi-Fi
-dev-tools-qps=Pseudo-localization
+devtools-wifi=DevTools via Wi-Fi
```
jankeromnes: Please keep in mind that changing `l10n-id`s causes re-translation work, which is unnecessary when there is no string change.

jryans: I am aware... so you'd prefer I revert the ID changes then?

apps/settings/locales/settings.en-US.properties
jryans: Hmm, is this a rule for the project? There seems to quite a mix of casing policies in this file, so it certainly doesn't seem obvious that there is "one way". Also, English [title casing rules][1] are typically used for labels like this. Really, I don't care which way it is, but what is the rule and why?

[1]: http://www.grammar-monster.com/lessons/capital_letters_title_case.htm

jankeromnes: This file localizes many different screens in settings, so let's look at the developer menu in particular. Judging from the settings there, the consensus appears to be "Sentence case" (caps only for first letter, letter after punctuation, acronyms like "Wi-Fi" or "ADB" and proper nouns like "Marketplace"). There are only two offenders right now: "Cards View: Screenshots" and "Update Channel".

There have been many back and forths in the past. If you feel like it, you could change the whole menu back to "Title Case" (as the [UX spec](https://bugzilla.mozilla.org/show_bug.cgi?id=995124) originally wanted), but I'll let you figure out what to do with smaller works like in "Show screen reader settings", "Shake to save system log" or "Geolocation output in ADB".

jryans: Alright, I'll update the strings in the developer panel. They won't need ID changes / re-translation, since it's an English-only update.

try-server-hook:
[jryans J. Ryan Stinnett (jryans)](https://github.com/jryans) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=64fd30d1a0f1)

apps/settings/locales/settings.en-US.properties
```diff
@@ -1236,18 +1154,102 @@ licensing-2      = All of the source code to this product is available under lic
...
-verbose-app-permissions=Verbose app permissions
-window-management=Window management
+software-home-button=Software Home Button
+home-gesture=Home Gesture
+continuous-transition=Continuous Transition
+debug-gaia-enabled=Gaia Debug Traces
+showScreenReaderSettings=Show Screen Reader Settings
+app-transition=App Transition
+app-suspending=App Suspending
+copypaste=Copy / Paste
```
jankeromnes: Nit: `Copy/Paste` or `Copy and Paste`.

jankeromnes:
Note: Please fix the conflict with commit e981284e3e909499f15be8b1341ac183789344e6 where Jessica Jong added the string `network-debugging=Network output in ADB`.

jryans:
Okay, I've fixed the `network-debugging` conflict. Also, I had misspelled `console-enabled` value, so I fixed that too.

apps/settings/locales/settings.en-US.properties
jryans: I suppose I leave it as `Copy/Paste`, as `and` could imply new translations and a new string ID.

jryans:
All nits resolved!

PR #28858, opened 2015-03-13
Bug 1141958 - Refine the change event handler to prevent some unexpected reasons. f=justindarc,mikeh

alphan102:
For the state of device storage, camera app only need to take care about "available", "unavailable", "shared".
By doing this, the state of device storage will be fine if there is a new added reason for change event.


try-server-hook:
[alphan102 alphan102](https://github.com/alphan102) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=82fcda0a0368)

PR #28863, opened 2015-03-13
bug 1140247 - support to wrap no name blob for sending file via bluetooth, r=djf, dominic

ian-liu:


autolander:

try-server-hook:
[ian-liu Ian Liu (ian-liu)](https://github.com/ian-liu) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=9565bf9806df)

apps/bluetooth/js/transfer.js
```diff
@@ -195,36 +196,89 @@ navigator.mozL10n.once(function showPanel() {
...
+            console.warn('Unable to write the file: ' + this.error);
+            defaultAdapter.sendFile(targetDevice.address, blob);
+            var msg = 'add name in file failed so that blob is sending ' +
+                      'without filename ' + this.error && this.error.name;
+            debug(msg);
+            if (--numberOfTasks === 0) {
+              transferred();
+            }
+          };
+        }
```
davidflanagan: There is lots of repeated code that could be factored out into a utility function to call sendFile(), print the debug message, decrement numberOfTasks, etc.

apps/bluetooth/js/transfer.js
```diff
@@ -195,36 +196,89 @@ navigator.mozL10n.once(function showPanel() {
...
+                transferred();
+              }
+            };
+
+            getRequest.onerror = function() {
+              defaultAdapter.sendFile(targetDevice.address, blob);
+              var msg = 'getFile failed after added name succeed so that ' +
+                        'blob is sending without filename ' +
+                        getRequest.error && getRequest.error.name;
+              debug(msg);
```
davidflanagan: Is it actually possible to send a blob without a filename? If it is, then why don't we just do that and not bother with the temporary file?  And if it is not possible to do that, then why even call sendFile() if you know it is going to fail?

apps/system/js/bluetooth_transfer.js
```diff
@@ -643,6 +645,33 @@ var BluetoothTransfer = {
...
     CustomDialog.show('cannotOpenFile', body, confirm, null, screen)
     .setAttribute('data-z-index-level', 'system-dialog');
-  }
+  },
 
+  _cleanupTempFileInDir: function bt__cleanupTempFileInDir(evt) {
+    // Try to get the file from bluetooth temp folder.
+    var filePath = '.bluetooth/tmp/' + evt.fileName;
+    var storageType = 'sdcard';
+    var storage = navigator.getDeviceStorage(storageType);
```
davidflanagan: during a long file transfer, the user might have time to go to the settings app and change the default sdcard on devices that have more than one. And if the do that, then I don't think this code will properly delete the file.  Is the actual blob stored in the transferInfo object? If so, you could get the full path from that and not have to re-create the path here.

apps/system/js/bluetooth_transfer.js
```diff
@@ -643,6 +645,33 @@ var BluetoothTransfer = {
     var body = {id: 'unknownMediaTypeToOpenFile', args: {fileName: fileName}};
     CustomDialog.show('cannotOpenFile', body, confirm, null, screen)
     .setAttribute('data-z-index-level', 'system-dialog');
-  }
+  },
 
+  _cleanupTempFileInDir: function bt__cleanupTempFileInDir(evt) {
```
davidflanagan: This function is declared to take an evt, but you're passing evt.transferInfo to it, which is confusing.

apps/bluetooth/js/transfer.js
```diff
@@ -195,36 +196,89 @@ navigator.mozL10n.once(function showPanel() {
...
+            if (--numberOfTasks === 0) {
+              transferred();
+            }
+          };
+        } else {
+          // There is no filepath, apply approach 2.
+          // Wrap the file with filename for sending.
+          // And the filename is better coming from the requsting app.
+          var TMPDIR = '.bluetooth/tmp';
+          var filename = TMPDIR + '/' + activity.source.data.filenames[index];
```
davidflanagan: What about the case where the app that initiates the activity does not specify a filenames array either?  In that case, you'll need to create a temporary filename.  It seems likely to me that an app that doesn't specify paths wouldn't specify filenames either.

apps/bluetooth/js/transfer.js
ian-liu: It is possible to send a blob without a filename. And the file will be named 'Unknown'. Above code in case of `onerror` is a worst case. At least, we can send a file successfully.

apps/system/js/bluetooth_transfer.js
ian-liu: Yes, it's a potential issue there. We don't keep the instance of device storage because the instance is possible out of useful in case of changed default media storage. I hit the problem before.
Bug 1110068 - [Settings]DUT can't receve file via bluetooth after you switch media location to internal.
Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1110068

Bug 1115496 - [DeviceStorage] Provide event from outdated DeviceStorage instance object to notify Gaia inconsistently.
https://bugzilla.mozilla.org/show_bug.cgi?id=1115496

We have to considerate all of the use case since default media storage is possible to be changed any time. Thanks for reminder here.

PR #28866, opened 2015-03-13
Bug 1129701 - RTL improvements on the search results r=mhenretty

albertopq:


try-server-hook:
[albertopq albertopq (albertopq)](https://github.com/albertopq) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=5dad901ffc20)

apps/search/style/search.css
```diff
@@ -17,13 +17,6 @@ html, body {
   font-size: 10px;
 }
 
-/* RTL fixes being tracked in bug 1008013 */
-/* Get rid of this rule when http://bugzil.la/1139925 is fixed,
- see http://bugzil.la/1137547 for details. */
-html[dir="rtl"] {
-  direction: ltr;
-}
-
```
mikehenrty: :+1:

apps/search/style/search.css
```diff
@@ -61,14 +56,16 @@ body, a {
...
+  -moz-padding-end: 2rem;
+  -moz-padding-start: 1rem;
 }
 
 #suggestions-provider::before {
   content: ' ';
-  position: absolute;
-  bottom: 0.2rem;
-  right: .8rem;
+  position: relative;
```
mikehenrty: This isn't really working. The icon needs to be displayed on the right in LTR mode, and the left in RTL. Relative positioning makes this hard, and right now it always displays on the left. One alternative is to use absolute positioning and use `offset-inline-end` (which is what Ahmed did). But I'll leave the final fix up to you.

apps/search/style/search.css
```diff
@@ -297,6 +299,14 @@ html[dir="rtl"] #suggestions-provider {
   padding: 0 1rem 0 2rem;
 }
 
+html[dir="rtl"] #suggestions li {
+  background-position: right 0.8rem center;
+}
+
+html[dir="rtl"] .local .result:after {
+  right: 0.5rem;
```
mikehenrty: Why do we need this right offset in RTL mode, but not in LTR mode? Ideally we would not need this RTL special rule, and instead we could just specify both `right` and `left` as `.5rem` and allow layout to take care of which rule to apply.

try-server-hook:
[albertopq albertopq (albertopq)](https://github.com/albertopq) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=c6753e571ace)

apps/search/style/search.css
albertopq: This rule avoided the regression with the horizontal scrolling, but you are right, setting both is the best option. Thanks!

apps/search/style/search.css
albertopq: Doing it as Ahmed did seems the right way to do it, but it didn't work for me (neither to him, because it caused the regression about the dropdown in LTR languages). I tried with relative positions, but you are right that is harder. I'm having some issues with RTL and pseudo-elements (as gmarty was pointing out in https://github.com/mozilla-b2g/gaia/pull/28866#discussion-diff-26511274L20) so I ended up using a new element.

PR #28870, opened 2015-03-13
Bug 1139437 - [Calendar] simplify the way elements are built on week and day views r=gaye

millermedeiros:
decided to do this because all the `setAttribute` calls was making it really hard to understand, this way it is at least easier to understand what is the HTML structure...

It did not improve as much as I expected, but at least I cleaned the test files and now it shold be a little bit easier to make updates to the markup structure..

### DIFF
--- a/apps/calendar/js/templates/multi_day.js
+++ b/apps/calendar/js/templates/multi_day.js
@@ -1,0 +1,96 @@
+define(function(require, exports, module) {
+'use strict';
+
+var DateSpan = require('./date_span');
+var colorUtils = require('utils/color');
+var create = require('template').create;
+
+var multiday = {};
+
+multiday.hour = function() {
+  var hour = this.arg('hour');
+
+  var html = DateSpan.hour.render({
+    hour: hour,
+    format: this.arg('format'),
+    addAmPmClass: this.arg('addAmPmClass'),
+    className: 'md__display-hour'
+  });
+
+  // we use this temporary element just to grab the textContent
+  var tmp = document.createElement('div');
+  tmp.innerHTML = html;
+  var text = tmp.textContent;
+
+  return `<li aria-label="${text}" class="md__hour md__hour-${hour}">
+    ${html}
+  </li>`;
+};
+
+multiday.day = function() {
+  return `<div class="md__day" data-date="${this.h('date')}"></div>`;
+};
+
+multiday.allday = function() {
+  return `<div class="md__allday" data-date="${this.h('date')}">
+    <h1 class="md__day-name" aria-level="2"
+      id="md__day-name-${this.h('id')}"></h1>
+    <div class="md__allday-events"></div>
+  </div>`;
+};
+
+multiday.event = function() {
+  var labels = [];
+  var id = this.h('id');
+  var instance = this.h('instance');
+  var color = this.h('color');
+  var bgColor = colorUtils.hexToBackground(color);
+
+  var titleId = `md__event-${id}-title-${instance}`;
+  labels.push(titleId);
+
+  var location = this.h('location');
+  var locationHtml = '';
+  if (location) {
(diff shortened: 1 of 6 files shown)


try-server-hook:
[millermedeiros Miller Medeiros (millermedeiros)](https://github.com/millermedeiros) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=197ac05255e9)

apps/calendar/js/utils/animated_scroll.js
```diff
@@ -0,0 +1,34 @@
+define(function(require, exports, module) {
+'use strict';
+
+module.exports = function animatedScroll(opts) {
+  var {container, content, scrollTop} = opts;
+
+  var maxScroll = container.scrollHeight - container.clientHeight;
+  scrollTop = Math.min(Math.max(scrollTop, 0), maxScroll);
```
gaye: Why do we need to take the max of `opts.scrollTop` and 0?

apps/calendar/js/utils/animated_scroll.js
```diff
@@ -0,0 +1,34 @@
...
+
+  var diff = container.scrollTop - scrollTop;
+  var seconds = Math.abs(diff) / 500;
```
gaye: Maybe we could expose a `pixelsPerSecond` option and have a default?

apps/calendar/js/templates/multi_day.js
```diff
@@ -0,0 +1,96 @@
+define(function(require, exports, module) {
+'use strict';
+
+var DateSpan = require('./date_span');
+var colorUtils = require('utils/color');
+var create = require('template').create;
+
+module.exports = create({
```
gaye: Do you think these would be cleaner if we did something like

```
var multiday = {};
multiday.hour = function() {};
multiday.day = function() {};
// ...
module.exports = create(multiday);
```

apps/calendar/js/templates/multi_day.js
```diff
@@ -0,0 +1,96 @@
...
+    var hour = this.arg('hour');
+
+    var html = DateSpan.hour.render({
+      hour: hour,
+      format: this.arg('format'),
+      addAmPmClass: this.arg('addAmPmClass'),
+      className: 'md__display-hour'
+    });
+
+    // we use this temporary element just to grab the textContent
```
gaye: This seems super sketchy

apps/calendar/js/utils/animated_scroll.js
millermedeiros: just in case the `scrollTop` is negative or `null` - I really miss a `Math.clamp(val, min, max)` function.

apps/calendar/js/utils/animated_scroll.js
millermedeiros: right now this is only really used on the month/week views.. so I think we should keep it simple and improve it later if needed.

apps/calendar/js/templates/multi_day.js
millermedeiros: makes sense. we would also have the benefit of decreasing 1 indent level.

apps/calendar/js/templates/multi_day.js
millermedeiros: I don't know how we would do this otherwise.. we need to add the `[aria-label]` and there is no easy way to construct this value unless we grab the whole `textContent`. This is basically the same thing we had before, the only difference is that it wasn't a temporary element, it was reading it from the busytime element itself.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=43cc284b5935cc9faf8cdd3022cc658a552d6da1

PR #28902, opened 2015-03-17
Bug 1143886 - Fix getElementById stubs in AppWindow tests

sfoster:
The 'front window should dispatch events on its element' test started to fail when I fixed the problem here. As far as I can tell, when an AppWindow has a rearWindow it should publish events on its element and so the listener on the parent (which is not a DOM parent in this case) should /not/ have its listener fired.


try-server-hook:
[sfoster Sam Foster (sfoster)](https://github.com/sfoster) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=e960505d1079)

PR #28908, opened 2015-03-17
Bug 1138718 - Update Camera's Notification API to use L10n API

zbraniecki:
https://bugzilla.mozilla.org/show_bug.cgi?id=1138718

### DIFF
--- a/apps/camera/js/controllers/settings.js
+++ b/apps/camera/js/controllers/settings.js
@@ -304,24 +304,24 @@ SettingsController.prototype.notify = function(setting, flashDeactivated) {
   if (dontNotify) { return; }
 
   var localizeOption = setting.get('optionsLocalizable') !== false;
-  var title = this.l10nGet(setting.get('title'));
-  var optionTitle = setting.selected('title');
-  var html;
-
+  var title = '<span data-l10n-id="' + setting.get('title') + '"></span>';
   // Localize option title only if not specified in the config
-  optionTitle = localizeOption ? this.l10nGet(optionTitle) : optionTitle;
+  var optionTitle = localizeOption ? 
+    '<span data-l10n-id="' + setting.selected('title') + '"></span>' :
+    '<span>' + optionTitle + '</span>';
+  var html;
 
   // Check if the `flashMode` setting is going to be deactivated as part
   // of the change in the `hdr` setting and display a specialized
   // notification if that is the case
   if (flashDeactivated) {
     html = title + ' ' + optionTitle + '<br/>' +
-      this.l10nGet('flash-deactivated');
+      '<span data-l10n-id="flash-deactivated"></span>';
   } else {
     html = title + '<br/>' + optionTitle;
   }
 
-  this.notification.display({ text: html });
+  this.notification.display({ text: {html: html} });
 };
 
 /**
--- a/apps/camera/js/views/notification.js
+++ b/apps/camera/js/views/notification.js
@@ -30,7 +30,12 @@ module.exports = View.extend({
    *
    * Options:
    *
-   *   - `text {String}`
+   *   - `text {L10nId}`
+   *         L10nId may be:
+   *         a string -> l10nId
+   *         an object -> {id: l10nId, args: l10nArgs}
+   *         an object -> {html: string}
+   *       read more at: http://mzl.la/1A76Fby
    *   - `className {String}`
    *   - `persistent {Boolean}`
    *
@@ -45,7 +50,16 @@ module.exports = View.extend({
 
     item.el = document.createElement('li');
     item.el.className = options.className || '';
-    item.el.innerHTML = '<span>' + options.text + '</span>';
+
+    var span = document.createElement('span');
+    if (typeof(options.text) === 'string') {
+      span.setAttribute('data-l10n-id', options.text);
+    } else {
+      if (options.text.html) {
+        span.innerHTML = options.text.html;
+      }
+    }
+    item.el.appendChild(span);
     if (options.attrs) { this.setAttributes(item.el, options.attrs); }
     // Notification should have the semantics of the status ARIA live region.
     item.el.setAttribute('role', 'status');
--- a/apps/camera/test/unit/controllers/settings_test.js
+++ b/apps/camera/test/unit/controllers/settings_test.js
@@ -217,17 +217,6 @@ suite('SettingsController#configureRecorderProfiles()', function() {
       this.controller.notify(this.setting);
       sinon.assert.notCalled(this.notification.display);
     });
-
-    test('It localizes the option title', function() {
-      this.controller.notify(this.setting);
-      sinon.assert.calledWith(this.app.l10nGet, 'l10n-key');
-    });
-
-    test('It doesn\'t localize the title if `localizeOption` is `false`', function() {
-      this.setting.get.withArgs('optionsLocalizable').returns(false);
-      this.controller.notify(this.setting);
-      assert.isFalse(this.app.l10nGet.calledWith('l10n-key'));
-    });
   });
 
   suite('SettingsController#validMenuItem()', function() {
--- a/apps/camera/js/controllers/battery.js
+++ b/apps/camera/js/controllers/battery.js
@@ -27,7 +27,6 @@ function BatteryController(app) {
   this.battery = app.battery || navigator.battery || navigator.mozBattery;
   this.levels = app.settings.battery.get('levels');
   this.notification = app.views.notification;
-  this.l10nGet = app.l10nGet;
   this.bindEvents();
   this.updateStatus();
   debug('initialized');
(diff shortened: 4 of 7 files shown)


try-server-hook:
[zbraniecki Zibi Braniecki (zbraniecki)](https://github.com/zbraniecki) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=b4aea10ac06f)

try-server-hook:
[zbraniecki Zibi Braniecki (zbraniecki)](https://github.com/zbraniecki) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=f363727bc716)

try-server-hook:
[zbraniecki Zibi Braniecki (zbraniecki)](https://github.com/zbraniecki) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=cbab752c2804)

try-server-hook:
[zbraniecki Zibi Braniecki (zbraniecki)](https://github.com/zbraniecki) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=227585ac0863)

try-server-hook:
[zbraniecki Zibi Braniecki (zbraniecki)](https://github.com/zbraniecki) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=884f15875b8c)

apps/camera/js/views/notification.js
```diff
@@ -45,7 +45,16 @@ module.exports = View.extend({
 
     item.el = document.createElement('li');
     item.el.className = options.className || '';
-    item.el.innerHTML = '<span>' + options.text + '</span>';
+
+    var span = document.createElement('span');
+    if (typeof(options.text) === 'string') {
+      span.setAttribute('data-l10n-id', options.text);
+    } else {
+      if (options.text.html) {
```
wilsonpage: Nit: `} else if (options.text.html) {`

apps/camera/js/views/notification.js
```diff
@@ -45,7 +45,16 @@ module.exports = View.extend({
 
     item.el = document.createElement('li');
     item.el.className = options.className || '';
-    item.el.innerHTML = '<span>' + options.text + '</span>';
+
+    var span = document.createElement('span');
+    if (typeof(options.text) === 'string') {
```
wilsonpage: NIT

```js
if (typeof options.text === 'string') {
```

apps/camera/js/controllers/settings.js
```diff
@@ -304,24 +304,24 @@ SettingsController.prototype.notify = function(setting, flashDeactivated) {
...
   if (flashDeactivated) {
     html = title + ' ' + optionTitle + '<br/>' +
-      this.l10nGet('flash-deactivated');
+      '<span data-l10n-id="flash-deactivated"></span>';
   } else {
     html = title + '<br/>' + optionTitle;
   }
 
-  this.notification.display({ text: html });
+  this.notification.display({ text: {html: html} });
```
wilsonpage: Seems a little convoluted:
```js
this.notification.display({ text: {html: html} });
```

I think it might be clearer to alter to API like so:

```js
this.notification.display({ l10nId: 'some-id' });
this.notification.display({ html: '<span data-icon="battery-low></span><span data-l10n-id="batter-low"></span>' });
```

apps/camera/test/unit/controllers/settings_test.js
```diff
@@ -217,17 +217,6 @@ suite('SettingsController#configureRecorderProfiles()', function() {
...
-    test('It localizes the option title', function() {
-      this.controller.notify(this.setting);
-      sinon.assert.calledWith(this.app.l10nGet, 'l10n-key');
-    });
-
-    test('It doesn\'t localize the title if `localizeOption` is `false`', function() {
-      this.setting.get.withArgs('optionsLocalizable').returns(false);
-      this.controller.notify(this.setting);
-      assert.isFalse(this.app.l10nGet.calledWith('l10n-key'));
-    });
```
wilsonpage: We still need a test to prove this works.

apps/camera/js/controllers/settings.js
zbraniecki: the reason why I went for this is to match https://developer.mozilla.org/en-US/Firefox_OS/Developing_Gaia/localization_code_best_practices#Writing_APIs_that_operate_on_L10nIDs

We don't specify it as a separate class, because that would cost mem, but in general, we use a L10nID object type that can be a string - 'l10n-id' or an object that looks like this:
 - {id: 'l10n-id', args: {}}
 - {raw: 'raw string'}
 - {html: 'html string'}

We start by supporting only types that you need, but this enables the API to be predictable and future compatible if you later will need to support more of it.

What you are suggesting here will work for this case, but will be different from what we're exposing everywhere.

What I'm suggesting in this patch is making "text" element of the options object be of that l10nId object type.
This way, if you will have another l10nId type (subText for example) later, you just treat it as L10nID object type and cover all scenarios.

Does it make sense?

apps/camera/test/unit/controllers/settings_test.js
zbraniecki: yeah, I'm not sure how. With the change, we don't call any functions that we could spy on, and the easiest way to test would be to reach out to DOM and see if the right attribute is set, but it seems to me that tests in controllers don't touch DOM.

What should I do?

apps/camera/js/controllers/settings.js
wilsonpage: This seems to make sense. Can you please leave a commenting pointing to that article to indicate the convention that is being followed.

wilsonpage: Although, from what I gathered from the article, shouldn't our API look more like:

```js
this.notification.display('my-l10n-id');
this.notification.display({ html: 'some <em>html</em>' });
this.notification.display({ raw: 'pre-localized text' });
this.notification.display({ id: 'something', args: 'something' });
```

The current approach, doesn't seem to conform with the suggested convention:

```js
this.notification.display({ text: {html: html} });
``

apps/camera/test/unit/controllers/settings_test.js
wilsonpage: You can spy on `notification.display()` and check that the HTML passed to it is correctly formed.

apps/camera/js/controllers/settings.js
zbraniecki: That would work if we're certain that display will need only one L10nId as a param and nothing else. But it seems to me that the there may be other params in the object passed to display() and I turned the one called "text" to be an L10nId.

Would you prefer me to turn it into the API you proposed?

wilsonpage: 9 times out of 10 the notification will just be an l10nId, the only case I can think of when it's not is the 'low battery' notifications which also come with an icon to indicate the battery level. I'm pretty sure `l10nId` with the option to use an HTML string for more complex notifications will fulfill all current and future use-cases.

Have I overlooked something?

zbraniecki: Hmmm... So, here's a case where notification.display takes more than l10nId:
https://github.com/mozilla-b2g/gaia/blob/master/apps/camera/js/controllers/battery.js#L114-L124

The current "text" and "attrs" should be turned into a single L10nId object {id: ..., args: ...}, but persistent and className are separate. So my approach is to turn it into:

`notification.display({
 text: l10nId,
 className: ...,
 persistent: ...
});`

and now the "text" may be a simple l10n-id:
`notification.display({
  text: 'my-string-id'
});`

or sth more complex:
`notification.display({
  text: {'id': 'my-string', args: {'name': 'John'}}
});`

it seems to me that your approach was more to do sth like this:

`notification.display({
  'id': ...,
  'args': ...,
  'persistent': ....,
  'className': ...,
});`

I'd prefer to keep L10nId object separately to keep it from potentially overlapping in a scenario like this:

1) We take your proposal
2) You add "raw" attribute next to persistent and className for some additional data
3) We want to provide "raw" variant of L10nId 
4) Attr name collision.

For that reason I'd keep text as an attribute.

Does it sound sane to you? It's your call, so if you prefer flat model, I'll write it your way.

wilsonpage: Ah I forgot about the `persistent` option. Can you remind me what `args` represents? I'm happy to just roll with your suggestion, it's no biggie.

zbraniecki: https://developer.mozilla.org/en-US/Firefox_OS/Developing_Gaia/localization_code_best_practices#Writing_APIs_that_operate_on_L10nIDs

Those are optional l10n-args, arguments passed from the developer to the localization code, like user name, download count etc.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=bc4cc514beaef590182f555565433b87645603cd

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7c76082b66e85ffd34e7e6bcf90ac7e4e113f7c3

PR #28938, opened 2015-03-18
Bug 1115304 - Add setting 'telephony.audiocompeting.disabled' to disable Audio Competing workaround.

weilonge:


apps/callscreen/js/audio_competing_helper.js
```diff
@@ -60,6 +60,23 @@
      * Init function.
      */
     init: function ach_init(appName) {
+      var self = this;
```
jaoo: Would be nice to default ```this._audioCompetingDisabled``` to false here.

apps/callscreen/js/audio_competing_helper.js
```diff
@@ -60,6 +60,23 @@
...
+      var req = window.navigator.mozSettings.createLock().get(
+        'telephony.audiocompeting.disabled');
+      req.onsuccess = function icc_getAudioCompetingDisabled() {
+        var value = req.result['telephony.audiocompeting.disabled'];
+
+        if (typeof value === 'undefined') {
+          navigator.mozSettings.createLock().set({
+            'telephony.audiocompeting.disabled': true
+          });
+          self._audioCompetingDisabled = true;
```
jaoo: Maybe default this to false if this lands on master.

apps/callscreen/js/audio_competing_helper.js
```diff
@@ -91,6 +113,11 @@
      * audio channel.
      */
     leaveCompetition: function ach_leaveCompetition() {
+      console.log('DEBUG leaveCompetition _audioCompetingDisabled ' +
+        this._audioCompetingDisabled);
+      if (this._audioCompetingDisabled) {
+        return ;
+      }
```
jaoo: As long as the ```compete()```call bails out early this check would not be necessary.

build/config/common-settings.json
```diff
@@ -199,6 +199,7 @@
    "support.callsupport2.title": "",
    "support.callsupport2.href": "",
    "telephony.speaker.enabled": false,
+   "telephony.audiocompeting.disabled": true,
```
jaoo: If this lands on master we should default this to false IMHO.

PR #28954, opened 2015-03-18
Bug 1144682 - Fix string substitution in bidi contexts. r=stas, r=smontagu

tedders1:


apps/ftu/locales/ftu.en-US.properties
```diff
@@ -237,20 +237,20 @@ SIM_mandatory_warm = Please, insert a SIM card and reboot your phone.
 #=#=#=#=#=#=#=#=#=#=#=#=#=#=#=#=#=#=#=#=#=#=#=#=#=#=#=#=#=#=#=#=#=#=#=#=#=#=#=#
 
 telemetry-sends2 = Telemetry sends Mozilla data about performance, hardware, configurations, customizations and app usage in your {{brandShortName}} device. We use this to make our products better, faster and more responsive.
-learn-more-telemetry2.innerHTML = Learn more at <a href="{{href}}" class="{{className}}">{{textContent}}</a>
-learn-more-information2.innerHTML = Learn how Mozilla handles your information at <a href="{{href}}" class="{{className}}">{{textContent}}</a>
+learn-more-telemetry2.innerHTML = Learn more at <a href="[[href]]" class="[[className]]">{{textContent}}</a>
```
sfoster: Nit: I think we need to increment all these string IDs here, right?

PR #28962, opened 2015-03-18
Bug 1141490 - [NFC] Refine NfcManager state transitions

tauzen:


apps/system/js/nfc_manager.js
```diff
@@ -20,7 +20,19 @@
 'use strict';
 
 (function(exports) {
-  var DEBUG = false;
+
+  const NfcHwEvents =
```
alivedise: Nit: we tend to use all upper case naming for const value.

apps/system/js/nfc_manager.js
```diff
@@ -20,7 +20,19 @@
...
+    ['enable', 'disable', 'enable-polling', 'disable-polling',
+     'hw-change-success', 'hw-change-failure'];
+
+  const NfcHwStateTable = {
+    'disabling': [null, null, null, null, 'disabled', 'enabled'],
+    'disabled': ['enabling', null, null, null, null, null],
+    'enabling': [null, null, null, null, 'enabled', 'disabled'],
+    'enabled': [null, 'disabling', 'polling-on', 'polling-off', null, null],
+    'polling-on': [null, 'disabling', null, 'polling-off', null, null],
```
alivedise: Could you explain what does polling mean?

apps/system/js/nfc_manager.js
```diff
@@ -200,134 +186,67 @@
...
-      this._debug('_nfcSettingsChanged, nfc.enabled: ' + enabled);
+      this._hwState = state;
+      this._notifyNfcStateChanged();
+      this._changeNfcHwState();
+    },
 
-      if (this.isActive() === enabled || this.isInTransition()) {
-        this._debug('_nfcSettingsChanged ignoring settings change');
-        return;
+    _notifyNfcStateChanged: function() {
```
alivedise: It seems to me what we do here could be merged to this._changeNfcHwState(); and then rename it to processStateChange...is there a reason that we have two functions?

apps/system/js/nfc_manager.js
```diff
@@ -200,134 +186,67 @@
...
-     * @param {string} state - new hardware state, one of
-     * {@link NfcManager#NFC_HW_STATE}
-     */
-    _changeHardwareState: function nm_changeHardwareState(state) {
-      this._debug('_changeHardwareState - state : ' + state);
-      this._hwState = state;
+    _changeNfcHwState: function() {
       var nfcdom = window.navigator.mozNfc;
       if (!nfcdom) {
+        this._doNfcStateTransition('hw-change-failure');
```
alivedise: We will always have navigator.mozNfc is NfcManager is loaded. So I wonder if we need this.

apps/system/js/nfc_manager.js
```diff
@@ -200,134 +186,67 @@
...
-      // event dispatching to handle statusbar change
-      // TODO remove in Bug 1103874
-      var event = new CustomEvent('nfc-state-changed', {
-        detail: {
-          active: isOn
-        }
-      });
-      window.dispatchEvent(event);
+      promise.then(() => this._doNfcStateTransition('hw-change-success'))
+      .catch(() => this._doNfcStateTransition('hw-change-failure'));
```
alivedise: Nice :)

apps/system/js/nfc_manager.js
```diff
@@ -358,21 +277,31 @@
...
-          window.dispatchEvent(new CustomEvent('shrinking-stop'));
+          this.debug('CheckP2PRegistration failed');
+          this._cleanP2PUI();
         }
       });
     },
 
+    _cleanP2PUI: function() {
+      window.removeEventListener('shrinking-sent', this._handleShrinkingSent);
+      window.dispatchEvent(new CustomEvent('shrinking-stop'));
```
alivedise: nit: this.publish('shrinking-stop', this, /* without prefix */ true);

apps/system/js/nfc_manager.js
```diff
@@ -358,21 +277,31 @@
...
+    },
+
+    _initP2PUI: function() {
+      window.dispatchEvent(new CustomEvent('shrinking-start'));
```
alivedise: nit: <code>this.publish('shrinking-start', this, /* without prefix */ true);</code>

apps/system/js/nfc_manager.js
```diff
@@ -358,21 +277,31 @@
...
+
+      this._handleShrinkingSent = () => {
+        this._cleanP2PUI();
+        this._dispatchP2PUserResponse();
+      };
+
+      window.addEventListener('shrinking-sent', this._handleShrinkingSent);
```
alivedise: I forgot why we need to dynamically add/remove event listener here. Is there any chance we could always listen to it?

apps/system/js/nfc_manager.js
tauzen: Sure, will add comments explaining this.

apps/system/js/nfc_manager.js
tauzen: Yes, I agree that that these functions can be merged. I actually separated them because of the different actions they perform. ```this._notifyNfcStateChanged``` notifies external components (icon, nfc toggle via ```nfc.status``` setting) about the Nfc state change. ```this._changeNfcHwState``` asks Gecko to perform HW state change.

apps/system/js/nfc_manager.js
tauzen: You're right (just made a quick look at ```core.js```), I will remove this.

apps/system/js/nfc_manager.js
tauzen: I think we should keep this the way it is for now. This is part of the (too-complicated) sharing flow. I think that If we would listen all the time for ```shrinking-sent``` this would mean that ```ShrinkingUI``` would be controlling the nfc sharing, and could trigger not necessary requests to Gecko (by mistake, or misuse).
Since we have ```onpeerfound``` available I think we could start to investigate using it in gaia apps, and let the apps control ```ShrinkingUI```. As result we could remove P2P related code from NfcManage. I'll file a followup bug for that.

PR #28976, opened 2015-03-19
Bug 1143594 - [RTL][Contacts]Apply block display bdi workaround for contact name ellipsis direction issue, r=arcturus

steveck-chung:
...

### DIFF
--- a/apps/communications/contacts/elements/details.html
+++ b/apps/communications/contacts/elements/details.html
@@ -1,7 +1,7 @@
 <element name="view-contact-details" extends="section">
   <template>
     <gaia-header id='details-view-header' action="back">
-      <h1 id='contact-name-title'></h1>
+      <h1 id='contact-name-title'><bdi class="ellipsis-dir-fix"></bdi></h1>
       <button id="edit-contact-button" data-icon="edit-contact" data-l10n-id="editButton"></button>
     </gaia-header>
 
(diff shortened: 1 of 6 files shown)


apps/communications/contacts/style/contacts.css
```diff
@@ -1088,6 +1088,13 @@ html[dir="rtl"] nav[data-type="scrollbar"] p {
   right: 10rem;
 }
 
+/* TODO: Remove this workaround once bug 883884 fixed */
+.ellipsis-dir-fix {
+  display: block;
+  text-overflow: ellipsis;
+  overflow: hidden;
```
julienw: do we need white-space: nowrap ? (or maybe we inherit it?)

apps/communications/contacts/elements/details.html
```diff
@@ -1,7 +1,7 @@
 <element name="view-contact-details" extends="section">
   <template>
     <gaia-header id='details-view-header' action="back">
-      <h1 id='contact-name-title'></h1>
+      <h1 id='contact-name-title'><bdi class="ellipsis-dir-fix"></bdi></h1>
```
millermedeiros: you could probably fix the issue by setting `[dir="auto"]` on the `<h1>` or setting `unicode-bidi: isolate` on the CSS

millermedeiros: nevermind.. it will probably make things harder and might cause undesired side effects. `<bdi>` is probably really the way to go.

apps/communications/contacts/style/contacts.css
steveck-chung: yep it's inherit from gaia-header h1

PR #28978, opened 2015-03-19
Bug 1135972 - Direction of 'Next' and 'Back' buttons contradict with page indicator on Guided Tour page +autolander

martapiekarska:


apps/privacy-panel/style/panels.css
```diff
@@ -278,6 +278,10 @@ section[role="region"] > div.gt-box {
   width: 100%;
 }
 
+html[dir=rtl] .gt-pager {
+  transform: rotate(180deg)
```
zbraniecki: I consulted #developers about it and they said that transform: rotate will do more than just flip horizontally, they recommend using scaleX(-1) or image-orientation: flip

PR #28979, opened 2015-03-19
Bug 1145097 - Clean up html syntax of ds-test/index.html R=michielbdejong

michielbdejong:
Move <html> tag into the right place. This was the reason to start this PR, I couldn't deploy to 5apps.com because of this.
But once I had the file open, I also made these changes:
- Fix indentation
- Lower-case 'html' in DOCTYPE
- Add missing lang="en" attribute
- Consistent empty tags. I had to look up which one is preferred, and it seems that the `< ... />` syntax was a thing only in XHTML; in html it [should now always just be `< ... >`](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Introduction#Tags).

UPDATE: also fixed typo "commad" -> "command"

### DIFF
--- a/dev_apps/ds-test/index.html
+++ b/dev_apps/ds-test/index.html
@@ -1,15 +1,15 @@
-  <!DOCTYPE HTML>
+<!DOCTYPE html>
+<html lang="en">
   <head>
-      <meta charset="utf-8" />
-<html>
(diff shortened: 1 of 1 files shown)


michielbdejong:
Thanks!

PR #28988, opened 2015-03-19
Bug 1145279 - fixed a typo for accessibility-key-hint.

yzen:


yzen:
Just a typo fix; merging without try-hook as an exception (confirmed with @KevinGrandon )

PR #28989, opened 2015-03-19
Hotfix for case when gaia_marionette_retry.js receives 0 test files

lambdabaa:


autolander:

PR #29010, opened 2015-03-20
Bug 944644 - [Messages][Refactoring] Replacing <br> with proper lists wh...

julienw:
...ile displaying error messages with the recipients case r=schung

### DIFF
--- a/apps/sms/js/dialog.js
+++ b/apps/sms/js/dialog.js
@@ -13,7 +13,7 @@
  localization_code_best_practices#Writing_APIs_that_operate_on_L10nIDs + 'raw'
  can be Node aw well.
 
- Options should follow the following structure:
+ The constructor parameter should follow the following structure:
 
  {
   title: { raw: 'Non localizable title' },
@@ -21,6 +21,7 @@
     id: 'localizableStringWithArgument',
     args: { n: count }
   },
+  classes: ['specific-class'], // optional
   options: {
     // Cancel is a mandatory option. You need to define at least the text.
     cancel: {
@@ -69,6 +70,10 @@ var Dialog = function(params) {
   this.form.tabIndex = -1;
   // Pick up option_menu.css styling
   this.form.dataset.subtype = 'menu';
+
+  if (params.classes) {
+    this.form.classList.add(...params.classes);
+  }
 
   var infoSection = document.createElement('section');
   infoSection.appendChild(createLocalizedElement('h1', params.title));
--- a/apps/sms/js/error_dialog.js
+++ b/apps/sms/js/error_dialog.js
@@ -36,15 +36,34 @@
       };
     }
 
-    var messageBodyParams = null;
-    if (errorDescription.showRecipient) {
-      messageBodyParams = {
-        n: options.recipients.length,
-        numbers: options.recipients.join('<br />')
-      };
-    }
+    var bodyL10nId = prefix + 'Body';
+    var body = null;
+    var additionalClasses = [];
 
-    if (errorDescription.showDsdsStatus) {
+    if (errorDescription.showRecipient) {
+      var fragment = document.createDocumentFragment();
+      var mainText = document.createElement('span');
+      navigator.mozL10n.setAttributes(
+        mainText, bodyL10nId,
+        { n: options.recipients.length }
+      );
+      fragment.appendChild(mainText);
+
+      var list = options.recipients.reduce((list, recipient) => {
+        var li = document.createElement('li');
+        li.textContent = recipient;
+
+        list.appendChild(li);
+
+        return list;
+      }, document.createElement('ul'));
+
+      fragment.appendChild(list);
+      body = { raw: fragment };
+      additionalClasses.push('error-dialog-show-recipient');
+    } else if (errorDescription.showDsdsStatus) {
+      var messageBodyParams = null;
+
       var mmsServiceId = Settings.mmsServiceId;
       if (mmsServiceId !== null) {
         // mmsServiceId = 0 => default Service is SIM 1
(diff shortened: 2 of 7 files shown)


try-server-hook:
[julienw Julien Wajsberg (julienw)](https://github.com/julienw) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=39830ec13834)

try-server-hook:
[julienw Julien Wajsberg (julienw)](https://github.com/julienw) started tests. [Results](https://treeherder.mozilla.org/ui/#/jobs?repo=gaia-try&revision=1bbb4bf36d0d)

apps/sms/js/errors.js
```diff
@@ -26,7 +26,7 @@
     });
 
     ERRORS.set('FdnCheckError',{
-      prefix: 'fdnBlocked2',
+      prefix: 'fdnBlocked3',
```
steveck-chung: could we refine this name like "fdnBlockedError" or something else?

apps/sms/locales/sms.en-US.properties
```diff
@@ -177,17 +177,17 @@ fdnBlocked2Title                 = FDN active
 # typed, and string in toaster will be 'Editing SMS 2' (editing 2nd segment).
 sms-counter-notice-label         = Editing SMS {{number}}
 
-#LOCALIZATION NOTE (fdnBlocked2Body): the English translation does not need the
+#LOCALIZATION NOTE (fdnBlocked3Body): the English translation does not need the
 # actual number of phone numbers. If your translation needs it, you can still add it
 # using the parameter {{ n }}. Also take into account that you could add HTML so be
 # careful with the content, it may affect to the whole app.
```
steveck-chung: Key changes for `fdnBlocked2Title` is missing here

apps/sms/js/dialog.js
```diff
@@ -21,6 +21,7 @@
     id: 'localizableStringWithArgument',
     args: { n: count }
   },
+  classes: ['specific-class'], // optional
```
steveck-chung: Just curious that if this  `classes` is optional, why don't we move it into `options` below?

apps/sms/js/error_dialog.js
```diff
@@ -36,15 +36,33 @@
...
+      var mainText = document.createElement('span');
+      navigator.mozL10n.setAttributes(
+        mainText, bodyL10nId,
+        { n: options.recipients.length }
+      );
+      fragment.appendChild(mainText);
+
+      var list = options.recipients.reduce((list, recipient) => {
+        var li = document.createElement('li');
+        li.setAttribute('dir', 'ltr'); // phone numbers are always LTR
```
steveck-chung: maybe we can move this to css since you already need a rule for li in FDN error dialog?

apps/sms/js/errors.js
julienw: yeah, even maybe "fdnError" :)

apps/sms/js/error_dialog.js
julienw: ah yeah good idea, I haven't thought of this

apps/sms/js/dialog.js
julienw: I thought that "options" is about the buttons; but I wasn't really sure actually. Do you think it would be better ?

apps/sms/locales/sms.en-US.properties
julienw: oops !

apps/sms/js/dialog.js
steveck-chung: ya the original options seems only for button(or we don't utilize it for other purpose before). I have no preferance about this so let's keep it.

apps/sms/style/sms.css
```diff
@@ -32,6 +32,23 @@ form[role="dialog"][data-type="confirm"] p {
...
 
+form[role="dialog"].error-dialog-show-recipient ul {
+  border: none;
+}
+
+form[role="dialog"].error-dialog-show-recipient ul li {
+  padding: 0;
+}
+
+.error-dialog-show-recipient li {
```
steveck-chung: Why do we need to split it into 2 rules?

julienw: to override building blocks :/

The other solution is to use `!important` :/

apps/sms/js/error_dialog.js
```diff
@@ -71,15 +90,18 @@
...
     Dialog.call(this, {
       title: prefix + 'Title',
-      body: {
-        id: prefix + 'Body',
-        args: messageBodyParams
-      },
-      options: dialogOptions
+      body: body || bodyL10nId,
+      options: dialogOptions,
+      classes: ['error-dialog-' + prefix, 'error-dialog', ...additionalClasses]
```
steveck-chung: Just thinking that do we really need all these classes? it seem that only `error-dialog-show-recipient` is the only rule with styling.

julienw: yes, but I thought it could be useful in the future. Having them would let us have more targeted CSS selectors instead of overriding BB (when we can do it of course...)

apps/sms/style/sms.css
steveck-chung: I see ...  maybe you can add a commet about the override part, or merge `.error-dialog-show-recipient li` into `form[role="dialog"].error-dialog-show-recipient ul li` with comment above the padding.

julienw: there is already a comment a little above, I thought it was enough, but I can add another one here too.

I also prefered to keep the sole BB overrides in its own rule, and have what didn't need to be overriden in the simpler rule... This way we clearly see what we override and what we don't override. And I still hope that we can override in a simpler way in the future

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8cc0e4eaa651c477819f89382bd6ce35c487ccdb

PR #29024, opened 2015-03-20
Bug 1134795 - Show icon when installing web app from web page

benfrancis:


apps/bookmark/js/bookmark_editor.js
```diff
@@ -28,6 +29,8 @@ var BookmarkEditor = {
     this.cancelButton = document.getElementById('cancel-button');
     this.saveButton = document.getElementById('done-button');
     this.appInstallationSection = document.getElementById('app-installation');
+    this.appIcon = document.getElementById('app-icon');
```
crdlc: Up to you, although I would mote these vars to https://github.com/mozilla-b2g/gaia/pull/29024/files#diff-32abdeb6061956598cbede801aa9bc1aR86 and https://github.com/mozilla-b2g/gaia/pull/29024/files#diff-32abdeb6061956598cbede801aa9bc1aR91 respectively

apps/bookmark/js/bookmark_editor.js
```diff
@@ -70,6 +73,25 @@ var BookmarkEditor = {
...
+    // Parse icon URL from app manifest
+    var iconURL = window.WebManifestHelper.iconURLForSize(manifest,
+      manifestURL, size);
+    if (!iconURL) {
+      return;
+    }
+    // Switch out placeholder for real icon when it loads
+    this.appIconListener = this._handleAppIconLoad.bind(this);
+    this.appIcon.addEventListener('load', this.appIconListener);
+    this.appIcon.setAttribute('src', iconURL);
```
crdlc: Just a quick question, why don't we delegate this to Icon object to render this one as it will appear in the homescreen if the user decided to install in the end (round shape, drop shadows, etc...)? Thanks a lot

apps/bookmark/style/save.css
```diff
@@ -69,10 +69,21 @@ body[data-mode="put"] #edit-bookmark-header {
...
 }
 
+#app-icon, #app-icon-placeholder {
+  display: block;
+  width: 6rem;
+  height: 6rem;
+  margin: 0.5rem auto 0 auto;
+}
+
+#app-icon.hidden, #app-icon-placeholder.hidden {
```
crdlc: What do you think about creating a rule like

```css
.hidden {
  display: none;
}
```

removing https://github.com/benfrancis/gaia/blob/1134795/apps/bookmark/style/save.css#L50 and these two new rules?

apps/bookmark/js/bookmark_editor.js
```diff
@@ -79,6 +101,7 @@ var BookmarkEditor = {
         this.installAppButtonListener = this._installApp.bind(this);
         this.installAppButton.addEventListener('click',
           this.installAppButtonListener);
+        this._renderAppIcon(manifestData, manifestURL, this.APP_ICON_SIZE);
```
crdlc: I would like to deal with the icon after showing the section and the name of the app

apps/bookmark/js/bookmark_editor.js
```diff
@@ -79,6 +101,7 @@ var BookmarkEditor = {
...
         this.appInstallationSection.classList.remove('hidden');
         this.appNameText.textContent = manifestData.short_name ||
```
crdlc: and also, if you can, please set the name before showing the section

crdlc:
Last thought, the WebManifestHelper is increasing regarding to code, what do you think to load this library once we know we have a manifestURL? Thanks a lot

apps/bookmark/js/bookmark_editor.js
benfrancis: Oh, does the homescreen do that to app icons? I thought that was only for bookmarks?

apps/bookmark/style/save.css
benfrancis: Unfortunately that wouldn't have enough specificity to have any effect, but I've moved them all into one rule including the IDs.

apps/bookmark/js/bookmark_editor.js
benfrancis: OK, I took a look at the code in /shared/js/homescreens/icon.js

As part of bug 1134657 we're going to want to change that code to use App.getIcon in Gecko to retrieve icons for installed apps rather than do an XHR from Gaia. That won't work for this patch because it's not an installed app so my conclusion is that it's probably better for this use to be implemented separately until such a time as we have a new API to let Gecko safely fetch the icon for not-yet-installed apps. I hope that's OK.

apps/bookmark/js/bookmark_editor.js
benfrancis: Done.

apps/bookmark/js/bookmark_editor.js
benfrancis: Done.

benfrancis:
Good idea about lazy loading the helper. If it gets any bigger let's do that, although I hope this helper will go away completely once we get proper platform support for some of these things.

PR #29030, opened 2015-03-21
Bug 1145875 - Guard against falsy this.element in show/hideScreenshotOverlay

sfoster:
Speculative patch, just guarding against a null/undefined this.element in the _showScreenshotOverlay and _hideScreenshotOverlay methods which were modified in bug 1107139

### DIFF
--- a/apps/system/js/app_window.js
+++ b/apps/system/js/app_window.js
@@ -1286,7 +1286,7 @@
         this.frontWindow._showScreenshotOverlay();
         return;
       }
-      if (!this.screenshotOverlay ||
+      if (!this.element || !this.screenshotOverlay ||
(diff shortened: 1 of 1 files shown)


apps/system/js/app_window.js
```diff
@@ -1286,7 +1286,7 @@
         this.frontWindow._showScreenshotOverlay();
         return;
       }
-      if (!this.screenshotOverlay ||
+      if (!this.element || !this.screenshotOverlay ||
```
geoelectric: Will this make a difference? Looks like `this.screenshotOverlay` was statically assigned from a selector query in line 672, rather than being a dynamic access. Or do we have magic so the query is lazy?

PR #29035, opened 2015-03-21
Update calendar.en-US.properties

orvi2014:


autolander:

PR #29042, opened 2015-03-23
test 1142371 - [Bluetooth] Replace getAdapter with service query

gasolin:
...alive
- use service query to get adapter
- remove bluetooth avaialable event


autolander:

PR #29055, opened 2015-03-23
Bug 1146330 - Try Server Hook not running tests for gaia pull requests

petemoore:
Here we go...


petemoore:
Hmm not sure why autolander has not responded...

petemoore:
Ah, I think it only responds when it cannot read the bug number from the commit message...

PR #29090, opened 2015-03-24
Bug 1125582 - Add wifi connection status check.

seinlin:


apps/settings/js/wifi.js
```diff
@@ -283,8 +283,17 @@ navigator.mozL10n.once(function wifiSettings() {
     // Show connection status
     icon.classList.add('wifi-signal');
     if (WifiHelper.isConnected(network)) {
-      localize(small, 'shortStatus-connected');
-      icon.classList.add('connected');
+      var conStatus = WifiHelper.getConnectionStatus(network);
+      if (conStatus == 'connected') {
```
EragonJ: plz use `===` instead.

apps/settings/js/wifi.js
```diff
@@ -283,8 +283,17 @@ navigator.mozL10n.once(function wifiSettings() {
...
+        localize(small, 'shortStatus-connected');
+        icon.classList.add('connected');
+      }
```
EragonJ: make the coding style consistent with this - `} else if {`

shared/js/wifi_helper.js
```diff
@@ -95,6 +95,13 @@ var WifiHelper = {
     return key === curkey;
   },
 
+  getConnectionStatus: function(network) {
+    if(this.isConnected(network)) {
```
EragonJ: `if (xxx)` we need one more space here.

apps/settings/js/wifi.js
```diff
@@ -283,8 +283,17 @@ navigator.mozL10n.once(function wifiSettings() {
...
+      else if (conStatus == 'connecting') {
+        localize(small, 'shortStatus-connecting');
+      }
+      else if (conStatus == 'associated') {
+        localize(small, 'shortStatus-associated');
+      }
```
EragonJ: is there any other chances that we would get other states here ? or only these three states would happen in this place ? (Just a question, if these are all possible situations, then please ignore this comment ;) )

PR #29388, opened 2015-04-08
Bug 1117450 - Add mozSettings and CalledNumber cache to optimize the updateCallNumber process

arvin-zhang:
As the time spent on displaying callscreen is much longer after the dial button clicked, we should do something to optimize the scenario:
1) To preload and observe the property 'ril.iccInfo.mbdn' via SettingsCache so that we'll no longer need to make request for every updateCallNumber;
2) Create a local buffer to store the contacts info of the latest TEN[1] dialing numbers and then the cached data can be used directly instead of calling mozContacts.find any more. Of course, we also implement the function to listen the change of mozContact and check to update the cache if needed. Besides, the local cache will always be pushed into the asyncStorage once updated occurred and we can get the cached data from asyncStorage easily when the device booted up next time.

With the above actions, we can solve the bottle-neck under the scenario most of the time. 

[1] TEN is a constant, which can be adjusted dynamically as needed,  stands for the cache size.

### DIFF
--- a/shared/js/dialer/contacts.js
+++ b/shared/js/dialer/contacts.js
@@ -11,6 +11,153 @@ var Contacts = {
     '/shared/js/fb/fb_reader_utils.js'
   ],
 
+  _calledNumberCache: [],
+
+  _CACHE_SIZE: 10,
+
+  init: function init() {
+    var self = this;
+    var mozContacts = window.navigator.mozContacts;
+
+    asyncStorage.getItem('contactsCache', function (value) {
+      if(value) {
+        self._calledNumberCache = JSON.parse(value);
+      }
+    });
+
+    if (mozContacts) {
+      LazyLoader.load(self._FB_FILES, function() {
+        mozContacts.oncontactchange = self._oncontactchange.bind(self);
+      });
+    }
+  },
+
+  _calledNumberCachePush: function (number, contact, matchingTel) {
+    if (this._calledNumberCache.length > this._CACHE_SIZE) {
+      this._calledNumberCache.shift();
+    }
+
+    this._calledNumberCache.push({
+      id: (contact ? contact.id : null),
+      number: number,
+      contact: contact,
+      matchingTel: matchingTel
+    });
+
+    var self = this;
+    setTimeout(function () {
+      asyncStorage.setItem('contactsCache', JSON.stringify(self._calledNumberCache));
+    },0);
+  },
+
+  _calledNumberCacheGet: function (number) {
+    for (var i = 0; i < this._CACHE_SIZE; i++) {
+      if (this._calledNumberCache[i]) {
+        if (number === this._calledNumberCache[i].number) {
+        return this._calledNumberCache[i];
+        }
+      }
+    }
+    return null;
+  },
+
+  _oncontactchange: function _oncontactchange(event) {
+    var self = this;
+    var reason = event.reason;
+    var contactID = event.contactID;
+    // Mark if the local cache has been modified on the event or not
+    var isCacheUpdated = false;
+    // To store the cache index should be deleted on the event and then remove
+    // it from cache after the traversal finished.
+    var tempDeletedIndex = null;
+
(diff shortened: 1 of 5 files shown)


shared/js/dialer/contacts.js
```diff
@@ -11,6 +11,153 @@ var Contacts = {
...
+  _calledNumberCache: [],
+
+  _CACHE_SIZE: 10,
+
+  init: function init() {
+    var self = this;
+    var mozContacts = window.navigator.mozContacts;
+
+    asyncStorage.getItem('contactsCache', function (value) {
+      if(value) {
```
gabrielesvelto: nit: Leave a space between ``if`` and ``(value)``

shared/js/dialer/contacts.js
```diff
@@ -11,6 +11,153 @@ var Contacts = {
...
+        self._calledNumberCache = JSON.parse(value);
+      }
+    });
+
+    if (mozContacts) {
+      LazyLoader.load(self._FB_FILES, function() {
+        mozContacts.oncontactchange = self._oncontactchange.bind(self);
```
gabrielesvelto: This is not enough to track changes properly. You also have to track the contact manager's revision number using ``mozContacts.getRevision()``. If the revision changed between two init() calls you have to invalidate the entire cache because the contacts will have changed when your ``oncontactchange`` is not registered (for example because the callscreen wasn't running).

shared/js/dialer/contacts.js
```diff
@@ -11,6 +11,153 @@ var Contacts = {
...
+
+    this._calledNumberCache.push({
+      id: (contact ? contact.id : null),
+      number: number,
+      contact: contact,
+      matchingTel: matchingTel
+    });
+
+    var self = this;
+    setTimeout(function () {
```
gabrielesvelto: Why are you using setTimeout() here? asyncStorage.setItem() is already asynchronous.

shared/js/dialer/contacts.js
```diff
@@ -11,6 +11,153 @@ var Contacts = {
...
+      asyncStorage.setItem('contactsCache', JSON.stringify(self._calledNumberCache));
+    },0);
+  },
+
+  _calledNumberCacheGet: function (number) {
```
gabrielesvelto: This and the ``init()`` function are not synchronized so it is possible that one might call this function before the contact cache has been populated. It's not a big deal because in that case we'd fall back on looking up the contact like we ordinarily do but it's worth keeping in mind.

shared/js/dialer/contacts.js
```diff
@@ -11,6 +11,153 @@ var Contacts = {
...
+    for (var i = 0; i < this._CACHE_SIZE; i++) {
+      if (this._calledNumberCache[i]) {
+        if (number === this._calledNumberCache[i].number) {
+        return this._calledNumberCache[i];
```
gabrielesvelto: nit: This should be indented two places to the right.

shared/js/dialer/contacts.js
```diff
@@ -233,3 +392,12 @@ var Contacts = {
...
 };
+
+window.addEventListener('load', function onLoad(event) {
+  window.removeEventListener('load', onLoad);
+  Contacts.init();
+
+  window.LazyL10n.get(function loadLazyFilesSet() {
+    console.log('moz1117450 --- LazyL10n is loaded.');
+  });
+});
```
gabrielesvelto: nit: Missing newline at the end of the file

shared/js/dialer/contacts.js
```diff
@@ -233,3 +392,12 @@ var Contacts = {
     }
   }
 };
+
+window.addEventListener('load', function onLoad(event) {
+  window.removeEventListener('load', onLoad);
+  Contacts.init();
+
+  window.LazyL10n.get(function loadLazyFilesSet() {
+    console.log('moz1117450 --- LazyL10n is loaded.');
```
gabrielesvelto: I'm not sure what you're trying to do here.

PR #29532, opened 2015-04-15
Bug 1080122 - [Gallery] Add gallery activities marionette tests

punamdahiya:


lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4ff9d60bc010efe1e92d0dba5756ffa7101db6f5

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8a93c1fdbc36c68d81aa5d38e74848ea518b8d54

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=994801d155a2bee78bac5d86cf6ec21456eb247b

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c15f59c077d70daeaf0e9382f437a635e81bfd9c

apps/gallery/test/marionette/activity_test.js
```diff
@@ -0,0 +1,130 @@
...
+    activityCallerApp = GalleryActivityCaller.create(client);
+  });
+
+  test('pick image using gallery pick activity', function() {
+    activityCallerApp.launch();
+    // Initiate pick activity by clicking pick image button
+    activityCallerApp.pickImage();
+
+    // Select image from thumbnail list in gallery app
+    galleryApp.selectImage();
```
davidflanagan: Do you need to do any kind of wait here? Is there a race condition or do we know that the gallery app has been launched?

If something goes wrong with launching the activity (e.g., if activityCallerApp initiates an "ick" activity instead of a "pick" activitiy) will we fail usefully here?  That is, do you want to have some sort of assertion here that the gallery app has actually opened up?

davidflanagan: Okay, I guess selectImage() does a wait, and presumably that is what would fail if the activity never launched the gallery.

apps/gallery/test/marionette/activity_test.js
```diff
@@ -0,0 +1,130 @@
...
+      return activityCallerApp.pickedImageName.displayed();
+    });
+
+    // Compare returned blob name and type with the
+    // image loaded in gallery app
+    var pickedImageName = activityCallerApp.pickedImageName.text();
+    assert.strictEqual(pickedImageName, imageInfo.name);
+
+    var pickedImageType = activityCallerApp.pickedImageType.text();
+    assert.strictEqual(pickedImageType, imageInfo.type);
```
davidflanagan: I'd suggest that you test the size as well as the name and type.

apps/gallery/test/marionette/activity_test.js
```diff
@@ -0,0 +1,130 @@
...
+    // Select image from thumbnail list in gallery app
+    galleryApp.selectImage();
+
+    client.waitFor(function(){
+      return galleryApp.editCropCanvas.displayed();
+    });
+
+    galleryApp.cropDoneButton.click();
+    activityCallerApp.switchTo();
+
```
davidflanagan: Can you add something here to verify that the gallery app exits or to wait until its window closes or something?

apps/gallery/test/marionette/activity_test.js
```diff
@@ -0,0 +1,130 @@
...
+    var pickedImageType = activityCallerApp.pickedImageType.text();
+    assert.strictEqual(pickedImageType, imageInfo.type);
+
+  });
+
+  test('share image using gallery share activity', function() {
+    galleryApp.launch();
+    galleryApp.tapFirstThumbnail();
+    galleryApp.shareImage('Fake Gallery Activity Caller');
+    activityCallerApp.switchTo();
```
davidflanagan: Does this switchTo() call also function as an assertion that the app was actually launched? If the activity fails to launch, will this fail somehow?

apps/gallery/test/marionette/activity_test.js
```diff
@@ -0,0 +1,130 @@
...
+    });
+
+    // Compare shared blob name and type with the
+    // image loaded in gallery app
+
+    var sharedImageName = activityCallerApp.sharedImageName.text();
+    assert.strictEqual(sharedImageName, imageInfo.name);
+
+    var sharedImageType = activityCallerApp.sharedImageType.text();
+    assert.strictEqual(sharedImageType, imageInfo.type);
```
davidflanagan: As above, please check the size of the transferred blob as well.

apps/gallery/test/marionette/activity_test.js
```diff
@@ -0,0 +1,130 @@
...
+    galleryApp.cropDoneButton.click();
+    activityCallerApp.switchTo();
+
+    client.waitFor(function(){
+      return activityCallerApp.pickedImageName.displayed();
+    });
+
+    // Tap picked image to view image using gallery open activity
+    activityCallerApp.tapPickedImage();
+    galleryApp.switchTo();
```
davidflanagan: This is a tricky test because you should have two different instances of the gallery app; one for the pick and one for the open. That seems like something that would be good to verify.  But maybe it would be easier if the activityCallerApp could get the image with its own device storage call instead of a pick activity. As it stands you're testing the pick activity twice here

apps/gallery/test/marionette/activity_test.js
```diff
@@ -0,0 +1,130 @@
...
+
+    // Check if the filename displayed in titlebar matches
+    // data sent by the initiating app
+    var title = galleryApp.openActivityImageTitle.text();
+    assert.strictEqual(title, 'firefoxOS.png');
+
+    // Check if save button is displayed
+    assert.ok(galleryApp.openActivitySaveButton.displayed());
```
davidflanagan: If you're going to test that, then you might want to also test the case where the save button is not displayed.  (I forget which way is the default)

apps/gallery/test/marionette/activity_test.js
```diff
@@ -0,0 +1,130 @@
...
+
+    client.waitFor(function(){
+      return galleryApp.openActivityImage.displayed();
+    });
+
+    // Check if passed blob is set as background in displayed image
+    assert.ok(galleryApp.hasBackgroundImageBlobURL());
```
davidflanagan: The comment indicates that you're testing whether the received blob is actually displayed. But the method name suggests that you're just testing whether there is some kind of blob:// URL displayed. I'd suggest you change one or the other.

apps/gallery/test/marionette/apps/fakeactivitycaller/index.html
```diff
@@ -0,0 +1,25 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Fake Gallery Activity Caller</title>
```
davidflanagan: This app is both a caller and a handler now, and it isn't actually faking anything: it is initiating and handling real activities. So this app name isn't quite right. It may not be worth the trouble to change, but you could call it "Gallery Activity Tester" or something like that.

apps/gallery/test/marionette/activity_test.js
davidflanagan: It might also be a good idea to test with both a File and a blob that is not a file.  You can create a valid image blob that is not a file by creating a canvas (you don't even have to draw anything to it) and calling toBlob() on it.

Since one of the major activity issues we've had in the past has involved memory-backed blobs, it would be good to have a test that uses them.

apps/gallery/test/marionette/activity_test.js
davidflanagan: Can you look at the imageblob property of the MediaFrame to get the blob itself? If so, you can check that its size is what you expect.

apps/gallery/test/marionette/activity_test.js
davidflanagan: Maybe you could test a memory-backed blob with the save button and a file-backed blob without the save button...

apps/gallery/test/marionette/apps/fakeactivitycaller/manifest.webapp
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "Fake Gallery Activity Caller",
+  "description": "Fake Gallery activity caller app",
+  "type": "certified",
```
davidflanagan: Try to use the lowest privilege necessary. It should work as type:privileged, I think, even if you use device storage to read a file for the open test.

apps/gallery/test/marionette/apps/fakeactivitycaller/manifest.webapp
```diff
@@ -0,0 +1,23 @@
...
+    "name": "The Gaia Team",
+    "url": "https://github.com/mozilla-b2g/gaia"
+  },
+  "activities": {
+    "share": {
+      "href": "/index.html#activity-share",
+      "filters": {
+        "type": ["image/*"],
+        "number": {
+          "max": 5
```
davidflanagan: Since you're not actually testing this number thing, I'd just leave it out of the manifest.

apps/gallery/test/marionette/apps/fakeactivitycaller/manifest.webapp
```diff
@@ -0,0 +1,23 @@
...
+        }
+       },
+      "disposition": "inline",
```
davidflanagan: It is probably more interesting to test the disposition:window case for share activities. And since your share activity handler never calls postResult() it doesn't really work as an inline activity, anyway.

apps/gallery/test/marionette/lib/fakeactivitycaller.js
```diff
@@ -0,0 +1,83 @@
...
+
+    create: function(client) {
+      var originURL = 'app://' + ORIGIN;
+
+      return {
+        get pickMenu() {
+          // Switch to the system app first.
+          client.switchToFrame();
+          return client.helper.waitForElement(SELECTORS.pickMenu);
+        },
```
davidflanagan: Does the caller that uses this .pickMenu property have to remember to switch back to an app after accessing pickMenu?  That seems error-prone. Is there a way to wait for the element, then switch back before returning it?

apps/gallery/test/marionette/lib/fakeactivitycaller.js
```diff
@@ -0,0 +1,83 @@
...
+
+        get sharedImageName() {
+          return client.helper.waitForElement(SELECTORS.sharedImageName);
+        },
+
+        get sharedImageType() {
+          return client.helper.waitForElement(SELECTORS.sharedImageType);
+        },
+
```
davidflanagan: It would be cool to have an integration test helper class that would take an object like your SELECTORS object and automatically generate getters like these for you.  Not suggesting you spend the time to do that right now, unless that kind of meta-programming seems cool to you. Just saying that it would be useful here.

apps/gallery/test/marionette/lib/fakeactivitycaller.js
davidflanagan: I'll add that getter functions that block seems counter-intuitive.  But I guess that is probably a common pattern in marionette tests.

apps/gallery/test/marionette/lib/fakeactivitycaller.js
```diff
@@ -0,0 +1,83 @@
...
+        switchTo: function() {
+          client.switchToFrame();
+          client.apps.switchToApp(originURL);
+        },
+
+        pickImage: function() {
+          client.helper.waitForElement(SELECTORS.pickImageButton).tap();
+          this.pickWithGalleryButton.tap();
+        },
```
davidflanagan: If I understand correctly, after calling pickImage() we've switched to the system app, and we're going to have to call switchToApp() to continue testing after this, right?  That might be worth a comment.

apps/gallery/test/marionette/lib/gallery.js
```diff
@@ -43,7 +44,15 @@ Gallery.Selector = Object.freeze({
...
-  fullscreenFrame3: '#frame3'
+  fullscreenFrame3: '#frame3',
+  cropDoneButton: '#crop-done-button',
+  editCropCanvas: '#edit-crop-canvas',
+  actionMenu: 'form[data-z-index-level="action-menu"]',
+  open: {
+      title: '#filename',
+      image: '#frame > div.image-view',
+      saveButton: '#save'
+    }
```
davidflanagan: I think it might be better to create a separate Open.Selector object to hold these rather than nesting them like this.  Or keep them here but prefix the property names with open?  openTitle, openImage, etc.? 

I say this because if someone was going to create a utility that would automatically generate a getter method for every selector, you'd want an object that was only one level deep...

apps/gallery/test/marionette/lib/gallery.js
```diff
@@ -240,6 +256,66 @@ Gallery.prototype = {
...
+   * @return {Marionette.Element} save button that saves image opened using
+   * gallery app open activity.
+   */
+  get openActivitySaveButton() {
+    return this.client.helper.waitForElement(Gallery.Selector.open.saveButton);
+  },
+
+  /**
+   * @return {Marionette.Element} that shows menu options on using web activity.
+   */
```
davidflanagan: Maybe add a comment to say that if you use this property you get switched to the system app and have to explicitly switch back to the gallery app if you want to keep using it?

apps/gallery/test/marionette/lib/gallery.js
```diff
@@ -240,6 +256,66 @@ Gallery.prototype = {
...
+  menuOptionButton: function(appName) {
+    var options = this.actionMenu.findElements('button');
+    for (var i = 0; i < options.length; i++) {
+      var menuOption = options[i];
+      if (menuOption.text() === appName) {
+        return menuOption;
+      }
+    }
+  },
```
davidflanagan: Consider putting both the actionMenu getter and the menuOptionButton() function into a separate library that can be used by both this file and fakeactivitycaller.js

apps/gallery/test/marionette/lib/gallery.js
```diff
@@ -310,6 +386,45 @@ Gallery.prototype = {
     }.bind(this));
   },
 
+  switchToGalleryFrame: function() {
+    var galleryFrame = this.client.findElement('iframe[src*="' +
+                                          Gallery.ORIGIN + '"]');
```
davidflanagan: Does this.client.findElement() wait for the element to be created, or does it fail if it does not exist when called?

If the gallery is being launched by an activity, it seems like there might be intermittent test failures because of variable timing if we're not really careful here.

apps/gallery/test/marionette/lib/gallery.js
```diff
@@ -310,6 +386,45 @@ Gallery.prototype = {
...
+    this.waitFor(galleryFrame);
```
davidflanagan: Now I'm just confused. If you just used findElement() to get the galleryFrame, what does it mean to wait for it? Does that wait for it to be visible?  Below you use waitFor() on a selector, not an element. Can you just do that here?

Probably I just dont' understand marionette well enough to get what is going on here.

apps/gallery/test/marionette/lib/gallery.js
```diff
@@ -310,6 +386,45 @@ Gallery.prototype = {
...
+    this.client.switchToFrame(galleryFrame);
```
davidflanagan: How is switchToFrame() different than switchToApp()?

Am I right in thinking that this switchToGalleryFrame() function can only be called if we have previously done a this.client.switchToFrame() to switch to the system app?  If so, please note that in a comment, or move the whole function into a separate library that represents the system app and action menu.

apps/gallery/test/marionette/lib/gallery.js
```diff
@@ -310,6 +386,45 @@ Gallery.prototype = {
...
+  },
+
+  tapFirstThumbnail: function() {
+    this.waitFor(Gallery.Selector.thumbnail);
+    var thumbnails = this.thumbnails;
+    this.waitFor(thumbnails[0]);
+    thumbnails[0].click();
```
davidflanagan: I think you can just use ```this.thumbnail.click()``` here.

davidflanagan: that is, I think you can replace the whole function with that one line.

apps/gallery/test/marionette/lib/gallery.js
```diff
@@ -310,6 +386,45 @@ Gallery.prototype = {
...
+  },
+
+  selectImage: function() {
+    this.switchToGalleryFrame();
+    this.tapFirstThumbnail();
+  },
```
davidflanagan: I think that this function starts off assuming we're in the system app, and ends up in the gallery app.  If so, then I'd suggest that it does not belong in this library, but should be a utility function in the test itself.  In the test, I'd expect to see something like:

```
system.switchToGallery()
gallery.tapFirstThumbnail()
```

apps/gallery/test/marionette/lib/gallery.js
```diff
@@ -310,6 +386,45 @@ Gallery.prototype = {
...
+
+  shareImage: function(appName) {
+    this.waitFor(Gallery.Selector.shareButton).tap();
+    this.menuOptionButton(appName).tap();
+  },
```
davidflanagan: This function also involves the gallery and systme apps, so maybe it should not be in this gallery-specific library.

Can you move it to the test itself and write something like this?

```
gallery.shareButton.tap();
system.activityMenuButton(appName).tap()
```

apps/gallery/test/marionette/lib/gallery.js
```diff
@@ -310,6 +386,45 @@ Gallery.prototype = {
...
+  hasBackgroundImageBlobURL: function() {
+    var url = 'blob:' + Gallery.ORIGIN;
+    return this.openActivityImage
+               .getAttribute('style').indexOf(url) > -1;
+  },
+
+  switchTo: function() {
+    this.client.switchToFrame();
+    this.client.apps.switchToApp(Gallery.ORIGIN);
+  },
```
davidflanagan: I don't understand how this function is different from switchToGalleryFrame(), and if I had to maintain this test, I'd just have to experiment with the two of them to figure out when to use each one.  Maybe you can clarify this with comments.

apps/gallery/test/marionette/lib/gallery.js
punamdahiya: Done

apps/gallery/test/marionette/lib/gallery.js
punamdahiya: Done

apps/gallery/test/marionette/lib/gallery.js
punamdahiya: Done

apps/gallery/test/marionette/activity_test.js
punamdahiya: Done

apps/gallery/test/marionette/activity_test.js
punamdahiya: Done

apps/gallery/test/marionette/activity_test.js
punamdahiya: Added waitForAppToDisappear function in system.js, which waits for the app to disapper before continuing

apps/gallery/test/marionette/activity_test.js
punamdahiya: Moved this switchTo call to system.js , which should fail if the app wasn't launched

apps/gallery/test/marionette/activity_test.js
punamdahiya: Added test to generate memory backed blob and open it using open activity with allowSave not set , which doesn't show save button

apps/gallery/test/marionette/activity_test.js
punamdahiya: save button not displayed is default, added test to open memory backed blob

apps/gallery/test/marionette/activity_test.js
punamdahiya: Updated comment to 'Check if displayed image src is set with blob url', I tried but it's not easy to access imageblob property of MediaFrame from activity test.  Need to explore on what's the best way to check the size of opened blob

apps/gallery/test/marionette/apps/fakeactivitycaller/index.html
punamdahiya: Updated App name and respective variable names to refer  'Gallery Activity Tester'

apps/gallery/test/marionette/apps/fakeactivitycaller/manifest.webapp
punamdahiya: Done

apps/gallery/test/marionette/apps/fakeactivitycaller/manifest.webapp
punamdahiya: Done

apps/gallery/test/marionette/apps/fakeactivitycaller/manifest.webapp
punamdahiya: Done

apps/gallery/test/marionette/lib/fakeactivitycaller.js
punamdahiya: With system.js managing action menu references, this method is no longer used and removed in updated patch

apps/gallery/test/marionette/lib/fakeactivitycaller.js
punamdahiya: +1

apps/gallery/test/marionette/lib/fakeactivitycaller.js
punamdahiya: Done

apps/gallery/test/marionette/lib/gallery.js
punamdahiya: Done and moved to system.js

apps/gallery/test/marionette/lib/gallery.js
punamdahiya: Done and moved to system.js

apps/gallery/test/marionette/lib/gallery.js
punamdahiya: I am equally confused, I have removed switchToGalleryFrame and now using single function switchToApp inside system.js for switching between apps and it works great. I am trying to remember why I needed to switch using selector but failed to replicate it.

apps/gallery/test/marionette/lib/gallery.js
punamdahiya: See above

apps/gallery/test/marionette/lib/gallery.js
punamdahiya: Done

apps/gallery/test/marionette/lib/gallery.js
punamdahiya: Done and now using single function switchToApp

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=92125a7fc1b39c1b582d2b56f25b7b5e92a01ce3

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5d1bb9ebbfd65f77012daea028c49dbdb8e5d57b

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d6c848d11d214f9553089393a80fa7ea5c40a5ab

apps/gallery/test/marionette/activity_test.js
```diff
@@ -0,0 +1,156 @@
...
+    activityTesterApp.launch();
+    // Initiate pick activity by clicking pick image button
+    activityTesterApp.tapPickImageButton();
+
+    // Tap Gallery button from Systm app Action Menu
+    system.menuOptionButton('Gallery').tap();
+
+    // We've switched to the system app, and we're calling
+    // switchToApp to switch to gallery
+    system.switchToApp(Gallery.ORIGIN);
```
davidflanagan: The tap() call above should cause Gallery to be launched for the pick activity. Does marionette's switchToApp() implicitly wait for that app to start running before switching to it?  I just want to make sure that switchToApp() isn't going to fail because of timing issues or launch a Gallery instance on its own.  I think I asked you that before though, so I assume this works.  Maybe add a comment saying that it waits.

(And also: do you need system.switchToApp(), or could you just do client.apps.switchToApp(Gallery.ORIGIN) from here?)

punamdahiya: I checked and switchToApp calls waitForApp inside it and throws error if it fails to switch. See
https://github.com/mozilla-b2g/gaia/blob/9dfedd7d35da00fa9b08dcffc3ab4f47da54e4f0/tests/jsmarionette/plugins/marionette-apps/lib/switchtoapp.js#L26

I will add comment to test explaining that switchToApp call waits. 

Also you are correct, since we have already switched to System app we don't need switchToFrame and can call directly client.apps.switchToApp(Gallery.ORIGIN) here.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=973d43435f2c0a90b811fd53c8b46c630c513a76

PR #29770, opened 2015-04-28
Bug 1158088 -  update storage information when 'panelready' occurs

mpizza:


apps/settings/js/app_storage.js
```diff
@@ -28,11 +28,13 @@ var AppStorage = (function AppStorage() {
   function attachListeners() {
     _appStorage.addEventListener('change', getSpaceInfo);
     window.addEventListener('localized', getSpaceInfo);
+    window.addEventListener('panelready', getSpaceInfo);
```
crh0716: We should only update the info when users are navigating to the application storage panel. This can be done by checking `e.detail.current` of the panel ready event.

apps/settings/js/app_storage.js
```diff
@@ -28,6 +28,12 @@ var AppStorage = (function AppStorage() {
   function attachListeners() {
     _appStorage.addEventListener('change', getSpaceInfo);
     window.addEventListener('localized', getSpaceInfo);
+    window.addEventListener('panelready', function(e){
```
crh0716: nit: Please add an extra space after `)`

crh0716: Note that we call to `attachListeners` when there is a visibility change and the visibility is true, so there is a chance that the listener will be added multiple times.
Please detach the handler in `detachListeners`.

PR #29843, opened 2015-05-03
Bug 917045 - [B2G][Email] Clicking on email address links in message bod...

rsajdok:
...ies results in confusing/misleading prompt 'Browse to URL?: mailto:'

### DIFF
--- a/apps/email/js/cards/compose.js
+++ b/apps/email/js/cards/compose.js
@@ -361,6 +361,18 @@ return [
       };
     },
 
+    haveAddressesChanged: function() {
+      var addrs = this.extractAddresses();
+      function addressesDiffer(a, b) {
+        return a.length !== b.length || a.some(function(e, i) {
+          return e !== b[i].address;
+        });
+      }
+      return addressesDiffer(this.composer.to, addrs.to) ||
+        addressesDiffer(this.composer.cc, addrs.cc) ||
+        addressesDiffer(this.composer.bcc, addrs.bcc);
+    },
+
     _saveStateToComposer: function() {
       var addrs = this.extractAddresses();
       this.composer.to = addrs.to;
@@ -380,17 +392,6 @@ return [
     },
 
     _saveNeeded: function() {
-      var self = this;
-      var checkAddressEmpty = function() {
-        var bubbles = self.querySelectorAll('.cmp-peep-bubble');
-        if (bubbles.length === 0 && !self.toNode.value && !self.ccNode.value &&
-            !self.bccNode.value) {
-          return true;
-        } else {
-          return false;
-        }
-      };
-
       // If no composer, then it means the card was destroyed before full
       // setup, which means there is nothing to save.
       if (!this.composer) {
@@ -405,7 +406,7 @@ return [
       // guy in which case we really want to provide the option to delete the
       // draft.
       return (this.subjectNode.value || hasNewContent ||
-          !checkAddressEmpty() || this.composer.attachments.length ||
+          this.haveAddressesChanged() || this.composer.attachments.length ||
           this.composer.hasDraft);
     },
 
--- a/apps/email/js/cards/message_reader.js
+++ b/apps/email/js/cards/message_reader.js
@@ -757,34 +757,44 @@ return [
     },
 
     onHyperlinkClick: function(event, linkNode, linkUrl, linkText) {
-      var dialog = msgBrowseConfirmNode.cloneNode(true);
-      var content = dialog.getElementsByTagName('p')[0];
-      mozL10n.setAttributes(content, 'browse-to-url-prompt', { url: linkUrl });
-      ConfirmDialog.show(dialog,
-        { // Confirm
-          id: 'msg-browse-ok',
-          handler: function() {
-            if (/^mailto:/i.test(linkUrl)) {
-              // Fast path to compose. Works better than an activity, since
-              // "canceling" the activity has freaky consequences: what does it
-              // mean to cancel ourselves? What is the sound of one hand
-              // clapping?
-              var data = queryURI(linkUrl);
-              cards.pushCard('compose', 'animate', {
-                composerData: {
-                  onComposer: function(composer, composeCard) {
-                    // Copy the to, cc, bcc, subject, body to the compose.
-                    // It is OK to do this blind key copy since queryURI
-                    // explicitly only populates expected fields, does not
-                    // blindly accept input from the outside, and the queryURI
-                    // properties match the property names allowed on composer.
-                    Object.keys(data).forEach(function(key) {
-                      composer[key] = data[key];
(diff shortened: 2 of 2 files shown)


lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2aeaa717fbbce37283f7fc70574ad4791e11f1ba

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9fbd1dfb6422b7c918e959865cf14bbf7313adc5

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b6688a6eeb8934ea142a5b42575e2f94e8d49a24

apps/email/js/cards/compose.js
```diff
@@ -382,6 +406,9 @@ return [
     _saveNeeded: function() {
       var self = this;
       var checkAddressEmpty = function() {
```
jrburke: It feels like the new checkAddressesPassed() function (or whatever the new name might be) serves the purpose of checkAddressEmpty(): we just want to know if an address changed, and if so save. In the case where the start state is no addresses then if none still, then that qualifies as no change. However, if the user started with addresses and removed them all that would be a save worth prompting.

So I would just call this.checkAddressEmpty() directly and remove this internal checkAddressEmpty function.

apps/email/js/cards/compose.js
```diff
@@ -361,6 +361,30 @@ return [
...
+
+      var to = frobAddressNode(this.toNode).toSource();
+      var cc = frobAddressNode(this.ccNode).toSource();
+      var bcc = frobAddressNode(this.bccNode).toSource();
+
+      return to == this.composer.to.toSource() &&
+        cc == this.composer.cc.toSource() &&
+        bcc == this.composer.bcc.toSource();
+    },
+
```
jrburke: This function could be named more precisely, as the current name does not indicate what test the addresses are passing. Perhaps something like haveAddressesChanged.

It is also tempting to try to just reuse extractAddresses() to do the address extraction, then do some array comparisons to figure out if they changed.

I would prefer to see an array iteration checking the values instead of using toSource(), since toSource is non-standard, and for native arrays could lead to misleading values.

I think it is enough for the check to be something that uses Array.some() to check the values and if there is a difference in the values, return true, and to confirm that the array sizes are the same. That array checking might be a helper function on the prototype since we need to run it for a few arrays.

If you found problems using extractAddresses() for this type of comparison it would be good to know, or if the array comparison suggestion needs more details, may have been a bit too brief in its description.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9d45117e2c9f5a72468580ab419074312cbd0587

apps/email/js/cards/compose.js
```diff
@@ -361,6 +361,18 @@ return [
       };
     },
 
+    haveAddressesChanged: function() {
+      var addrs = this.extractAddresses();
+      function compareAddresses(a, b) {
+        return a.length == b.length && !a.some(function(e, i) {
```
jrburke: Seems best to indicate what this function returns since it is the opposite of what `haveAddressesChanged` returns, maybe via a different function name.

Although thinking about it more, perhaps change the function to match the boolean state of `haveAddressesChanged`, so something like `addressesDiffer`, then this check becomes something like `a.length !== b.length || a.some(` where the function passed to some returns true if it detects a difference.

Then, the final return for `haveAddressesChanged` would be an `||` of those differ checks.

apps/email/js/cards/compose.js
```diff
@@ -361,6 +361,18 @@ return [
...
+          return typeof b[i].address === 'undefined' || e !== b[i].address;
```
jrburke: I think it is enough here to just return `e !== b[i].address` for this. 

Something I have not checked, just doing code review without running the code, is if we need to test for e.address, for cases where user tapped Compose button, selected addresses via the Contacts + activity, saved the card, then goes to drafts, edits the draft, then goes back -- does this function work in that case?

If it turns out it is best for all addresses coming into this card to be of form {name: ..., address: ...}, then we could look at just creating that object without the name part in the message_reader change where it does the `model.api.parseMailbox(addr).address` normalization.

apps/email/js/cards/compose.js
```diff
@@ -382,12 +394,8 @@ return [
     _saveNeeded: function() {
       var self = this;
       var checkAddressEmpty = function() {
-        var bubbles = self.querySelectorAll('.cmp-peep-bubble');
-        if (bubbles.length === 0 && !self.toNode.value && !self.ccNode.value &&
-            !self.bccNode.value) {
+        if (self.haveAddressesChanged()) {
```
jrburke: It is now fine to remove the `checkAddressEmpty` function here and just directly call `this.haveAddressesChanged()` in the final return check for `_saveNeeded`.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a323ea127578b18f8811cc696cdcfd89f3f18fbe

apps/email/js/cards/compose.js
rsajdok: *...does this function work in that case?*
I have tested it and it seems to work in the same way as working in the version without my corrections, before I have started working on it.


*If it turns out it is best for all addresses coming into this card to be of form {name: ..., address: ...}, then we could look at just creating that object without  the name part in the message_reader change where it does the model.api.parseMailbox(addr).address normalization.*
I do not understand what do you mean. Could you describe it in another way?

apps/email/js/cards/compose.js
```diff
@@ -405,7 +407,7 @@ return [
       // guy in which case we really want to provide the option to delete the
       // draft.
       return (this.subjectNode.value || hasNewContent ||
-          !checkAddressEmpty() || this.composer.attachments.length ||
+          self.haveAddressesChanged() || this.composer.attachments.length ||
```
jrburke: Now that there is no inner function in this method, we can remove the var self = this, and just use `this` instead of `self` here.

apps/email/js/cards/compose.js
jrburke: > I do not understand what do you mean. Could you describe it in another way?

If it was better that we make sure the data is normalized to `{name: ..., address: ...}` (just without the name property in this case), then we could adapt the ` model.api.parseMailbox(addr).address` normalization to do that. However, since the other pathway from a compose/activity is working then we do not need to do that.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d2d908831a94dee8c091328f5fb4154fafaec5f5

PR #29859, opened 2015-05-04
Bug 1160416 - Fix silder in settings app sound panel

shamenchens:


apps/settings/js/sound.js
```diff
@@ -90,31 +87,38 @@
         // or we will see the queued callbacks try to update the slider's value
         // which we are unable to avoid and make bad ux for the users.
         if (isFirstInput) {
+          // Stop observing when the user is adjusting the slider, this is to
+          // get better ux that the slider won't be updated by both the observer
+          // and the ui.
+          SettingsListener.unobserve(channelKey, setSliderValue);
```
crh0716: I found that even I moved this line to the handler of `touchstart` the bug still resolved. Could you help confirm it?

shamenchens: Yeah you are right, we don't have to move unobserve, `isFirstInput` is equivalent to `touchstart`. I will put that back.

PR #30144, opened 2015-05-20
Bug 1168441 - [Messages][NG] Implement ActivityShim

azasypkin:


autolander:

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=abcfe5356fcae52242dec62e6cf6c97e0ad10c7b

apps/sms/index.html
```diff
@@ -84,6 +84,9 @@
     <script defer src="views/shared/js/settings.js"></script>
     <script defer src="services/js/message_manager.js"></script>
     <script defer src="views/shared/js/app.js"></script>
+    <script defer src="services/activity/threads.js"></script>
```
azasypkin: We should decide where we put threads.js

apps/sms/services/activity/client.js
```diff
@@ -0,0 +1,62 @@
...
+   * Name of the service that is responsible for managing activities.
+   * @type {string}
+   */
+  const SERVICE_NAME = 'activity-service';
+
+  /**
+   * Definition of all available private members.
+   * @type {Object}
+   */
+  const priv = Object.freeze({
```
azasypkin: You probably remember that I wanted to try Symbol instead of WeakMap for private variables - here we're :)

apps/sms/services/activity/client.js
```diff
@@ -0,0 +1,62 @@
...
+    client: Symbol('ActivityClient.client'),
+
+    onActivityRequest: Symbol('ActivityClient.onActivityRequest')
+  });
+
+  var ActivityClient = function() {
+    var client = this[priv.client] = threads.client(SERVICE_NAME);
```
azasypkin: I divided service into two parts - client and service. It's clear why we need dedicated file for service part - because it can be used in worker/iframe, but here I experimenting with dedicated file for client to have one more abstraction around threads.js. The main reason is that I didn't want to use text method names (```client.method("my-method")```) in views especially if we have several places where client can be used as it's a bit harder to search for usages and do renames in case we need them.

But still don't have strong opinion on that, also not sure if we want singleton, factory or class here.

What do you guys think?

apps/sms/services/activity/service.js
```diff
@@ -0,0 +1,85 @@
...
+    onActivityRequest: Symbol('ActivityService.onActivityRequest')
+  });
+
+  var ActivityService = function() {
+    var service = this[priv.service] = threads.service(SERVICE_NAME);
+
+    service.method('postResult', this[priv.onPostResult].bind(this));
+    service.method('postError', this[priv.onPostError].bind(this));
+
+    exports.navigator.mozSetMessageHandler(
```
azasypkin: I suspect system messages will likely be replaced by "something" in service worker, but don't know how soon we'll get it as well as support for system messages in iframes with different URLs, while we need to load conversation in separate iframe (with dedicated URL) now...

apps/sms/views/conversation/js/conversation.js
```diff
@@ -476,8 +477,10 @@ var ConversationView = {
    * visible.
    */
   beforeEnter: function conv_beforeEnter(args) {
+    this.inActivityMode = !!args.isInActivity;
```
azasypkin: In this WIP I made this very ugly workaround to pass ```isInActivity``` flag, but the idea for real patch is to indicate about activity mode through parameters so that view don't have to query ActivityHandler or any other thing for it.

apps/sms/services/activity/service.js
```diff
@@ -0,0 +1,85 @@
...
+  /**
+   * Handler that fires once app receives activity request via system message.
+   * @param {ActivityRequestHandler} request Activity request.
+   * @private
+   */
+  ActivityService.prototype[priv.onActivityRequest] = function(request) {
+    this[priv.activityRequest] = request;
+
+    // Talk to Wilson to understand how to fix this.
+    setTimeout(() => {
```
azasypkin: Otherwise message isn't dispatched, likely because of https://github.com/gaia-components/threads/blob/4a440ddf3b448ce235f709bae06440a787076447/threads.js#L1749

apps/sms/views/shared/js/activity_handler.js
```diff
@@ -21,22 +23,27 @@ const ActivityDataType = {
...
+      'new-activity-request', this._onNewActivity.bind(this)
+    );
+
+    this._activityClient.on(
+      'share-activity-request', this._onShareActivity.bind(this)
     );
 
+    // Currently both client and service for activity resides in the same
+    // context, so we instantiate both.
+    (new ActivityService());
```
azasypkin: This also doesn't work good when we have two messages instances... will dig it if we're good with the solution

apps/sms/services/activity/client.js
julienw: If I understand correctly, it's a middle term between weak maps and prefixing `_`, right ?

apps/sms/services/activity/client.js
julienw: I think we also need 2 parts to make our "business" code easier to read/write/maintain.

as for singleton/factory/class I think we clearly want a singleton (be it like we always did or using a factory method)... unless you want one different client class for each activity? I don't think this is really needed (activity handler could be just one function) but this could be

apps/sms/services/activity/service.js
julienw: yeah, let's not count on this. I think the approach you have here (handling in the main view) is best for the current time.

apps/sms/index.html
julienw: and name :) "threads" is not good for us, I think Steve chose "bridge" which fits more our app.

Maybe we should have a simple script to update it (using browserify etc); the script shoud better be js based so that we could call it during the build in build.js.

azasypkin: Yep, I remember you wrote that you were discussing this with Wilson. Did you want to have "threads.js" sources somewhere in shared (so that everyone can use it and update new versions only there) and we use browserify as build step? Or you want to have our own browserified version that we should update by ourselves?

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1084,7 +1087,7 @@ var ConversationView = {
   close: function conv_close() {
     return this._onNavigatingBack().then(() => {
       this.cleanFields();
-      ActivityHandler.leaveActivity();
+      return (new ActivityClient()).postResult();
```
julienw: we miss the `{success: true}` argument, but I don't know if it's really useful.

apps/sms/views/conversation/js/conversation.js
```diff
@@ -302,10 +303,10 @@ var ConversationView = {
    * @private
    */
   backOrClose: function conv_backOrClose() {
-    var inActivity = ActivityHandler.isInActivity();
     var isComposer = Navigation.isCurrentPanel('composer');
     var isThread = Navigation.isCurrentPanel('thread');
-    var action = inActivity && (isComposer || isThread) ? 'close' : 'back';
+    var action = this.isInActivityMode() && (isComposer || isThread) ?
```
julienw: nit: I'd keep the "inActivity" boolean just to make this line easier to read

apps/sms/services/activity/client.js
azasypkin: Yeah, maybe it's just me, but I feel like Symbol suits "private case" better. Some time ago I wanted to switch to it after reading:

"Symbols are a new, special kind of object that can be used as a unique property name in objects.
Using symbols instead of strings allows different modules to create properties that don't conflict with one another.
Symbols can also be made private, so that their properties can't be accessed by anyone who doesn't already have direct access to the symbol." [1]

But something changed since then and symbols aren't really private and can be accessed via "getOwnPropertySymbols", but still looks ok for our purpose + we can add private prototype method simpler and add jsdoc for every private member if we want.

I use it in my personal projects and eager to try in sms, but only if you guys agree :)

[1] http://tc39wiki.calculist.org/es6/symbols/

apps/sms/views/conversation/js/conversation.js
azasypkin: I'm adding it in server part if result is not specified, not sure if it's the best place though :)

apps/sms/views/shared/js/activity_handler.js
julienw: where will the service live later on ? in a worker ? in the main view ?

I'm afraid this won't work like this... the "threads" library is supposed to spawn the service itself, either in a worker, in a shared worker or in window (which is a hidden iframe in the main view). But for the activity, we actually want the service to live in the main view itself so that we can handle the system message... at least until the various bugs around system messages are fixed.

So either we can augment the "threads" lib to handle this case, or we'll need some sort of adhoc mechanism (using broadcast channel ?).

azasypkin: Let me describe what problem we have currently:

Case №1:
1. When we run Messages app we create service instance with a name "A";
2. Then we create client instance bound to service "A";

Everything works fine in this case; Hope it's by intention :)

Case №2:
1. When we run Messages app we create service #1 with a name "A";
2. Then we create client#1 instance bound to service "A";
3. Then we run one more instance via activity system message and create service#2 "A" again;
4. Then one more client#2 for service "A".

I haven't debug threads.js yet, but that's how I think it works currently:
Service#2 is ignored, like when we create service#2, we see that we already have such service (via BroadcastChannel?) and when client#2 calls service method, then service#1 processes this message only.

If my guess is correct I think we can solve it in the following ways:
* If app is run not as activity (mozHasPendingMessage?) we don't create activity service - everthing should be fine here;
* ActivityService and ActvivityClient can use "context-bound" service name (like threads.service('activity-service-' + App.Id)) where App.Id maybe just Date.now() called on app startup;
* We can improve threads.js to say that service we register is in-process only, so clients from another processes can't access it.

I'll debug and check if my guess isn't stupid :)

apps/sms/views/conversation/js/conversation.js
julienw: ah yeah I missed this :)

apps/sms/services/activity/client.js
julienw: I also missed that calling `Symbol("foo")` twice gives 2 different symbols.

So yeah, it looks like a good fit.

TBH I never really liked weak maps for this purpose...

apps/sms/index.html
julienw: if we want to change the name then I guess we'll need our own version...

julienw: unless everyone agrees with the name "bridge". (I think Wilson didn't like it but I don't remember why)

apps/sms/views/shared/js/activity_handler.js
julienw: sorry, my comment was not related to your comment, but it was related to the same line of code :)

apps/sms/services/activity/client.js
azasypkin: Yeah, there is also [1] to always get the same symbol for the name. Haven't tried it though

[1] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/for

apps/sms/views/conversation/js/conversation.js
azasypkin: Okay :)

apps/sms/services/activity/client.js
azasypkin: > unless you want one different client class for each activity? I don't think this is really needed (activity handler could be just one function) but this could be

Agree, also right now reading threads.js code and see that every client instance creates its own BroadcastChannel - so it's better to have only one client to not waste memory :)

autolander:

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=44e1c547a458bdfe7679f955bf94cb5714fe475c

autolander:

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b3306b394e4126ca3f3d311b6156d1ee5d4de169

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6f498c6cdb16da622e1038f3f209bb46284941e4

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ca5e199bf16e2bdb49a693a8aa6adc74b232e928

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=45e976ebfa9d6fda60199c5781b0937fb8dee395

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=208ee52baa71a4672552a4c3ef1064dc6c2722a4

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a1a617701b5059784b52df7e7985aad4406d7a09

apps/sms/services/js/activity/client.js
```diff
@@ -0,0 +1,85 @@
...
+
+      hasPendingRequest = false;
+      return client.method('postError', error);
+    },
+
+    /**
+     * Checks if client has pending activity request.
+     * @returns {boolean}
+     */
+    hasPendingRequest() {
```
azasypkin: I need some sync (and for the case when service isn't created) way to determine if view is in activity mode

apps/sms/services/js/activity/service.js
```diff
@@ -0,0 +1,93 @@
...
+      service.method('postError', onPostError);
+
+      navigator.mozSetMessageHandler(SYSTEM_MESSAGE_NAME, onActivityRequest);
+    },
+
+    /**
+     * Checks if activity service has pending activity request.
+     */
+    hasPendingRequest() {
+      return navigator.mozHasPendingMessage(SYSTEM_MESSAGE_NAME);
```
azasypkin: Just want to hide everything related to activity&system message dependency inside service wrapper.

apps/sms/services/test/unit/mock_activity_client.js
```diff
@@ -0,0 +1,13 @@
+/* exported MockActivityClient */
+
+(function(exports) {
+  'use strict';
+
+  exports.MockActivityClient = {
+    init: () => null,
```
julienw: I'd do `() => {}` that is closer to the reality.

apps/sms/services/test/unit/mock_activity_service.js
```diff
@@ -0,0 +1,10 @@
+/* exported MockActivityService */
+
+(function(exports) {
+  'use strict';
+
+  exports.MockActivityService = {
+    init: () => null,
```
julienw: same for here

apps/sms/services/test/unit/mock_bridge.js
```diff
@@ -0,0 +1,10 @@
+/* exported bridge */
+
+(function(exports) {
+  'use strict';
+
+  exports.Mockbridge = {
+    client: () => null,
+    service: () => null
```
julienw: same for these

apps/sms/views/conversation/js/conversation.js
```diff
@@ -5,13 +5,14 @@
...
-         ActivityHandler, TimeHeaders, ContactRenderer, Draft, Drafts,
+         TimeHeaders, ContactRenderer, Draft, Drafts,
          MultiSimActionButton, Navigation, Promise, LazyLoader,
          SharedComponents,
          Errors,
          EventDispatcher,
          SelectionHandler,
-         TaskRunner
+         TaskRunner,
+         ActivityClient
```
julienw: nit: put ActivityClient just before Errors

apps/sms/views/conversation/test/unit/conversation_test.js
```diff
@@ -14,7 +14,8 @@
          AssetsHelper,
          SMIL,
          TaskRunner,
-         Thread
+         Thread,
+         ActivityClient
```
julienw: same here, put ActivityClient before AssetsHelper (trying to have them alphabetically organized for the new ones...); you can move AssetsHelper before DocumentFragment too...

apps/sms/services/js/activity/client.js
```diff
@@ -0,0 +1,85 @@
...
+
+  var ActivityClient = {
+    /**
+     * Initialized activity service client bridge.
+     */
+    init() {
+      client = bridge.client(SERVICE_NAME);
+
+      client.on('activity-request', (request) => {
+        hasPendingRequest = true;
```
julienw: This works because our activities are all inline -- but maybe you could postError automatically if hasPendingRequest is true already ? Or you prefer that hasPendingRequest is just a way to know that an activity is happening ?

apps/sms/services/js/activity/service.js
```diff
@@ -0,0 +1,93 @@
...
+   * Reference to active bridge service instance.
+   * @type {Service}
+   */
+  var service = null;
+
+  /**
+   * Reference to currently active activity request.
+   * @type {ActivityRequestHandler}
+   */
+  var activityRequest = null;
```
julienw: because this will eventually live in the service worker, we can't have only one singleton for this, we need a way to track several requests.

julienw:
Tell me if I'm wrong here.

I think we need something more.

Here is how I see this thing (ultimately):

* we need a ActivityWrapper that handles listening to the 'activity' system message, and calls postResult/postError. Note that we'll need to make sure this wrapper is not killed until the activity finishes, I'm not sure how easy it is :/ I also don't really know if the wrapper can be long living (can it answer to several requests ? related to my previous comment)
* ActivityService living in the SharedWorker -- this will be called directly by the wrapper when a new activity comes in. It's especially responsible for calling "openWindow" with the right window. Another possibility is that the right window is already opened because it's the endpoint for this activity... in that case likely we don't need this service at all.
* ActivityClient -- not sure it's really useful but why not.

Please tell me how you see this !

azasypkin:
Yeah, I have a bit different understanding what we'll have ultimately :)

Since "openWindows" and "push notifications" (let's assume that something like this will replace system messages) are only available in ServiceWorkers (SW) I don't see any useful application for SharedWorkers (ShW) here. I see only need in Window (View/Client) <---> SW (Service) channel.

Let me write pseudo code just from the top of my head:

```
.....push-notification-handler.js (to be used in SW context).....
interface PushNotificationHandler {
	canHandle(notificationName: string): boolean;
	handle(notification: any): Promise<any>;
}
```
```
.....sw.js (our main SW "controller" file).....
const pushNotificationHandlers = [
  ActivityService,
  NotificationService
];

self.addEventListener('push-notification', (e) => {
	var handler = pushNotificationHandlers.find((h) => h.canHandle(e.name));

	if (!handler) {
	  ... we can respond that we can't process this....
	}

	handler.handle(e.params).then(() => {
	  ... somehow successfully respond to request ...;
	}).catch(() => {
      ... say that we've failed ...
	});
})
```
```
.....activity-service.js (to be used in SW context).....
class ActivityService implements PushNotificationHandler {
	constructor(clients: Clients) {
      this.clients = clients;
	}

	canHandle(notificationName) {
		// Or we can just have "frozen" "activity<--->handler" map in sw.js, 
		// but as with Cache middleware we may need several push notification handlers on
		// "stop-once-found-that-can-handle" basis.
		return notificationName === 'activity';
	}

	handle(activity) {
      // So depending on what platform provides we may have different ways
      // to determine if it's not the first activity request;

      // If we don't have service, let spawn one to support activity client;
      if (!service) {
        service = bridge.service('activity-service');
        service.method('postResult', () => {
           .... resolve "handle" promise with result ....

           // Don't have idea about this method, but bridge has to likely support
           // smth like this.
           if (!service.hasOtherClients()) {
             service.shutDown();
             service = null;
           }
        });
      } else {
        // we can either return Error from "handle" or support multiple requests
        // For now I don't think we should care about multiple request case at all.
      }

      // So we have "supporting" service, let's open window
      this.clients.openWindow(.....);

      // Opened window will create activity client (bridge.client) to respond to
      // activity request. But we may need to pass some unique clientId if we're
      // going to support several activity requests at the same time, but as I said
      // above I don't think we should care right now.

      // Let's pass the data to the newly created client
      service.broadcast('activity', activity, ....clientId?....);
   }
}
```
```
.....activity-client.js (to be used in view/window context).....
class ActivityClient {
	constructor() {
	  this.client = bridge.client('activity-service');
	  this.client.on('activity', (activity) => {
	    // Emit "public" event to the view.
	    this.emit(activity.name + '-request');
	  });
	}

       postResult(result) {
         return this.client.method('postResult', result);
       }
}
```

I use service/client because it feels natural to me especially that it corresponds to bridge.service/bridge.client; I wrap ```bridge.service``` and ```bridge.client``` into activity-service and activity-client, just because I don't want to deal with text-based methods in views/sw. But I'm fine with any name for it.

So guys do you see any flaw in my thoughts?

apps/sms/services/js/activity/service.js
azasypkin: Replied below in discussions :)

apps/sms/services/js/activity/client.js
azasypkin: Yeah, I can postError if you want :)

The only my concern about the code that is not used now is that it's confusing when you read it later on, but it's still not used or case it covers, still doesn't exist - I usually spend time to find the case it covers (and eager to remove if can't find :D) :) Just say me what way you prefer!

apps/sms/services/test/unit/mock_activity_client.js
azasypkin: Yep! Not sure why I return ```nulll``` :)

apps/sms/services/js/activity/client.js
julienw: Well, you usually check arguments and throw errors in methods -- I see it's the same here :) 

So if hasPendingRequest is already true, console.error(...), and call postError (I think postError automatically close the activity?). And add a comment to explain that it should not happen because our activities are inline

azasypkin: Okay :)

apps/sms/views/shared/js/activity_handler.js
```diff
@@ -21,63 +23,42 @@ const ActivityDataType = {
...
-    if (typeof handler === 'function') {
-      this.setActivity(activity);
+      ActivityClient.on(
+        'share-activity-request', this._onShareActivity.bind(this)
+      );
 
-      handler.call(this, activity);
+      // Currently both client and service for activity reside in the same
+      // context, so we spin up both in the same context/window.
+      ActivityService.init();
```
steveck-chung: So it's not working if the client and service is in different context, right?  How about move this `ActivityService.init` in as a method of bridge and called in ActivityClient.init(but this means we'll need to spin up the service earlier and ActivityService.init only for register system message )?

azasypkin: > So it's not working if the client and service is in different context, right?

Nope, it should work. I mean if we define service in iframe/worker (if it were possible) and client will be in the main window - it should not have any problems. Or what do you mean?

steveck-chung: I mean, if ActivityService does not live in the same context as activityHandler, we could not access the ActivityService directly. The only we to access the service is though bridge(and we're likely to wrap methods in ActivityClient).

steveck-chung: BTW I'm ok if you plan to move this out in follow up

steveck-chung:
@azasypkin Firstly I was thinking why need another wrapper for the bridge.client, but I changed my mind while writing the messaging service for some reasons: Exposing the text-based methods is really not that pleasant, and it brings more limitation because it could not passing object that could not be serialized. That means we need to do more rewrite in view or handler. So I'm fine with view <-> client wrapper <-> service communication and views can only call the method from client wrapper/middleware.

For several activity requests issue, I don't think we handled all the requests originally because the previous activity will force exit. So I think current implementation should work for my POV.

azasypkin:
> Firstly I was thinking why need another wrapper for the bridge.client, but I changed my mind while writing the messaging service for some reasons: Exposing the text-based methods is really not that pleasant, and it brings more limitation because it could not passing object that could not be serialized. That means we need to do more rewrite in view or handler. So I'm fine with view <-> client wrapper <-> service communication and views can only call the method from client wrapper/middleware.

Yep, I had the same thoughts when I started with this service/shim :)

> For several activity requests issue, I don't think we handled all the requests originally because the previous activity will force exit. So I think current implementation should work for my POV.

Agree on this. Julien's point was that it would be better if we keep working correctly even when Gecko/System starts supporting more than one activity. So I'll try to do this simple change and you'll tell me if it's better to do in the follow up :)

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a0c8b18a4e550f02b247d62233daebb7f916ecfb

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3b3eaa9a8d7497e4c2343576d38c878952dfd2a0

apps/sms/services/test/unit/mock_activity_client.js
azasypkin: Done

apps/sms/services/test/unit/mock_activity_service.js
azasypkin: Done

apps/sms/services/test/unit/mock_bridge.js
azasypkin: Done

apps/sms/views/conversation/js/conversation.js
azasypkin: Done

apps/sms/views/conversation/test/unit/conversation_test.js
azasypkin: Done

apps/sms/services/js/activity/client.js
azasypkin: Done

azasypkin:
Today I was thinking about handling activities in "separated" views. As we were discussing earlier we'll likely use iframe for this. So let me just dump what I have currently in mind:

Let's say Gallery initiates share activity request. In Messages app we use dedicated HTML file to handle activities (e.g. ```views/activity.html```). This file will just contain 100% height/width iframe and few JS files. Its main JS will be similar to current activity handler: it will receive system message and expose bridge service (maybe with the help of ActivityShim). Then it will decide which view to load (via setting of nested iframe src).

Now more complex part: we should somehow pass activity parameters to iframe. If it were just string/number we could just use query string parameters for the iframe's URL, but we have can have blobs. Currently ActivityHandler converts every Blob to Attachment, but we can't pass custom Attachment type via ```postMessages/BroadcastChannelAPI```. That means that "Blob+FileName ---> Attachment" conversion should be made inside view itself.

So the only way I can imagine right now is to pass some sort of string ```activity token``` parameter via iframe URL. When Conversation/NewMessage view sees such parameter in ```beforeEnter``` it initializes ActivityClient with this token (at the moment I see this token as ```appInstanceId``` I use to create 1-to-1 connection between bridge service and client) _and_ requests the activity data with something like:

```
 // We need both blob and its name that come separately; 
interface ActivityBlob {
	blob: Blob; // activityData.blobs[i]
	name: string; // activityData.filenames[i]
}

interface ActivityData {
	number?: string; // number or email;
	body?: string | Array<ActivityBlob>
}

interface ActivityClient {
	init: (appInstanceId: string) => void;
	getData: () => Promise<ActivityData>;
        ......
}
```

Once the activity data is reiceved we can continue with ```handleActivity``` as before with the intermediate "Blob+FileName ---> Attachment" conversion step.

So what do you @steveck-chung @julienw  think? Does it sound about right or you have other ideas?

Btw, question unrelated to the post above: what if we use this top-level + single nested iframe structure not only for the activity. So that top level doc will be just host for all of our shims?

julienw:
I can't think of anything else that could work; but I'll raise the issue with Vivien. Ideally I'd like to be able to change the page easily or to use window.open or SW's openWindow. But that's not possible given the UX and technical constraints.

steveck-chung:
@azasypkin yeah I think this implementation that accessing the data through activity client make sense before we have any other better solution that could navigate to certain view with data directly(I thought we should have more generic way for system message or other similar mechanism in the future, not not now).

About the iframe for shim, I don't see why we need to put shims into different iframe separately, since we'll likely to keep all the shims on all the time instead of close specific shim manually in certain view.

apps/sms/services/test/unit/activity/shim_test.js
```diff
@@ -0,0 +1,139 @@
...
+    this.sinon.stub(navigator, 'mozSetMessageHandler');
+    this.sinon.stub(navigator, 'mozHasPendingMessage');
+
+    // It should throw if app instance id is not specified.
+    assert.throws(() => ActivityShim.init());
+
+    ActivityShim.init(appInstanceId = Date.now());
+  });
+
+  test('bridge client is correctly initialized', function() {
```
steveck-chung: nit: bridge "service"?

azasypkin:
> About the iframe for shim, I don't see why we need to put shims into different iframe separately, since we'll likely to keep all the shims on all the time instead of close specific shim manually in certain view.

I meant that we'll need "something" that will choose what view to load for the activity case, that why I was thinking about top-level window and view inside iframe

apps/sms/services/test/unit/activity/shim_test.js
azasypkin: yep!

apps/sms/views/shared/js/activity_handler.js
```diff
@@ -3,7 +3,10 @@
 
 /*global Utils, MessageManager, Compose, NotificationHelper,
          Attachment, Notify, SilentSms, Threads, SMIL, Contacts,
-         ConversationView, Notification, Settings, Navigation */
+         ConversationView, Notification, Settings, Navigation,
+         ActivityClient,
+         ActivityShim
```
julienw: nit: put the Shim before Client :)

azasypkin: I thought you wanted it to be in alphabetical order :)

apps/sms/views/shared/js/activity_handler.js
```diff
@@ -21,63 +24,46 @@ const ActivityDataType = {
...
-    if (typeof handler === 'function') {
-      this.setActivity(activity);
+      ActivityClient.on(
+        'share-activity-request', this._onShareActivity.bind(this)
+      );
 
-      handler.call(this, activity);
+      // Currently both client and service for activity reside in the same
+      // context, so we spin up both in the same context/window.
+      ActivityShim.init(appInstanceId);
```
julienw: I wonder if you'll be able to pass the instanceId to the shim once it lives in another context ?

julienw: unrelated: should we run the shim.init before the client.init or it doesn't matter ?

azasypkin: > I wonder if you'll be able to pass the instanceId to the shim once it lives in another context ?

I've explained one of the ways in https://github.com/mozilla-b2g/gaia/pull/30144#issuecomment-110022887, and we won't need it when we live in the Shared or Service Worker

apps/sms/services/js/activity/client.js
```diff
@@ -0,0 +1,102 @@
...
+*/
+
+(function(exports) {
+  'use strict';
+
+  /**
+   * Name of the service that is responsible for managing activities.
+   * @type {string}
+   */
+  const SERVICE_NAME = 'activity-service';
```
julienw: should we rename this as well? Not sure, tell me what you think

apps/sms/views/shared/js/activity_handler.js
azasypkin: > unrelated: should we run the shim.init before the client.init or it doesn't matter ?

IIRC it won't work right now, i.e. service will broadcast earlier than client listen for it, need to check

apps/sms/views/shared/test/unit/activity_handler_test.js
```diff
@@ -113,211 +122,196 @@ suite('ActivityHandler', function() {
...
-      var shareURLActivity = {
-        source: {
-          name: 'share',
-          data: {
-            type: 'url',
-            url: 'test_url'
-          }
-        },
-        postResult: sinon.stub()
+      function() {
```
julienw: nit: I now think the "function() {" part should be 2 pixels to the left so that we clearly see where the function code starts.

apps/sms/views/shared/js/activity_handler.js
julienw: Oops, you're right, I compared "h" and "l" but forgot the "C" and "S" that are before. Forget me.

apps/sms/services/js/activity/client.js
azasypkin: I have mixed feelings, if we're going to leave client and just replace shim with service hosted in ServiceWorker in the future, then I'd say we need to use ```service``` term, so that client doesn't notice any change, but if we're going to live with the shim for some long term maybe we can use ```shim``` prefix instead to make it clearer for developers.

Both options work for me, just let me know what you prefer.

apps/sms/views/shared/js/activity_handler.js
julienw: OK, no need to decide now; it's just weird to me that we have a client before we have a service.

apps/sms/views/shared/test/unit/activity_handler_test.js
azasypkin: Right!

apps/sms/views/shared/js/activity_handler.js
azasypkin: Yeah, I agree

apps/sms/services/js/activity/client.js
julienw: keep it like this for now, we'll see over time how it goes.

apps/sms/services/js/activity/client.js
```diff
@@ -0,0 +1,102 @@
...
+
+  /**
+   * Reference to active bridge client instance.
```
julienw: maybe add a comment explaining why we can have only one client ?

apps/sms/services/js/activity/client.js
```diff
@@ -0,0 +1,102 @@
...
+   */
+  var client;
+
+  /**
+   * Indicates if there is pending activity request.
+   * @type {boolean}
+   */
+  var hasPendingRequest;
+
+  var ActivityClient = {
```
julienw: can you add a jsdoc explaining what this does?

apps/sms/services/js/activity/shim.js
```diff
@@ -0,0 +1,100 @@
...
+  function onActivityRequest(request) {
+    activityRequest = request;
+
+    // Talk to Wilson to understand how to fix this.
+    setTimeout(() => {
+      service.broadcast('activity-request', request.source);
+    });
+  }
+
+  var ActivityShim = {
```
julienw: can you add a jsdoc explanint what it does?

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6b11e28edcd73eac01c464054062412b8750c2c7

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=854b87dcfe31e01b58bb50d863d50c32fa5c7bbc

PR #30157, opened 2015-05-20
Bug 1104667 - Remove the LazyL10n code and refactored the callscreen to use modern L10n primitives r=drs,gandalf

gabrielesvelto:


lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=426cd673e1e77808fd11bb900e009d1a1e13f2f0

shared/locales/phone_types/phone_types.en-US.l20n
```diff
@@ -0,0 +1,25 @@
...
+<_phone_type[$type] {
+  *mobile:    "{{ phone_type.mobile }}",
+  home:      "{{ phone_type.home }}",
+  work:      "{{ phone_type.work }}",
+  personal:  "{{ phone_type.personal }}",
+  faxHome:   "{{ phone_type.fax_home }}",
+  faxOffice: "{{ phone_type.fax_office }}",
+  faxOther:  "{{ phone_type.fax_other }}",
+  other:     "{{ phone_type.other }}"
+}>
```
zbraniecki: Why do you need both, _phone_type and phone_type?

shared/locales/phone_types/phone_types.en-US.l20n
```diff
@@ -0,0 +1,25 @@
...
+<phone_type_and_carrier "{{ _phone_type }}, {{ carrier }}">
+<custom_phone_type_and_carrier "{{ $type }}, {{ carrier }}">
+<custom_phone_type "{{ $type }}">
```
zbraniecki: In those two, you are passing the argument variable $type instead of a localized value to the string. Why?

shared/locales/phone_types/phone_types.en-US.l20n
```diff
@@ -0,0 +1,25 @@
...
+<phone_number "{{ number }}">
```
zbraniecki: Here, you are referencing an L10nID "number". Did you want to use a variable "$number"?

shared/locales/phone_types/phone_types.en-US.l20n
stasm: While I'm eager to start using more L20n syntax in localization, in this case we should have 8 separate strings here.

zbraniecki: But then we need a way to select the right one out of those, instead of $phone_type[$type].

shared/locales/phone_types/phone_types.en-US.l20n
zbraniecki: Also, "carrier" is an entity reference as well. You probably want "$carrier" which is a variable reference.

shared/locales/phone_types/phone_types.en-US.l20n
zbraniecki: Quite honestly, the more I think about it the more I lean toward simple set of entities:

<phone_type_mobile "Mobile, {{ $carrier }}">
<phone_type_fax "Fax, {{ $carrier }}">

The reason for that is that those are not really *variants* of the same entity (like formFactor stuff). They are separate entities that should be trackable (so that if a localizer does not translate phone_type_fax, we know that it's a missing entity).

If you want to make words like Mobile, Fax etc. reusable, you could do:

<phone_type_mobile "Mobile">
<phone_type_fax "Fax">
<phone_type_and_carrier_mobile "{{ phone_type_mobile }}, {{ $carrier }}">
<phone_type_and_carrier_fax "{{ phone_type_fax }}, {{ $carrier }}">

But I wouldn't use hashes here. Hashes are more for plural variants, or formFactor variants *of the same string*. Those are separate strings I believe.

shared/locales/phone_types/phone_types.en-US.l20n
gabrielesvelto: When a custom phone type is given we should not localize it but just print it out as-is. carrier should have been $carrier obviously.

shared/locales/phone_types/phone_types.en-US.l20n
gabrielesvelto: Yeah, my mistake

shared/locales/phone_types/phone_types.en-US.l20n
gabrielesvelto: OK, thanks to both for clarifying that; then I'll probably need 16 different strings here, 8 for the phone types (these can be alone) plus 8 for the phone-type-comma-carrier forms. I suppose I can reuse the strings for the phone types as described above.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f125275ca5964406b6cfae6c35423737da4c4b44

apps/callscreen/js/call_screen.js
```diff
@@ -372,14 +372,15 @@ var CallScreen = {
...
-      var timeFormat = window.navigator.mozHour12 ? _('shortTimeFormat12') :
-                                                    _('shortTimeFormat24');
+      var timeFormat =
+        window.navigator.mozHour12 ? navigator.mozL10n.get('shortTimeFormat12')
+                                   : navigator.mozL10n.get('shortTimeFormat24');
       // FIXME/bug 1060333: Replace span with hidden mechanism.
       // Don't show am/pm (for 12 or 24 time) in the callscreen
       timeFormat = timeFormat.replace('%p', '<span>%p</span>');
-      var dateFormat = _('longDateFormat');
+      var dateFormat = navigator.mozL10n.get('longDateFormat');
```
zbraniecki: if you want to use the opportunity to remove sync .get, you can use mozL10n.formatValue here. This will be compatible with the upcoming v3 API then :)

apps/callscreen/js/calls_handler.js
```diff
@@ -277,45 +277,47 @@ var CallsHandler = (function callsHandler() {
...
   function handleCallWaiting(call) {
-    LazyL10n.get(function localized(_) {
-      var number = call.secondId ? call.secondId.number : call.id.number;
+    var number = call.secondId ? call.secondId.number : call.id.number;
+
+    if (!number) {
+      navigator.mozL10n.setAttributes(CallScreen.incomingNumber,
+                                      'withheld-number');
+      // Force a synchronous translation so we can resize the string
+      navigator.mozL10n.translateFragment(CallScreen.incomingNumber);
```
zbraniecki: this is risky because translateFragment may not be sync (and in API 3 it is not sync) https://github.com/stasm/l20n.js/blob/lightweight/src/bindings/html/dom.js#L36

While in API3 you can use the returned promise, it will still only resize once. If we retranslate later, you will not resize.

If you need to resize when the value changes, why not set up MutationObserver on the element?

apps/callscreen/js/handled_call.js
```diff
@@ -292,29 +274,31 @@ HandledCall.prototype.formatPhoneNumber =
...
 
 HandledCall.prototype.updateDirection = function hc_updateDirection() {
-  var self = this;
   var classList = this.node.classList;
   if (this._initialState == 'incoming') {
     classList.add('incoming');
-    LazyL10n.get(function localized(_) {
-      self.node.setAttribute('aria-label', _('incoming'));
-    });
+    this.node.setAttribute('aria-label', 'incoming');
```
zbraniecki: how is this supposed to work? I think what you want here is

this.node.setAttribute('data-l10n-id', 'incoming');

and in your l10n resources:

<incoming
  ariaLabel: "Foo">

apps/callscreen/test/unit/call_screen_test.js
```diff
@@ -807,6 +806,7 @@ suite('call screen', function() {
 
     setup(function() {
       this.sinon.useFakeTimers();
+      this.sinon.spy(navigator.mozL10n, 'setAttributes');
```
zbraniecki: so, you can spy on setAttributes in your tests, but you can also just check if the data-l10n-id/data-l10n-args is set:

var l10nAttrs = navigator.mozL10n.getAttributes(element);
assert.equal(l10nAttrs.id, 'foo');
assert.deepEqual(l10nAttrs.args, {'n': 5});

I have no preference here, feel free to leave it as is, just offering as an option :)

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=29ca386bf675a12811c5835cee048f8db5050c1d

shared/test/unit/l10n_helper.js
```diff
@@ -0,0 +1,39 @@
...
+ *   matches the'html' field.
+ */
+function l10nAssert(node, expectedL10n) {
+  if (typeof(expectedL10n) === 'string') {
+    assert.isTrue(node.hasAttribute('data-l10n-id'));
+    assert.equal(node.getAttribute('data-l10n-id'), expectedL10n);
+  } else if (expectedL10n.id) {
+    assert.isTrue(node.hasAttribute('data-l10n-id'));
+    assert.equal(node.getAttribute('data-l10n-id'), expectedL10n.id);
+    assert.isTrue(node.hasAttribute('data-l10n-args'));
```
zbraniecki: I love this idea!

The only change, can you test if expectedL10n.args exist before testing it on the node? {id: 'l10nId} should work.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=08d67ba8fe017096964c2a44c1e3c27388ba5d61

apps/callscreen/index.html
```diff
@@ -17,20 +17,19 @@
...
     <!-- Localization -->
     <script defer type="text/javascript" src="/shared/js/lazy_loader.js"></script>
-    <script defer type="application/javascript" src="/shared/js/lazy_l10n.js"></script>
     <meta name="defaultLanguage" content="en-US">
     <meta name="availableLanguages" content="en-US">
     <link rel="localization" href="../communications/dialer/locales/dialer.{locale}.properties">
     <link rel="localization" href="../communications/dialer/locales/shared.{locale}.properties">
     <link rel="localization" href="shared/locales/date/date.{locale}.properties">
     <link rel="localization" href="shared/locales/phone_types/phone_types.{locale}.properties">
+    <link rel="localization" href="shared/locales/phone_types/phone_types.{locale}.l20n">
```
zbraniecki: not needed anymore?

gabrielesvelto: good point :)

shared/test/unit/l10n_helper.js
gabrielesvelto: Will do

apps/callscreen/js/calls_handler.js
gabrielesvelto: I'm trying to address a similar issue in bug 1155901 so I hope that the fix for that will apply here too.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ae1bc750c52923a6dadbb0dbcbb0cbe3c03d92a3

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=314dc4fc692143871a5d298c4f8ead2bdad2f7cc

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b93c612e040be5f9e6d0037a98e0895098523dc5

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=24fee9ed714dae59d85d26f22ed94260fdc3d0a5

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4ee5fd6bacbbd28ae080866460bb46456165950b

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5bc418b7c991ace25add13f9f8e2c70757c54b69

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7f4a875b8ab4dca6e0a0040ed81775bfa5ffcf53

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7087bda34e17ebbb68186efb594d0270bafee0f0

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3e9cf3082967abcffe7e559cfe8b145305535720

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b45acacbd17001359b410f1374ea1f973d733148

apps/callscreen/index.html
```diff
@@ -17,20 +17,18 @@
...
     <link rel="localization" href="shared/locales/lockscreen/lockscreen.{locale}.properties">
     <link rel="localization" href="shared/locales/keypad/keypad.{locale}.properties">
     <link rel="localization" href="shared/locales/telephony/telephony.{locale}.properties">
-    <!-- Making sure the shared resources will be copy into the zip
-    <script type="application/javascript" src="/shared/js/l10n.js"></script>
-    <script type="application/javascript" src="/shared/js/l10n_date.js"></script>
-    -->
+
+    <script defer type="application/javascript" src="/shared/js/l10n.js"></script>
+    <script defer type="application/javascript" src="/shared/js/l10n_date.js"></script>
```
zbraniecki: can you remove this file in your patch? I removed all uses and forgot to remove the reference :)

apps/callscreen/js/call_screen.js
```diff
@@ -379,25 +377,32 @@ var CallScreen = {
...
       var timeText =
         f.localeFormat(now, timeFormat).replace(/\s<span>.*<\/span>/, '');
       var dateText = f.localeFormat(now, dateFormat);
 
-      this.lockedClockTime.textContent = timeText;
-      this.lockedDate.textContent = dateText;
-    }.bind(this));
+      self.lockedClockTime.textContent = timeText;
+      self.lockedDate.textContent = dateText;
+    });
```
zbraniecki: showClock has been updated in my lockscreen Intl patch so this change is not needed

apps/callscreen/js/call_screen.js
```diff
@@ -495,7 +500,7 @@ var CallScreen = {
     durationChildNode.textContent = '00:00';
     durationNode.classList.add('isTimer');
 
-    LazyL10n.get(function localized(_) {
+    navigator.mozL10n.ready(function() {
```
zbraniecki: `Utils.prettyDuration` is actually not using mozL10n.get, so you don't need ready here.

apps/callscreen/js/handled_call.js
```diff
@@ -455,10 +440,26 @@ HandledCall.prototype.connected = function hc_connected() {
...
+      var id;
+
+      switch (this._cachedInfo && this._cachedInfo.id) {
+        case 'withheld-number': id = 'withheld-number-left-call';  break;
+        case 'voiceMail':       id = 'voicemail-left-call';        break;
+      }
+
+      CallScreen.showStatusMessage({
+        id: id,
+        args: { caller: this._cachedInfo ? this._cachedInfo.toString() : '' }
```
zbraniecki: I'm not sure if I understand this line.

In my patch I used formatValue for that scenario when you need to resolve an entity and then pass it as an argument as a string.

https://github.com/mozilla-b2g/gaia/pull/31104/files#diff-0cc741478a6ca5d2f75dd44f7f7ab18bR468

How is the toString supposed to work here?

shared/test/unit/mocks/mock_l10n.js
```diff
@@ -8,7 +8,7 @@
     get: stringify,
 
     formatValue: function(id, args) {
-      return Promise.resolve(stringify(id, args));
+      return Promise.resolve(stringify(id, args || ''));
```
zbraniecki: why do you need this change?

gabrielesvelto: I can't remember... I've re-run all the unit-tests and they pass w/o this change so it might have been related to something that was fixed in another bug (the lockscreen maybe?).

zbraniecki: I actually know now - if args is undefined then it gets `tostringed` in stringify and then

`{
  id: 'foo',
  args: undefined
}`

gets turned into `fooundefined` instead of `foo`. Keep it :)

apps/callscreen/js/handled_call.js
gabrielesvelto: That's a leftover from the old code, there's no need to pass an id/args couple there, I'll remove it.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f681c452b48bfe9d240cc467f9ebd5e18a8f06f1

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c26d33b57f2d413f7ad3acef9e4592a6c07b4198

apps/callscreen/js/conference_group_handler.js
```diff
@@ -112,6 +107,26 @@ var ConferenceGroupHandler = (function() {
...
   }
 
+  /* Observe changes to the node that holds the group call duration, this is
+   * also used for displaying the call ended string and needs to be resized
+   * dynamically to accomodate for certain locales. */
+  function observeMutation(mutations) {
+    if (groupDurationChildNode.hasAttribute('data-l10n-id') &&
+        groupDurationChildNode.getAttribute('data-l10n-id') === 'callEnded') {
+      /* Disable the observer to prevent it from being called recursively
+       * while we modify the DOM tree, re-enable it once we're done . */
```
DouglasSherk: Nit: stray space here.

apps/callscreen/js/conference_group_handler.js
```diff
@@ -112,6 +107,26 @@ var ConferenceGroupHandler = (function() {
...
+      mutationObserver.disconnect();
+      FontSizeManager.adaptToSpace(
+        CallScreen.getScenario(), groupLabel, false, 'end');
+      mutationObserver.observe(groupDurationChildNode, mutationObserverConfig);
```
DouglasSherk: This is referring to ```mutationObserverConfig``` which hasn't been instantiated yet, at least in order of prose. It would be better to refactor this into a function that instantiates the config and sets the ```MutationObserver```.

apps/callscreen/js/handled_call.js
```diff
@@ -77,22 +76,19 @@ function HandledCall(aCall) {
...
-      var n = this.call.serviceId + 1;
-      this.viaSimNode.textContent = _('via-sim', { n: n });
-      this.simNumberNode.textContent = _('sim-number', { n: n });
-    } else {
-      this.viaSimNode.hidden = true;
-      this.simNumberNode.hidden = true;
-    }
-  }).bind(this));
+  navigator.mozL10n.setAttributes(this.durationChildNode,
+    (this.call.state == 'incoming') ? 'incoming' : 'connecting');
```
DouglasSherk: ```===```

apps/callscreen/js/handled_call.js
```diff
@@ -289,65 +265,70 @@ HandledCall.prototype.updateCallNumber = function hc_updateCallNumber() {
...
-      var contactCopy = {
-        id: contact.id,
-        name: contact.name,
-        org: contact.org,
-        tel: contact.tel
-      };
       if (primaryInfo) {
-        node.textContent = primaryInfo;
-        self._cachedInfo = primaryInfo;
+        self.replacePhoneNumber({ raw: primaryInfo[0] }, 'end');
```
DouglasSherk: ?

apps/callscreen/js/handled_call.js
```diff
@@ -456,10 +443,23 @@ HandledCall.prototype.connected = function hc_connected() {
...
-    CallScreen.showStatusMessage({
-      id: 'caller-left-call',
-      args: { caller: this._cachedInfo.toString() }
-    });
+    if (typeof(this._cachedInfo) === 'string') {
+      CallScreen.showStatusMessage(this._cachedInfo);
+    } else if (this._cachedInfo.id) {
+      var id;
+
+      switch (this._cachedInfo && this._cachedInfo.id) {
```
DouglasSherk: You could use an associative array / object map here, instead.

apps/callscreen/test/unit/call_screen_test.js
```diff
@@ -912,6 +906,7 @@ suite('call screen', function() {
     var timeNode;
     setup(function() {
       this.sinon.useFakeTimers();
+      this.sinon.stub(navigator.mozL10n, 'ready', function(cb) { cb(); });
```
DouglasSherk: You can use ```callsArg(0)``` here. See http://sinonjs.org/docs/

shared/js/dialer/utils.js
```diff
@@ -139,5 +139,76 @@ var Utils = {
...
+    var carrier = matchingTel.carrier;
+
+    if (Array.isArray(type)) {
+      type = type[0];
+    }
+
+    var id;
+    var args = {};
+
+    switch (type) {
```
DouglasSherk: This should definitely be an associative array instead.

shared/js/dialer/utils.js
```diff
@@ -139,5 +139,76 @@ var Utils = {
...
+      'home',
+      'work',
+      'personal',
+      'faxHome',
+      'faxOffice',
+      'faxOther',
+      'other'
+    ];
+
+    return phoneTypes.some(element => element == type);
```
DouglasSherk: ```===``` .

apps/callscreen/test/unit/call_screen_test.js
gabrielesvelto: I just realized I don't need that anymore as I've removed the last reference to mozL10n.ready() from the code

apps/callscreen/js/handled_call.js
gabrielesvelto: The primaryInfo is basically the name field of the associated MozContact, this field is an array and we're taking the first entry by default: https://developer.mozilla.org/en-US/docs/Web/API/MozContact/name#Value

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ec94d336f24b5a08085bbde9d830fa41957673fb

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5316606aabb17ef5eec6fffbd8d15339a84920e0

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ec7356343b6208cd70a9a1d63d0f2d1db22d39f5

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3be2f819739ab7844bed24c94aa704250b28aa35

apps/callscreen/js/calls_handler.js
zbraniecki: so, do you need this here? I'm ok with this landing, just wanted to know if it's needed since you added the MutationObserver

apps/callscreen/js/handled_call.js
```diff
@@ -382,30 +362,37 @@ HandledCall.prototype.formatPhoneNumber =
...
 HandledCall.prototype.updateDirection = function hc_updateDirection() {
-  var self = this;
   var classList = this.node.classList;
   if (this._initialState === 'incoming') {
     classList.add('incoming');
-    LazyL10n.get(function localized(_) {
-      self.node.setAttribute('aria-label', _('incoming'));
-    });
+    // XXX: This should be replaced with mozL10n.setAttribute whenever possible
+    this.node.setAttribute('aria-label', 'incoming');
```
zbraniecki: this will not do any l10n, right? Why not move the strings to `incoming.ariaLabel` and `outgoing.ariaLabel` and set `data-l10n-id` to `incoming`|`outgoing` here? You don't need to change the ID because our tools will catch this change.

apps/callscreen/test/unit/call_screen_test.js
```diff
@@ -830,10 +827,7 @@ suite('call screen', function() {
     });
 
     test('should show the text', function() {
-      assert.equal(
-        statusMessage.querySelector('p').getAttribute('data-l10n-id'),
-        'message'
-      );
+      l10nAssert(statusMessage.querySelector('p'), 'message');
```
zbraniecki: how does it work?

apps/callscreen/js/handled_call.js
gabrielesvelto: Yes, no l10n and we have both incoming/outgoing and incoming.ariaLabel/outgoing.ariaLabel strings. That's fixable but since the dialer/callscreen properties files are a mess right now and it's unclear what's shared and what's not I'd stick to this for now and fix it later.

apps/callscreen/test/unit/call_screen_test.js
gabrielesvelto: If you pass a string to l10nAssert it'll check for the data-l10n-id only, see here: https://github.com/mozilla-b2g/gaia/pull/30157/files#diff-32497f9d76f1a7e98c97a5a07ed5898eR23

zbraniecki: ah! I somehow skipped the l10nAssert. Love it :)

apps/callscreen/js/handled_call.js
zbraniecki: It seems to me that nothing except of callscreen uses outgoing/incoming l10nId. Why not just remove it then from dialer.properties and use the .ariaLabel only?

PR #30163, opened 2015-05-20
Bug 1164148 - Make sure the progress doesn't show -0:01 if we don't have a duration.

hfiguiere:


lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c402806a9f344305907f163a71ad0a266662ee26

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e51a52e87dd423cf308976047288bc2a590727ad

apps/music/js/ui/views/player_view.js
```diff
@@ -763,22 +763,17 @@ var PlayerView = {
...
+    this.setSeekBar(this.audio.duration, this.audio.currentTime);
   },
 
-  setSeekBar: function pv_setSeekBar(startTime, endTime, currentTime) {
-    if (!isFinite(endTime)) {
-      endTime = Math.max(this.seekBar.max, currentTime);
-    }
-
+  setSeekBar: function pv_setSeekBar(endTime, currentTime) {
     if (this.seekBar.max != endTime) {
```
davidflanagan: In the previous code, endTime was guaranteed to be finite at this point. But now it is not. So maybe you should add the isFinite(endTime) test here, too?

davidflanagan: Same thing at line 782 below. If endTime is not finite, then maybe you should just not set aria-valuemax at all.

apps/music/js/ui/views/player_view.js
```diff
@@ -788,7 +783,8 @@ var PlayerView = {
     this.seekSlider.setAttribute('aria-valuenow', currentTime);
 
     // if endTime is 0, that's a reset of seekBar
-    var ratio = (endTime !== 0) ? (currentTime / endTime) : 0;
+    var ratio = (isFinite(endTime) && endTime !== 0) ?
+        (currentTime / endTime) : 0;
```
davidflanagan: I think this change effectively revert the workaround from 1150322. If gecko gives us an infinte duration for some reason (like if we ever add support for audio streams) then we will have the symptoms from 1150322 again. But 1150322 was resolved with a gecko fix, so maybe that is okay.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6ed25c91a89e48bf675372da2575984f92f03aa7

PR #30305, opened 2015-05-29
Bug 1163889 - Implement an activity for enabling full developer mode r=eragonj

crh0716:
https://bugzilla.mozilla.org/show_bug.cgi?id=1163889

### DIFF
--- a/apps/settings/elements/full_developer_mode.html
+++ b/apps/settings/elements/full_developer_mode.html
@@ -1,0 +1,39 @@
+<element name="full-developer-mode" extends="section">
+  <template>
+    <div class="full-dev-mode-warning" hidden>
+      <gaia-confirm>
+        <h1 data-l10n-id="enable-full-dev-mode-first-warning-title"></h1>
+        <p>
+          <span data-l10n-id="enable-full-dev-mode-first-warning-msg-1"></span>
+          <br/>
+          <br/>
+          <span data-l10n-id="enable-full-dev-mode-first-warning-msg-2"></span>
+          <br/>
+          <br/>
+          <span data-l10n-id="enable-full-dev-mode-first-warning-msg-3"></span>
+        </p>
+        <gaia-buttons skin="dark">
+          <button type="reset" data-l10n-id="cancel"></button>
+          <button type="submit" data-l10n-id="enable" class="danger"></button>
+        </gaia-buttons>
+      </gaia-confirm>
+    </div>
+
+    <div class="dev-menu-disabled-warning" hidden>
+      <gaia-confirm>
+        <h1 data-l10n-id="developer-menu-disabled-warning-title"></h1>
+        <p>
+          <span data-l10n-id="developer-menu-disabled-warning-msg-1"></span>
+          <br/>
+          <br/>
+          <span data-l10n-id="developer-menu-disabled-warning-msg-2"></span>
+        </p>
+        <gaia-buttons skin="dark">
+          <button type="submit" data-l10n-id="close"></button>
+        </gaia-buttons>
+      </gaia-confirm>
+    </div>
+
+    <panel data-path="panels/full_developer_mode/panel"></panel>
+  </template>
+</element>
--- a/apps/settings/elements/root.html
+++ b/apps/settings/elements/root.html
@@ -3,7 +3,7 @@
 
     <gaia-header id="main-header">
       <h1 id="main-list-title" data-l10n-id="settings"></h1>
-      <a id="activityDoneButton" href="#home" role="menuitem" data-l10n-id="done">Done</a>
+      <a id="activityDoneButton" role="menuitem" data-l10n-id="done"></a>
     </gaia-header>
 
     <div>
--- a/apps/settings/js/panels/full_developer_mode/panel.js
+++ b/apps/settings/js/panels/full_developer_mode/panel.js
@@ -1,0 +1,56 @@
+define(function(require) {
+  'use strict';
+
+  var SettingsService = require('modules/settings_service');
+  var SettingsCache = require('modules/settings_cache');
+  var DialogService = require('modules/dialog_service');
+  var Panel = require('modules/panel');
+
+  /**
+   * The panel can be invoked by triggering an activity:
+   * new MozActivity({
+   *   name: 'configure',
+   *   data: {
+   *     target: 'device',
+   *     section: 'full-developer-mode'
+   *   }
+   * })
+   */
+  return function ctor_full_dev_panel() {
+    return Panel({
+      onInit: function(panel) {
+        this._panel = panel;
+        SettingsCache.getSettings((results) => {
+          var devMenuEnabled = !!results['developer.menu.enabled'];
+          if (devMenuEnabled) {
+            this._showFullDevModeWarning();
+          } else {
+            this._showEnableDevMenuWarning();
+          }
+        });
+      },
+      _close: function() {
+        SettingsService.back();
+      },
+      _showFullDevModeWarning: function() {
+        var warningDialog = this._panel.querySelector('.full-dev-mode-warning');
+        var okBtn = warningDialog.querySelector('button[type="submit"]');
+        var cancelBtn = warningDialog.querySelector('button[type="reset"]');
+
+        warningDialog.hidden = false;
+        cancelBtn.onclick = this._close;
+        okBtn.onclick = () => {
+          DialogService.show('full-developer-mode-final-warning');
+        };
+      },
+      _showEnableDevMenuWarning: function() {
+        var warningDialog =
+          this._panel.querySelector('.dev-menu-disabled-warning');
+        var closeBtn = warningDialog.querySelector('button[type="submit"]');
+
+        warningDialog.hidden = false;
+        closeBtn.onclick = this._close;
+      }
+    });
+  };
+});
--- a/apps/settings/js/panels/full_developer_mode_final_warning/panel.js
+++ b/apps/settings/js/panels/full_developer_mode_final_warning/panel.js
@@ -1,0 +1,58 @@
+define(function(require) {
+  'use strict';
+
+  var SettingsService = require('modules/settings_service');
+  var DialogPanel = require('modules/dialog_panel');
+
+  return function ctor_full_dev_warning() {
+    return DialogPanel({
+      onInit: function(panel) {
+        this._warningDialog = {
+          root: panel.querySelector('.warning-dialog'),
+          okBtn: panel.querySelector('.warning-dialog button[type="submit"]'),
+          cancelBtn:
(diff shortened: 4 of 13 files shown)


lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6c78a693ac106de48935d4a440933e8727d2ff9c

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d7fda724c3df746c35831eaf911e2eee2ad9829b

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d23757c02758f436230addd647c0ad9e464dfb47

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c95acde4e8932aa6fec152fd4c5b4384bbbc035f

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=14b6ab0c7b129959c347c9248299788f9ab88fbc

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ef20dc9646456a4ac889a60a72f1a7fa172b809f

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3ed1e316c351174905fb157c6ceb3fe6694dadb7

apps/settings/elements/full_developer_mode.html
```diff
@@ -0,0 +1,39 @@
...
+    <div class="full-dev-mode-warning" hidden>
+      <gaia-confirm>
+        <h1 data-l10n-id="enable-full-dev-mode-first-warning-title"></h1>
+        <p>
+          <span data-l10n-id="enable-full-dev-mode-first-warning-msg-1"></span>
+          <br/>
+          <br/>
+          <span data-l10n-id="enable-full-dev-mode-first-warning-msg-2"></span>
+          <br/>
+          <br/>
```
EragonJ: Let's not use `<br/>` here, how about using specific CSS rules to make do the same thing ? For me, `<br/>` is too general to be used.

apps/settings/js/panels/full_developer_mode_final_warning/panel.js
```diff
@@ -0,0 +1,57 @@
+define(function(require) {
+  'use strict';
+
+  var SettingsService = require('modules/settings_service');
+  var DialogPanel = require('modules/dialog_panel');
+
+  return function() {
```
EragonJ: `return function ctor_..`

apps/settings/js/panels/full_developer_mode/panel.js
```diff
@@ -0,0 +1,45 @@
...
+          } else {
+            this._showEnableDevMenuWarning();
+          }
+        });
+      },
+      _close: function() {
+        SettingsService.navigate('home');
+      },
+      _showFullDevModeWarning: function() {
+        var warning = this._panel.querySelector('.full-dev-mode-warning');
```
EragonJ: how about `warningDialog` ?

apps/settings/js/panels/full_developer_mode/panel.js
```diff
@@ -0,0 +1,45 @@
...
+          var devMenuEnabled = !!results['developer.menu.enabled'];
+          if (devMenuEnabled) {
+            this._showFullDevModeWarning();
+          } else {
+            this._showEnableDevMenuWarning();
+          }
+        });
+      },
+      _close: function() {
+        SettingsService.navigate('home');
```
EragonJ: Because this panel is opened as webActivity, why we have to navigate to `home` page after user clicking on cancel button ? Shouldn't we directly dismiss this webActivity and user will be redirected to original app ?

apps/settings/js/panels/full_developer_mode_final_warning/panel.js
```diff
@@ -0,0 +1,57 @@
...
+        this._close();
+      },
+      _updateWarningInfo: function(count) {
+        window.navigator.mozL10n.setAttributes(this._warningDialog.warningInfo,
+          'enable-full-dev-mode-final-warning-msg', {
+            count: count
+          });
+      },
+      _close: function() {
+        SettingsService.navigate('home');
```
EragonJ: Same question, why we have to control the panel flow in a dialog ? Why not just close it ?

EragonJ: ah... is this because we want to hit `postResult` to dismiss webActivity ? If so, can we add some comments on this ?

apps/settings/js/panels/full_developer_mode/panel.js
EragonJ: Ok, I know the reason why you want to do this, can you check related comments on `full_developer_mode_final_warning/panel.js` ? Thanks !

apps/settings/js/panels/full_developer_mode_final_warning/panel.js
```diff
@@ -0,0 +1,57 @@
...
+      onInit: function(panel) {
+        this._warningDialog = {
+          root: panel.querySelector('.warning-dialog'),
+          okBtn: panel.querySelector('.warning-dialog button[type="submit"]'),
+          cancelBtn:
+            panel.querySelector('.warning-dialog button[type="reset"]'),
+          warningInfo: panel.querySelector('.warning-dialog .warning-info')
+        };
+
+        this._count = 10;
```
EragonJ: I am thinking about the case that if in the future, this dialog got a chance to be opened several times (I remembered there would be the other entry point for this based on Doug's comment, right ?) If so, we should re-initialize `this._count` each time in `onBeforeShow` to make sure for each round, users have to confirm for 10 times.

apps/settings/js/panels/full_developer_mode_final_warning/panel.js
EragonJ: Because it's not a good idea to call navgation API from a dialog, normally the dialog is a tool that used to ask for user's answer of something, for dialog itself, it should not realize the existence of the `panel`, to be honest, this should be done in caller side instead of callee.

apps/settings/elements/full_developer_mode.html
crh0716: This is inside `gaia-confirm` and it expects a paragraph element, so I prefer not to introduce specific style, what do you think?

apps/settings/js/panels/full_developer_mode_final_warning/panel.js
crh0716: I would say this is a special case for better UX experience. If we close the activity in the caller, the users sees the dialog dismissed and then the activity closed.

apps/settings/js/panels/full_developer_mode/panel.js
crh0716: In typical use cases a panel knows its parent panel, but there is no parent panel in this case. Therefore we need to have a virtual panel representing this case and which is `home`.
I agree `home` is not a good name. Maybe we should have a new API `close#SettingsService` that can be invoked in all panels.

apps/settings/js/panels/full_developer_mode/panel.js
crh0716: ok~

apps/settings/elements/full_developer_mode.html
crh0716: The other option would be using multiple paragraph elements, but there will be lines between the elements.

apps/settings/js/panels/full_developer_mode_final_warning/panel.js
crh0716: Good point!  I'll reset the count in `onBeforeShow`.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c91c1ab789438a46d16fda05d9de363af966bfc0

apps/settings/elements/full_developer_mode.html
EragonJ: ok, let's just keep it !

apps/settings/elements/root.html
```diff
@@ -3,7 +3,7 @@
 
     <gaia-header id="main-header">
       <h1 id="main-list-title" data-l10n-id="settings"></h1>
-      <a id="activityDoneButton" href="#home" role="menuitem" data-l10n-id="done">Done</a>
+      <a id="activityDoneButton" role="menuitem" data-l10n-id="done">Done</a>
```
EragonJ: remove `done` here.

apps/settings/js/panels/full_developer_mode/panel.js
```diff
@@ -0,0 +1,56 @@
...
+      _close: function() {
+        SettingsService.back();
+      },
+      _showFullDevModeWarning: function() {
+        var warningDialog = this._panel.querySelector('.full-dev-mode-warning');
+        var okBtn = warningDialog.querySelector('button[type="submit"]');
+        var cancelBtn = warningDialog.querySelector('button[type="reset"]');
+
+        warningDialog.hidden = false;
+        cancelBtn.onclick = this._close.bind(this);
```
EragonJ: I think we don't have to rebind scope to `this` here because in `_close` method, we don't use anything about `this`.

apps/settings/js/panels/full_developer_mode/panel.js
```diff
@@ -0,0 +1,56 @@
...
+          DialogService.show('full-developer-mode-final-warning');
+        };
+      },
+      _showEnableDevMenuWarning: function() {
+        var warningDialog =
+          this._panel.querySelector('.dev-menu-disabled-warning');
+        var closeBtn = warningDialog.querySelector('button[type="submit"]');
+
+        warningDialog.hidden = false;
+        closeBtn.onclick = this._close.bind(this);
```
EragonJ: same here.

apps/settings/elements/root.html
crh0716: removed.

apps/settings/js/panels/full_developer_mode/panel.js
crh0716: ok~

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2ebb270098e9ad5f594dabf86653040c93f4aaaa

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c46cbacb7b73e7eda4161a071651843f11072218

PR #30381, opened 2015-06-03
Bug 1162028 - [Messages][NG] Extract views/conversation/conversation.html

azasypkin:


lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f9f6e1ab64bea6170dbd09b61c7dd6265ae1d821

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=01d54455cfe0b0b6923b290bb4e6e004746d9daf

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d9a07c0697cef51a50998e01e79f673eff375c79

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=718f544d3e4f4f6e231c161ea906e82a55ee4733

apps/sms/views/conversation/js/compose.js
```diff
@@ -16,14 +16,14 @@
...
 
-const TYPES = ['image', 'audio', 'video'];
-
 /**
  * Handle UI specifics of message composition. Namely,
  * resetting (auto manages placeholder text), getting
  * message content, and message size
  */
 var Compose = (function() {
+  const TYPES = ['image', 'audio', 'video'];
```
azasypkin: Reading my comments again :) There were vanished after rebasing....

Just moved here to not pollute global scope and IIRC FxDesktop was complaining about redefining of constant somehow.

apps/sms/views/conversation/js/conversation.js
```diff
@@ -480,7 +480,6 @@ var ConversationView = {
     this.clearConvertNoticeBanners();
     this.setHeaderAction(ActivityClient.hasPendingRequest() ? 'close' : 'back');
 
-    Recipients.View.isFocusable = true;
```
azasypkin: Looks like it's Composer only, so moved to ```beforeEnterComposer```

apps/sms/views/conversation/js/conversation.js
```diff
@@ -563,6 +562,8 @@ var ConversationView = {
       this.recipients.focus();
     }
 
+    this.emit('visually-loaded');
```
azasypkin: Don't have better names in mind, using the same that raptor uses.

apps/sms/views/conversation/js/startup.js
```diff
@@ -0,0 +1,87 @@
...
+          Utils,
+          InterInstanceEventDispatcher
+*/
+
+(function(exports) {
+  'use strict';
+
+  const DEFAULT_PANEL = 'composer';
+
+  const LAZY_DEPENDENCIES = [
```
azasypkin: Tried to re-use this list and ```initHeaders``` in the main app startup - but it looked more like a mess for me - so I went "duplication" way here.

apps/sms/views/conversation/js/startup.js
```diff
@@ -0,0 +1,87 @@
...
+      header.removeAttribute('no-font-fit');
+    }
+  }
+
+  function initShims() {
+    exports.InboxView = Object.freeze({
+      whenReady: () => Promise.resolve(),
+      markReadUnread: () => {},
+      updateThread: () => {},
+      isShim: true
```
azasypkin: This property is for possible navigation.js shim that can detect separated/monolithic view environment by probing behavior object existence (e.g. if ```panel.behaviour && !panel.behaviour.isShim```) - maybe it's just a bad idea :)

apps/sms/views/conversation/js/startup.js
```diff
@@ -0,0 +1,87 @@
...
+    });
+
+    var deferred = Utils.Promise.defer();
+
+    MessageManager.getThreads({
```
azasypkin: Using the simplest, but slow approach here :)

apps/sms/views/conversation/style/conversation.css
```diff
@@ -157,3 +157,11 @@
 #messages-header-text .ellipsis-dir-fix {
   width: auto;
 }
+
+.conversation-view .panel {
```
azasypkin: Haven't found smarter way that will work for both hybrid(separated inbox, but multi-functional conversation)/monolithic views yet :) Maybe you have better idea?

apps/sms/views/shared/js/navigation.js
```diff
@@ -72,11 +71,11 @@ var Navigation = {
     this.transitionPromise = null;
 
     currentPanel = null;
+  },
 
-    Startup.on('post-initialize', function() {
```
azasypkin: Removed direct dependency since it's not available in separated view context. But not really happy with ```setReady```....

apps/sms/views/shared/js/navigation.js
```diff
@@ -106,7 +105,7 @@ var Navigation = {
     var currentArgs = currentPanel.args || {};
 
     for (var arg in args) {
-      if (args[arg] !== currentArgs[arg]) {
+      if (args[arg] != currentArgs[arg]) {
```
azasypkin: That's the problem I don't have good solution yet - in views we use  ```isCurrentPanel``` with _number_ thread id, but when we load panel separately and pass thread id as parameter it's parsed as string...

I was thinking about JSONable query string parameters that will solve this, but it's kind of ugly... Don't know really.

What do you think?

apps/sms/views/shared/js/startup.js
```diff
@@ -108,16 +98,28 @@ var Startup = {
 
       MessageManager.init();
       Navigation.init();
+      InboxView.init();
```
azasypkin: This changes flow for the activity case, previously we didn't call ```InboxView.init``` before lazy loading, but I believe it should not have big impact.

I just didn't want to call init in two palces :)

shared/js/event_dispatcher.js
```diff
@@ -101,6 +101,22 @@
     },
 
     /**
+     * Registers listener function to be executed only first time when event
+     * occurs.
+     * @param {string} eventName Name of the event to listen for.
+     * @param {function} handler Handler to be executed once event occurs.
+     */
+    once: function(eventName, handler) {
```
azasypkin: I defined it to simplify patch, but I can switch back to ```on+off``` to get rid of shared peer review dependency :)

shared/js/event_dispatcher.js
```diff
@@ -199,7 +215,7 @@
             'Object to mix into already has "' + method + '" property defined!'
           );
         }
-        target[method] = eventDispatcher[method].bind(this);
+        this[method] = target[method] = eventDispatcher[method].bind(this);
```
azasypkin: Otherwise we ```this.once``` doesn't have access to ```this.off/this.on``` - haven't found better way yet.

apps/sms/views/conversation/index.html
```diff
@@ -0,0 +1,558 @@
...
+  <script defer src="/views/conversation/js/link_helper.js"></script>
+  <script defer src="/shared/js/image_utils.js"></script>
+  <script defer src="/views/conversation/js/attachment.js"></script>
+  <script defer src="/views/conversation/js/attachment_renderer.js"></script>
+  <script defer src="/views/conversation/js/subject_composer.js"></script>
+  <script defer src="/views/conversation/js/compose.js"></script>
+  <script defer src="/shared/js/gesture_detector.js"></script>
+  <script defer src="/lib/bridge.js"></script>
+  <script defer src="/services/js/activity/activity_client.js"></script>
+  <script defer src="/services/js/activity/activity_service.js"></script>
```
azasypkin: Forgot to update ```service``` to ```shim``` here, but maybe we don't need neither ```shim``` nor ```activity_handler``` here at all, since nobody uses it.

apps/sms/views/conversation/style/conversation.css
julienw: In my navigation rewrite, I set a "panel-active" class on the panel section that is currently active...

With the current code I don't have a better idea though.

shared/js/event_dispatcher.js
julienw: I did this in the past: https://github.com/julienw/studio/blob/autotheme/js/libs/event_dispatcher.js#L103-L110
(never filed a bug to land it, sorry :p)

basically calling eventDispatcher.on.call(this...).

shared/js/event_dispatcher.js
```diff
@@ -101,6 +101,22 @@
...
+     * Registers listener function to be executed only first time when event
+     * occurs.
+     * @param {string} eventName Name of the event to listen for.
+     * @param {function} handler Handler to be executed once event occurs.
+     */
+    once: function(eventName, handler) {
+      var once = (...args) => {
+        this.off(eventName, once);
+
+        handler(args);
```
julienw: you miss `...`, right ?

apps/sms/views/shared/js/navigation.js
julienw: can't we just use force threadId to be a number after parsing it as string? threadId = +threadId ?

apps/sms/views/shared/js/navigation.js
julienw: it's temporary code anyway, I don't think we'll need this isReady anymore once we have splitted panels

apps/sms/views/conversation/js/conversation.js
julienw: I like this, this makes the code in startup.js really clearer and cleaner.

apps/sms/views/conversation/js/conversation.js
julienw: yep I think you're right

apps/sms/views/conversation/js/compose.js
julienw: sounds good

apps/sms/views/conversation/js/startup.js
julienw: good for me

apps/sms/views/conversation/js/startup.js
julienw: This is how I resolved this issue in my navigation WIP:

https://github.com/julienw/gaia-navigator/blob/sms-like-example-2/examples/sms-like/navigation.js#L23
https://github.com/julienw/gaia-navigator/blob/sms-like-example-2/examples/sms-like/navigation.js#L12 (I don't actually use this yet -- but I could use it in https://github.com/julienw/gaia-navigator/blob/sms-like-example-2/examples/sms-like/navigation.js#L59, not sure yet... maybe it's just duplicated data I should get rid of :) )

julienw: so in short: I don't think we need isShim

apps/sms/views/conversation/js/startup.js
julienw: works for me until we use services

julienw: although maybe we could set only the id we need :)

apps/sms/views/conversation/js/startup.js
```diff
@@ -0,0 +1,87 @@
...
+  ConversationView.init();
+
+  ConversationView.once('visually-loaded', () => {
+    initLazyDependencies();
+    initHeaders();
+  });
+
+  initShims().then(() => {
+    Navigation.defaultPanel = Navigation.getPanelName() || DEFAULT_PANEL;
+    Navigation.toDefaultPanel(Utils.params(exports.location.hash));
```
julienw: can you put a comment explaining this will be figured out in bug 1162030 	 ?

apps/sms/views/conversation/js/startup.js
azasypkin: I'm happy if we don't need it! :)

apps/sms/views/conversation/js/startup.js
azasypkin: I'd rather be on the safe side :) I'm not sure what will happen when we tap on the number link and choose "Send message" (same in Participants view) and we already have thread for this number - so we don't need to change document and re-run startup....

apps/sms/views/conversation/js/startup.js
azasypkin: Yep!

apps/sms/views/conversation/style/conversation.css
azasypkin: Then I'll go with what I have now, unless we find something better during review.

apps/sms/views/shared/js/navigation.js
azasypkin: Yep

apps/sms/views/shared/js/navigation.js
azasypkin: You mean in ```beforeEnter/afterEnter/beforeLeavee/afterLeave``` (i.e. in places where we use ```args.id```)?

shared/js/event_dispatcher.js
azasypkin: Hmm, looks like I don't need ```...``` at all (not sure why I added it for the arguments) :)  since we always have only one parameter

shared/js/event_dispatcher.js
azasypkin: :D So, what approach do you prefer?

apps/sms/views/shared/js/navigation.js
julienw: everywhere we use isCurrentPanel and know for sure this is a number.

Another possibility is to use parseInt and keep the number if the result is not NaN... but this has a perf penalty, right ?

shared/js/event_dispatcher.js
julienw: I prefer mine, which one do you prefer ? :p

azasypkin: Ha-ha, I prefer mine just because I don't like to deal with ```call/apply``` if I can avoid that :) But it doesn't matter for me in the end, so will use your way :)

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=04e3132bb08911ca38999322ab28816405a50909

apps/sms/views/shared/js/navigation.js
azasypkin: > everywhere we use isCurrentPanel and know for sure this is a number.

But the problem is that ```isCurrentPanel``` uses ```currentPanel.args``` internally where ```threadId``` is the string. To make it work like you suggest, we should convert threadId to string instead, every time we pass it to navigation ```toPanel, isCurrentPanel```

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1c00b70f4cf738c3c7962f0ee757dfb9bce96a59

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=afc6c1768defb67c74199ea8bfee81914280c97f

apps/sms/views/inbox/js/inbox.js
```diff
@@ -124,6 +125,16 @@ var InboxView = {
       groupThreadTitleMaxLength: (window.innerWidth - 100) / 10
     });
 
+    this.once('fully-loaded', () => {
```
azasypkin: I felt like these actions ```ensureReadAheadSetting``` and ```initStickyHeader``` are more related to the view itself rather then ```renderThreads``` methods - so moved them here. What do you think?

apps/sms/views/inbox/js/inbox.js
```diff
@@ -124,6 +125,16 @@ var InboxView = {
...
+      this.ensureReadAheadSetting();
+
+      this.readyDeferred.resolve();
```
azasypkin: Btw, this ```readyDeferred``` is used only in ```InboxView.whenReady``` that is used only by ConversationView :) And it basically is resolved when ```App.whenReady``` is resolved. so we probably can get rid of this method entirely and switch ConversationView to ```App.whenReady``` instead.

apps/sms/views/inbox/test/unit/startup_test.js
```diff
@@ -0,0 +1,90 @@
...
+
+    var gaiaHeader = document.createElement('gaia-header');
+    gaiaHeader.setAttribute('no-font-fit', '');
+    document.body.appendChild(gaiaHeader);
+
+    Startup.init();
+  });
+
+  test('correctly initializes dependencies', function() {
+    sinon.assert.calledOnce(MessageManager.init);
```
azasypkin: Maybe these checks are too verbose and not required, tell me if you don't want simpler tests that test only something important.

apps/sms/views/shared/js/startup.js
```diff
@@ -108,16 +98,33 @@ var Startup = {
...
+
+      InboxView.once('fully-loaded', () => {
+        window.performance.mark('fullyLoaded');
+
+        App.setReady();
+      });
+
+      InboxView.once('visually-loaded', () => {
```
azasypkin: We can probably call ```_lazyLoadInit``` here and remove one more listener below, for this we'll have to change ```_lazyLoadInit``` a bit so that it can check if lazy loading already started.

I was thinking that having two listeners should not affect performance much

shared/js/event_dispatcher.js
```diff
@@ -101,6 +101,26 @@
     },
 
     /**
+     * Registers listener function to be executed only first time when event
+     * occurs.
+     * @param {string} eventName Name of the event to listen for.
+     * @param {function} handler Handler to be executed once event occurs.
+     */
+    once: function(eventName, handler) {
+      ensureValidEventName(eventName);
```
azasypkin: In fact we need only ```ensureValidHandler(handler);``` here since ```ensureAllowedEventName``` and ```ensureValidEventName``` will be performed by ```eventDispatcher.on``` - so just kept it for the consistency. What do you think?

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=46bc33c2f3bfaa9bcc849004d5727aa9cea28c98

apps/sms/views/conversation/style/conversation.css
steveck-chung: +1 that we can simply keep this as it is until navigation ready

shared/js/event_dispatcher.js
steveck-chung: I might not check these again simply for consistency reason, but up to you

julienw: I think it's good enough to call ensureValidHandler only

apps/sms/views/shared/js/startup.js
julienw: it's better like this. I doubt these listeners affect performance at all.

shared/js/event_dispatcher.js
azasypkin: Okay, guys, will leave only ```ensureValidHandler``` :)

apps/sms/views/shared/js/navigation.js
julienw: let's keep the loose equality for now; I don't have other better ideas...

apps/sms/views/inbox/test/unit/startup_test.js
julienw: I think it's fine

apps/sms/views/inbox/js/inbox.js
julienw: yeah I think it actually makes a lot of sense

apps/sms/views/inbox/js/inbox.js
julienw: yep why not

apps/sms/views/conversation/js/startup.js
julienw: I think we'll want to change the document in the future though, so that it's easy to go back :)

julienw:
I see one unit test is failing

julienw:
I found an issue, I'm sure you'll love it. I don't seem to reproduce on master.

STR:
1. send a SMS to the device.
2. Force close the SMS app.
3. swipe the notification so that it's removed.
4. Launch the SMS app
=> all items are duplicated in the inbox :)

Maybe other STR trigger the same issue, I'l llet you dig ;)

julienw:
>  3. swipe the notification so that it's removed.

What happens here is that we get the "notification" system message and the app is launched.

apps/sms/views/conversation/js/startup.js
azasypkin: Ah yeah makes sense

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d2a34b378c66814e420d2b2df152ee3a7a00ba26

apps/sms/views/inbox/js/inbox.js
```diff
@@ -140,6 +147,10 @@ var InboxView = {
     if (args.notifyAboutSavedDraft) {
       this.showDraftSavedBanner();
     }
+
+    if (!args.meta.prev) {
+      this.renderThreads();
+    }
```
julienw: wondering if we can not fix it here instead?

Some ideas:
* force that renderThreads runs only once -- `if (!renderingOrRendered) { this.renderThreads() }`
* do not call it here at all (for now) -- use apps/sms/views/conversation/js/startup.js instead

Both these ideas seem better than your current fix.

What do you think ?

azasypkin: Well, I'm happy to have better fix! But to give more meaningful alternatives, could you please tell me what exactly feels wrong in the current solution to be on the same page?

My current solution was based on two assumptions:

* ```Navigation.once('navigated')``` in startup.js is a more or less long-living hack that will go away - start rendering treads in another panel when "some" undetermined other panel is loaded.
* ```Navigation.ensureCurrentPanel``` is hack either for the old notification API that doesn't correspond to the spec (notification on close action), that also will go away.

So I just wanted to keep all those hacks together, but anyway, I agree it's a hack :)

> force that renderThreads runs only once -- if (!renderingOrRendered) { this.renderThreads() }

This will make file with what we're going to live with in the future more complex for nothing. So I'd rather not do it.

> do not call it here at all (for now) -- use apps/sms/views/conversation/js/startup.js instead

I suspect you meant ```apps/sms/views/inbox/js/startup.js``` and ```apps/sms/views/shared/js/startup.js```? If so, yeah we can call ```renderThreads``` directly in startup files for now if we can't figure out better alternatives.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5f4e16f8165af30f6ddb4c1fb1083a10c8d5e675

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=83370df6c157b8bb4d8a78493228eeaa3674c20d

PR #30395, opened 2015-06-03
Bug 1171206 - Update SMS to be ready for L10n API v3

zbraniecki:
https://bugzilla.mozilla.org/show_bug.cgi?id=1171206

### DIFF
--- a/apps/sms/index.html
+++ b/apps/sms/index.html
@@ -63,8 +63,7 @@
     -->
     <!-- Shared code -->
     <script defer src="shared/js/l10n.js"></script>
-    <script defer src="shared/js/l10n_date.js"></script>
-    <!-- Should be removed while gecko support navigator.mozHour12 API -->
+    <!-- Should be removed when bug 1172732 lands -->
     <script defer src="shared/js/date_time_helper.js"></script>
     <script defer src="shared/js/template.js"></script>
     <script defer src="shared/js/async_storage.js"></script>
@@ -271,7 +270,7 @@
                 </div>
                 <div class="composer-message-input-container">
                   <div id="messages-input" contentEditable="true" name="message"
-                       x-inputmode="-moz-sms" class="js-l10n-placeholder"
+                       x-inputmode="-moz-sms" data-l10n-id="messagesInput"
                        role="textbox" aria-multiline="true">
                   </div>
                 </div>
@@ -445,8 +444,7 @@
       <span data-l10n-id="${titleL10n}"></span>
       <span class="timestamp l10n-contains-date"
             data-l10n-date="${timestamp}"
-            data-l10n-date-format12="${messageL10nDateFormat12}"
-            data-l10n-date-format24="${messageL10nDateFormat24}"
+            data-l10n-date-format="${messageL10nDateFormat}"
             >${reportDateL10n}</span>
       -->
     </div>
--- a/apps/sms/views/conversation/js/information.js
+++ b/apps/sms/views/conversation/js/information.js
@@ -81,20 +81,24 @@ const MESSAGE_EVENTS = [
   'message-sending'
 ];
 
+const reportDateTimeFormatOptions = {
+  month: 'long',
+  day: '2-digit',
+  year: 'numeric',
+  hour: 'numeric',
+  minute: 'numeric'
+};
+
 function completeLocaleFormat(timestamp) {
-  return Utils.date.format.localeFormat(
-    new Date(+timestamp),
-    navigator.mozL10n.get(
-      navigator.mozHour12 ? 'report-dateTimeFormat12' :
-        'report-dateTimeFormat24'
-    )
+  return new Date(+timestamp).toLocaleString(
+    navigator.languages,
+    reportDateTimeFormatOptions
   );
 }
 
 function l10nContainsDateSetup(element, timestamp) {
   element.dataset.l10nDate = timestamp;
-  element.dataset.l10nDateFormat12 = 'report-dateTimeFormat12';
-  element.dataset.l10nDateFormat24 = 'report-dateTimeFormat24';
+  element.dataset.l10nDateFormat = JSON.stringify(reportDateTimeFormatOptions);
   element.textContent = completeLocaleFormat(timestamp);
 }
 
@@ -108,9 +112,8 @@ function createReportDiv(reports) {
   var data = {
     titleL10n: '',
     reportDateL10n: '',
-    timestamp: '',
-    messageL10nDateFormat12: 'report-dateTimeFormat12',
-    messageL10nDateFormat24: 'report-dateTimeFormat24'
+    messageL10nDateFormat: JSON.stringify(reportDateTimeFormatOptions),
+    timestamp: ''
   };
   var status;
   var deliveryStatus = reports.deliveryStatus;
--- a/apps/sms/views/conversation/test/unit/conversation_test.js
+++ b/apps/sms/views/conversation/test/unit/conversation_test.js
@@ -2723,10 +2723,6 @@ suite('conversation.js >', function() {
     }
 
     setup(function() {
-      // we do this here because of sinon fake timers
-      this.sinon.stub(Utils.date.format, 'localeFormat', function() {
-        return 'date_stub';
-      });
       this.sinon.stub(MessageManager, 'retrieveMMS', function() {
         return {};
       });
@@ -2771,15 +2767,24 @@ suite('conversation.js >', function() {
         assert.equal(notDownloadedMessage.dataset.l10nId,
           'tobedownloaded-attachment',
           'localization id set correctly');
+        var formatter = new Intl.DateTimeFormat(navigator.languages, {
+          weekday: 'long',
+          month: 'short',
+          day: '2-digit',
+        });
         assert.equal(notDownloadedMessage.dataset.l10nArgs,
-          '{"date":"date_stub"}',
+          JSON.stringify({'date': formatter.format(message.expiryDate)}),
           'localization arguments set correctly');
       });
       test('date is correctly determined', function() {
-        assert.equal(+Utils.date.format.localeFormat.args[0][0],
-          message.expiryDate);
-        assert.equal(Utils.date.format.localeFormat.args[0][1],
-          'expiry-date-format');
+        assert.deepEqual(
+          JSON.parse(notDownloadedMessage.dataset.l10nDateFormat),
+          {
+            weekday: 'long',
+            month: 'short',
+            day: '2-digit',
+          }
+        );
       });
       test('button text is correct', function() {
         assert.equal(button.dataset.l10nId, 'downloading-attachment');
@@ -2829,15 +2834,24 @@ suite('conversation.js >', function() {
         assert.equal(notDownloadedMessage.dataset.l10nId,
           'tobedownloaded-attachment',
           'localization id set correctly');
+        var formatter = new Intl.DateTimeFormat(navigator.languages, {
+          weekday: 'long',
+          month: 'short',
+          day: '2-digit',
+        });
         assert.equal(notDownloadedMessage.dataset.l10nArgs,
-          '{"date":"date_stub"}',
+          JSON.stringify({'date': formatter.format(message.expiryDate)}),
           'localization arguments set correctly');
       });
       test('date is correctly determined', function() {
-        assert.equal(+Utils.date.format.localeFormat.args[0][0],
-          message.expiryDate);
-        assert.equal(Utils.date.format.localeFormat.args[0][1],
-          'expiry-date-format');
+        assert.deepEqual(
+          JSON.parse(notDownloadedMessage.dataset.l10nDateFormat),
+          {
+            weekday: 'long',
+            month: 'short',
+            day: '2-digit',
+          }
+        );
       });
       test('button text is correct', function() {
         assert.equal(button.dataset.l10nId, 'download-attachment');
@@ -3028,27 +3042,34 @@ suite('conversation.js >', function() {
         assert.isFalse(element.classList.contains('pending'));
       });
       test('message is correct', function() {
-        assert.equal(
-          notDownloadedMessage.dataset.l10nId,
+        assert.equal(notDownloadedMessage.dataset.l10nId,
           'tobedownloaded-attachment',
-          'localization id set correctly'
-        );
-        assert.equal(
-          notDownloadedMessage.dataset.l10nArgs,
-          '{"date":"date_stub"}',
-          'localization arguments set correctly'
-        );
+          'localization id set correctly');
+        var l10nFormat = {
+          weekday: 'long',
+          month: 'short',
+          day: '2-digit',
+        };
+        var formatter =
+          new Intl.DateTimeFormat(navigator.languages, l10nFormat);
+        assert.equal(notDownloadedMessage.dataset.l10nArgs,
+          JSON.stringify({'date': formatter.format(message.expiryDate)}),
+          'localization arguments set correctly');
         assert.equal(
           notDownloadedMessage.dataset.l10nDateFormat,
-          'expiry-date-format',
+          JSON.stringify(l10nFormat),
           'localization date format set correctly'
         );
       });
       test('date is correctly determined', function() {
-        assert.equal(+Utils.date.format.localeFormat.args[0][0],
-          message.expiryDate);
-        assert.equal(Utils.date.format.localeFormat.args[0][1],
-          'expiry-date-format');
+        assert.deepEqual(
+          JSON.parse(notDownloadedMessage.dataset.l10nDateFormat),
+          {
+            weekday: 'long',
+            month: 'short',
+            day: '2-digit',
+          }
+        );
       });
       test('button text is correct', function() {
         assert.equal(button.dataset.l10nId, 'download-attachment');
@@ -3140,15 +3161,26 @@ suite('conversation.js >', function() {
         assert.equal(notDownloadedMessage.dataset.l10nId,
           'expired-attachment',
           'localization id set correctly');
+        var l10nFormat = {
+          weekday: 'long',
+          month: 'short',
+          day: '2-digit',
+        };
+        var formatter =
+          new Intl.DateTimeFormat(navigator.languages, l10nFormat);
         assert.equal(notDownloadedMessage.dataset.l10nArgs,
-          '{"date":"date_stub"}',
+          JSON.stringify({'date': formatter.format(message.expiryDate)}),
           'localization arguments set correctly');
       });
       test('date is correctly determined', function() {
-        assert.equal(+Utils.date.format.localeFormat.args[0][0],
-          message.expiryDate);
-        assert.equal(Utils.date.format.localeFormat.args[0][1],
-          'expiry-date-format');
+        assert.deepEqual(
+          JSON.parse(notDownloadedMessage.dataset.l10nDateFormat),
+          {
+            weekday: 'long',
+            month: 'short',
+            day: '2-digit',
+          }
+        );
       });
       suite('clicking', function() {
         setup(function() {
--- a/apps/sms/views/conversation/test/unit/information_test.js
+++ b/apps/sms/views/conversation/test/unit/information_test.js
@@ -302,7 +302,6 @@ suite('Information view', function() {
         };
         return request;
       });
-      this.sinon.spy(Utils.date.format, 'localeFormat');
       reportView.beforeEnter();
     });
 
@@ -558,18 +557,18 @@ suite('Information view', function() {
           assert.isFalse(
             reportView.container.classList.contains('no-valid-sent-timestamp')
           );
+
+          var formatter = new Intl.DateTimeFormat(
+            navigator.languages,
+            JSON.parse(sentTimestampNode.dataset.l10nDateFormat));
+
           assert.equal(
             sentTimestampNode.textContent,
-            Utils.date.format.localeFormat(
-              new Date(messageOpts.sentTimestamp),
-              'report-dateTimeFormat' + hourPostfix
-            )
+            formatter.format(new Date(messageOpts.sentTimestamp))
           );
           assert.equal(
             sentTimestampNode.dataset.l10nDate, messageOpts.sentTimestamp
           );
-          assert.ok(sentTimestampNode.dataset.l10nDateFormat12);
-          assert.ok(sentTimestampNode.dataset.l10nDateFormat24);
         });
       });
     });
@@ -735,8 +734,13 @@ suite('Information view', function() {
           titleL10n: '',
           reportDateL10n: '',
           timestamp: '',
-          messageL10nDateFormat12: 'report-dateTimeFormat12',
-          messageL10nDateFormat24: 'report-dateTimeFormat24'
+          messageL10nDateFormat: JSON.stringify({
+            month: 'long',
+            day: '2-digit',
+            year: 'numeric',
+            hour: 'numeric',
+            minute: 'numeric'
+          })
         };
       });
 
@@ -792,10 +796,13 @@ suite('Information view', function() {
             reportView.render();
 
             data.titleL10n = 'report-status-delivered';
-            data.reportDateL10n = Utils.date.format.localeFormat(
-              new Date(messageOpts.deliveryTimestamp),
-              navigator.mozL10n.get('report-dateTimeFormat' + hourPostfix)
-            );
+
+            var formatter = new Intl.DateTimeFormat(
+              navigator.languages,
+              JSON.parse(data.messageL10nDateFormat));
+
+            data.reportDateL10n = formatter.format(
+              new Date(messageOpts.deliveryTimestamp));
             data.timestamp = '' + messageOpts.deliveryTimestamp;
             sinon.assert.calledWith(Template.prototype.interpolate, data);
             reportDiv = getInfoBlock(reportView.renderContactList);
@@ -848,8 +855,13 @@ suite('Information view', function() {
           titleL10n: '',
           reportDateL10n: sinon.match.any,
           timestamp: '',
-          messageL10nDateFormat12: 'report-dateTimeFormat12',
-          messageL10nDateFormat24: 'report-dateTimeFormat24'
+          messageL10nDateFormat: JSON.stringify({
+            month: 'long',
+            day: '2-digit',
+            year: 'numeric',
+            hour: 'numeric',
+            minute: 'numeric'
+          })
         };
       });
 
@@ -978,10 +990,13 @@ suite('Information view', function() {
               block = getInfoBlock(reportView.renderContactList);
               data.titleL10n = 'report-status-read';
               data.timestamp = '' + deliveryInfo.deliveryTimestamp;
-              data.reportDateL10n = Utils.date.format.localeFormat(
-                new Date(messageOpts.deliveryInfo[0].readTimestamp),
-                navigator.mozL10n.get('report-dateTimeFormat' + hourPostfix)
-              );
+
+              var formatter = new Intl.DateTimeFormat(
+                navigator.languages,
+                JSON.parse(data.messageL10nDateFormat));
+
+              data.reportDateL10n = formatter.format(
+                new Date(messageOpts.deliveryInfo[0].readTimestamp));
               data.timestamp = '' + messageOpts.deliveryInfo[0].readTimestamp;
 
               assert.equal(block.dataset.deliveryStatus, 'read');
--- a/apps/sms/views/shared/js/activity_handler.js
+++ b/apps/sms/views/shared/js/activity_handler.js
@@ -213,17 +213,15 @@ var ActivityHandler = {
       });
     };
 
-    navigator.mozL10n.once(function waitLocalized() {
-      if (!document.hidden) {
-        // Case of calling from Notification
-        navigateToView();
-        return;
-      }
+    if (!document.hidden) {
+      // Case of calling from Notification
+      navigateToView();
+      return;
+    }
 
-      document.addEventListener('visibilitychange', function waitVisibility() {
-        document.removeEventListener('visibilitychange', waitVisibility);
-        navigateToView();
-      });
+    document.addEventListener('visibilitychange', function waitVisibility() {
+      document.removeEventListener('visibilitychange', waitVisibility);
+      navigateToView();
     });
   },
 
--- a/apps/sms/views/shared/js/localization_helper.js
+++ b/apps/sms/views/shared/js/localization_helper.js
@@ -1,5 +1,3 @@
-/* global Utils */
-
 /* exported LocalizationHelper */
 (function(exports) {
   'use strict';
@@ -25,32 +23,29 @@
    * changed and we need to update only elements with time aware formats.
    */
   function localizeDateTime(isTimeFormatChanged) {
-    var dateTimeElements = document.querySelectorAll(isTimeFormatChanged ?
-      '.l10n-contains-date[data-l10n-date-format12][data-l10n-date-format24]' :
-      '.l10n-contains-date'
-    );
+    var dateTimeElements = document.querySelectorAll('.l10n-contains-date');
+
     Array.forEach(dateTimeElements, function(element) {
       if (!element.dataset.l10nDate) {
         return;
       }
 
-      // If element contains time-aware format then we should respect current
-      // time format setting
       var formatL10nId = element.dataset.l10nDateFormat;
-      if (element.dataset.l10nDateFormat12 &&
-          element.dataset.l10nDateFormat24) {
-        formatL10nId = navigator.mozHour12 ?
-          element.dataset.l10nDateFormat12 : element.dataset.l10nDateFormat24;
-      }
 
       if (!formatL10nId) {
         return;
       }
 
-      var localeData = Utils.date.format.localeFormat(
-        new Date(+element.dataset.l10nDate),
-        navigator.mozL10n.get(formatL10nId)
-      );
+      formatL10nId = JSON.parse(formatL10nId);
+      
+      if (isTimeFormatChanged && !formatL10nId.hour) {
+        return;
+      }
+      formatL10nId.hour12 = navigator.mozHour12;
+
+      var formatter = 
+        new Intl.DateTimeFormat(navigator.languages, formatL10nId);
+      var localeData = formatter.format(new Date(+element.dataset.l10nDate));
 
       if (element.hasAttribute('data-l10n-id') &&
           element.hasAttribute('data-l10n-args')) {
@@ -63,18 +58,6 @@
     });
   }
 
-   /**
-   * Re-translate the placeholder messages
-   */
-  function localizePlaceholders() {
-    var placeholderElements = document.querySelectorAll('.js-l10n-placeholder');
-    Array.forEach(placeholderElements, function(element) {
-      element.dataset.placeholder = navigator.mozL10n.get(
-        Utils.camelCase(element.id) + '_placeholder'
-      );
-    });
-  }
-
   exports.LocalizationHelper = {
     init: function () {
       // This will be called during startup, and every time the language is
@@ -82,7 +65,6 @@
       navigator.mozL10n.ready(function localized() {
         localizeIFrames();
         localizeDateTime();
-        localizePlaceholders();
       });
 
       // This event is fired when time format (12h/24h) is changed
--- a/apps/sms/views/shared/js/time_headers.js
+++ b/apps/sms/views/shared/js/time_headers.js
@@ -43,23 +43,26 @@
         return;
       }
 
-      var newElement;
-
       // only date
       if (element.dataset.dateOnly === 'true') {
-        newElement = Utils.getHeaderDate(ts);
+        Utils.setHeaderDate({
+          time: ts,
+          element: element,
+          withTime: false
+        });
 
       // only time
       } else if (element.dataset.timeOnly === 'true') {
-        newElement = Utils.getFormattedHour(ts);
+        element.removeAttribute('data-l10n-id');
+        element.textContent = Utils.getFormattedHour(ts);
 
       // date + time
       } else {
-        newElement = Utils.getHeaderDate(ts) + ' ' + Utils.getFormattedHour(ts);
-      }
-
-      if (newElement !== element.textContent) {
-        element.textContent = newElement;
+        Utils.setHeaderDate({
+          time: ts,
+          element: element,
+          withTime: true
+        });
       }
     }
   };
--- a/apps/sms/views/shared/js/utils.js
+++ b/apps/sms/views/shared/js/utils.js
@@ -35,40 +35,91 @@
     },
     getFormattedHour: function ut_getFormattedHour(time) {
       this.date.shared.setTime(+time);
-      return this.date.format.localeFormat(
-        this.date.shared,
-        navigator.mozL10n.get(
-          navigator.mozHour12 ? 'shortTimeFormat12' : 'shortTimeFormat24'
-        )
-      );
+      return this.date.shared.toLocaleString(navigator.languages, {
+        hour12: navigator.mozHour12,
+        hour: 'numeric',
+        minute: 'numeric',
+      });
     },
     getDayDate: function re_getDayDate(time) {
       this.date.shared.setTime(+time);
       this.date.shared.setHours(0, 0, 0, 0);
       return this.date.shared.getTime();
     },
-    getHeaderDate: function ut_giveHeaderDate(time) {
-      var _ = navigator.mozL10n.get;
+    _getFormatter: function ut_getFormatter({options, withTime}) {
+      if (withTime) {
+        options.hour12 = navigator.mozHour12;
+        options.hour = 'numeric';
+        options.minute = 'numeric';
+      }
+      return new Intl.DateTimeFormat(navigator.languages, options);
+    },
+    setHeaderDate: function ut_setHeaderDate({time, element, withTime}) {
+      var formatter;
       var today = Utils.getDayDate(Date.now());
       var otherDay = Utils.getDayDate(time);
       var dayDiff = (today - otherDay) / 86400000;
       this.date.shared.setTime(+time);
 
       if (isNaN(dayDiff)) {
-        return _('incorrectDate');
+        element.setAttribute('data-l10n-id', 'incorrectDate');
+        return;
       }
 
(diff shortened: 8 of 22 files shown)


lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1affdca7a2321191b6786e93e7d5f085ab4468fa

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=965fae5886c2c526c6b6597cb9b4aff3edaf1b3c

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6e12b0c65877a47340584229bb7dac007f13de0e

apps/sms/views/shared/test/unit/activity_handler_test.js
```diff
@@ -460,7 +460,7 @@ suite('ActivityHandler', function() {
       });
     });
 
-    suite('Close notification', function() {
+    suite.skip('Close notification', function() {
```
zbraniecki: I don't know how to test this. The notification object is returned by the NotificationHelper promise and I'm not sure how to access it here.

apps/sms/views/conversation/test/unit/information_test.js
```diff
@@ -427,7 +426,7 @@ suite('Information view', function() {
       });
     });
 
-    test('Outgoing Message report(MMS w/ subject)', function() {
+    test.skip('Outgoing Message report(MMS w/ subject)', function() {
```
zbraniecki: I don't know why those two tests fail. The code complains about missing reports.readTimestamp and I'm not sure where it was supposed to be added.

apps/sms/views/conversation/test/unit/conversation_test.js
```diff
@@ -3165,15 +3186,27 @@ suite('conversation.js >', function() {
...
-          '{"date":"date_stub"}',
+          JSON.stringify({'date': formatter.format(message.expiryDate)}),
           'localization arguments set correctly');
       });
-      test('date is correctly determined', function() {
-        assert.equal(+Utils.date.format.localeFormat.args[0][0],
-          message.expiryDate);
-        assert.equal(Utils.date.format.localeFormat.args[0][1],
-          'expiry-date-format');
+      test.skip('date is correctly determined', function() {
```
zbraniecki: I'm skipping this test because I don't understand why in the next line we expect dataset to be empty.

apps/sms/views/conversation/js/attachment.js
```diff
@@ -26,8 +26,7 @@
   function Attachment(blob, options) {
     options = options || {};
     this.blob = blob;
-    this.name = blob.name || options.name ||
-      navigator.mozL10n.get('unnamed-attachment');
+    this.name = blob.name || options.name || 'undefined';
```
julienw: I think we still need to translate this ?

apps/sms/views/conversation/test/unit/conversation_test.js
julienw: can you point me where this test is in the original code ?

zbraniecki: line 3172 in apps/sms/views/conversation/test/unit/conversation_test.js

apps/sms/views/conversation/js/attachment.js
zbraniecki: I don't know. In all cases where Attachment is being used the name is there. The only place where there is no name is in tests where it just tests if it's a string.

So my thinking is that it should be '' or undefined and if you want to show something, we should do:
```
if (attachment.name) {
  element.textContent = attachment.name;
} else {
  element.setAttribute('data-l10n-id', 'unknown');
}
```
I just don't know where is the code that displays the name.

julienw: from what I read in the code (basically `git grep 'new Attachment`) in most cases the name comes from external sources (activities). So we're never sure we'll have it properly. So this can definitely happen in some cases and we need to localize it.

Currently, it's used in these locations:
* https://github.com/mozilla-b2g/gaia/blob/956982413ef87c192ebc778012f6c085e678570b/apps/sms/views/conversation/js/conversation.js#L51 (this is then used when generating the SMIL markup when we send the MMS -- happens only after a user action)
* https://github.com/mozilla-b2g/gaia/blob/956982413ef87c192ebc778012f6c085e678570b/apps/sms/views/conversation/js/compose.js#L750 (used when displaying the file in gallery -- happens only after a user action)
* https://github.com/mozilla-b2g/gaia/blob/956982413ef87c192ebc778012f6c085e678570b/apps/sms/views/conversation/js/attachment_renderer.js#L278 (used when displaying the file in the app itself -- it's shown only when it's not an image). Note that this is called from `attachment.render()` so we can pretty well control when it's called.
* another place in attachment.js but it won't matter if we localize it or not in this place.

So here is my idea to make it work:
```javascript
function Attachment(...) {
  ...
  this.name = blob.name || options.name;
  if (this.name) {
    this.namePromise = Promise.resolve(this.name);
  } else {
    this.name = 'untitled';
    this.namePromise = navigator.mozL10n.formatValue('unnamed-attachment').then(
      (val) => this.name = val
    );
  }
  ...
}

...
render: function(readyCallback) {
  var attachmentRenderer = this._getAttachmentRenderer();

  this.namePromise.then(() => attachmentRenderer.render()).catch(function(e) {
    console.error('Error occurred while rendering attachment.', e);
  }).then(readyCallback);

  // We still need this for the case where we render a list of attachments
  // and right order is important.
  return attachmentRenderer.getAttachmentContainer();
},
```

I think all cases are covered with this?

apps/sms/views/shared/js/utils.js
```diff
@@ -33,40 +33,93 @@
...
+      return formatter.format(this.date.shared);
     },
     getDayDate: function re_getDayDate(time) {
       this.date.shared.setTime(+time);
       this.date.shared.setHours(0, 0, 0, 0);
       return this.date.shared.getTime();
     },
-    getHeaderDate: function ut_giveHeaderDate(time) {
-      var _ = navigator.mozL10n.get;
+    setHeaderDate: function ut_setHeaderDate(time, element, withTime) {
```
julienw: I don't like booleans parameters; I'd rather use a parameter object here, but we can use ES6 constructs to make it easier to use here:
```
  setHeaderDate({ time, element, withTime }) {
    ...
  }
```

shared/js/multi_sim_action_button.js
```diff
@@ -117,11 +117,9 @@ MultiSimActionButton.prototype._updateUI = function() {
       var self = this;
       var l10nId = this._simIndication.dataset.l10nId ||
                    'gaia-sim-picker-button';
-      navigator.mozL10n.ready(function() {
```
julienw: be careful as this is used in Dialer (the "call" button) and Contacts (the "dial" button in a contact page) too... So maybe we'll want to keep it

apps/sms/views/shared/js/localization_helper.js
```diff
@@ -25,32 +23,29 @@
...
-      var localeData = Utils.date.format.localeFormat(
-        new Date(+element.dataset.l10nDate),
-        navigator.mozL10n.get(formatL10nId)
-      );
+      formatL10nId = JSON.parse(formatL10nId);
+      
+      if (isTimeFormatChanged && !formatL10nId.hour) {
+        return;
+      }
+      formatL10nId.hour12 = navigator.mozHour12;
```
julienw: do you think this will eventually be handled automatically by Intl.DateTimeFormat ?

julienw: if yes, do we have a bug for this? If yes, can you add the bug number in a comment ? If no, is it worth filing one bug ?

apps/sms/views/shared/js/time_headers.js
```diff
@@ -43,23 +43,18 @@
         return;
```
julienw: thanks for making me look at this file again... I think there are a lot of things to eventually change here :)

apps/sms/views/shared/js/activity_handler.js
```diff
@@ -340,11 +338,13 @@ var ActivityHandler = {
           var titlePromise;
 
           if (Settings.hasSeveralSim() && message.iccId) {
-            var simName = Settings.getSimNameByIccId(message.iccId);
-            titlePromise = formatValue(
-              'dsds-notification-title-with-sim',
-              { sim: simName, sender: sender }
-            );
+            titlePromise = Settings.getSimNameByIccId(message.iccId).then(
```
julienw: nit: use arrow functions:
```javascript
titlePromise = Settings.getSimNameByIccId(message.iccId).then((simName) => {
  return {
    id: 'dsds-notification-title-with-sim',
    args: { sim: simName, sender: sender }
  };
});
```

If the first line is too long:
```javascript
titlePromise = Settings.getSimNameByIccId(message.iccId).then(
  (simName) => {
    return {
      id: 'dsds-notification-title-with-sim',
      args: { sim: simName, sender: sender }
    };
  }
);
```

apps/sms/views/shared/js/activity_handler.js
```diff
@@ -356,20 +356,22 @@ var ActivityHandler = {
...
-            // Close notification if we are already in thread view and view
-            // become visible.
-            if (document.hidden && threadId === Threads.currentId) {
-              document.addEventListener('visibilitychange',
-                function onVisible() {
-                  document.removeEventListener('visibilitychange', onVisible);
-                  notification.close();
-              });
-            }
+          titlePromise.then(function(title) {
```
julienw: not sure why you changed this :)

apps/sms/views/shared/js/activity_handler.js
```diff
@@ -356,20 +356,22 @@ var ActivityHandler = {
...
+            NotificationHelper.send(title, options).then(
```
julienw: Because we don't need the title inside the notification's promise chain, I'd do something with less indentation like this:

```javascript
return titlePromise.then(
  (title) => NotificationHelper.send(title, options)
).then((notification) => {
  notification.addEventListener(...);
  ...
});
```

(we return because it's good practice ... and could make unit tests easier in the future when we'll return promises from all asynchronous functions, which is not the case yet here)

apps/sms/views/conversation/test/unit/conversation_test.js
julienw: But I don't see this assertion in the original test line 3172 ?

apps/sms/views/conversation/test/unit/information_test.js
julienw: The error I get is different, I get: `RangeError: date value is not finite in DateTimeFormat.format()`

Does it ring any bell ?

note: test.skip doesn't work with our old version of mocha... it gives a javascript error and skips everything after that. And no, this is not caught by anything. Yay. :)

julienw: full stack is:
```
     RangeError: date value is not finite in DateTimeFormat.format()
      at completeLocaleFormat (app://sms.gaiamobile.org/views/conversation/js/information.js:89:10)
      at createReportDiv (app://sms.gaiamobile.org/views/conversation/js/information.js:137:29)
      at (anonymous) (app://sms.gaiamobile.org/views/conversation/js/information.js:196:29)
      at createListWithMsgInfo (app://sms.gaiamobile.org/views/conversation/js/information.js:194:5)
      at (anonymous) (app://sms.gaiamobile.org/views/conversation/js/information.js:327:32)
      at onsuccess (app://sms.gaiamobile.org/views/conversation/test/unit/information_test.js:294:13)
      at renderReport (app://sms.gaiamobile.org/views/conversation/js/information.js:264:28)
      at (anonymous) (app://sms.gaiamobile.org/views/conversation/test/unit/information_test.js:440:7)
```

julienw: OK, i think we need to add readTimestamp in https://github.com/mozilla-b2g/gaia/blob/master/apps/sms/views/shared/test/unit/mock_messages.js#L54
According to https://dxr.mozilla.org/mozilla-central/source/dom/webidl/MozMmsMessage.webidl it's never absent so the mock should be correct.

julienw: looks like DateTimeFormat was more permissive :)

apps/sms/index.html
```diff
@@ -62,9 +62,6 @@
     -->
     <!-- Shared code -->
     <script defer src="shared/js/l10n.js"></script>
-    <script defer src="shared/js/l10n_date.js"></script>
-    <!-- Should be removed while gecko support navigator.mozHour12 API -->
-    <script defer src="shared/js/date_time_helper.js"></script>
```
julienw: does it mean navigator.mozHour12 exists without this lib ?

julienw: unrelated: can you also do the change in views/inbox/index.html ? It's not used yet but you can try to load it in Firefox in a DEBUG=1 DESKTOP=0 profile.

apps/sms/views/shared/test/unit/activity_handler_test.js
```diff
@@ -357,8 +357,8 @@ suite('ActivityHandler', function() {
 
       test('passes contact name in plain text', function() {
         sinon.assert.called(sendSpy);
-        var notification = sendSpy.firstCall.thisValue;
-        assert.equal(notification.title, contactName);
+        var notification = sendSpy.firstCall;
+        assert.isTrue(notification.calledWith(contactName));
```
julienw: nit: use `sinon.assert.calledWith(notification, contactName)`

apps/sms/views/shared/test/unit/activity_handler_test.js
```diff
@@ -396,8 +396,8 @@ suite('ActivityHandler', function() {
 
       test('phone in notification title when contact without name', function() {
         sinon.assert.called(sendSpy);
-        var notification = sendSpy.firstCall.thisValue;
-        assert.equal(notification.title, phoneNumber);
+        var notification = sendSpy.firstCall;
+        assert.equal(notification.args[0], phoneNumber);
```
julienw: nit: use `sinon.assert.calledWith(sendSpy, phoneNumber)`. You can even remove the first line as well.

apps/sms/views/shared/test/unit/activity_handler_test.js
```diff
@@ -418,28 +418,28 @@ suite('ActivityHandler', function() {
 
       test('email in notification title when contact without name', function() {
         sinon.assert.called(sendSpy);
-        var notification = sendSpy.firstCall.thisValue;
-        assert.equal(notification.title, emailAddress);
+        var notification = sendSpy.firstCall;
+        assert.isTrue(notification.calledWith(emailAddress));
```
julienw: nit: use sinon.assert.calledWith

apps/sms/views/shared/test/unit/activity_handler_test.js
```diff
@@ -418,28 +418,28 @@ suite('ActivityHandler', function() {
...
-        assert.equal(notification.body, message.body);
-        assert.deepEqual(notification.data, {
+        var notification = sendSpy.firstCall;
+        assert.equal(notification.args[1].body, message.body);
+        assert.deepEqual(notification.args[1].data, {
           id: message.id,
           threadId: message.threadId
         });
-        assert.equal(notification.icon, 'sms');
+        assert.equal(notification.args[1].icon, 'sms');
```
julienw: nit: use: `sinon.assert.calledWith(sendSpy, { body: ..., data: ..., icon: ... });` (maybe you'll need calledWithMatch)

apps/sms/views/shared/test/unit/activity_handler_test.js
julienw: Right, the object returnd by the mock for NotificationHelper is not really easy to work with.
You can replace MockNotificationHelper.send current code with:
```javascript
  send: function(titleL10n, options) {
    this.mOptions = options;
    this.mTitleL10n = titleL10n;
    var notification = {
      addEventListener: function(type, cb) {
        self.mEvents[type] = cb;
      },
      close: sinon.stub()
    };
    sinon.stub(notification, 'addEventListener');

    return Promise.resolve(notification);
  },
```

Then from your test:

```javascript
// in setup or at the start of test
this.sinon.spy(NotificationHelper, 'send');

... test stuff

NotificationHelper.send.firstCall.returnValue.then((notification) => {
  sinon.assert.called(notification.close);
}).then(done, done);
```

apps/sms/views/shared/test/unit/settings_test.js
```diff
@@ -36,8 +36,12 @@ suite('Settings >', function() {
...
   });
 
-  test('getSimNameByIccId returns the empty string before init', function() {
-    assert.equal('', Settings.getSimNameByIccId('anything'));
+  test('getSimNameByIccId returns the empty string before init',
+    function(done) {
+      Settings.getSimNameByIccId('anything').then(function(ret) {
+        assert.equal('', ret);
+        done();
+      });
```
julienw: nit: use the following construct:
```javascript
functionReturningPromise.then(() => {
  ... assert stuff
}).then(done, done);
```

Then "done" correctly catch any assertion error and forward it to mocha; otherwise you lose it and it just times out

apps/sms/views/shared/test/unit/settings_test.js
```diff
@@ -80,8 +84,11 @@ suite('Settings >', function() {
       assert.equal('', Settings.getOperatorByIccId('anything'));
     });
 
-    test('getSimNameByIccId returns the empty string', function() {
-      assert.equal('', Settings.getSimNameByIccId('anything'));
+    test('getSimNameByIccId returns the empty string', function(done) {
+      Settings.getSimNameByIccId('anything').then(function(ret) {
+        assert.equal('', ret);
+        done();
+      });
```
julienw: same

apps/sms/views/shared/test/unit/settings_test.js
```diff
@@ -238,16 +245,19 @@ suite('Settings >', function() {
...
+        });
+        var test2 = Settings.getSimNameByIccId('SIM 2').then(function(ret) {
+          assert.equal('sim-id-label{"id":2}', ret);
+        });
+        var test3 = Settings.getSimNameByIccId('SIM 3').then(function(ret) {
+          assert.equal('', ret);
+        });
+        Promise.all([test1, test2, test3]).then(function() {
+          done();
+        });
```
julienw: use `.then(() => done(), done)`

apps/sms/views/shared/test/unit/time_headers_test.js
```diff
@@ -94,7 +94,9 @@ suite('TimeHeaders > ', function() {
 
         var headers = document.querySelectorAll('header');
         for (i = 0, l = headers.length; i < l; i++) {
-          assert.equal(headers[i].textContent, existingHeaders[i]);
+          assert.deepEqual(
+            navigator.mozL10n.getAttributes(headers[i]),
+            existingHeaders[i]);
```
julienw: nit: put he closing parens on the next line

apps/sms/views/shared/test/unit/time_headers_test.js
```diff
@@ -117,7 +119,9 @@ suite('TimeHeaders > ', function() {
...
```
julienw: same

apps/sms/views/shared/test/unit/time_headers_test.js
```diff
@@ -139,7 +143,9 @@ suite('TimeHeaders > ', function() {
 
         var headers = document.querySelectorAll('header');
         for (i = 0, l = headers.length; i < l; i++) {
-          assert.notEqual(headers[i].textContent, existingHeaders[i]);
+          assert.notDeepEqual(
+            navigator.mozL10n.getAttributes(headers[i]),
+            existingHeaders[i]);
```
julienw: same

apps/sms/views/shared/test/unit/time_headers_test.js
```diff
@@ -168,7 +174,9 @@ suite('TimeHeaders > ', function() {
 
         var headers = document.querySelectorAll('header');
         for (i = 0, l = headers.length; i < l; i++) {
-          assert.equal(headers[i].textContent, existingHeaders[i]);
+          assert.deepEqual(
+            navigator.mozL10n.getAttributes(headers[i]),
+            existingHeaders[i]);
```
julienw: same

apps/sms/views/shared/test/unit/time_headers_test.js
```diff
@@ -193,7 +201,9 @@ suite('TimeHeaders > ', function() {
...
```
julienw: same

apps/sms/views/shared/test/unit/time_headers_test.js
```diff
@@ -218,11 +228,9 @@ suite('TimeHeaders > ', function() {
...
         var headers = document.querySelectorAll('header');
         for (i = 0, l = headers.length; i < l; i++) {
-          assert.notEqual(headers[i].textContent, existingHeaders[i]);
-        }
-
-        for (i = 0, l = timeElements.length; i < l; i++) {
-          assert.notEqual(timeElements[i].textContent, existingTimes[i]);
+          assert.notDeepEqual(
+            navigator.mozL10n.getAttributes(headers[i]),
+            existingHeaders[i]);
```
julienw: same

apps/sms/views/shared/test/unit/time_headers_test.js
```diff
@@ -233,43 +241,59 @@ suite('TimeHeaders > ', function() {
...
+        hour: 'numeric',
+        minute: 'numeric'
+      });
+
       var content = subject.textContent;
-      assert.include(content, formattedTime);
-      assert.isTrue(content.indexOf(formattedDate) === -1,
-                    'subject.textContent should not contain formattedDate');
+      assert.equal(content, formatter.format(time),
+          'subject.textContent should not contain formattedDate');
```
julienw: nit: please use the following style:
```javascript
assert.equal(
  content, formater.format(time),
  'subject...'
);
```

How you arrange things between () is not very strict, things should be as logic as possible; but if you need several line for arguments, then please have the closing and ending parens like this.

apps/sms/views/conversation/js/information.js
```diff
@@ -73,20 +73,25 @@ const MESSAGE_EVENTS = [
...
-  return Utils.date.format.localeFormat(
-    new Date(+timestamp),
-    navigator.mozL10n.get(
-      navigator.mozHour12 ? 'report-dateTimeFormat12' :
-        'report-dateTimeFormat24'
-    )
-  );
+  var formatter = new Intl.DateTimeFormat(
+    navigator.languages,
+    reportDateTimeFormatOptions);
```
julienw: nit: please put the ending parens on the next line

julienw:
I tried on a device, here are my findings:
* I think we should use "2-digit" for months and days, instead of "numeric"
* the format stays at month/day even when I select the french locale, is that expected? Mmm likely it's because I don't flash the french locale, is there a pseudo locale that supports the day/month format?
* when I change the "hour" setting between 12/24 hour it doesn't seem to change

julienw:
I think you need to rebase because of bug 1168441 that just landed

apps/sms/index.html
zbraniecki: Agh, it's not :) It'll be soon (bug 1172732), but it's not yet.

apps/sms/views/conversation/js/attachment.js
zbraniecki: ugh. Let's handle this in a follow up patch. I'm not sure if I know how to implement your solution.
Also, I believe that it's a bad habit to store localized string in the data model, since it's a view role to display it.

So I'm more on the side of:
```
function Attachment(name) {
  this.name = name;
}
[...]

function render() {
  if (!attachment.name) {
    element.textContent = attachment.name;
  } else {
    element.setAttribute('data-l10n-id', 'unnamed-attachment');
  }
}
```

But since it seems complex enough, let's come back to it in a separate patch.

apps/sms/views/conversation/test/unit/conversation_test.js
zbraniecki: lol. I have no idea what I saw and how I got it. Fixed :)

apps/sms/views/conversation/test/unit/information_test.js
zbraniecki: I uncommented and it just worked. I don't get it. But I added the readTimestamp to the mock anyway

zbraniecki: > note: test.skip doesn't work with our old version of mocha... it gives a javascript error and skips everything after that. And no, this is not caught by anything. Yay. :)

I don't think I wanted to know that.

apps/sms/views/shared/js/activity_handler.js
zbraniecki: same here. fixed :)

apps/sms/views/shared/js/activity_handler.js
zbraniecki: Yeah, I like it!

apps/sms/views/shared/js/localization_helper.js
zbraniecki: Not anytime soon. We first have to move mozHour12 to the platform (bug 1172732), then standardize it, then we can talk about incorporating it into Intl.

For now we can write wrappers to do that.

apps/sms/views/shared/test/unit/activity_handler_test.js
zbraniecki: Error: send(<&>, { body: "body", data: { id: 1, threadId: 1 }, icon: "sms", tag: "threadId:1" }) => { then: function () {} } is not a function

apps/sms/views/shared/test/unit/activity_handler_test.js
zbraniecki: Hmm, tried this:

```
        sinon.assert.calledWithMatch(sendSpy, {
          body: message.body,
          data: {
            id: message.id,
            threadId: message.threadId
          },
          icon: 'sms',
          tag: "threadId:1"
        });
```

and the error is: 
```
  1) [sms-views/shared/test/unit/activity_handler_test.js] ActivityHandler sms received after getSelf a notification is sent:
     Error: expected send to be called with match { body: "body", data: { id: 1, threadId: 1 }, icon: "sms", tag: "threadId:1" }
    send(Pepito O'Hare, { body: "body", data: { id: 1, threadId: 1 }, icon: "sms", tag: "threadId:1" }) => { then: function () {} }
```

which is weird because it's exactly how it is supposed to work :/

julienw: look closer: you miss the first argument "Pepito O'Hare" :)

apps/sms/views/shared/js/localization_helper.js
julienw: ok ! is it worth filing a bug about incorporating mozHour12 to Intl ? Or is it too early ?

apps/sms/views/conversation/test/unit/information_test.js
julienw: > I uncommented and it just worked. I don't get it. But I added the readTimestamp to the mock anyway

Weird because when I uncommented it didn't for me :)

apps/sms/views/conversation/js/attachment.js
julienw: Why not, but I think some code would break if this.name is not a string; so let's at least assign an empty string in the constructor.

* For the first case (SMIL generation) I think it's correct if the name stays blank.
* for the gallery viewing, I think we need to show something better, but this should be handled by the gallery app instead (maybe they do it already).
* for the rendering inside the SMS app, I think you're right, and this could be done in attachment_renderer.
* for the last case, we don't really care but we need at least a string.

This is enough of a edge case that it could be handled in a follow-up, but if you want to do this here I'd be happy :)

shared/js/multi_sim_action_button.js
zbraniecki: that doesn't matter. The code inside is only setting data-l10n-id, data-l10n-args. Everything else is async.

And nothing in l10nArgs changes when language change so re-firing it on language change doesn't make sense.

Or am I wrong here?

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b9ba7aa40c7aefa450c91dcc0fd8dec8412e962d

apps/sms/views/shared/js/localization_helper.js
zbraniecki: too early. Intl is regulated by ECMA 402. There are several things we'd like to add in edition 2, but we need to first get navigator.hour12 before we can argue for including it into ECMA.

apps/sms/views/shared/test/unit/activity_handler_test.js
zbraniecki: geez, I was hoping calledWithMatch would allow me to only match for the selected argument. Thanks!

apps/sms/views/shared/test/unit/activity_handler_test.js
```diff
@@ -455,38 +452,38 @@ suite('ActivityHandler', function() {
...
-          sinon.assert.called(closeSpy);
-        }).then(done, done);
+        MockNavigatormozApps.mTriggerLastRequestSuccess().then(
+          () => NotificationHelper.send.firstCall.returnValue
+        ).then((notification) => {
+          sinon.assert.called(notification.addEventListener);
+          sinon.assert.notCalled(notification.close);
+          notification.addEventListener.yield();
+          sinon.assert.called(notification.close);
+        }).then(done, done);*/
```
zbraniecki: I tried to run this, but this errored.
First it errored with:
```TypeError: MessageManager.getMessage(...).then is not a function```

so I went to MessageManager mockup which doesn't return a Promise in getMessage. I changed that.

That changed the error to:

```Error: expected stub to have been called at least once but was never called```

Can you help me here?

File: apps/sms/views/shared/test/unit/activity_handler_test.js (hunk 16 above)
Comment (julienw) [inline apps/sms/views/shared/test/unit/activity_handler_test.js]: calledWith and calledWithMatch allow you to skip remaining arguments, but if you want to skip the first ones, you can use sinon.match.any. See all predefined values here: http://sinonjs.org/docs/#sinon-match-api

Matching allows you to put only some of the properties. I think that for objects if you don't use match you need a reference equality, while with match it compares the properties values ... to be checked :)

File: shared/js/multi_sim_action_button.js (hunk 6 above)
Comment (julienw) [inline shared/js/multi_sim_action_button.js]: mmm no I think you're right !

**Comment by zbraniecki** [comment]
> I think we should use "2-digit" for months and days, instead of "numeric"

Sure!

> the format stays at month/day even when I select the french locale, is that expected? Mmm likely it's because I don't flash the french locale, is there a pseudo locale that supports the day/month format?

No, it's because we don't relocalize on language/hour12 change. We should listen to timeformatchange and languagechange and update those strings.

I believe I preserved the behavior of the original code. For example in conversation.js, we didn't update the code when language changed, or when hour12 changed, so the date/time portion would stay the same even when language changed. I think I didn't regress anything.

> when I change the "hour" setting between 12/24 hour it doesn't seem to change

Same reason.

**Comment by lightsofapollo** [comment]
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=327605f23f2ef1fd94c768e16734ca9918bfa10d

**Comment by zbraniecki** [comment]
Just to clarify, I'd be happy to work on getting datetime/hour12 updated when language change, I just think it's a follow up material since it was not there in the original code and this patch is already huge :)

File: apps/sms/views/conversation/js/attachment.js (hunk 4 above)
Comment (zbraniecki) [inline apps/sms/views/conversation/js/attachment.js]: sweet. Follow up material then.

File: apps/sms/views/shared/test/unit/activity_handler_test.js [hunk 29]
```diff
@@ -455,38 +452,38 @@ suite('ActivityHandler', function() {
...
-          sinon.assert.notCalled(closeSpy);
-          document.addEventListener.yield();
-          sinon.assert.called(closeSpy);
-        }).then(done, done);
+        MockNavigatormozApps.mTriggerLastRequestSuccess().then(
+          () => NotificationHelper.send.firstCall.returnValue
+        ).then((notification) => {
+          sinon.assert.called(notification.addEventListener);
+          sinon.assert.notCalled(notification.close);
+          notification.addEventListener.yield();
```
julienw: this is not the same here: we actually want to yield the "visibilitychange"  event.

Maybe to make it clearer: `document.addEventListener.withArgs('visibilitychange').yield()`

apps/sms/views/shared/test/unit/activity_handler_test.js
```diff
@@ -455,38 +452,38 @@ suite('ActivityHandler', function() {
...
-        MockNavigatormozApps.mTriggerLastRequestSuccess().then(() => {
-          sinon.assert.called(document.addEventListener);
-          sinon.assert.notCalled(closeSpy);
-          document.addEventListener.yield();
-          sinon.assert.called(closeSpy);
-        }).then(done, done);
+        MockNavigatormozApps.mTriggerLastRequestSuccess().then(
+          () => NotificationHelper.send.firstCall.returnValue
+        ).then((notification) => {
+          sinon.assert.called(notification.addEventListener);
```
julienw: same here: we want to assert that `document.addEventListener.withArgs('visibilitychange')` is called... But you can just remove this line, as this is implied by the yield we have later on.

apps/sms/views/shared/test/unit/activity_handler_test.js
julienw: I agree that the implementation of getMessage in the mock is _really_ weird... and utterly wrong. I think we should just change this to an empty function but let's not do that here if we can avoid.

apps/sms/views/shared/test/unit/activity_handler_test.js
zbraniecki: sweet. fixed!

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b63818f5784289436643f24c2fccb31ace953d90

zbraniecki:
Actually, in localization_helper does listen to timeformatchange and I see it updating the 12/24h in message list view when I toggle the setting.

julienw:
Looks like you need to update information_test.js for the numeric => 2-digit change too: https://treeherder.mozilla.org/logviewer.html#?job_id=1349902&repo=gaia

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=782c6acb4fb7d92eb97f9fc1bdb97f8c35a470b5

apps/sms/views/shared/js/utils.js
```diff
@@ -33,40 +33,93 @@
...
+          year: 'numeric',
+        };
+        if (withTime) {
+          options.hour12 = navigator.mozHour12;
+          options.hour = 'numeric';
+          options.minute = 'numeric';
+        }
+        element.removeAttribute('data-l10n-id');
+        formatter = new Intl.DateTimeFormat(navigator.languages, options);
+        element.textContent = formatter.format(this.date.shared);
```
julienw: This is basically the same code than the future time block ahead -- I think we should factorize this to an inner function. Maybe even factorize to include the block for `dayDiff < 6` (for example using 2 functions that both call a 3rd function).

What do you think?

julienw:
Do you know what  I need to make TODAY show up localized in french ? I tried fetching sms.fr.properties and date.fr.properties (in shared) from gaia-l10n, and I even added todayWithTime and yesterdayWithTime to sms.fr.properties.

apps/sms/views/shared/js/utils.js
zbraniecki: yeah, sure!

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3713deeadbdfa1d332e283b75a8761c94850d029

zbraniecki:
So, you have ./apps/sms/locales//sms.en-US.properties:today and you need to translate it. I just plugged to webide and did navigator.mozL10n.get('today') and it returns localized "today".

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=41626cdb101b2bf0670b51bdb04d6a3ab0a99be3

apps/sms/views/shared/js/utils.js
```diff
@@ -33,40 +33,91 @@
...
+        });
+        element.removeAttribute('data-l10n-id');
+        element.textContent = formatter.format(this.date.shared);
+        return;
       }
 
-      return dayDiff === 0 && _('today') ||
-        dayDiff === 1 && _('yesterday') ||
-        dayDiff < 6 && this.date.format.localeFormat(this.date.shared, '%A') ||
-        this.date.format.localeFormat(this.date.shared, '%x');
```
julienw: see here: we were using %x which translates to %m/%d/%Y in en-US. and %d uses a leading 0 (%e uses a leading space), %m uses a leading 0 as well (I don't see any command character that uses a leading space). Also note that %b is for short _names_, not digits.

So, do yo know why  "numeric" in Intl yields leading spaces but our previous translations yields leading zeros ? Is it specified in the Intl spec ?

Don't worry, this is not holding this patch, but I want to be sure we don't have an underlying bug in Intl.

zbraniecki: We don't control it in Intl. It's ICU provided data (which is nice because it standardizes the patterns between systems and browsers).

I'd point out at Ecma 402 spec, but the ecma-international.org website is down :)

julienw: yeah I'm looking at the drafts in http://wiki.ecmascript.org/doku.php?id=globalization:specification_drafts (but ecmascript specs are so difficult to understand).

OK then we'll try to settle this later

julienw:
With the current l10n framework it seems impossible to just throw a file sms.fr.properties in sms/locales/ and take it into account. This is a bit sad :(

zbraniecki:
The list of available languages is defined in <meta> of the document:
```
    <meta name="defaultLanguage" content="en-US">
    <meta name="availableLanguages" content="en-US, fr">
```

we do this at build time for you - https://developer.mozilla.org/en-US/docs/Mozilla/Firefox_OS/Building#Building_multilocale

If we didn't know what languages are available, we wouldn't be able to do any language negotiation (between available, and user preferences) and we'd have to blindly XHR for resources until we get the one we want. :) That wouldn't be very efficient or smart.

apps/sms/views/shared/js/utils.js
zbraniecki: The bottom line is, I don't think there's any room for a bug in Intl implementation around this. That's coming straight from ICU. There may be bugs in ICU (and we're updating our ICU fairly often (bug 992079 and bug 1075758), but there shouldn't be too many, and with Safari, Chrome, Windows and MacOS all using the same base technology as us, we should get more consistent and error-free experience.

One benefit of this change is that we're moving from localizer-defined formatting to developer-defined formatting. Which means that if next month you'll decide to switch something from 2-digit to numeric or long, you're free to do that at no cost to localizers.

And since we're not close to any release, we'll have a lot of time to ensure the formatting looks good.

julienw:
Yeah I know I could do all this, but just putting a file in locales/ should ideally work as well -- I mean this could be discovered at build time too.

But well, out of the scope of this patch anyway. It just makes things harder to test.

julienw:
So I don't know why yet but the current patch breaks the app now. I get these errors:

E/Messages(18325): [JavaScript Error: "too much recursion"]
E/Messages(18325): [JavaScript Error: "ReferenceError: Dialog is not defined" {file: "app://sms.gaiamobile.org/views/shared/js/error_dialog.js" line: 108}]

But I don't have more information than this; I'm digging...

zbraniecki:
Long story short, we can't yet autodiscover, because we do build-time merging. So if we see a flag to build 'fr' locale and there's no sms.fr.properties we will create it based on fallback chain (probably using en-US). Once we switch to client-side, we could aim at build time detection.

zbraniecki:
Do you have STR?

julienw:
It's just launching the app.

I built a new B2G and flashed and I don't have the issue anymore now...

apps/sms/index.html
```diff
@@ -62,8 +62,7 @@
     -->
     <!-- Shared code -->
     <script defer src="shared/js/l10n.js"></script>
-    <script defer src="shared/js/l10n_date.js"></script>
```
julienw: please remove this from views/inbox/index.html as well

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a3afaca054d3bedca29418e6c2acb00682db2d36

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c87c053300c03c6197ca225015582d3ef9170d31

apps/sms/index.html
```diff
@@ -445,8 +444,7 @@ <h1 data-l10n-id="message-report"></h1>
       <span data-l10n-id="${titleL10n}"></span>
       <span class="timestamp l10n-contains-date"
             data-l10n-date="${timestamp}"
-            data-l10n-date-format12="${messageL10nDateFormat12}"
-            data-l10n-date-format24="${messageL10nDateFormat24}"
+            data-l10n-date-format="${messageL10nDateFormat}"
```
julienw: please put this change in views/conversation/index.html as well

apps/sms/index.html
```diff
@@ -271,7 +270,7 @@ <h1 id="messages-header-text" aria-hidden="true"></h1>
                 </div>
                 <div class="composer-message-input-container">
                   <div id="messages-input" contentEditable="true" name="message"
-                       x-inputmode="-moz-sms" class="js-l10n-placeholder"
+                       x-inputmode="-moz-sms" data-l10n-id="messagesInput"
```
julienw: please put these changes in views/conversation/index.html as well

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0a37b6655bab6ecf6482ace3326603f27a3f649e

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0df8d88320492897f0b4eb9e522099e103f09274

PR #30420, opened 2015-06-04
Bug 1162040 - updating accessibility for gaia-icons used in settings …

yzen:
…app.

Note: some id's are added because aria-describedby spec only allows for idrefs to be used for values.

### DIFF
--- a/apps/settings/elements/root.html
+++ b/apps/settings/elements/root.html
@@ -24,37 +24,37 @@
           </label>
         </li>
         <li data-category="connectivity">
-          <a id="menuItem-wifi" class="menu-item" data-icon="wifi-4" href="#wifi">
+          <a id="menuItem-wifi" class="menu-item" data-icon="wifi-4" href="#wifi" data-l10n-id="wifiSection" aria-describedby="wifi-desc">
             <span data-l10n-id="wifi">Wi-Fi</span>
             <small id="wifi-desc" class="menu-item-desc"></small>
           </a>
         </li>
         <li id="simCardManager-settings">
-          <a id="menuItem-simManager" class="menu-item" data-icon="sim" href="#sim-manager">
+          <a id="menuItem-simManager" class="menu-item" data-icon="sim" href="#sim-manager" data-l10n-id="simManagerSection" aria-describedby="simmanager-desc">
             <span data-l10n-id="simManager">SIM Manager</span>
             <small id="simmanager-desc" class="menu-item-desc"></small>
           </a>
         </li>
         <li id="call-settings">
-          <a id="menuItem-callSettings" class="menu-item" data-icon="call" href="#call">
+          <a id="menuItem-callSettings" class="menu-item" data-icon="call" href="#call" data-l10n-id="callSettingsSection" aria-describedby="call-desc">
             <span data-l10n-id="callSettings">Call Settings</span>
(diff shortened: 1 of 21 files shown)


lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=484b7fd8e8582367fed143d90b6350cd97c56570

apps/settings/elements/root.html
```diff
@@ -24,37 +24,37 @@ <h2 data-l10n-id="networkAndConnectivity">Network &amp; Connectivity</h2>
           </label>
         </li>
         <li data-category="connectivity">
-          <a id="menuItem-wifi" class="menu-item" data-icon="wifi-4" href="#wifi">
+          <a id="menuItem-wifi" class="menu-item" data-icon="wifi-4" href="#wifi" data-l10n-id="wifiSection" aria-describedby="wifi-desc">
```
crh0716: I'm a little bit confused on this one. Does this mean screen reader would read both the strings of `wifiSection` and `wifi-desc`?

crh0716: `aria-hidden="true"` seems make more sense to me as the icon actually presents the menu item described by the following span element. What do you think?

apps/settings/elements/root.html
```diff
@@ -190,25 +184,25 @@ <h2 data-l10n-id="storage">Storage</h2>
       </header>
       <ul>
         <li class="pack-split usb-item">
-          <label class="pack-switch">
-            <input class="usb-switch" data-ignore type="checkbox" disabled>
+          <label class="pack-switch" role="presentation">
```
crh0716: What is `role="presentation"` for?

apps/settings/elements/root.html
yzen: Unfortunately we can't use aria-hidden on the menu item with an icon because it will make the whole sub-tree invisible to the screen reader. One thing to notice is that if aria-label (which is what data-l10n-id is for) is present for elements with link or role semantics, it automatically overrides what screen reader otherwise would read from the content of the menu item. Which means we would be loosing the description items and thus we need to use aria-describedby explicitly to make sure the screen reader user gets them.

apps/settings/elements/root.html
yzen: Here I want to drop label semantics since it's unnecessary and confusing. presentation role tells the screen reader to just treat the element as a generic container with not additional semantics (but not remote its sub-tree from being visible, like aria-hidden="true" does). so without it screen reader would say something like: `USB Storage checked check button label`
and with the role presentation it will say: `USB Storage checked check button`
which is more clear to the user

apps/settings/elements/root.html
yzen: Also since the aria-label will override all of the link's content text, it will not read the aria-label + main text in content (twice) and will only read the aria-label.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3c2d44a42ca68f16cb8c32bf98a09a2164cf3627

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b6e268b119607581dd45c9e42882355aed005ae0

PR #30431, opened 2015-06-04
Bug 1162040 - updating accessibility for gaia-icons used in sms app.

yzen:


lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=77334a4a25c45bf642f2cc1c51d0ea47a06e8c4b

apps/sms/locales/sms.en-US.properties
```diff
@@ -315,6 +315,10 @@ contact-photo.ariaLabel = Contact photo
 unread-thread.ariaLabel = Unread
 has-draft.ariaLabel = Has draft
 is-draft.ariaLabel = Draft
+optionsButton.ariaLabel = Options
+settingsButton.ariaLabel = Settings
+composeLink.ariaLabel = Compose
+callButton.ariaLabel = Call
```
julienw: Should we have longer labels ?
Like:
* options => Display available actions
* settings => Open the settings panel
* compose => Compose a new message
* call => Call this contact/phone number (depending on whether we have a known contact) -- but maybe it's not easy and in that case "Call this contact" could work in both cases IMO

yzen: @julienw : 
Generally we are trying to be as terse as possible since there's a lot of verbal content that the screen reader user needs to comprehend, thus we are trying to be short as possible. Also there's a lot of contextual information that the user has since they know where they are located within the UI. If what I have is too long how about this:
* options => Show actions
* I would leave settings as is because it's a settings button and having settings panel button is a little confusing
* compose => Compose message
* call => Call contact
What do you think?

julienw: OK I understand what you say;

Because the buttons are actions, how about having always a verb:
* show options
* show settings
* compose (message) <= optional; but I have an issue here, is that the button is really to compose a message to a recipient we choose after; as opposed to choosing an existing conversation and composing a message to _this_ recipient. I'd like a verb to show this difference. Maybe "New message" ? "new" is not used as a verb these days according to wiktionary :) I'd accept either "compose", "compose message" or "new message", unless you find something better, so please pick your prefered one !
* call <= it's pretty clear we want to call this recipient

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ea2a149f0546d216ad6cf81a6cdd47a2ff5ea7f5

apps/sms/locales/sms.en-US.properties
```diff
@@ -315,6 +315,10 @@ contact-photo.ariaLabel = Contact photo
 unread-thread.ariaLabel = Unread
 has-draft.ariaLabel = Has draft
 is-draft.ariaLabel = Draft
+optionsButton.ariaLabel = Options
```
julienw: please replace this one with Actions

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2bfaea7ffbd50756b40de362808cf2aae2006097

PR #30436, opened 2015-06-05
Bug 1161225 - Remove css transitions from software home button

jedireza:


lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=45eff5ed20bb0f221484acefdb79d0f8fa84c92f

apps/system/style/software_button_manager/software_button_manager.css
```diff
@@ -8,22 +8,15 @@
...
-#software-buttons.no-transition {
-  transition: none;
-}
-
 #screen:not(.locked) #software-buttons.visible,
 #screen.locked.attention #software-buttons.visible,
 #screen.locked.secure-app #software-buttons.visible {
-  transform: translateY(-100%);
   visibility: visible;
+  bottom: 0px;
```
mikehenrty: We can probably just get rid of the `bottom: calc(var(--software-home-button-height) * -1);` rule above and make `bottom: 0` the default since we are not animating anymore.

apps/system/style/software_button_manager/software_button_manager.css
```diff
@@ -122,5 +110,5 @@
 }
 
 #screen:not(.fullscreen-layout-app) #software-buttons-fullscreen-layout {
-  visibility: hidden;
+  visibility: visible;
```
mikehenrty: What's this rule for? AFAIK we don't currently use the fullscreen layout SHB anywhere.

jedireza: In landscape mode the menu bar on the right wasn't visible without this change.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=aeb075edd1dab1ba91afcc7d164588478ad221fa

apps/system/style/software_button_manager/software_button_manager.css
jedireza: Fixed.

apps/system/style/software_button_manager/software_button_manager.css
mikehenrty: Hmmm, that's strange because this css rule only effects a version of the SHB that we currently never display. Or at least are not supposed to. In any case, this causes the fullscreen-layout SHB to display on the homescreen, which shows a slight drop-shadow. So we need to come up with a different approach for the landscape fix.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=046b6e2e99903f0052a736a5426626f18d8a49a3

apps/system/js/software_button_manager.js
```diff
@@ -332,10 +332,10 @@
           var isPortrait = this._currentOrientation.contains('portrait');
           var newOrientation = Service.query('fetchCurrentOrientation');
           if (isPortrait && newOrientation.contains('landscape')) {
-            this.element.style.right = this.element.style.bottom;
+            this.element.style.right = 0;
```
jedireza: While debugging I noticed that the `this.element.style.bottom` value was `""` instead of `0`. I first tried setting the `bottom: 0;` style via the css file so this code change wouldn't be needed, but that didn't seem to work.

lightsofapollo:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d50af8df8fbb9748d70bf1ea428c58a5d3e421d5

PR #30728, opened 2015-06-27
Bug 1178024 - Remove overlay detection during parsing. r=gandalf

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1178024


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=dad51e771e09e72d170717561d24b1f4fdce1ca4

shared/js/l10n.js
```diff
@@ -2357,7 +2324,7 @@
     }
 
     if (typeof value === 'string') {
-      if (!entity.overlay) {
+      if (value.indexOf('<') === -1 && value.indexOf('&') === -1) {
```
zbraniecki: this will overlay strings like "Warsaw < San Francisco" and "Mail & News". I think it would make sense to test the next char not being space here.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=767f6afa3926a2057f75928edc0bd77cc45be2d5

shared/js/l10n.js
stasm: Did you mean `Warsaw <3 San Francisco`? :)

I think we should support things like `< span >` which is valid HTML.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a9c1637d884110c3cc4ee772bb5b31443269557d

shared/js/l10n.js
```diff
@@ -2278,6 +2245,8 @@
   }
 
 
+  var reHTMLEntity = /&.+;/;
```
zbraniecki: how about /\<|&\w+;/; and then we could skip the indexOf?

stasm: Yeah, sounds like a good idea.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6855196c9fb18d969cdd78236af8c09928767e42

PR #30745, opened 2015-06-30
Bug 1176926 - Introduce (again) a timeout before hiding the keyboard

timdream:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8cd682d023bfe8837d99dd0e97816c0642a99d28

apps/system/js/input_window_manager.js
```diff
@@ -422,6 +422,8 @@
     this._currentWindow = nextWindow;
   };
 
+  InputWindowManager.prototype.HIDE_INPUT_WINDOW_TIMEOUT = 200;
```
mnjul: Quick FYI: the old timeout we used at KeyboardManager for this matter was 100. See: https://github.com/mozilla-b2g/gaia/commit/35fd27531d52a371835f5188a2a2cef446c299e7

timdream: I just want to pick a larger timeout to be on the safe side :)

PR #30752, opened 2015-06-30
Bug 1155534 - [Messages][NG] Extract NewMessage view from Conversation view.

azasypkin:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1632f42c31a3025873bbc6165edb2cd1f9c30e14

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+         EventDispatcher,
+         SelectionHandler,
+         TaskRunner
+*/
+/*exported ConversationView */
+
+(function(exports) {
+'use strict';
+
+var attachmentMap = new WeakMap();
```
azasypkin: This specific to Conversation and will be left in ConversationView.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+
+function conv_mmsAttachmentClick(target) {
```
azasypkin: ```conv_mmsAttachmentClick``` specific to Conversation and will be left in ConversationView.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+  attachment.view({
+    allowSave: true
+  });
+
+  return true;
+}
+
+// reduce the Composer.getContent() into slide format used by SMIL.generate some
+// day in the future, we should make the SMIL and Compose use the same format
+function conv_generateSmilSlides(slides, content) {
```
azasypkin: I'd propose to move ```conv_generateSmilSlides``` to MessageManager/SMIL/Compose (in my another WIP it's in MessageManager right before we pass this to SMIL).

What do you think?

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    slides.push({
+      blob: content.blob,
+      name: content.name
+    });
+  }
+  return slides;
+}
+
+var ConversationView = {
+  CHUNK_SIZE: 10,
```
azasypkin: ```CHUNK_SIZE``` specific to Conversation and will be left in ConversationView.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+  // duration of the notification that message type was converted
+  CONVERTED_MESSAGE_DURATION: 3000,
```
azasypkin: All timeouts related to app notifications are shared between ConversationView and NewMessageView, I think we need to move it to Compose as Steve is trying to do in https://github.com/mozilla-b2g/gaia/pull/29944. Later on we can generalize app notification code further, but moving it in a single place would be a good first step.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+  // Toast duration when you write a long text and need more than one SMS
+  // to send it
+  ANOTHER_SMS_TOAST_DURATION: 3000,
+
+  // when sending an sms to several recipients in activity, we'll exit the
+  // activity after this delay after moving to the thread list.
+  LEAVE_ACTIVITY_DELAY: 3000,
+
+  draft: null,
+  recipients: null,
```
azasypkin: ```recipients``` specific to NewMessage and will be moved to NewMessageView.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+  // Set to |true| when in edit mode
+  inEditMode: false,
```
azasypkin: ```inEditMode``` specific to Conversation and will be left in ConversationView.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+  isNewMessageNoticeShown: false,
```
azasypkin: ```isNewMessageNoticeShown``` specific to Conversation and will be left in ConversationView.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+  shouldChangePanelNextEvent: false,
```
azasypkin: ```shouldChangePanelNextEvent``` specific to NewMessage and will be moved to NewMessageView.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+  showErrorInFailedEvent: '',
+  previousSegment: 0,
+
+  timeouts: {
+    update: null,
```
azasypkin: ```update``` is not used anymore.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    subjectLengthNotice: null
```
azasypkin: Move to ```Compose``` as mentioned above.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+  },
+
+  multiSimActionButton: null,
+  init: function conv_init() {
+    var templateIds = [
```
azasypkin: ```'recipient'``` is the only template needed in NewMessageView.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+      'new-message-notice', 'subject-max-length-notice', 'sms-counter-notice',
+      'recipient-suggestions'
+    ].forEach(function(id) {
+      this[Utils.camelCase(id)] = document.getElementById('messages-' + id);
+    }, this);
+
+    this.mainWrapper = document.getElementById('main-wrapper');
+    this.threadMessages = document.getElementById('thread-messages');
+
+    window.addEventListener('resize', this.resizeHandler.bind(this));
```
azasypkin: Resize handler is used in both views, but handler is simple and can be repeated separately for every view (for ConversationView it will scroll messages and composer content, for NewMessageView it will scroll only composer content).

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+
+    // binding so that we can remove this listener later
+    this.onVisibilityChange = this.onVisibilityChange.bind(this);
```
azasypkin: Same for ```visibilitychange``` handler - can be separated into two methods as they work a bit differently for different views.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    document.addEventListener('visibilitychange',
+                              this.onVisibilityChange);
+
+    this.toField.addEventListener(
```
azasypkin: Everything related to ```toField``` is needed only in NewMessageView.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+      'input', this.toFieldInput.bind(this), true
+    );
+
+    this.toField.addEventListener(
+      'focus', this.toFieldInput.bind(this), true
+    );
+
+    this.sendButton.addEventListener(
```
azasypkin: Looks like we don't need any handlers on sendButton anymore, since we ```activate``` Compose event that does the same (assimilate recipients). And in case something goes wrong and button is clicked when it should have been disabled, MessagingService should prevent sending.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+      'click', this.onSendClick.bind(this)
+    );
+
+    this.sendButton.addEventListener(
+      'contextmenu', this.onSendClick.bind(this)
+    );
+
+    this.container.addEventListener(
```
azasypkin: Container scroll event handler is specific to Conversation and will be left in ConversationView.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+      'scroll', this.manageScroll.bind(this)
+    );
+
+    this.editHeader.addEventListener(
```
azasypkin: ```editHeader``` is specific to Conversation and will be left in ConversationView.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+      'action', this.cancelEdit.bind(this)
+    );
+
+    this.header.addEventListener(
```
azasypkin: Header ```action``` handler is used in both views, but can be separately defined in every view.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+      'action', this.onHeaderAction.bind(this)
+    );
+
+    this.optionsButton.addEventListener(
+      'click', this.showOptions.bind(this)
+    );
+
+    this.callNumberButton.addEventListener('click', function() {
```
azasypkin: ```callNumberButton``` is specific to Conversation and will be left in ConversationView.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+      ActivityPicker.dial(Threads.active.participants[0]);
+    });
+
+    this.deleteButton.addEventListener(
```
azasypkin: ```deleteButton``` is specific to Conversation and will be left in ConversationView.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+      'click', this.delete.bind(this)
+    );
+
+    this.headerText.addEventListener(
+      'click', this.onHeaderActivation.bind(this)
+    );
+
+    this.newMessageNotice.addEventListener(
```
azasypkin: ```newMessageNotice``` is specific to Conversation and will be left in ConversationView.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+
+    this.callNumberButton.addEventListener('click', function() {
+      ActivityPicker.dial(Threads.active.participants[0]);
+    });
+
+    this.deleteButton.addEventListener(
+      'click', this.delete.bind(this)
+    );
+
+    this.headerText.addEventListener(
```
azasypkin: ```headerText``` is managed by both views, but the logic is quite different, so I think it can be just separated.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+      'click', this.onHeaderActivation.bind(this)
+    );
+
+    this.newMessageNotice.addEventListener(
+      'click', this.onNewMessageNoticeClick.bind(this)
+    );
+
+    // These events will be handled in handleEvent function
+    this.container.addEventListener('click', this);
```
azasypkin: Both container ```click``` and ```contextmenu``` events are necessary for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    this.container.addEventListener('contextmenu', this);
+    this.editForm.addEventListener('submit', this);
+    this.composeForm.addEventListener('submit', this);
```
azasypkin: Compose form ```submit``` handler only prevents default submit action, looks like we can move that to Compose.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+
+    // For picking a contact from Contacts. It's mouse down for
+    // avoiding weird effect of keyboard, as in 'send' button.
+    this.contactPickButton.addEventListener(
```
azasypkin: ```contactPickButton``` specific to NewMessage and will be moved to NewMessageView.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+
+    // Avoid click event propagate to recipient view, otherwise Recipients.View
+    // constructor will attach click event on the messages-to-field element.
+    this.contactPickButton.addEventListener(
+      'click', function onClick(event) {
+        event.stopPropagation();
+      }
+    );
+
+    navigator.mozContacts.addEventListener(
```
azasypkin: We don't update recipients in recipients input on ```contactchange``` so we can leave this in ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+      'contactchange',
+      this.updateHeaderData.bind(this)
+    );
+
+    this.recipientSuggestions.addEventListener(
```
azasypkin: ```recipientSuggestions``` specific to NewMessage and will be moved to NewMessageView.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+      'click',
+      this.onRecipientSuggestionClick.bind(this)
+    );
+
+    MessageManager.on('message-sending', this.onMessageSending.bind(this));
```
azasypkin: We need ```'message-sending'``` in both views, can be easily separated.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    MessageManager.on('message-sent', this.onMessageSent.bind(this));
```
azasypkin: ```'message-sent', 'message-received', 'message-delivered', 'message-read'``` are specific to Conversation and will be left in ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    MessageManager.on('message-received', this.onMessageReceived.bind(this));
+    MessageManager.on(
```
azasypkin: Not sure about 'message-failed-to-send' event in NewMessageView, if it happens before ```message-sending```, what should we do (don't clear message input, subject, recipients?)?

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    this.tmpl = templateIds.reduce(function(tmpls, name) {
+      tmpls[Utils.camelCase(name)] =
+        Template('messages-' + name + '-tmpl');
+      return tmpls;
+    }, {});
+
+    Compose.init('messages-compose-form');
+
+    // In case of input, we have to resize the input following UX Specs.
+    Compose.on('input', this.messageComposerInputHandler.bind(this));
```
azasypkin: Once we move app notifications to Compose, we'll need these handlers only to update ```draft.isEdited``` property, so we can declare these simple event handlers for both views until we decide how to improve drafts handling further.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    //  1. Focus will jump from the recipient input to the
+    //      message input
+    //  2. 1 or 2 characters may remain in the recipient
+    //      editable, which will be "assimilated"
+    //  3. If a user has made it past 1 & 2, any attempts to
+    //      select a contact from contact search results
+    //      will also jump focus to the message input field
+    //
+    // So we assimilate recipients if user starts to interact with Composer
+    Compose.on('interact', this.assimilateRecipients.bind(this));
```
azasypkin: ```Compose.on('interact', this.assimilateRecipients.bind(this));``` NewMessage specific.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+
+    this.shouldChangePanelNextEvent = false;
+
+    this.showErrorInFailedEvent = '';
+
+    // Bound methods to be detachables
+    this.onMessageTypeChange = this.onMessageTypeChange.bind(this);
+  },
+
+  onVisibilityChange: function conv_onVisibilityChange(e) {
```
azasypkin: As I mentioned above - this method can be split into two: Conversation and NewMessage specific.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+  backOrClose: function conv_backOrClose() {
+    var inActivity = ActivityClient.hasPendingRequest();
+    var isComposer = Navigation.isCurrentPanel('composer');
+    var isThread = Navigation.isCurrentPanel('thread');
+    var action = inActivity && (isComposer || isThread) ? 'close' : 'back';
+    this[action]();
+  },
+
+  // Initialize Recipients list and Recipients.View (DOM)
+  initRecipients: function conv_initRecipients() {
```
azasypkin: ```initRecipients``` is NewMessageView specific.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+        navigator.mozSettings.addObserver(this.sentAudioKey, (function(e) {
+          this.sentAudioEnabled = e.settingValue;
+        }).bind(this));
+      } catch (e) {
+        this.sentAudioEnabled = false;
+      }
+    }
+  },
+
+  getIdIterator: function conv_getIdIterator() {
```
azasypkin: ```getIdIterator``` is ConversationView specific.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    return Threads.active.messages.keys();
+  },
+
+  setHeaderAction: function conv_setHeaderAction(icon) {
+    this.header.setAttribute('action', icon);
+  },
+
+  messageComposerInputHandler: function conv_messageInputHandler(event) {
```
azasypkin: ```messageComposerInputHandler``` will be used to update ```draft.isEdited``` only, so can be repeated for every view.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+      }
+
+      this._resizeNoticeTimeout = setTimeout(function hideResizeNotice() {
+        this.resizeNotice.classList.add('hide');
+        this._resizeNoticeTimeout = null;
+      }.bind(this), this.IMAGE_RESIZE_DURATION);
+    }
+  },
+
+  showMaxLengthNotice: function conv_showMaxLengthNotice(opts) {
```
azasypkin: Compose-related app notifications can be moved to Compose.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    this.timeouts.subjectLengthNotice &&
+      clearTimeout(this.timeouts.subjectLengthNotice);
+  },
+
+  /*
+   * This function will be called before we slide to the thread or composer
+   * panel. This way it can change things in the panel before the panel is
+   * visible.
+   */
+  beforeEnter: function conv_beforeEnter(args) {
```
azasypkin: ```beforeEnter``` method can be easily separated.

We won't need ```clearConvertNoticeBanners``` for separated views.

```this.setHeaderAction(ActivityClient.hasPendingRequest() ? 'close' : 'back');``` will be needed for both views, but doesn't look like a big deal to repeat it for every view.

What about moving ```multiSimActionButton``` to Compose?

File: apps/sms/views/new_message/js/new_message.js [hunk 40]
```diff
@@ -0,0 +1,3031 @@
...
+    if (Threads.active.participants.length === 1 &&
+        (!Settings.supportEmailRecipient ||
+         !Utils.isEmailAddress(Threads.active.participants[0]))) {
+      this.callNumberButton.classList.remove('hide');
+    }
+
+    return this.updateHeaderData();
+  },
+
+  afterEnter: function conv_afterEnter(args) {
```
azasypkin: ```afterEnter``` method can be easily separated.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    }
+
+    if (args.focusComposer) {
+      Compose.focus();
+    }
+
+    return Utils.closeNotificationsForThread(threadId);
+  },
+
+  beforeLeave: function conv_beforeLeave(args) {
```
azasypkin: ```disableConvertNoticeBanners``` is not needed for separated views.

```cancelEdit``` and attachment disposal code are specific to ConversationView only.

```cleanFields``` are for both.

So looks like can be easily separated between two views (if we really need these cleanups for the case when views are separated :) since entire view will be unloaded).

File: apps/sms/views/new_message/js/new_message.js [hunk 42]
```diff
@@ -0,0 +1,3031 @@
...
+      });
+    }
+
+    // TODO move most of back() here: Bug 1010223
+    if (!this.isConversationPanel(Threads.currentId, nextPanel)) {
+      this.cleanFields();
+    }
+  },
+
+  afterLeave: function conv_afterLeave(args) {
```
azasypkin: ```afterLeave``` looks like unnecessary method for separated views.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+      this.toggleRecipientSuggestions();
+    }
+
+    if (!Navigation.isCurrentPanel('thread')) {
+      this.threadMessages.classList.remove('has-carrier');
+      this.callNumberButton.classList.add('hide');
+    }
+  },
+
+  handleActivity: function conv_handleActivity(params) {
```
azasypkin: ```handleActivity``` is for NewMessageView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+        // Compose event that sets Recipients.View.isFocusable to false.
+        Recipients.View.isFocusable = true;
+        this.recipients.focus();
+      }
+    });
+  },
+
+  // recalling draft for composer only
+  // Bug 1164431 might use it for thread drafts too
+  handleDraft: function conv_handleDraft(draftId) {
```
azasypkin: ```handleDraft``` is for NewMessageView only. We can improve/unify this code further after split.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    this.initRecipients();
+    this.updateComposerHeader();
+    this.container.textContent = '';
+    this.threadMessages.classList.add('new');
+
+    // not strictly necessary but being consistent
+    return Promise.resolve();
+  },
+
+  assimilateRecipients: function conv_assimilateRecipients() {
```
azasypkin: ```assimilateRecipients``` is for NewMessageView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+
+          break;
+        }
+      }
+    } while ((node = node.previousSibling));
+  },
+
+  // Function for handling when a new message (sent/received)
+  // is detected
+  onMessage: function onMessage(message) {
```
azasypkin: ```onMessage``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    TimeHeaders.updateAll('header[data-time-update]');
+  },
+
+  /**
+   * Checks if specified conversation id is currently active. It can be true for
+   * either conversation, participants or report panels.
+   * @param {number} conversationId Id of the conversation.
+   * @returns {boolean}
+   */
+  isCurrentConversation: function conv_isCurrentConversation(conversationId) {
```
azasypkin: ```isCurrentConversation``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+  },
+
+  /**
+   * Checks if specified panel corresponds to the specified conversation id. It
+   * can be true for either conversation, participants or report panels.
+   * @param {number} conversationId Id of the conversation.
+   * @param {Object} panel Panel description object to compare against.
+   * @returns {boolean}
+   */
+  isConversationPanel:
```
azasypkin: ```isConversationPanel``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    if (!panel) {
+      return false;
+    }
+
+    return panel.panel === 'thread' && panel.args.id === conversationId ||
+      panel.panel === 'report-view' && panel.args.threadId === conversationId ||
+      panel.panel === 'group-view' && panel.args.id === conversationId;
+  },
+
+  onMessageReceived: function conv_onMessageReceived(e) {
```
azasypkin: ```onMessageReceived``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+      this.forceScrollViewToBottom();
+    } else {
+      if (this.shouldChangePanelNextEvent) {
+        Navigation.toPanel('thread', { id: message.threadId });
+        this.shouldChangePanelNextEvent = false;
+      }
+    }
+  },
+
+  onNewMessageNoticeClick: function conv_onNewMessageNoticeClick(event) {
```
azasypkin: ```onNewMessageNoticeClick``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    MessageManager.markMessagesRead([message.id]);
+
+    this.onMessage(message);
+    this.scrollViewToBottom();
+    if (this.isScrolledManually) {
+      this.showNewMessageNotice(message);
+    }
+  },
+
+  onMessageSending: function conv_onMessageReceived(e) {
```
azasypkin: ```onMessageSending``` for both views, but can be easily separated.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+  onNewMessageNoticeClick: function conv_onNewMessageNoticeClick(event) {
+    event.preventDefault();
+    this.hideNewMessageNotice();
+    this.forceScrollViewToBottom();
+  },
+
+  /**
+   * Fires once user clicks on any recipient in the suggestions list.
+   */
+  onRecipientSuggestionClick: function(event) {
```
azasypkin: ```onRecipientSuggestionClick``` is for NewMessageView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+
+    // Since the "dataset" DOMStringMap property is essentially
+    // just an object of properties that exactly match the properties
+    // used for recipients, push the whole dataset object into
+    // the current recipients list as a new entry.
+    this.recipients.add(event.target.dataset).focus();
+  },
+
+  // Message composer type changed:
+  onMessageTypeChange: function conv_onMessageTypeChange() {
```
azasypkin: ```onMessageTypeChange``` likely will be moved to Compose

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+
+  enableConvertNoticeBanners: function conv_enableConvertNoticeBanner() {
+    Compose.on('type', this.onMessageTypeChange);
+  },
+
+  disableConvertNoticeBanners: function conv_disableConvertNoticeBanner() {
+    Compose.off('type', this.onMessageTypeChange);
+  },
+
+  onSubjectChange: function conv_onSubjectChange() {
```
azasypkin: ```onSubjectChange``` will be used for updating ```draft.isEdited``` only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    if (Compose.isSubjectVisible && Compose.isSubjectMaxLength()) {
+      this.showSubjectMaxLengthNotice();
+    } else {
+      this.hideSubjectMaxLengthNotice();
+    }
+  },
+
+  // Triggered when the onscreen keyboard appears/disappears.
+  resizeHandler: function conv_resizeHandler() {
```
azasypkin: ```resizeHandler``` can be easily separated.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    // Scroll to bottom
+    this.scrollViewToBottom();
+    // Make sure the caret in the "Compose" area is visible
+    Compose.scrollMessageContent();
+  },
+
+  // Create a recipient from contacts activity.
+  requestContact: function conv_requestContact() {
```
azasypkin: ```requestContact``` is for NewMessageView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+
+    activity.onerror = (function(e) {
+      Recipients.View.isFocusable = true;
+
+      console.log('WebActivities unavailable? : ' + e);
+    }).bind(this);
+  },
+
+  // Method for updating the header when needed
+  updateComposerHeader: function conv_updateComposerHeader() {
```
azasypkin: ```updateComposerHeader``` is for NewMessageView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+      this.setHeaderContent('newMessage');
+    }
+  },
+
+  // scroll position is considered as "manual" if the view is not completely
+  // scrolled to the bottom
+  isScrolledManually: false,
+
+  // We define an edge for showing the following chunk of elements
+  manageScroll: function conv_manageScroll(oEvent) {
```
azasypkin: ```manageScroll``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+      // We update the scroll to the previous position
+      // taking into account the previous offset to top
+      // and the current height due to we have added a new
+      // chunk of visible messages
+      this.container.scrollTop =
+        (this.container.scrollHeight - scrollHeight) + scrollTop;
+    }
+  },
+
+  scrollViewToBottom: function conv_scrollViewToBottom() {
```
azasypkin: ```scrollViewToBottom``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    if (!this.isScrolledManually &&
+        this.container.lastElementChild &&
+        Navigation.isCurrentPanel('thread')) {
+      this.container.lastElementChild.scrollIntoView(false);
+    }
+  },
+
+  forceScrollViewToBottom: function conv_forceScrollViewToBottom() {
```
azasypkin: ```forceScrollViewToBottom``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    this.isScrolledManually = false;
+    this.scrollViewToBottom();
+  },
+
+  showNewMessageNotice: function conv_showNewMessageNotice(message) {
```
azasypkin: ```showNewMessageNotice``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+      var notice = this.newMessageNotice;
+      var newMessageContactNode = notice.querySelector('.contact');
+      newMessageContactNode.textContent = sender;
+
+      this.isNewMessageNoticeShown = true;
+      notice.classList.remove('hide');
+    });
+  },
+
+  hideNewMessageNotice: function conv_hideNewMessageNotice() {
```
azasypkin: ```hideNewMessageNotice``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+          window.clearTimeout(setTimer);
+          window.removeEventListener('resize', keyboardHidden);
+          resolve();
+        });
+      });
+    }
+    return Promise.resolve();
+  },
+
+  _showMessageSaveOrDiscardPrompt: function() {
```
azasypkin: ```_showMessageSaveOrDiscardPrompt``` is going to be removed, so I don't really care if generalize it somehow or just duplicate in both views for now.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+
+  // Toast duration when you write a long text and need more than one SMS
+  // to send it
+  ANOTHER_SMS_TOAST_DURATION: 3000,
+
+  // when sending an sms to several recipients in activity, we'll exit the
+  // activity after this delay after moving to the thread list.
+  LEAVE_ACTIVITY_DELAY: 3000,
+
+  draft: null,
```
azasypkin: I'd propose to duplicate drafts handling code for now. It has some subtle differences between two views anyway. I think once we improve drafts handling we'll just have smth like draft-get, draft-save and draft-discard.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+
+    this.editHeader.addEventListener(
+      'action', this.cancelEdit.bind(this)
+    );
+
+    this.header.addEventListener(
+      'action', this.onHeaderAction.bind(this)
+    );
+
+    this.optionsButton.addEventListener(
```
azasypkin: ```showOptions``` can be separated with only duplication related to show/hide subject menu item... imho it doesn't look like a big problem to duplicate this.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+        this.threadMessages.classList.toggle(
+          'multiline-recipients-mode',
+           mode === 'multiline-mode'
+        );
+      }.bind(this));
+    }
+    this.toggleRecipientSuggestions();
+  },
+
+  initSentAudio: function conv_initSentAudio() {
```
azasypkin: What about this, should we move this to a separate ```AudioNotification``` module/file?

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    });
+  },
+
+  hideNewMessageNotice: function conv_hideNewMessageNotice() {
+    this.isNewMessageNoticeShown = false;
+    //Hide the new message's banner
+    this.newMessageNotice.classList.add('hide');
+  },
+
+  close: function conv_close() {
```
azasypkin: Don't have good plan for ```close, back, _onNavigatingBack, isKeyboardDisplayed, _ensureKeyboardIsHidden``` yet.

``` isKeyboardDisplayed, _ensureKeyboardIsHidden``` can probably be somewhere in shared code, utils maybe (although don't really want to make utils.js catch-all module)?

```close, back, _onNavigatingBack``` - once we remove draft discard/save prompt, the code will become easier and won't look that scary/big to be duplicated.

File: apps/sms/views/new_message/js/new_message.js [hunk 68]
```diff
@@ -0,0 +1,3031 @@
...
+      return Promise.reject(e);
+    });
+  },
+
+  /**
+   * Navigates user to Composer or Thread panel with custom parameters.
+   * @param {*} parameters Optional navigation parameters.
+   * @returns {Promise} Promise that is resolved once navigation is completed.
+   */
+  navigateToComposer: function(parameters) {
```
azasypkin: ```navigateToComposer``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+        }
+      ]
+    };
+
+    new OptionMenu(options).show();
+
+    return deferred.promise;
+  },
+
+  onSegmentInfoChange: function conv_onSegmentInfoChange() {
```
azasypkin: ```onSegementInfoChange``` handle will disappear if we move Compose-related notifications to Compose module.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+        { number: currentSegment }
+      );
+      this.smsCounterNotice.classList.remove('hide');
+      window.setTimeout(function() {
+        this.smsCounterNotice.classList.add('hide');
+      }.bind(this), this.ANOTHER_SMS_TOAST_DURATION);
+    }
+  },
+
+  checkMessageSize: function conv_checkMessageSize() {
```
azasypkin: same here

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+        return true;
+      }
+    }
+
+    this.hideMaxLengthNotice();
+    return true;
+  },
+
+  // Adds a new grouping header if necessary (today, tomorrow, ...)
+  getMessageContainer:
```
azasypkin: ```getMessageContainer``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    } else {
+      TimeHeaders.update(header);
+    }
+
+    this._insertTimestampedNodeToContainer(messageDateGroup, this.container);
+
+    return messageContainer;
+  },
+
+  updateCarrier: function conv_updateCarrier(thread, contacts) {
```
azasypkin: ```updateCarrier``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+        threadMessages.classList.remove('has-carrier');
+      }
+    } else {
+      // Hide carrier tag in group message or unknown contact cases.
+      threadMessages.classList.remove('has-carrier');
+    }
+  },
+
+  // Method for updating the header with the info retrieved from Contacts API
+  updateHeaderData: function conv_updateHeaderData() {
```
azasypkin: ```updateHeaderData``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+   * Updates header content since it's used for different panels and should be
+   * carefully handled for every case. In Thread panel header contains HTML
+   * markup to support bidirectional content, but other panels still use it with
+   * mozL10n.setAttributes as it would contain only localizable text. We should
+   * get rid of this method once bug 961572 and bug 1011085 are landed.
+   * @param {string|{ html: string }|{id: string, args: Object }} contentL10n
+   * Should be either safe HTML string or l10n properties.
+   * @public
+   */
+  setHeaderContent: function conv_setHeaderContent(contentL10n) {
```
azasypkin: Looking at the method comment, looks like we'll have to get rid of it once we move to separated panels :)

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    }
+
+    if (contentL10n.html) {
+      this.headerText.removeAttribute('data-l10n-id');
+      this.headerText.innerHTML = contentL10n.html;
+      return;
+    }
+  },
+
+  initializeRendering: function conv_initializeRendering() {
```
azasypkin: ```initializeRendering``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    // Clean list of messages
+    this.container.innerHTML = '';
+    // Initialize infinite scroll params
+    this.messageIndex = 0;
+    // reset stopRendering boolean
+    this._stopRenderingNextStep = false;
+  },
+
+  // Method for stopping the rendering when clicking back
+  stopRendering: function conv_stopRendering() {
```
azasypkin: ```stopRendering``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    this._stopRenderingNextStep = true;
+  },
+
+  // Method for rendering the first chunk at the beginning
+  showFirstChunk: function conv_showFirstChunk() {
```
azasypkin: ```showFirstChunk``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    this.showChunkOfMessages(this.CHUNK_SIZE);
+    // Boot update of headers
+    TimeHeaders.updateAll('header[data-time-update]');
+    // Go to Bottom
+    this.scrollViewToBottom();
+
+    this.emit('visually-loaded');
+  },
+
+  createMmsContent: function conv_createMmsContent(dataArray) {
```
azasypkin: ```createMmsContent``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+
+        textElement.innerHTML = text;
+        container.appendChild(textElement);
+      }
+    });
+    return container;
+  },
+
+  // Method for rendering the list of messages using infinite scroll
+  renderMessages: function conv_renderMessages(threadId, callback) {
```
azasypkin: ```renderMessages``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+
+    MessageManager.getMessages(renderingOptions);
+
+    // force the next scroll to bottom
+    this.isScrolledManually = false;
+  },
+
+  // generates the html for not-downloaded messages - pushes class names into
+  // the classNames array also passed in, returns an HTML string
+  _createNotDownloadedHTML:
```
azasypkin: ```_createNotDownloadedHTML``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+      messageL10nDate: message.expiryDate.toString(),
+      messageL10nDateFormat: 'expiry-date-format',
+      downloadL10nId: downloadL10nId
+    });
+  },
+
+  // Check deliveryStatus for both single and multiple recipient case.
+  // In multiple recipient case, we return true only when all the recipients
+  // deliveryStatus set to success.
+  shouldShowDeliveryStatus: function conv_shouldShowDeliveryStatus(message) {
```
azasypkin: ```shouldShowDeliveryStatus``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+      });
+    } else {
+      return message.deliveryStatus === 'success';
+    }
+  },
+
+  // Check readStatus for both single and multiple recipient case.
+  // In multiple recipient case, we return true only when all the recipients
+  // deliveryStatus set to success.
+  shouldShowReadStatus: function conv_shouldShowReadStatus(message) {
```
azasypkin: ```shouldShowReadStatus``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+      !message.deliveryInfo) {
+      return false;
+    }
+
+    return message.deliveryInfo.every(function(info) {
+      return info.readStatus === 'success';
+    });
+  },
+
+  buildMessageDOM: function conv_buildMessageDOM(message, hidden) {
```
azasypkin: ```buildMessageDOM``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+      return this.mmsContentParser(message).then((mmsContent) => {
+        pElement.appendChild(mmsContent);
+        return messageDOM;
+      });
+    }
+
+    return Promise.resolve(messageDOM);
+  },
+
+  mmsContentParser: function conv_mmsContentParser(message) {
```
azasypkin: ```mmsContentParser``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    return new Promise((resolver) => {
+      SMIL.parse(message, (slideArray) => {
+        resolver(this.createMmsContent(slideArray));
+      });
+    });
+  },
+
+  getMessageStatusMarkup: function(status) {
```
azasypkin: ```getMessageStatusMarkup``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    return ['read', 'delivered', 'sending', 'error'].indexOf(status) >= 0 ?
+      this.tmpl.messageStatus.prepare({
+        statusL10nId: 'message-delivery-status-' + status
+      }) : '';
+  },
+
+  appendMessage: function conv_appendMessage(message, hidden) {
```
azasypkin: ```appendMessage``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+  },
+
+  /**
+   * Inserts DOM node to the container respecting 'timestamp' data attribute of
+   * the node to insert and sibling nodes in ascending order.
+   * @param {Node} nodeToInsert DOM node to insert.
+   * @param {Node} container Container DOM node to insert to.
+   * @private
+   */
+  _insertTimestampedNodeToContainer: function(nodeToInsert, container) {
```
azasypkin: ```_insertTimestampedNodeToContainer``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+
+    while (currentNode && nodeTimestamp > +currentNode.dataset.timestamp) {
+      currentNode = currentNode.nextElementSibling;
+    }
+
+    // With this function, "inserting before 'null'" means "appending"
+    container.insertBefore(nodeToInsert, currentNode || null);
+  },
+
+  showChunkOfMessages: function conv_showChunkOfMessages(number) {
```
azasypkin: ```showChunkOfMessages``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+
+    Array.slice(elements, -number).forEach((element) => {
+      element.classList.remove('hidden');
+      if (element.tagName === 'HEADER') {
+        TimeHeaders.update(element);
+      }
+    });
+  },
+
+  showOptions: function conv_showOptions() {
```
azasypkin: I think we can just split this method with minimal duplication related to subject handling.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    // Last item is the Cancel button
+    params.items.push({
+      l10nId: 'cancel',
+      incomplete: true
+    });
+
+    new OptionMenu(params).show();
+  },
+
+  startEdit: function conv_edit() {
```
azasypkin: ```startEdit``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+          isInEditMode: this.isInEditMode.bind(this)
+        });
+        editModeSetup.call(this);
+      });
+    } else {
+      editModeSetup.call(this);
+    }
+  },
+
+  isInEditMode: function conv_isInEditMode() {
```
azasypkin: ```isInEditMode``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    return this.inEditMode;
+  },
+
+  deleteUIMessages: function conv_deleteUIMessages(list, callback) {
```
azasypkin: ```deleteUIMessages``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+        InboxView.updateThread(request.result, { deleted: true });
+      };
+      request.onerror = function() {
+        console.error('Error when updating the list of threads');
+        callback();
+      };
+    }
+  },
+
+  delete: function conv_delete() {
```
azasypkin: ```delete``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+      },
+      null,
+      {
+        text: 'delete',
+        className: 'danger'
+      }
+    ).then(performDeletion.bind(this));
+  },
+
+  cancelEdit: function conv_cancelEdit() {
```
azasypkin: ```cancelEdit``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    if (this.inEditMode) {
+      this.inEditMode = false;
+      this.mainWrapper.classList.remove('edit');
+    }
+  },
+
+  updateSelectionStatus: function conv_updateSelectionStatus() {
```
azasypkin: ```updateSelectionStatus``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+      this.deleteButton.disabled = false;
+      navigator.mozL10n.setAttributes(this.editMode, 'selected-messages',
+        {n: selected});
+    } else {
+      this.deleteButton.disabled = true;
+      navigator.mozL10n.setAttributes(this.editMode, 'deleteMessages-title');
+    }
+  },
+
+  handleMessageClick: function conv_handleMessageClick(evt) {
```
azasypkin: ```handleMessageClick``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+    // retrieve message again in this edge case.
+    if (elems.message.classList.contains('invalid-empty-content')) {
+      return;
+    }
+
+    // Click events originating from a "message-status" aside of an error
+    // message should trigger a prompt for retransmission.
+    if (elems.message.classList.contains('error') && elems.messageStatus) {
+      Utils.confirm('resend-confirmation').then(() => {
+        this.re
```
azasypkin: ```getMessageBubble``` is for ConversationView only.

azasypkin: Considering that we don't need container's ```click``` and ```contextmenu``` events for NewMessageView and compose form ```submit``` handler can be moved to Compose - we need ```handleEvent``` only in ConversationView.

azasypkin: ```cleanFields``` is easy to split.

azasypkin: If I'm not mistaken we don't need this ```onSendClick``` anymore.

azasypkin: We'll likely have two separate and bit simplified ```sendMessage``` methods for every view, haven't tried to split it yet.

azasypkin: ```onMessageSent``` is for ConversationView only.

azasypkin: It's small, so can be duplicated I think.

azasypkin: Not sure about ```onMessageFailed```, asked about it above.

azasypkin: ```onDeliverySuccess``` is for ConversationView only.

azasypkin: ```onReadSuccess``` is for ConversationView only.

azasypkin: ```removeMessageDOM``` is for ConversationView only.

azasypkin: ```setMessageStatus``` is for ConversationView only.

azasypkin: ```retrieveMMS``` is for ConversationView only.

azasypkin: ```resendMessage``` is for ConversationView/ReportView only.

azasypkin: ```toFieldKeypress``` is for NewMessageView only.

azasypkin: ```toFieldInput``` is for NewMessageView only.

azasypkin: ```exactContact``` is for NewMessageView only.

azasypkin: ```searchContact``` is for NewMessageView only.

azasypkin: ```validateContact``` is for NewMessageView only.

azasypkin: ```listContacts``` is for NewMessageView only.

azasypkin: ```onHeaderActivation``` is for ConversationView only.

azasypkin: ```promptContacts``` is for ConversationView/ParticipantsView only.

azasypkin: ```prompt``` is for ConversationView/ParticipantsView only.

azasypkin: ```discardDraft``` is for both views, but I believe it can be separated into two methods without much duplication.

azasypkin: same here

azasypkin: ```toggleRecipientSuggestions``` is for NewMessageView only.

azasypkin: ```enableBubbleSelection``` is for ConversationView only.

apps/sms/views/new_message/js/new_message.js
julienw: yep I agree; not sure what is the right place... anyway it seems weird that we need to do this work (as explained in the comment)

apps/sms/views/new_message/js/new_message.js
julienw: I'm not so sure the notification should move to Compose. I'd rather see a generalized notification system. This could be a very simple lib that would display a message in a notification bubble with some id (the id could be "calculated" from the notification name -- and with the notification name we'd have the duration using an array).

Just from the top of my head here; not sure this is the best solution. But moving to Compose is not a solution from my pov.

apps/sms/views/new_message/js/new_message.js
julienw: something I had in my mind for a long time: wondering if in ConversationView "recipients" could hold the recipients for this conversation.

julienw: not sure how useful it is with the current code though

apps/sms/views/new_message/js/new_message.js
julienw: agreed

apps/sms/views/new_message/js/new_message.js
julienw: we'll know soon enough if we still need it :p

apps/sms/views/new_message/js/new_message.js
julienw: yep

apps/sms/views/new_message/js/new_message.js
julienw: actually, to prevent submit, we just need a `<button type="button">`. But more importantly, for accessibility, maybe we'd like the "submit" event to actually run the action (instead of listening for the click event on a button)

apps/sms/views/new_message/js/new_message.js
julienw: ideally I'd see this move to the recipient panel, actually; but not in this patch

apps/sms/views/new_message/js/new_message.js
julienw: same here, I'd see this in the recipient panel ideally, in long term

apps/sms/views/new_message/js/new_message.js
julienw: yep; there is a bug about this actually :) https://bugzilla.mozilla.org/show_bug.cgi?id=1136147

apps/sms/views/new_message/js/new_message.js
julienw: > We won't need clearConvertNoticeBanners for separated views.

But we may need it at first since we'll use the same JS both in split and non-split modes.

julienw: > What about moving multiSimActionButton to Compose and expose send event?

not sure; it's not really a big piece of code anyway.

apps/sms/views/new_message/js/new_message.js
julienw: yep (and I'm not even sure we'll be able to call it anyway)

apps/sms/views/new_message/js/new_message.js
julienw: yeah, the structure of the function itself shows it _should_ be separated ;)

apps/sms/views/new_message/js/new_message.js
julienw: should we start by removing this first ? :)

apps/sms/views/new_message/js/new_message.js
julienw: yep, why not

apps/sms/views/new_message/js/new_message.js
julienw: agreed;  showOptions is only configuration for Optionmenu

apps/sms/views/new_message/js/new_message.js
julienw: IMO this could just move to notify.js

apps/sms/views/new_message/js/new_message.js
julienw: > isKeyboardDisplayed, _ensureKeyboardIsHidden can probably be somewhere in shared code, utils maybe (although don't really want to make utils.js catch-all module)?

Agreed these are more common utils.

> close, back, _onNavigatingBack - once we remove draft discard/save prompt, the code will become easier and won't look that scary/big to be duplicated.

Wondering if we could have a mixin for draft handling (just throwing the idea, I don't have much idea about how it could look like -- or even if it's needed given we want to remove the prompt).

maybe we should remove the prompt first, especially that Jenny was OK with it before she left.

apps/sms/views/new_message/js/new_message.js
julienw: yeah, main use of this method was to be called by other panels... and we don't want that :)

apps/sms/views/new_message/js/new_message.js
steveck-chung: It might make sense to move to MessageManager (or Messaging service in the future)

apps/sms/views/new_message/js/new_message.js
steveck-chung: Not quite understand this part: In the ConversationView we would only need an array of recipient and it will contain in thread object. But the "recipients" here means the recipients module, and we don't really need most of the functionalities in conversation view.

apps/sms/views/new_message/js/new_message.js
steveck-chung: Just wondering why we still keep this after introducing MultiSimActionButton

apps/sms/views/new_message/js/new_message.js
```diff
@@ -0,0 +1,3031 @@
...
+      'message-status',
+      'not-downloaded',
+      'recipient',
+      'date-group',
+      'header',
+      'group-header'
+    ];
+
+    // Fields with 'messages' label
+    [
```
steveck-chung: Clearly we will have different set of ids between 2 views, but I think you definitely know :)

apps/sms/views/new_message/js/new_message.js
steveck-chung: +1 that move to recipient in long term

apps/sms/views/new_message/js/new_message.js
steveck-chung: It's interesting that it might be good if we can update recipient when contact update, but not really important.

apps/sms/views/new_message/js/new_message.js
steveck-chung: No matter where it would be, moving it out from view +1

apps/sms/views/new_message/js/new_message.js
steveck-chung: > We won't need clearConvertNoticeBanners for separated views.

So how about navigating to another conversation thread?

apps/sms/views/new_message/js/new_message.js
steveck-chung: I thought about moving `isKeyboardDisplayed/ _ensureKeyboardIsHidden` to utils as well. And not sure about how to simplify this part either. Maybe removing the draft save prompt could be a good start.

apps/sms/views/new_message/js/new_message.js
azasypkin: Okay, let's say we move it to MessageManager  and then to MessagingService.

apps/sms/views/new_message/js/new_message.js
azasypkin: Well, the problem is that we have a bunch of supporting code that I don't really know how to generalize. For example:

1. When subject limit is reached - we show banner that will go away after timeout, but we should hide it as soon as length doesn't exceed limit anymore - so here we should keep reference to "notification" to be able to hide it;

2. When message is resized we show "persistent" notification that should be removed (in our current logic we don't hide it immediately like in example#1, but with with timeout :() only when resize is finished. Also when we said that notification should be hidden in two seconds, it may happen that resizing is happening again, so that we should stop "hiding" :)

So as you see it has quite a bit code that either should be duplicated in both views or moved to Compose. What do you think?

Btw, I have a draft of ```AppNotification``` (still don't like some pieces of it) that can cover all our cases I think: https://github.com/mozilla-b2g/gaia/pull/30794/

apps/sms/views/new_message/js/new_message.js
azasypkin: > Not quite understand this part: In the ConversationView we would only need an array of recipient and it will contain in thread object. But the "recipients" here means the recipients module, and we don't really need most of the functionalities in conversation view.

Yeah, I think the same... @julienw for what case/reason you wanted to have a copy of ```ConversationView.conversation.participants``` in  ```ConversationView.recipients```? I think that could be useful for non split views, when we save draft or send message so that we can avoid different ```if```, but not sure if it makes sense for split views?

apps/sms/views/new_message/js/new_message.js
azasypkin: Yeah, some of them are even duplicated right now :) Like ```new-message-notice``` that mentioned twice :)

apps/sms/views/new_message/js/new_message.js
azasypkin: > Just wondering why we still keep this after introducing MultiSimActionButton

The only case I can imagine: we wanted to assimilate recipients when user taps on _disabled_ send button (multi sim action callback won't be called in this case), but now we have ```activate``` event that we can use for this purpose.

apps/sms/views/new_message/js/new_message.js
azasypkin: > actually, to prevent submit, we just need a <button type="button">

Need to check if we have something specific about ```[type=submit]``` or ```[type=button]```` in BB, if not - then it's good.

 > But more importantly, for accessibility, maybe we'd like the "submit" event to actually run the action (instead of listening for the click event on a button)

We don't directly listen for ```click``` event on send button we listen for ```mulisim action callback``` instead...

apps/sms/views/new_message/js/new_message.js
azasypkin: Yep, agree, but in the separate bug.

apps/sms/views/new_message/js/new_message.js
azasypkin: Yeah, I think it'd be nice to have eventually - fled papercut for this https://bugzilla.mozilla.org/show_bug.cgi?id=1179719

apps/sms/views/new_message/js/new_message.js
azasypkin: Yep

apps/sms/views/new_message/js/new_message.js
azasypkin: ahhhh, nice!

apps/sms/views/new_message/js/new_message.js
azasypkin: Good, filed https://bugzilla.mozilla.org/show_bug.cgi?id=1179727

apps/sms/views/new_message/js/new_message.js
azasypkin: > But we may need it at first since we'll use the same JS both in split and non-split modes.

Yeah, I'm currently thinking how to handle this, what I meant is that we don't have think too much about this piece since it will be removed :)

> So how about navigating to another conversation thread?

You mean conversation#1---->conversation#2 case when we separate views? If yes that's a good question, we talked with @julienw about this some time ago - whether we want to have real navigation from say ```..../conversation/?id=1``` to ```..../conversation/?id=2```  or just update current view with new conversation, at that time I was thinking about the latter, but now I think real navigation is more preferable :)

apps/sms/views/new_message/js/new_message.js
azasypkin: Yeah :)

apps/sms/views/new_message/js/new_message.js
azasypkin: Okay moving ```isKeyboardDisplayed, _ensureKeyboardIsHidden``` to utils then.

> maybe we should remove the prompt first, especially that Jenny was OK with it before she left.

> Maybe removing the draft save prompt could be a good start.

Yes and yes :) Added this bug as dependency!

apps/sms/views/new_message/js/new_message.js
azasypkin: Yes!

apps/sms/views/new_message/js/new_message.js
julienw: I think it was part of my solution for making Compose instanciable; but because we went to another direction for this, it's maybe not useful now

apps/sms/views/new_message/js/new_message.js
julienw: I don't think we get the "click" event when the button is disabled.

Not really sure if this is still needed, maybe this is here for historical reason only.

apps/sms/views/new_message/js/new_message.js
julienw: yep right

apps/sms/views/new_message/js/new_message.js
julienw: maybe this will get obsolete if you land this patch before that patch is ready

apps/sms/views/new_message/js/new_message.js
julienw: yep, I think real navigation is better, more consistent

apps/sms/views/new_message/js/new_message.js
azasypkin: > I don't think we get the "click" event when the button is disabled.

mm, you're right, then it doesn't have any value now :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5565c09f8275a3a5810a2b29276100e9a4000ae4

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=dd8aa86179e80d327999f19a56d507f301ffbb03

PR #30753, opened 2015-06-30
Bug 1169538 - [Stingray] Sort cards before playing filter animation

dwi2:
https://bugzilla.mozilla.org/show_bug.cgi?id=1169538

### DIFF
--- a/tv_apps/smart-home/js/filter_manager.js
+++ b/tv_apps/smart-home/js/filter_manager.js
@@ -138,8 +138,14 @@
       }
       this._isFilterChanging = true;
       this._cardListElem.setAttribute('smart-bubbles-direction', 'down');
-      this._smartBubblesElem.play(
-        document.querySelectorAll('#card-list > .card > .app-button'));
+
+      // Notice that methods like document.querySelectorAll or
+      // document.getElementsByClassName returns elements based on the
+      // sequence of time that element added to DOM structure. But we should
+      // play animation in the left-to-right order. Card scrollable keeps the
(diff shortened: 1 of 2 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e16268c345140f59fa4cc908efe8a95c5a6b199a

tv_apps/smart-home/js/filter_manager.js
```diff
@@ -138,8 +138,26 @@
       }
       this._isFilterChanging = true;
       this._cardListElem.setAttribute('smart-bubbles-direction', 'down');
-      this._smartBubblesElem.play(
-        document.querySelectorAll('#card-list > .card > .app-button'));
+
+      // Notice that document.querySelectorAll returns elements based on the
```
rexboy7: Oh, thanks for addressing this problem.
I would consider it as a mismatch between how `scrollable.js` works and DOM structure itself.
And since the logic that really cause the order mismatch with its real DOM tree order is inside `scrollable.js`, I think it would be better to move the code below into `scrollable.js` as a function called `sortNodes`, and then call it here.
How do you think?

dwi2: Yeh, I think it is a good idea. I'll change that and then set review to you.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5dbd660148a7f0b2c7115e2847ff2a5280ff53eb

PR #30766, opened 2015-07-01
System app ACL integration

mikehenrty:


apps/system/index.html
```diff
@@ -14,6 +14,7 @@
     <script src="shared/js/l10n_date.js"></script>
     <script src="shared/js/idletimer.js"></script>
     <script src="shared/js/dump.js"></script>
+    <script defer src="js/acl_manager.js"></script>
```
KevinGrandon: This would need to change when going to master as we want to avoid boot slowdown. Need to lazyload it using the new loader stuff.

apps/system/js/acl_manager.js
```diff
@@ -0,0 +1,283 @@
...
+  function debug() {
+    DEBUG && dump('[ACL] ' +
+      Array.prototype.slice.call(arguments).join(', ') + '\n');
+  }
+
+  function getRandomString() {
+    return Math.random().toString(36).replace(/[^a-z]+/g, '');
+  }
+
+  var ACLManager = function() {
```
KevinGrandon: You're going to hate me for this, but I'm going to ask you to rename this. IMO `ACL` typically implies access control list in software. Using this abbreviation is confusing to me. I suggest: `app_compatibility.js`, and `var AppCompatibility = function() {`

apps/system/js/notifications.js
```diff
@@ -369,6 +371,14 @@ var NotificationScreen = {
     return date;
   },
 
+  /**
+   * Check if notification is from system and marked silent in icon url.
+   */
+  isSilentNotification: function ns_isSilentNotication(detail) {
+    return detail.manifestURL === this.SYSTEM_MANIFEST_URL &&
+      detail.icon && detail.icon.indexOf('silent=1') !== -1;
```
KevinGrandon: silent=1 for an icon url really? Is there no way to make this some kind of detail property?

apps/system/js/acl_manager.js
```diff
@@ -0,0 +1,283 @@
...
+  var HOST = '127.0.0.1';
+  var PORT = 33334;
+  var RETRY_TIMEOUT = 1000;
+  var POLL_TIMEOUT = 60000 * 5; // 5 minutes
+  var SETTINGS_KEY = 'acl.enabled';
+  var NOTIFICATION_PREFIX = 'ACL_NOTIFICATION';
+
+  function debug() {
+    DEBUG && dump('[ACL] ' +
+      Array.prototype.slice.call(arguments).join(', ') + '\n');
```
KevinGrandon: This can just be `Array.slice(arguments)`

apps/system/js/acl_manager.js
```diff
@@ -0,0 +1,283 @@
...
+    }
+    this.socket.send(this.serializeMessage(msg));
+  };
+
+  ACLManager.prototype.getAppFromMessage = function(msg) {
+    var app = null;
+    if (msg.manifestURL) {
+      app = applications.getByManifestURL(msg.manifestURL);
+    }
+    if (!app && msg.origin) {
```
KevinGrandon: nit: maybe make this an `else if`?

apps/system/js/acl_manager.js
```diff
@@ -0,0 +1,283 @@
...
+  ACLManager.prototype.removeNotification = function(id) {
+    if (!id || !this.notifications[id]) {
+      debug('Unable to find notification for removal', id);
+    } else {
+      this.notifications[id].close();
+      delete this.notifications[id];
+    }
+  };
+
+  exports.aclManager = new ACLManager();
```
KevinGrandon: nit: I think that these two lines should be in bootstrap.js? Generally they're placed there to help with unit testing and such.

dev_apps/acl-daemon/style/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  background-color: black;
+  font-family: "Lucida Console", Monaco, monospace;
+  color: green;
+}
+
+button {
+  font-size: 30px;
```
KevinGrandon: nit: I know it's just a dev app, but I'd still prefer to just use rems because we're supposed to use them everywhere.

apps/system/js/acl_manager.js
mikehenrty: Will do.

apps/system/js/notifications.js
mikehenrty: Not in v2.1 and 2.2 :( on master this will be much nicer.

apps/system/js/acl_manager.js
mikehenrty: We don't want these to be `else` since the above functions will return null if they can't find apps with that manifest or origin.

Makefile
```diff
@@ -93,6 +93,9 @@ NODE_MODULES_SRC?=modules.tar
 # tv
 GAIA_DEVICE_TYPE?=phone
 
+# Android Compatibility Layer
+ENABLE_ACL?=0
```
KevinGrandon: Is it possible to rename this to `ENABLE_APP_COMPATIBILITY`?

apps/system/js/app_compatibility.js
```diff
@@ -0,0 +1,285 @@
...
+      }
+    }.bind(this));
+  };
+
+  AppCompatibility.prototype.checkForExternalApps = function() {
+    var req = navigator.mozApps.mgmt.getAll();
+    req.onsuccess = function() {
+      var apps = req.result;
+      var lastApp = null;
+      var found = apps.some(function(app) {
```
KevinGrandon: Arrow functions will make your life easier - no need to update though.

dev_apps/acl-daemon/style/style.css
```diff
@@ -0,0 +1,19 @@
...
+  font-family: "Lucida Console", Monaco, monospace;
+  color: green;
+}
+
+button {
+  font-size: 2rem;
+}
+
+.label {
+  margin-bottom: 2px;
```
KevinGrandon: nit: don't use a mix of px and rem in this file :-p

apps/system/js/app_compatibility.js
mikehenrty: I wasn't sure if we had array functions in 2.0. In master, I'll use 'em.

PR #30770, opened 2015-07-01
Bug 1169573 - [Messages][NG] Lay out Messaging service structure, r=julienw, azasypkin

steveck-chung:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8af6a642e974449c5e99c853408b64521e0002bc

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
```diff
@@ -0,0 +1,187 @@
...
+
+  var MessagingService = {
+    /**
+     * Initializes our service using the Service mixin's initService and
+     * requests mozMobileMessage shim as client.
+     */
+    init() {
+      this.initService();
+
+      if (!('bridge' in self)) {
```
steveck-chung: Since threads.js should be imported in this.initService(), I dup the checking and import for sync initService. Not sure whether we should check this again for client initialization(maybe I could move this part into initClient() ).

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
```diff
@@ -0,0 +1,187 @@
...
+      var i = 0;
+      var requestResult = { hasError: false, return: [] };
+      var sendOpts = getSendOptionsFromServiceId(serviceId);
+      var requests = [];
+
+      recipients.forEach((recipient) => {
+        requests.push(this.client('send', recipient, content, sendOpts));
+      });
+
+      return Promise.all(requests).then((values) =>
```
steveck-chung: I simplify the return data to only the array of request result. Maybe I can simply return target/recipient and let client side to check result by themselves.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2e5fa28587e2c83cfb544b52029299d1ee5fa047

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
```diff
@@ -0,0 +1,179 @@
...
+        request = this.client('sendMMS', {
+          receivers: message.receivers,
+          subject: message.subject,
+          smil: message.smil,
+          attachments: message.attachments
+        }, sendOpts);
+      }
+
+      return request.then((evt) => {
+        this.client('delete', message.id);
```
steveck-chung: Not sure we should make delete shim and request into promise chain since these 2 actions seems have no relationship. Wondering if we can move out delete shim from request.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=92aefd766ef708472eab7e40dd3941002f940328

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
```diff
@@ -0,0 +1,176 @@
...
+      var recipients = opts.recipients,
+          subject = opts.subject,
+          content = opts.content,
+          serviceId = opts.serviceId = sanitizeServiceId(opts.serviceId),
+
+      if (!Array.isArray(recipients)) {
+        recipients = [recipients];
+      }
+
+      var message = SMIL.generate(content);
```
steveck-chung: Do we need to move `SMIL` as another service or simply imported it here? If we move it as service, we might have some problem to run this module in worker because SMIL.parse will use DOMParser and it might not work in worker(but it's still have some concern if we only imported the script. It might work just because we don't call parse here).

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
julienw: yeah, I don't really understand why you do that ?

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
julienw: simply importing it is good, we're already in a worker here so no problem.

I think DOMParser is available in a worker, but not sure.

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
```diff
@@ -0,0 +1,176 @@
...
+     * Initializes our service using the Service mixin's initService and
+     * requests mozMobileMessage shim as client.
+     */
+    init() {
+      this.initService();
+
+      if (!('bridge' in self)) {
+        importScripts('/lib/bridge.js');
+      }
+      this.client = bridge.client(CLIENT_NAME);
```
julienw: I'd call this variable differently; messagingApiClient ?

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
julienw: oh I think you mean to do that for consistency because the client needs the bridge ? then I'd do a separate function "ensureBridge" that would import it, and I'd call it at the start of init().

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
```diff
@@ -0,0 +1,176 @@
...
+        recipients = [recipients];
+      }
+
+      var i = 0;
+      var sendOpts = getSendOptionsFromServiceId(serviceId);
+      var requests = [];
+
+      recipients.forEach((recipient) => {
+        requests.push(this.client('send', recipient, content, sendOpts));
+      });
```
julienw: nit: you can use `var resuts = recipients.map((recipient) => this.client('send', recipient, content, sendOpts));`

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
```diff
@@ -0,0 +1,176 @@
...
+
+      return Promise.all(requests).then((values) =>
+        values.map((value, idx) => ({
+          success: !event.target.error,
+          result: event.target.result,
+          code: event.target.error,
+          recipient: recipients[idx]
+        }));
```
julienw: oh, I didn't know this trick of wrapping an object in parenthesis :) good !

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
```diff
@@ -0,0 +1,176 @@
...
+      return this.client('sendMMS', {
+        receivers: recipients,
+        subject: subject,
+        smil: message.smil,
+        attachments: message.attachments
+      }, sendOpts).then(
+        (event) => event.target.result
+      ).catch(
+        (event) => event.target.error
+      );
```
julienw: nit: you can use the 2-parameter `then` syntax here

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
julienw: IMO this is part of the "resendMessage" behavior

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
```diff
@@ -0,0 +1,176 @@
+/*global BridgeServiceMixin, Settings */
```
azasypkin: typo: ```messaging_service``` (in the name of the file). I don't have strong opinion, but I'd rather have separate folder for messaging service.

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
```diff
@@ -0,0 +1,176 @@
...
+'use strict';
+
+if (!('BridgeServiceMixin' in self)) {
+  importScripts('/services/js/bridge_service_mixin.js');
+}
+
+(function(exports) {
+  const SERVICE_NAME = 'messaging-service';
```
azasypkin: nit: I remember we agreed to not use 2 space indent for module pattern :)

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
```diff
@@ -0,0 +1,176 @@
...
+  const CLIENT_NAME = 'mozMobileMessageShim';
```
azasypkin: nit: maybe smth like ```MOBILE_MESSAGE_CLIENT_NAME``` or smth similar so that it's clear what client is supposed to be used?

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
azasypkin: > then I'd do a separate function "ensureBridge" that would import it, and I'd call it at the start of init().

Yeah, I think that makes sense

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
```diff
@@ -0,0 +1,176 @@
...
+     * Send sms by using send API shim.
+     * @param {*} opts Options for sending sms contains following properties:
+     * - recipients (string or array of string): contains the list of
+     *   recipients for this message.
+     * - content (string): the message's body.
+     * - serviceId (optional long or string): the SIM serviceId we use to send
+     *   the message.
+     * @returns {Promise} Promise with array of the request result.
+     */
+    sendSMS: function mm_send(opts) {
```
azasypkin: nit: ```sendSMS() {```

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
```diff
@@ -0,0 +1,176 @@
...
+      var recipients = opts.recipients || [],
+          content = opts.content,
+          serviceId = sanitizeServiceId(opts.serviceId);
+
+      if (!Array.isArray(recipients)) {
+        recipients = [recipients];
+      }
+
+      var i = 0;
```
azasypkin: nit: ```i``` is unused

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
```diff
@@ -0,0 +1,176 @@
...
+      var sendOpts = getSendOptionsFromServiceId(serviceId);
+      var requests = [];
+
+      recipients.forEach((recipient) => {
+        requests.push(this.client('send', recipient, content, sendOpts));
```
azasypkin: Interesting, if ```mozMobileShim``` you return ```DOMRequest``` from service method, but it's thenable and ```then``` callback will be called with ```request.result``` instead, I'm wondering what type of value you'll have here (I suspect threads lib can use this ```then``` to wait for result to return to the client).

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
```diff
@@ -0,0 +1,176 @@
...
+      });
+
+      return Promise.all(requests).then((values) =>
+        values.map((value, idx) => ({
+          success: !event.target.error,
```
azasypkin: mmm something is wrong here - I don't see where ```event``` is defined....

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
azasypkin: My IDE still complains :) But, yeah, it looks like a valid thing (btw I've learned that at https://hacks.mozilla.org/2015/06/es6-in-depth-arrow-functions/) :)

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
azasypkin: Mmm, I'm afraid not, at least with this I don't see that it's available in Aurora:
```
var blob = new Blob(
  ['self.onmessage = function(e) {self.postMessage("DOMParser" in self);};'], 
  { type: "text/javascript" }
);

var worker = new Worker(window.URL.createObjectURL(blob));
worker.onmessage = (e) => alert(e.data);
worker.postMessage("hello");
```

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
```diff
@@ -0,0 +1,176 @@
...
+        attachments: message.attachments
+      }, sendOpts).then(
+        (event) => event.target.result
+      ).catch(
+        (event) => event.target.error
+      );
+    },
+
+    /**
+     * Resend a message by send API shim with exsiting message information.
```
azasypkin: typo: ```existing```

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
```diff
@@ -0,0 +1,176 @@
...
+     * @param {Message} a formatted message in case you happen to have one.
```
azasypkin: nit: ```{Message} message A formatted```

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
azasypkin: > IMO this is part of the "resendMessage" behavior

Yep, I think so as well

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
```diff
@@ -0,0 +1,176 @@
...
+if (!('BridgeServiceMixin' in self)) {
+  importScripts('/services/js/bridge_service_mixin.js');
+}
+
+(function(exports) {
+  const SERVICE_NAME = 'messaging-service';
+  const CLIENT_NAME = 'mozMobileMessageShim';
+
+  const METHODS = Object.freeze([
+    'sendSMS', 'sendMMS', 'resendMessage'
```
azasypkin: Btw, what about ```retrieveMMS```? I'm fine if we plan to implement it in a separate bug.

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
julienw: ok then I think it was just planned and "possible"... Maybe we can file a bug about it if we need it.

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
steveck-chung: I thought separate function as well, but I'm not sure where we can set this function.  So we'll create 'ensureBridge' in BridgeServiceMixin?

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
steveck-chung: I'm thinking if it's possible to split the parse and generate into 2 js and we only need generate part for sending mms and parse for rendering in conversation view and composer.

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
julienw: I'd add it in this file instead; or maybe we need a BridgeClientMixin that would help calling the functions more easily? Not sure this is so useful yet but I can be convinced.

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
julienw: ideally generate and parse should be done in the same "layer" (or same object); but this could be a longer-term goal. I'm fine with a more pragmatic approach for now.

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
steveck-chung: You mean, we should clarify what is returned from `this.client('send', recipient, content, sendOpts)` because it might be simply the DOMRequest, not the result of the DOMRequest?

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
steveck-chung: aar right it should be value instead of event...

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
steveck-chung: yeah `retrieveMMS` should be in this service

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
azasypkin: Btw, just found useful link: https://developer.mozilla.org/en-US/docs/Web/API/Worker/Functions_and_classes_available_to_workers :)

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
azasypkin: I mean that when we do smth like this:
```
service.method('send', () => {
  return Promise.resolve(result);
});

// client will get "result"
```


So I'm wondering if threads.js considers ```DOMRequest.then``` as ```Promises.then``` somewhere in [1]. Maybe I'm just wrong.
```
service.method('send', () => {
  return mozMobileMessage.send(); // or mozMobileMessage.send().then((result) => result); 
});
```

[1] https://github.com/gaia-components/threads/blob/bbd610cd3b8d9b25c75925eba4f7d1e49f794dc8/lib/message.js#L175

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
steveck-chung: Thanks! I also found we already have a bug about this: https://bugzilla.mozilla.org/show_bug.cgi?id=677123

julienw: and now I can know how to find this directly from the webidl:
https://dxr.mozilla.org/mozilla-central/source/dom/webidl/MessageChannel.webidl#11

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
julienw: > So I'm wondering if threads.js considers DOMRequest.then as Promises.then somewhere in [1]. Maybe I'm just wrong.

Well, `Promise.resolve()` works the same with all "then-able", so works the same with Promises and DOMRequests :)

julienw: so YES :)

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
julienw: > Thanks! I also found we already have a bug about this: https://bugzilla.mozilla.org/show_bug.cgi?id=677123

OK, I don't think we'll have it soon ;)

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
steveck-chung: Ok, maybe it's time to give some experiment for our messageShim to test how the lib treat our DOMRequest

apps/sms/services/js/moz_mobile_message/messaging_serivce.js
steveck-chung: Second thought, since `retrieveMMS` already wrapped in shim, do you think we could could call the shim directly from view, or we need to access it though one more messaging service wrap up because they should be in the same group?

azasypkin: > we need to access it though one more messaging service wrap up because they should be in the same group?

Personally I'd go this way at least for the consistency sake :)

julienw: yes, me too, especially for this method which does not need to be so performant.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d17f9df8107609a9c7fbded30006798029a8bc76

apps/sms/services/js/moz_mobile_message/messaging_service.js
```diff
@@ -0,0 +1,191 @@
...
+  /**
+   * Initializes our service using the Service mixin's initService and
+   * requests mozMobileMessage shim as client.
+   */
+  init() {
+    ensureBridge();
+
+    this.initService();
+
+    this.mozMobileMessageClient = bridge.client(MOBILE_MESSAGE_CLIENT_NAME);
```
steveck-chung: I choose this name because this client is based on mozMobileMessage shim

apps/sms/services/js/moz_mobile_message/messaging_service.js
```diff
@@ -0,0 +1,191 @@
...
+(function(exports) {
+
+const SERVICE_NAME = 'messaging-service';
+const MOBILE_MESSAGE_CLIENT_NAME = 'mozMobileMessageShim';
+
+const METHODS = Object.freeze([
+  'sendSMS', 'sendMMS', 'resendMessage'
+]);
+
+function ensureBridge() {
```
steveck-chung: I move it to here but I think we might this as utils for all the services in worker

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=95d53d6e5991d1216abdc643e265f8fb41cb2d40

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=048f23c9683a746304fcdb8e0e8d643d5df5a67a

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c7faa0bf1f7fea8d808fff4c3629d0ee26e49f6f

apps/sms/services/js/moz_mobile_message/messaging_service.js
```diff
@@ -0,0 +1,193 @@
...
+
+    if (!Array.isArray(recipients)) {
+      recipients = [recipients];
+    }
+
+    var sendOpts = getSendOptionsFromServiceId(serviceId);
+    var results = recipients.map(
+      (recipient) => this.mozMobileMessageClient(
+        'send', recipient, content, sendOpts
+      ).then((evt) => evt, (evt) => evt)
```
steveck-chung: Since we don't have `finally` yet, not sure if there is other better way for this because I want aggregate all the success and error result instead of separate to resolve and reject .

apps/sms/services/js/moz_mobile_message/messaging_service.js
```diff
@@ -0,0 +1,193 @@
...
+    var sendOpts = getSendOptionsFromServiceId(serviceId);
+
+    return this.mozMobileMessageClient('sendMMS', {
+      receivers: recipients,
+      subject: subject,
+      smil: message.smil,
+      attachments: message.attachments
+    }, sendOpts).then(
+      (evt) => evt.target.result,
+      (evt) => {
```
steveck-chung: If I return the error like success originally, sendMMS will return both success result and error in resolve. So I change it to throw the error. But not sure if it's good for all the error handling because all the error will be mixed within reject.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=730a15ce97c3f1ee3048c1d2fdb8d4e0ea17365f

apps/sms/services/js/moz_mobile_message/messaging_service.js
```diff
@@ -0,0 +1,193 @@
...
+const SERVICE_NAME = 'messaging-service';
+const MOBILE_MESSAGE_CLIENT_NAME = 'mozMobileMessageShim';
+
+const METHODS = Object.freeze([
+  'sendSMS', 'sendMMS', 'resendMessage', 'retrieveMMS'
+]);
+
+function ensureBridge() {
+  if (!('bridge' in self)) {
+    importScripts('/lib/bridge.js');
```
azasypkin: ```lib/bridge/bridge.js````

apps/sms/services/js/moz_mobile_message/messaging_service.js
azasypkin: optional nit: since ```mozMobileMessageClient``` is private and we started to use Symbols, we can use smth like ```const priv = Object.freeze({ mozMobileMessageClient: Symbol("mozMobileMessageClient")})```. But it's veeeery optional, so you can ignore if you want :)

apps/sms/services/js/moz_mobile_message/messaging_service.js
```diff
@@ -0,0 +1,193 @@
+/*global BridgeServiceMixin, Settings */
```
azasypkin: I'm still not sure that ```messaging_service``` should live inside ```moz_mobile_message``` folder, what was your rationale?

apps/sms/services/js/moz_mobile_message/messaging_service.js
```diff
@@ -0,0 +1,193 @@
...
+function ensureBridge() {
+  if (!('bridge' in self)) {
+    importScripts('/lib/bridge.js');
+  }
+}
+
+// 0 is a valid value so we need to take care to not consider it as a falsy
+// value. We want to return null for anything that's not a number or a
+// string containing a number.
+function sanitizeServiceId(serviceId) {
```
azasypkin: nit: wrap method comment into ```/* */```

apps/sms/services/js/moz_mobile_message/messaging_service.js
```diff
@@ -0,0 +1,193 @@
...
+
+    this.initService();
+
+    this.mozMobileMessageClient = bridge.client(MOBILE_MESSAGE_CLIENT_NAME);
+  },
+
+  /**
+   * Send sms by using send API shim.
+   * @param {*} opts Options for sending sms contains following properties:
+   * - recipients (string or array of string): contains the list of
```
azasypkin: nit: you can use http://usejsdoc.org/tags-param.html#parameters-with-properties here, I see @julienw started to use this notation in the new code :)

apps/sms/services/js/moz_mobile_message/messaging_service.js
```diff
@@ -0,0 +1,193 @@
...
+        content = opts.content,
+        serviceId = sanitizeServiceId(opts.serviceId);
+
+    if (!Array.isArray(recipients)) {
+      recipients = [recipients];
+    }
+
+    var sendOpts = getSendOptionsFromServiceId(serviceId);
+    var results = recipients.map(
+      (recipient) => this.mozMobileMessageClient(
```
azasypkin: should not it be ```this.mozMobileMessageClient.method(....)```?

apps/sms/services/js/moz_mobile_message/messaging_service.js
azasypkin: ```bridge``` is not in ```/* global```` - one of the reason why JSHINT job is failing currently :)

apps/sms/services/js/moz_mobile_message/messaging_service.js
azasypkin: Btw, did you have chance to check if ```evt``` is ```DOMRequest``` or actual result as we discussed earlier?

apps/sms/services/js/moz_mobile_message/messaging_service.js
```diff
@@ -0,0 +1,193 @@
...
+
+    return Promise.all(results).then((values) =>
+      values.map((event, idx) => ({
+        success: !event.target.error,
+        result: event.target.result,
+        code: event.target.error,
+        recipient: recipients[idx]
+      }))
+    );
+
```
azasypkin: nit: redundant empty line

apps/sms/services/js/moz_mobile_message/messaging_service.js
```diff
@@ -0,0 +1,193 @@
...
+   * Send sms by using send API shim.
+   * @param {*} opts Options for sending sms contains following properties:
+   * - recipients (string or array of string): contains the list of
+   *   recipients for this message.
+   * - content (string): the message's body.
+   * - serviceId (optional long or string): the SIM serviceId we use to send
+   *   the message.
+   * @returns {Promise} Promise with array of the request result.
+   */
+  sendSMS(opts) {
```
azasypkin: optional nit: ```sendSMS({ recipients, content, serviceId })```?

apps/sms/services/js/moz_mobile_message/messaging_service.js
```diff
@@ -0,0 +1,193 @@
...
+        code: event.target.error,
+        recipient: recipients[idx]
+      }))
+    );
+
+  },
+
+  /**
+   * Send mms by using sendMMS API shim.
+   * @param {*} opts Options for sending mms contains following properties:
```
azasypkin: nit: same jsdoc nit about parameter with properties as above.

apps/sms/services/js/moz_mobile_message/messaging_service.js
```diff
@@ -0,0 +1,193 @@
...
+   * - content (array of SMIL slides): this is the content for the message (
+   *   see ConversationView for more information).
+   * - serviceId (optional long or string): the SIM that should be used for
+   *   sending this message. If this is not the current default configuration
+   *   for sending MMS, then we'll first switch the configuration to this
+   *   serviceId, and only then send the message. That means that the
+   *   "sending" event will come quite late in this case.
+   * @returns {Promise} DOMRequest promise with success result or error.
+   */
+  sendMMS(opts) {
```
azasypkin: optional nit: ```sendMMS({ recipients, content, subject, serviceId })```?

apps/sms/services/js/moz_mobile_message/messaging_service.js
```diff
@@ -0,0 +1,193 @@
...
+    var recipients = opts.recipients,
+        subject = opts.subject,
+        content = opts.content,
+        serviceId = opts.serviceId = sanitizeServiceId(opts.serviceId);
+
+    if (!Array.isArray(recipients)) {
+      recipients = [recipients];
+    }
+
+    var message = SMIL.generate(content);
```
azasypkin: nit: JSHINT error ```SMIL is not defined```

apps/sms/services/js/moz_mobile_message/messaging_service.js
azasypkin: nit: can be onliner with ```Promise.reject(evt.target.error)```

apps/sms/services/js/moz_mobile_message/messaging_service.js
```diff
@@ -0,0 +1,193 @@
...
+
+    var sendOpts = getSendOptionsFromServiceId(serviceId);
+
+    return this.mozMobileMessageClient('sendMMS', {
```
azasypkin: ```this.mozMobileMessageClient.method(...)``` ?

apps/sms/services/js/moz_mobile_message/messaging_service.js
```diff
@@ -0,0 +1,193 @@
...
+      (evt) => evt.target.result,
+      (evt) => {
+        throw evt.target.error;
+      }
+    );
+  },
+
+  /**
+   * Resend a message by send API shim with existing message information.
+   * @param {Message} message Formatted message in case you happen to have one.
```
azasypkin: ```in case you happen to have one``` - looks like ```message``` parameter is required :)

apps/sms/services/js/moz_mobile_message/messaging_service.js
```diff
@@ -0,0 +1,193 @@
...
+    if (!message) {
+      throw new Error('Message to resend is not defined.');
+    }
+
+    var serviceId = Settings.getServiceIdByIccId(message.iccId);
+    var sendOpts = getSendOptionsFromServiceId(serviceId);
+    var request;
+
+    if (message.type === 'sms') {
+      request = this.mozMobileMessageClient(
```
azasypkin: ```this.mozMobileMessageClient.method(...)``` ?

apps/sms/services/js/moz_mobile_message/messaging_service.js
```diff
@@ -0,0 +1,193 @@
...
+        'send', message.receiver, message.body, sendOpts
+      );
+    }
+    if (message.type === 'mms') {
+      request = this.mozMobileMessageClient('sendMMS', {
```
azasypkin: ```this.mozMobileMessageClient.method(...)``` ?

apps/sms/services/js/moz_mobile_message/messaging_service.js
```diff
@@ -0,0 +1,193 @@
...
+        receivers: message.receivers,
+        subject: message.subject,
+        smil: message.smil,
+        attachments: message.attachments
+      }, sendOpts);
+    }
+
+    return this.mozMobileMessageClient('delete', message.id).then(() =>
```
azasypkin: ```this.mozMobileMessageClient.method(...)``` ?

apps/sms/services/js/moz_mobile_message/messaging_service.js
```diff
@@ -0,0 +1,193 @@
...
+    });
+  },
+
+  /**
+   * Retrieve a MMS by retrieveMMS API shim by message ID.
+   * @param {Number} id mms message id for retrieval.
+   * @returns {Promise} DOMRequest promise with success result or error.
+   */
+  retrieveMMS(id) {
+    return this.mozMobileMessageClient('retrieveMMS', id).then(
```
azasypkin: ```this.mozMobileMessageClient.method(...)``` ?

apps/sms/services/js/moz_mobile_message/messaging_service.js
```diff
@@ -0,0 +1,193 @@
...
+      (evt) => evt.target.result,
+      (evt) => { throw evt.target.error }
```
azasypkin: nit: either add ```;``` before ```}``` or just use ```Promise.reject(evt.target.error)``` without curly brackets

apps/sms/services/test/unit/moz_mobile_message/messaging_service_test.js
```diff
@@ -0,0 +1,670 @@
...
+    require('/services/js/moz_mobile_message/messaging_service.js', done);
+  });
+
+  suiteTeardown(function() {
+    delete window.self;
+    delete window.importScripts;
+  });
+
+  setup(function() {
+    MessagingService =self.MessagingService;
```
azasypkin: nit: space after ```=```

apps/sms/services/js/moz_mobile_message/messaging_service.js
```diff
@@ -0,0 +1,193 @@
+/*global BridgeServiceMixin, Settings */
+'use strict';
+
+if (!('BridgeServiceMixin' in self)) {
+  importScripts('/services/js/bridge_service_mixin.js');
```
azasypkin: wondering if it would be better to use ```self.importScripts``` - will be a bit easier to mock this function in tests :)

julienw: why do you think it will be easier ?

My current thought is that we will never call importScripts in tests because the necessary object will always be there before.

apps/sms/services/js/moz_mobile_message/messaging_service.js
julienw: > optional nit: since mozMobileMessageClient is private and we started to use Symbols, we can use smth like const priv = Object.freeze({ mozMobileMessageClient: Symbol("mozMobileMessageClient")}). But it's veeeery optional, so you can ignore if you want :)

Because the file will be in its separate window, I don't think we need to be extra-careful here. I'd say simpler is better in this case :)

I agree with you about the logic to find the endpoint; is there a plugin for this in threads.js or do we need to implement ourselves. For sure there was an implementation of this in the previous version of threads.js

apps/sms/services/js/moz_mobile_message/messaging_service.js
julienw: Would it make sense to have different types of error ? WebAPIError vs normal Errors ? In Java that's what we'd do :)

apps/sms/services/test/unit/moz_mobile_message/messaging_service_test.js
```diff
@@ -0,0 +1,670 @@
...
+
+suite('Messaging service >', function() {
+
+  MocksHelperForAttachment.attachTestHelpers();
+
+  suiteSetup(function(done) {
+    window.self = {};
+    window.importScripts = sinon.stub();
+
+    require('/services/js/moz_mobile_message/messaging_service.js', done);
```
julienw: I think you can require the js at the top ? and I'm not sure we need to mock importScripts... or maybe you can mock it and make it throw because it should not be called :)

apps/sms/services/test/unit/moz_mobile_message/messaging_service_test.js
```diff
@@ -0,0 +1,670 @@
...
+]).init();
+
+var MessagingService, successEvt, errorEvt;
+
+suite('Messaging service >', function() {
+
+  MocksHelperForAttachment.attachTestHelpers();
+
+  suiteSetup(function(done) {
+    window.self = {};
```
julienw: is it really necessary ? `self === window` in a window context so if we have the right global objects it should work fine ?

apps/sms/services/test/unit/moz_mobile_message/messaging_service_test.js
```diff
@@ -0,0 +1,670 @@
...
+    delete window.importScripts;
+  });
+
+  setup(function() {
+    MessagingService =self.MessagingService;
+    successEvt = {
+      target: {
+        result: {}
+      }
+    };
```
julienw: maybe use a real event ? like a CustomEvent or even a real one ?

apps/sms/services/test/unit/moz_mobile_message/messaging_service_test.js
```diff
@@ -0,0 +1,670 @@
...
+        broadcast: () => {}
+      });
+
+      sinon.stub(bridge, 'service').returns(serviceStub);
+      sinon.stub(bridge, 'client');
+      MessagingService.init();
+    });
+
+    test('service mixing and client are initialized', function() {
+      sinon.assert.called(importScripts, '/lib/bridge.js');
```
julienw: I don't get how this works, because the bridge mock should have been already injected before you call init?

apps/sms/services/test/unit/moz_mobile_message/messaging_service_test.js
```diff
@@ -0,0 +1,670 @@
...
+      sinon.assert.calledWith(bridge.service, 'messaging-service');
+      sinon.assert.calledWith(bridge.client, 'mozMobileMessageShim');
+    });
+  });
+
+  suite('sendSMS', function() {
+    var smsOpts;
+
+    setup(function() {
+      MessagingService.mozMobileMessageClient = sinon.stub();
```
julienw: why don't you use this.sinon.stub ?

apps/sms/services/js/moz_mobile_message/messaging_service.js
azasypkin: > My current thought is that we will never call importScripts in tests because the necessary object will always be there before.

ah, nice catch, you're right!

apps/sms/services/js/moz_mobile_message/messaging_service.js
azasypkin: I think ideally - yes, but I'd rather do it later as enhancement/continuous-improvement once this service works and used by at least by one "real" consumer so that we can do smart things easier :)

apps/sms/services/test/unit/moz_mobile_message/messaging_service_test.js
azasypkin: Agree, I think in my past tests for SharedWorker, I used this approach (requiring script in ```suiteSetup```), because I was adding event listeners to ```self/window``` object, but now we have ```bridge``` abstraction that does this, so we can use usual flow here.

apps/sms/services/test/unit/moz_mobile_message/messaging_service_test.js
azasypkin: just a guess: currently ```setup``` is failing, so there is a chance that this test just isn't run :)

apps/sms/services/js/moz_mobile_message/messaging_service.js
steveck-chung: I'm not sure either. The reason why I put them into single folder is because I want centralize the messaging API related services into single place instead of disperse to separated folders, but I'm fine with other possible solution once we have agreement. @julienw wdyt?

apps/sms/services/js/moz_mobile_message/messaging_service.js
steveck-chung: My main reason is just like @julienw said, since the property could not be exposed if it's not defined in service, that's way I didn't make it into real "private" in the end. But still thanks for the suggestion because it should be essential pattern at least for shared component for view.

For the iframe endpoint, yeah I agree that it would be a problem if `endpoint` is necessary for returning the client. Implementing this logic might be better in my POV because service -> shim(iframe) communication is quite common currently. Making it into plugin is a good suggestion sounds a good idea since not all the service will need this, and hopefully we could get rid of the shim in the future(not the near future anyway...).

apps/sms/services/test/unit/moz_mobile_message/messaging_service_test.js
steveck-chung: yeah you're right that we don't need to mock the importScripts since it's not being called in th test actually.

apps/sms/services/js/moz_mobile_message/messaging_service.js
julienw: I think MessagingService should be in its own directory -- it's not part of the moz_mobile_message ship, it's only using it.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2f430ae825012f0b6da398c74159f503ce14d7a6

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b6e26275962189e86f2a0dc027b157a6e927fe9d

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ce920f9517c2b36bf1db59f3707727998c5db6da

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a3cbfcc84b935ccfc811e79a3014bdbb2c25e48c

apps/sms/services/js/messaging/messaging_service.js
```diff
@@ -0,0 +1,182 @@
...
+const METHODS = Object.freeze([
+  'sendSMS', 'sendMMS', 'resendMessage', 'retrieveMMS'
+]);
+
+function ensureBridge() {
+  if (!('bridge' in self)) {
+    importScripts('/lib/bridge/bridge.js');
+  }
+}
+/**
```
azasypkin: nit: empty line

apps/sms/services/js/messaging/messaging_service.js
```diff
@@ -0,0 +1,182 @@
...
+    recipients = recipients || [];
+    serviceId = sanitizeServiceId(serviceId);
+
+    if (!Array.isArray(recipients)) {
+      recipients = [recipients];
+    }
+
+    var sendOpts = getSendOptionsFromServiceId(serviceId);
+    var results = recipients.map(
+      (recipient) => this.mozMobileMessageClient.method(
```
azasypkin: optional nit: up to you, but I'd rather use outer ```then(successCallback, errorCallback)``` instead of inner ```then(successCallback)``` and outer ```catch(errorCallback)```.

apps/sms/services/js/messaging/messaging_service.js
```diff
@@ -0,0 +1,182 @@
...
+      receivers: recipients,
+      subject: subject,
+      smil: message.smil,
+      attachments: message.attachments
+    }, sendOpts);
+  },
+
+  /**
+   * Resend a message by send API shim with existing message information.
+   * @param {Message} message Target message that need to be resended.
```
azasypkin: typo: ```resent```

apps/sms/services/test/unit/messaging/messaging_service_test.js
```diff
@@ -0,0 +1,628 @@
...
+  'Settings',
+  'SMIL'
+]).init();
+
+suite('Messaging service >', function() {
+  var MessagingService, successResult, errorResult;
+
+  MocksHelperForAttachment.attachTestHelpers();
+
+  suiteSetup(function(done) {
```
azasypkin: mmm, I remember we agreed to get rid of ```suiteSetup``` and ```suiteTeardown``` enterily and move ```require('/services/js/messaging/messaging_service.js')``` to the place where the rest of ```require``` is located? :)

apps/sms/services/test/unit/messaging/messaging_service_test.js
```diff
@@ -0,0 +1,628 @@
...
+
+    require('/services/js/messaging/messaging_service.js', done);
+  });
+
+  suiteTeardown(function() {
+    delete window.importScripts;
+  });
+
+  setup(function() {
+    MessagingService = self.MessagingService;
```
azasypkin: nit: ```self``` is global context aka window, so need to redefine, we can just reference ```MessagingService``` directly.

Also can't we just leverage this setup to stub both client and service in the same way and don't abuse private ```MessagingService.mozMobileMessageClient```, like this (basically we'll just have ```'service mixing and client are initialized'``` test instead of ```init``` suite)?:
```
setup(function() {
  successResult = {};
  errorResult = new Error();

  serviceStub = sinon.stub({
    method: () => {},
    stream: () => {},
    broadcast: () => {},
    listen: () => {},
    plugin: () => {}
  });

  mozMobileMessageClientStub =  {
    method: sinon.stub()
  };

  this.sinon.stub(bridge, 'service').returns(serviceStub);
  this.sinon.stub(bridge, 'client').returns(mozMobileMessageClientStub);

  MessagingService.init();
});
```

apps/sms/services/test/unit/messaging/messaging_service_test.js
```diff
@@ -0,0 +1,628 @@
...
+        content: 'hola'
+      };
+
+      MessagingService.mozMobileMessageClient.method
+        .withArgs('send').returns(Promise.resolve(successResult));
+      MessagingService.sendSMS(smsOpts).then((results) => {
+        sinon.assert.calledTwice(
+          MessagingService.mozMobileMessageClient.method
+        );
+        assert.deepEqual(
```
azasypkin: optional nit: I'd rather replace ```deepEqual vs ....method.firstCall.args``` with just:
```
sinon.assert.calledTwice(mozMobileMessageClientStub.method);

sinon.assert.calledWithExactly(
  mozMobileMessageClientStub.method,
  'send', smsOpts.recipients[0], smsOpts.content, undefined
);

sinon.assert.calledWithExactly(
  mozMobileMessageClientStub.method,
  'send', smsOpts.recipients[1], smsOpts.content, undefined
);
```

apps/sms/services/test/unit/messaging/messaging_service_test.js
```diff
@@ -0,0 +1,628 @@
...
+        content: 'hola'
+      };
+
+      MessagingService.mozMobileMessageClient.method
+        .withArgs('send').returns(Promise.reject(errorResult));
+      MessagingService.sendSMS(smsOpts).then((results) => {
+        sinon.assert.calledTwice(
+          MessagingService.mozMobileMessageClient.method
+        );
+        assert.deepEqual(
```
azasypkin: optional nit: same as above

apps/sms/services/test/unit/messaging/messaging_service_test.js
```diff
@@ -0,0 +1,628 @@
...
+      this.sinon.stub(Settings, 'hasSeveralSim').returns(true);
+      smsOpts = {
+        recipients: '123',
+        content: 'hola',
+        serviceId: 0 // we use 0 because it's falsy, to test it still works
+      };
+
+      MessagingService.mozMobileMessageClient.method
+        .withArgs('send').returns(Promise.resolve(successResult));
+      MessagingService.sendSMS(smsOpts).then((results) => {
```
azasypkin: nit: ```results``` parameter is not used, should we assert result or just remove unused parameter (here and in a few places below)?

apps/sms/services/test/unit/messaging/messaging_service_test.js
```diff
@@ -0,0 +1,628 @@
...
+          'send', smsOpts.recipients, smsOpts.content, undefined
+        );
+      }).then(done, done);
+    });
+  });
+
+  suite('sendMMS', function() {
+
+    setup(function() {
+      this.sinon.spy(SMIL, 'generate');
```
azasypkin: nit: wondering if we should just stub it with predefined return result and avoid ```SMIL.generate.firstCall.returnValue```?

apps/sms/services/test/unit/messaging/messaging_service_test.js
```diff
@@ -0,0 +1,628 @@
...
+    test('send to one recipient successfully', function(done) {
+      var mmsOpts = {
+        recipients: '123',
+        subject: null,
+        content: 'hola'
+      };
+
+      MessagingService.mozMobileMessageClient.method
+        .withArgs('sendMMS').returns(Promise.resolve(successResult));
+      MessagingService.sendMMS(mmsOpts).then((result) => {
```
azasypkin: nit: ```result``` parameter is not used, should we assert result or just remove unused parameter (here and in a few places below)?

apps/sms/services/test/unit/messaging/messaging_service_test.js
steveck-chung: Will remove the unused parameter because we already verified the results (if service method called correctly) in previous test.

apps/sms/services/test/unit/messaging/messaging_service_test.js
steveck-chung: I think we need result in this test, but not necessary in the following tests for serviceId part in deed

apps/sms/services/test/unit/messaging/messaging_service_test.js
steveck-chung: This part is just copied from message_manager_test, but you're right that we can simply stub it here.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7c17c57366bb4b3f80bcec704723bbfffda63830

apps/sms/services/test/unit/messaging/messaging_service_test.js
```diff
@@ -0,0 +1,614 @@
...
+  suite('sendSMS', function() {
+    var smsOpts;
+
+    test('send to one recipient successfully', function(done) {
+      smsOpts = {
+        recipients: '123',
+        content: 'hola'
+      };
+
+      MessagingService.mozMobileMessageClient.method
```
azasypkin: optional nit: now you can replace ```MessagingService.mozMobileMessageClient.method```with ```clientStub.method``` everywhere, should be much shorter :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=95632a07d008257917338efe26a14e2437a0ec6e

PR #30793, opened 2015-07-02
Bug 1167507 - Wrap the input of the FDN Contact to MozIcc.updateContact() as a mozContact.

bevis-tseng:
This is to have explicit data type 'mozContact' in MozIcc.updateContact() instead of 'any'.
After review the implementation in Gaia, there are 2 modules infected
1. Contacts App for the ADN contacts in UICC.
2. Settings App for the FDN contacts in UICC.
However Contacts App already apply mozContact as the input of MozIcc.updateContact().
What we have to do is to have FDN contacts wrapped with mozContact.

### DIFF
--- a/apps/settings/js/modules/fdn_context.js
+++ b/apps/settings/js/modules/fdn_context.js
@@ -83,7 +83,7 @@ define(function() {
      * @param {String} action
      * @param {Number} options.cardIndex
      * @param {Object} options.contact
-     * @return {Object}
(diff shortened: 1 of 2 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4c7e6ba39b240cd08a39408d03cf4bf41479e4a8

PR #30798, opened 2015-07-02
Bug 1167144 - [Messages] Reduce the use of Threads.active and Threads currentId in conversation view, r=julienw

steveck-chung:


apps/sms/views/conversation/js/conversation.js
```diff
@@ -407,6 +407,7 @@ var ConversationView = {
   },
 
   getIdIterator: function conv_getIdIterator() {
+    // TODO: Will need to replace with Conversation service method.
```
julienw: not sure; I think it makes sense that the view itself keeps a local JS model once it's loaded

apps/sms/views/conversation/js/conversation.js
```diff
@@ -513,6 +514,9 @@ var ConversationView = {
     // TODO should we implement hooks to Navigation so that Threads could
     // get an event whenever the panel changes?
     Threads.currentId = args.id;
+    this.currentThreadId = args.id;
+    // TODO: Will need to replace with Conversation service method.
+    this.currentThread = Threads.active;
```
julienw: wondering if `activeThread` would be a better name (nitpicking...)

julienw: also wondering if we actually need to keep Threads.active at all. What do you think ?

apps/sms/views/conversation/js/conversation.js
```diff
@@ -747,6 +752,7 @@ var ConversationView = {
     // TODO add the activity/forward/draft stuff here
     // instead of in afterEnter: Bug 1010223
 
+    // TODO: Will need to replace with Conversation service method.
```
julienw: I don't think this makes sense here ?

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1159,7 +1165,8 @@ var ConversationView = {
       // If we're leaving a thread's message view,
       // ensure that the thread object's unreadCount
       // value is current (set = 0)
-      if (Threads.active) {
+      if (this.currentThread) {
+        // TODO: Will need to replace with Conversation service method.
```
julienw: not sure if this condition will be useful in the future actually -- I think it's here especially because we want the "unread" mark to disappear from the thread list even if we haven't marked the thread as read yet (because it's not finished loading)

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1168,7 +1175,7 @@ var ConversationView = {
       // do not prompt to save a draft and remove saved drafts
       // as the user deleted them manually
       if (Compose.isEmpty() &&
-        (Threads.active || this.recipients.length === 0)) {
+        (this.currentThread || this.recipients.length === 0)) {
```
julienw: in this case, I think it really means "isCurrentPanel('thread')"

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1180,7 +1187,7 @@ var ConversationView = {
         // Thread-less drafts are orphaned at this point
         // so they need to be resaved for persistence
         // Otherwise, clear the draft directly before leaving
-        if (!Threads.currentId) {
+        if (!this.currentThreadId) {
```
julienw: same for this, I think

julienw: we can tell draft support was done before we had a navigation

apps/sms/views/conversation/js/conversation.js
julienw: and wondering if we need `currentThreadId`, as maybe `this.activeThread.id` works just as fine ?

apps/sms/views/conversation/js/conversation.js
```diff
@@ -642,6 +646,7 @@ var ConversationView = {
     if (Navigation.isCurrentPanel('thread-list')) {
       this.container.textContent = '';
       this.cleanFields();
+      // TODO: Will need to replace with Conversation service method.
```
julienw: I don't think so :)

apps/sms/views/conversation/js/conversation.js
```diff
@@ -2915,9 +2923,9 @@ var ConversationView = {
     }
 
     // Update thread timestamp. Will be removed in bug 958105.
-    var thread = Threads.active;
-    if (thread) {
-      thread.timestamp = Date.now();
+    if (this.currentThread) {
+      // TODO: Will need to replace with Conversation service method.
```
julienw: we decided we don't want to change the timestamp anyway (https://bugzilla.mozilla.org/show_bug.cgi?id=958105)

apps/sms/views/conversation/js/conversation.js
steveck-chung: Well, it might depend on where we cache the messages map. It was cached in Threads originally, so it's possible to have a method in Conversation service to get the messages (iterator). Or you prefer to cache the messages in conversation view?

apps/sms/views/conversation/js/conversation.js
steveck-chung: I'm ok with the `activeThread`, and yes, `this.activeThread.id` should just work here.

We still need Threads before Conversation service ready, so you can see Threads.currentId still set here(and `Threads.active` is actually a getter that returning thread object once you set the valid currentId). So you can see `Threads.active` is still in used in somewhere and I just add a comment for it that it might be replace by Conversation service method.

apps/sms/views/conversation/js/conversation.js
steveck-chung: Ok, I think the question here would be: do we still need to cache currentId in conversation service? I think we can remove the dependency in Composer, but not sure about the activity part. If we still need currentId cross the view, it might be in conversation service and we'll need a method to cache it.

apps/sms/views/conversation/js/conversation.js
steveck-chung: So maybe we can just set thread as read while entering the panel and unify with https://github.com/steveck-chung/gaia/blob/message-remove-threads-dependency/apps/sms/views/conversation/js/conversation.js#L604. It will become a single entry that calling a conversation service method and change the thread property, then fire a thread change event to inbox view for UI update, wdyt?

apps/sms/views/conversation/js/conversation.js
steveck-chung: Seems possible to be new message panel without recipient?

apps/sms/views/conversation/js/conversation.js
steveck-chung: Same question about how we deal with currentId (https://github.com/mozilla-b2g/gaia/pull/30798#discussion_r33837856)

apps/sms/views/conversation/js/conversation.js
steveck-chung: Ok, so we can just remove this part right in this patch, or leave this for bug 958105? I'm fine to address this since the change should be relatively small.

apps/sms/views/conversation/js/conversation.js
steveck-chung: > we can tell draft support was done before we had a navigation 

Sorry but I'm not quite sure what you mean about this...

apps/sms/views/conversation/js/conversation.js
julienw: I think we'll have the method in ConversationService to get all messages, but that we'll have a local array as well; but I'm not sure of course.

apps/sms/views/conversation/js/conversation.js
julienw: ah yeah if Threads.active is used elsewhere then we need to keep it of course.

apps/sms/views/conversation/js/conversation.js
julienw: I don't think we need currentId in the service ? currentId is really a "view" characteristic: which view is currently displayed in this window.

I think the id is passed through the beforeEnter method, and can be kept locally in the view. I'm fine if we have another object that can hold the view's local model (including the id then) if we need to access it in a separate object, but we can also pass it as parameter instead when we need it.

I see this is used:
* in conversation.js -- ok, we remove it in this patch
* in utils.js/closeNotificationsForThread -- I see we always pass the argument so we can remove the code that uses currentId
* activity_handler.js -- we can replace by Navigation.isCurrentPanel
* information.js -- same, and I even think I do it in my navigation patch

apps/sms/views/conversation/js/conversation.js
julienw: previously we did that but then we decided to move it to after rendering so that rendering coud be as fast as possible.

All this code has been modified billion times :) maybe we can rationalize it a little more

apps/sms/views/conversation/js/conversation.js
julienw: yep, sorry, I meant to replace only the first part of the condition

apps/sms/views/conversation/js/conversation.js
julienw: maybe address bug 958105 first to be more consistent ?

apps/sms/views/conversation/js/conversation.js
julienw: so I mean: `if (!Threads.currentId)` really means `if (Navigation.isCurrentPanel('thread'))`; but because we did the support for "drafts" before the `Navigation` object existed, we had to check the panel using `Threads.currentId` instead of using a clean method on `Navigation`.

So it was only a comment saying this :) hope this is clearer

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=607948effec41ee1726f2069f73658331b502ede

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=993e2b2759fa6d2aec1fce7884dadebc104d6fdb

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=21187831c6bcee476dc4e0128ee2aad663e3e974

apps/sms/views/conversation/js/conversation.js
steveck-chung: Caching the ids might increase the complexity for maintenance since the messages could be updated frequently and we have to make sure this cache is synced with messages in service all the time. That means we need to update the cache while receiving the message event and I'm not sure if it's a proper way to do so in conversation view.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f23243a309ee905468c5d3066323a5dea35459c4

julienw:
It seems integration tests are failing for some reason. For example the case of launching the app using the notification. Unfortunately I can't try to reproduce manually right now because my SIM card has not been paid and thus doesn't work right now... Can you have a look ?

apps/sms/views/conversation/js/conversation.js
```diff
@@ -672,7 +676,7 @@ var ConversationView = {
     }
 
     // TODO move most of back() here: Bug 1010223
-    if (!this.isConversationPanel(Threads.currentId, nextPanel)) {
+    if (!this.isConversationPanel(this.activeThread, nextPanel)) {
```
julienw: you need to pass the id here; so either `Threads.currentId` or `this.activeThread.id` (I prefer the second one because this is one dependency less to Threads)

julienw: I think this is why the integration tests are failing

apps/sms/views/conversation/js/conversation.js
```diff
@@ -681,6 +685,8 @@ var ConversationView = {
     if (Navigation.isCurrentPanel('thread-list')) {
       this.container.textContent = '';
       this.cleanFields();
+      // TODO: Revisit the necessity of currentId once the Navigation patch is
+      // ready.
```
julienw: maybe remove "once the navigation patch is ready"

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1203,13 +1211,15 @@ var ConversationView = {
 
     // We're waiting for the keyboard to disappear before animating back
     return this._ensureKeyboardIsHidden().then(function() {
+      var isThread = Navigation.isCurrentPanel('thread');
```
julienw: nit: "isThread" => "inConversation" ? (we'll eventually move away from the term "thread" so better start now)

apps/sms/views/conversation/test/unit/conversation_test.js
```diff
@@ -143,6 +143,14 @@ suite('conversation.js >', function() {
       id: id,
       participants: participants
     });
+
+    ConversationView.activeThread = Threads.active;
```
julienw: what we really need (not in this patch though) is to properly call beforeEnter()/afterEnter() so that the view is in a correct state.

apps/sms/views/conversation/js/conversation.js
```diff
@@ -440,6 +440,8 @@ var ConversationView = {
   },
 
   getIdIterator: function conv_getIdIterator() {
+    // TODO: Will need to replace with Conversation service method or caching
+    // the ids locally.
     return Threads.active.messages.keys();
```
julienw: but can't we use this.activeThread now even if we replace this later ?

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1203,13 +1211,15 @@ var ConversationView = {
...
       // Need to assimilate recipients in order to check if any entered
       this.assimilateRecipients();
 
       // If we're leaving a thread's message view,
       // ensure that the thread object's unreadCount
       // value is current (set = 0)
-      if (Threads.active) {
+      if (isThread) {
+        // TODO: Will need to replace with Conversation service method.
         Threads.active.unreadCount = 0;
```
julienw: same here: even if we replace this later we can still use this.activeThread here now, right ?

apps/sms/views/conversation/test/unit/conversation_test.js
steveck-chung: Yeah I agree

apps/sms/views/conversation/js/conversation.js
steveck-chung: The reason why I don't use this.activeThread now is because we it might be forgotten easily once the service is ready and we do cache the this.activeThread in the view. No error will be displayed if we don't use the service method to change conversation properties. But yes we can use this.activeThread right now with no problem and I'm fine if you think leaving the TODO comments here is enough.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=24e74e6a8314bff8839e5bec2923f86f88115f8b

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ce7d084305e53aa51842b63a5601239b75dbf0d8

apps/sms/views/conversation/js/conversation.js
steveck-chung: You're right for this part, and we might need to check if `this.activeThread` exist first since new message will not have `this.activeThread`

apps/sms/views/conversation/js/conversation.js
```diff
@@ -2160,13 +2170,13 @@ var ConversationView = {
           },
           {
             l10nId: 'view-message-report',
-            method: function showMessageReport(messageId) {
+            method: (messageId) => {
```
steveck-chung: One more missing part that `this` didn't set correctly for the method

apps/sms/views/conversation/js/conversation.js
julienw: I think we can use it now; we'll especially need the service method when applying the changes from the select form, but not necessarily here. Maybe we'll need to be more asynchronous as well (or maybe not, I don't really know yet :D).

One possibility is having a "Thread" object in this.activeThread; this thread could have all message "ids" already, but not the messages' content, that could be fetched lazily.

So in the end, maybe we'll have `this.activeThread.messageIds()` here later.

All that to say... we don't really know how this will be later, so better use this.activeThread here now to simplify later things :)

apps/sms/views/conversation/js/conversation.js
julienw: good catch

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1203,13 +1211,15 @@ var ConversationView = {
...
       // Need to assimilate recipients in order to check if any entered
       this.assimilateRecipients();
 
       // If we're leaving a thread's message view,
       // ensure that the thread object's unreadCount
       // value is current (set = 0)
-      if (Threads.active) {
+      if (inConversation) {
+        // TODO: Will need to replace with Conversation service method.
         Threads.active.unreadCount = 0;
```
julienw: you missed my comment here :) can we replace `Threads.active` by `this.activeThread` here ?

apps/sms/views/conversation/js/conversation.js
```diff
@@ -672,7 +676,8 @@ var ConversationView = {
     }
 
     // TODO move most of back() here: Bug 1010223
-    if (!this.isConversationPanel(Threads.currentId, nextPanel)) {
+    if (this.activeThread &&
+      !this.isConversationPanel(this.activeThread.id, nextPanel)) {
```
julienw: nit: can you indent this so that `!this` is just below `this` ? (so 2 characters right -- not 1)

julienw:
One integration test is still failing

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3934f2d9af4d6421170c90ad29e1e246aae36948

apps/sms/views/conversation/js/conversation.js
```diff
@@ -157,8 +157,8 @@ var ConversationView = {
       'click', this.showOptions.bind(this)
     );
 
-    this.callNumberButton.addEventListener('click', function() {
```
steveck-chung: Another missing `this` - thanks for the integration test :p

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c007451321175bd92d4f9ca46aa822a65f186a1f

apps/sms/views/shared/test/unit/sms_test.js
```diff
@@ -114,7 +114,7 @@ suite('SMS App Unit-Test', function() {
     return elements;
   }
 
-  function threadSetupAppend() {
+  function threadSetup() {
```
steveck-chung: I admit the the unit test here looks not really qualified :/ ... But I simply make the test runnable instead of rewriting this because we'll going to refactor this later.

julienw:
of course you'll need a rebase now, because of bug 1162030; but I did it locally and it's quite easy indeed

apps/sms/views/conversation/js/conversation.js
```diff
@@ -2837,13 +2847,8 @@ var ConversationView = {
...
   prompt: function conv_prompt(opt) {
-    var complete = (function complete() {
-      if (!Navigation.isCurrentPanel('thread')) {
-        Navigation.toPanel('thread', { id: Threads.currentId });
-      }
-    }).bind(this);
-
-    var thread = Threads.active;
+    var currentId = this.activeThread.id;
+    var thread = this.activeThread;
```
julienw: nit: maybe reverse these 2 lines, so that you can have `var currentId = thread.id` ?

apps/sms/views/conversation/js/conversation.js
```diff
@@ -549,11 +551,13 @@ var ConversationView = {
     // TODO should we implement hooks to Navigation so that Threads could
     // get an event whenever the panel changes?
     Threads.currentId = args.id;
+    // TODO: Will need to replace with Conversation service method.
+    this.activeThread = Threads.active;
```
julienw: so I see that `Threads.active` is used only here, and we only assign to `Threads.currentId` and read only in `Threads.active` accessor...

So what do you think of removing both `Threads.active` and `Threads.currentId`, and replace with `Threads.get(args.id)` here ?

If you think this should be in a separate patch, it could be, but I think it's just as easy to do it now here, and this is quite the same purpose. What do you think ?

steveck-chung: Let's clean up `Threads.active` and `Threads.currentId` right in this patch

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d222d7ed4c3276e5eaa432a564949bf5a98ca90c

apps/sms/views/shared/js/activity_handler.js
```diff
@@ -353,7 +353,8 @@ var ActivityHandler = {
```
julienw: you need to remove `Threads` from the globals line, see https://treeherder.mozilla.org/logviewer.html#?job_id=1912132&repo=gaia

apps/sms/views/shared/js/activity_handler.js
```diff
@@ -353,7 +353,8 @@ var ActivityHandler = {
 
             // Close notification if we are already in thread view and view
             // become visible.
-            if (document.hidden && threadId === Threads.currentId) {
+            if (document.hidden &&
+              Navigation.isCurrentPanel('thread', { id: threadId })) {
```
julienw: nit: please move the line 2 spaces to the right, so that "Navigation" is just below "document".

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d3a35bf6f48e37d568fb3d01640a6538e1fb521f

PR #30843, opened 2015-07-06
Bug 1180771 - When navigating to a page on the same domain, don't set the default icon

gmarty:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c30b023c7999bb7f921788e59b7bc7de6beb3abe

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f83922b39244623284f05a99db46ce0d79b2139f

apps/system/js/app_chrome.js
```diff
@@ -709,6 +710,16 @@
 
       this.updateAddToHomeButton();
 
+      // We only change the icon back to the default one if the new page is not
+      // on the same domain than the previous one.
+      // In both cases, we look for the best icon after `mozbrowserloadend`.
+      var origin = this.getOriginFromURL(evt.detail);
```
benfrancis: var origin = new URL(evt.detail).origin;

apps/system/js/app_chrome.js
```diff
@@ -39,6 +39,7 @@
     this._themeChanged = false;
     this._titleTimeout = null;
     this.scrollable = app.browserContainer;
+    this._previousOrigin = '';
```
benfrancis: nit: In a resting state, this is actually current origin, I'd probably call it this._currentOrigin. We would probably change this to this._currentScope once we support app scope.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2256322f1ccb429a4ce981d71a283bac4c02822f

apps/system/js/app_chrome.js
gmarty: I tried this but there is a difference between `URL` and the `<a>` tag method: `new URL()` throws a TypeError if the string is not a valid URL. I could wrap it in a try/catch block but it might have performance issue. Also `URL` only support absolute URL (scheme-relative URL like `//example.com` are not supported).
Some unit tests fail because of that change.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8cb5e164056ee878d2c5fcdb0a9e6b102b30592e

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d37d8c9678b414afb61cce7b5670dbf3d43c12bf

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8ebe62390e9270adbc4a273d8f34bda09c9cdbc6

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a897f4566357563302c9cc60ceaaf910a57ef104

apps/system/js/app_chrome.js
benfrancis: My understanding is that the new URL() approach performs better than creating a whole node object. Is there a simple way to work around the //example.com by prepending the scheme from a base URL somehow do you think?

gmarty: Yes, AFAIK, `new URL()` performs better than the `<a>` tag. That's the reason why I create and reuse the same node object, `this.dummyATag`, over and over again in `getOriginFromURL()`, to avoid possible delays due to object instantiation and garbage collection.
For robustness, I'd definitely recommend use the `<a>` tag method. We don't want to risk throwing exception in this part of the code.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=fd23ea48f0577ca69487205357eac46bc60dd614

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=407e639895a5b5ba141e2c8d1d3b55625c5f970e

PR #30847, opened 2015-07-06
Bug 1180819 - Add features in imagecompare to help automating RTL acceptance tests

nojunpark:
following features are added to imagecompare:
- specify locale as gaiatest parameter
- put custom name per screenshot
- move failed diffs into mismatch folder

and a script is created to demo the RTL test automation in settings app

### DIFF
--- a/tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
+++ b/tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
@@ -1,6 +1,6 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
-# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 from marionette_driver import expected, By, Wait
 
@@ -10,6 +10,12 @@ from gaiatest.apps.base import Base
 class Settings(Base):
 
     name = 'Settings'
+
+    _bluetooth_l10n_locator = (By.CSS_SELECTOR, '[data-l10n-id="bt-status-turnoff"]')
+
+    _header_locator = (By.CSS_SELECTOR, '.current gaia-header')
+    _non_gaia_header_locator = (By.CSS_SELECTOR, '.current header')
+    _page_locator = (By.ID, 'root')
 
     _header_text_locator = (By.CSS_SELECTOR, '#root > gaia-header > h1')
     _data_text_locator = (By.ID, 'data-desc')
@@ -30,29 +36,57 @@ class Settings(Base):
     _usb_storage_confirm_button_locator = (By.CSS_SELECTOR, "button.ums-confirm-option")
     _gps_enabled_locator = (By.XPATH, "//input[@name='geolocation.enabled']")
     _gps_switch_locator = (By.XPATH, "//input[@name='geolocation.enabled']/..")
-    _accessibility_menu_item_locator = (By.ID, 'menuItem-accessibility')
+
+    # Following menu item list matches the menu item order in the settings app
+    _wifi_menu_item_locator = (By.ID, 'menuItem-wifi')
+    _sim_manager_menu_item_locator = (By.ID, 'menuItem-simManager')
+    _call_settings_menu_item_locator = (By.ID, 'menuItem-callSettings')
+    _message_settings_menu_item_locator = (By.ID, 'menuItem-messagingSettings')
     _cell_data_menu_item_locator = (By.ID, 'menuItem-cellularAndData')
     _bluetooth_menu_item_locator = (By.CSS_SELECTOR, '.menuItem-bluetooth')
+    _internet_sharing_menu_item_locator = (By.ID, 'menuItem-internetSharing')
+
     _sound_menu_item_locator = (By.ID, 'menuItem-sound')
-    _keyboard_menu_item_locator = (By.ID, "menuItem-keyboard")
-    _language_menu_item_locator = (By.CSS_SELECTOR, '.menuItem-languageAndRegion')
-    _do_not_track_menu_item_locator = (By.ID, 'menuItem-doNotTrack')
-    _media_storage_menu_item_locator = (By.CSS_SELECTOR, '.menuItem-mediaStorage')
-    _screen_lock_menu_item_locator = (By.ID, 'menuItem-screenLock')
     _display_menu_item_locator = (By.ID, 'menuItem-display')
-    _wifi_menu_item_locator = (By.ID, 'menuItem-wifi')
-    _device_info_menu_item_locator = (By.ID, 'menuItem-deviceInfo')
-    _battery_menu_item_locator = (By.CSS_SELECTOR, '.menuItem-battery')
-    _sim_manager_menu_item_locator = (By.ID, 'menuItem-simManager')
-    _date_and_time_menu_item_locator = (By.ID, 'menuItem-dateAndTime')
     _homescreen_menu_item_locator = (By.ID, 'menuItem-homescreen')
-    _browsing_privacy_item_locator = (By.ID, 'menuItem-browsingPrivacy')
+    _search_menu_item_locator = (By.ID, 'menuItem-search')
+    _navigation_menu_item_locator = (By.ID, 'menuItem-navigation')
+    _notification_menu_item_locator = (By.ID, 'menuItem-notifications')
+    _date_and_time_menu_item_locator = (By.ID, 'menuItem-dateAndTime')
+    _language_menu_item_locator = (By.CSS_SELECTOR, '.menuItem-languageAndRegion')
+    _keyboard_menu_item_locator = (By.ID, "menuItem-keyboard")
+    _theme_menu_item_locator = (By.ID, "menuItem-themes")
+    _addon_menu_item_locator = (By.ID, "menuItem-addons")
+    _achievements_menu_item_locator = (By.ID, "achievements-section")
+
+    _firefox_accounts_menu_item_locator = (By.ID, "menuItem-fxa")
     _findmydevice_locator = (By.ID, 'menuItem-findmydevice')
+
+    _screen_lock_menu_item_locator = (By.ID, 'menuItem-screenLock')
+    _app_permission_menu_item_locator = (By.ID, 'menuItem-appPermissions')
+    _do_not_track_menu_item_locator = (By.ID, 'menuItem-doNotTrack')
+    _browsing_privacy_item_locator = (By.ID, 'menuItem-browsingPrivacy')
+    _media_storage_menu_item_locator = (By.CSS_SELECTOR, '.menuItem-mediaStorage')
+    _application_storage_menu_item_locator = (By.CSS_SELECTOR, '.menuItem-applicationStorage')
+
+    _device_info_menu_item_locator = (By.ID, 'menuItem-deviceInfo')
+    _downloads_menu_item_locator = (By.ID, 'menuItem-downloads')
+    _battery_menu_item_locator = (By.CSS_SELECTOR, '.menuItem-battery')
+    _accessibility_menu_item_locator = (By.ID, 'menuItem-accessibility')
+    _developer_menu_item_locator = (By.ID, 'menuItem-developer')
+    _improve_menu_item_locator = (By.ID, 'menuItem-improveBrowserOS')
+    _help_menu_item_locator = (By.ID, 'menuItem-help')
+
+    _main_title_locator = (By.ID, 'main-list-title')
 
     def launch(self):
         Base.launch(self)
         Wait(self.marionette).until(
             expected.element_present(*self._app_loaded_locator))
+
+        # this is located at the end of the page.  If this is shown, everything is rendered.
+        Wait(self.marionette).until(
+            expected.element_present(*self._bluetooth_l10n_locator))
 
     def switch_to_settings_app(self):
         Wait(self.marionette).until(
@@ -144,97 +178,131 @@ class Settings(Base):
     def bluetooth_menu_item_description(self):
         return self.marionette.find_element(*self._bluetooth_text_locator).text
 
-    def a11y_open_accessibility_settings(self):
-        from gaiatest.apps.settings.regions.accessibility import Accessibility
-        self._a11y_click_menu_item(self._accessibility_menu_item_locator)
-        return Accessibility(self.marionette)
+    @property
+    def screen_element(self):
+        return self.marionette.find_element(*self._page_locator)
 
-    def open_cell_and_data_settings(self):
-        from gaiatest.apps.settings.regions.cell_data import CellData
-        self._tap_menu_item(self._cell_data_menu_item_locator)
-        return CellData(self.marionette)
+    @property
+    def title(self):
+        return self.marionette.find_element(*self._main_title_locator)
 
-    def open_bluetooth_settings(self):
-        from gaiatest.apps.settings.regions.bluetooth import Bluetooth
-        self._tap_menu_item(self._bluetooth_menu_item_locator)
-        return Bluetooth(self.marionette)
+    def open_wifi(self):
+        return self._open_subpage(self._wifi_menu_item_locator, 'wifi', 'Wifi')
 
-    def open_sound_settings(self):
-        from gaiatest.apps.settings.regions.sound import Sound
-        self._tap_menu_item(self._sound_menu_item_locator)
-        return Sound(self.marionette)
+    def open_sim_manager(self):
+        return self._open_subpage(self._sim_manager_menu_item_locator, 'sim_manager', 'SimManager')
 
-    def open_keyboard_settings(self):
-        from gaiatest.apps.settings.regions.keyboard import Keyboard
-        self._tap_menu_item(self._keyboard_menu_item_locator)
-        return Keyboard(self.marionette)
+    def open_call(self):
+        return self._open_subpage(self._call_settings_menu_item_locator)
 
-    def open_language_settings(self):
-        from gaiatest.apps.settings.regions.language import Language
-        self._tap_menu_item(self._language_menu_item_locator)
-        language_menu = Language(self.marionette)
+    def open_message(self):
+        return self._open_subpage(self._message_settings_menu_item_locator)
+
+    def open_cell_and_data(self):
+        return self._open_subpage(self._cell_data_menu_item_locator, 'cell_data', 'CellData')
+
+    def open_bluetooth(self):
+        return self._open_subpage(self._bluetooth_menu_item_locator, 'bluetooth', 'Bluetooth')
+
+    def open_internet_sharing(self):
+        return self._open_subpage(self._internet_sharing_menu_item_locator)
+
+    def open_sound(self):
+        return self._open_subpage(self._sound_menu_item_locator, 'sound', 'Sound')
+
+    def open_display(self):
+        return self._open_subpage(self._display_menu_item_locator, 'display', 'Display')
+
+    def open_homescreen(self):
+        return self._open_subpage(self._homescreen_menu_item_locator, 'homescreen_settings',
+                                           'HomescreenSettings')
+
+    def open_search(self):
+        return self._open_subpage(self._search_menu_item_locator)
+
+    def open_navigation(self):
+        return self._open_subpage(self._navigation_menu_item_locator)
+
+    def open_notification(self):
+        return self._open_subpage(self._notification_menu_item_locator)
+
+    def open_date_and_time(self):
+        return self._open_subpage(self._date_and_time_menu_item_locator, 'date_and_time', 'DateAndTime')
+
+    def open_language(self):
+        language_menu = self._open_subpage(self._language_menu_item_locator, 'language', 'Language')
         language_menu.wait_for_languages_to_load()
         return language_menu
 
-    def open_do_not_track_settings(self):
-        from gaiatest.apps.settings.regions.do_not_track import DoNotTrack
-        self._tap_menu_item(self._do_not_track_menu_item_locator)
-        return DoNotTrack(self.marionette)
+    def open_keyboard(self):
+        return self._open_subpage(self._keyboard_menu_item_locator, 'keyboard', 'Keyboard')
 
-    def open_media_storage_settings(self):
-        from gaiatest.apps.settings.regions.media_storage import MediaStorage
-        self._tap_menu_item(self._media_storage_menu_item_locator)
-        return MediaStorage(self.marionette)
+    def open_themes(self):
+        return self._open_subpage(self._theme_menu_item_locator)
 
-    def open_screen_lock_settings(self):
-        from gaiatest.apps.settings.regions.screen_lock import ScreenLock
-        self._tap_menu_item(self._screen_lock_menu_item_locator)
-        return ScreenLock(self.marionette)
+    def open_addons(self):
+        return self._open_subpage(self._addon_menu_item_locator)
 
-    def open_display_settings(self):
-        from gaiatest.apps.settings.regions.display import Display
-        self._tap_menu_item(self._display_menu_item_locator)
-        return Display(self.marionette)
+    def open_achievements(self):
+        return self._open_subpage(self._achievements_menu_item_locator)
 
-    def open_wifi_settings(self):
-        from gaiatest.apps.settings.regions.wifi import Wifi
-        self._tap_menu_item(self._wifi_menu_item_locator)
-        return Wifi(self.marionette)
-
-    def open_date_and_time_settings(self):
-        from gaiatest.apps.settings.regions.date_and_time import DateAndTime
-        self._tap_menu_item(self._date_and_time_menu_item_locator)
-        return DateAndTime(self.marionette)
+    def open_firefox_accounts(self):
+        return self._open_subpage(self._firefox_accounts_menu_item_locator)
 
     def open_findmydevice(self):
-        from gaiatest.apps.settings.regions.findmydevice import FindMyDevice
-        self._tap_menu_item(self._findmydevice_locator)
-        return FindMyDevice(self.marionette)
+        return self._open_subpage(self._findmydevice_locator, 'findmydevice', 'FindMyDevice')
 
-    def open_device_info_settings(self):
-        from gaiatest.apps.settings.regions.device_info import DeviceInfo
-        self._tap_menu_item(self._device_info_menu_item_locator)
-        return DeviceInfo(self.marionette)
+    def open_screen_lock(self):
+        return self._open_subpage(self._screen_lock_menu_item_locator, 'screen_lock', 'ScreenLock')
 
-    def open_battery_settings(self):
-        from gaiatest.apps.settings.regions.battery import Battery
-        self._tap_menu_item(self._battery_menu_item_locator)
-        return Battery(self.marionette)
+    def open_app_permissions(self):
+        return self._open_subpage(self._app_permission_menu_item_locator)
 
-    def open_sim_manager_settings(self):
-        from gaiatest.apps.settings.regions.sim_manager import SimManager
-        self._tap_menu_item(self._sim_manager_menu_item_locator)
-        return SimManager(self.marionette)
+    def open_do_not_track(self):
+        return self._open_subpage(self._do_not_track_menu_item_locator, 'do_not_track', 'DoNotTrack')
 
-    def open_homescreen_settings(self):
-        from gaiatest.apps.settings.regions.homescreen_settings import HomescreenSettings
-        self._tap_menu_item(self._homescreen_menu_item_locator)
-        return HomescreenSettings(self.marionette)
+    def open_browsing_privacy(self):
+        return self._open_subpage(self._browsing_privacy_item_locator, 'browsing_privacy', 'BrowsingPrivacy')
 
-    def open_browsing_privacy_settings(self):
-        from gaiatest.apps.settings.regions.browsing_privacy import BrowsingPrivacy
-        self._tap_menu_item(self._browsing_privacy_item_locator)
-        return BrowsingPrivacy(self.marionette)
+    def open_media_storage(self):
+        return self._open_subpage(self._media_storage_menu_item_locator, 'media_storage', 'MediaStorage')
+
+    def open_application_storage(self):
+        return self._open_subpage(self._application_storage_menu_item_locator)
+
+    def open_device_info(self):
+        return self._open_subpage(self._device_info_menu_item_locator, 'device_info', 'DeviceInfo')
+
+    def open_downloads(self):
+        return self._open_subpage(self._downloads_menu_item_locator)
+
+    def open_battery(self):
+        return self._open_subpage(self._battery_menu_item_locator, 'battery', 'Battery')
+
+    def open_accessibility(self):
+        return self._open_subpage(self._accessibility_menu_item_locator, 'accessibility', 'Accessibility')
+
+    def open_developer(self):
+        return self._open_subpage(self._developer_menu_item_locator)
+
+    def open_improve(self):
+        return self._open_subpage(self._improve_menu_item_locator)
+
+    def open_help(self):
+        return self._open_subpage(self._help_menu_item_locator)
+
+    def _open_subpage(self, locator, file_name=None, class_name=None):
+        self._tap_menu_item(locator)
+        class_object = self._get_class_by_name(file_name, class_name)
+        if class_object is not None:
+            return class_object(self.marionette)
+
+    def _get_class_by_name(self, file_name=None, class_name=None):
+        if file_name is not None:
+            package_path = 'gaiatest.apps.settings.regions.{}'.format(file_name.lower())
+            mod = __import__(package_path, fromlist=[class_name])
+            class_defn = getattr(mod, class_name)
+            return class_defn
 
     @property
     def is_airplane_mode_visible(self):
@@ -276,3 +344,13 @@ class Settings(Base):
     def _wait_for_toggle_ready(self, by, locator):
         checkbox = self.marionette.find_element(by, locator)
         Wait(self.marionette).until(expected.element_enabled(checkbox))
+
+    # this method is a copy of go_back() method in regions/keyboard.py
+    def return_to_prev_menu(self, parent_view, gaia_header=True):
+
+        # TODO: remove tap with coordinates after Bug 1061698 is fixed
+        if gaia_header:
+            self.marionette.find_element(*self._header_locator).tap(25, 25)
+        else:
+            self.marionette.find_element(*self._non_gaia_header_locator).tap(25, 25)
+        Wait(self.marionette).until(expected.element_displayed(parent_view))
--- a/tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/device_info.py
+++ b/tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/device_info.py
@@ -9,15 +9,26 @@ from gaiatest.apps.base import Base
 
 class DeviceInfo(Base):
 
+    _page_locator = (By.ID, 'about')
     _phone_number_locator = (By.CSS_SELECTOR, '.deviceInfo-msisdns')
     _model_locator = (By.CSS_SELECTOR, '#about small[data-name="deviceinfo.product_model"]')
     _software_locator = (By.CSS_SELECTOR, '#about small[data-name="deviceinfo.software"]')
     _more_info_button_locator = (By.CSS_SELECTOR, 'a[href="#about-moreInfo"]')
+    _reset_button_locator = (By.CLASS_NAME, 'reset-phone')
+    _reset_confirm_locator = (By.CLASS_NAME, 'confirm-reset-phone')
+    _reset_cancel_locator = (By.CLASS_NAME, 'cancel-reset-phone')
+
+    _update_frequency_locator = (By.NAME, 'app.update.interval')
+    _update_ok_button_locator = (By.CLASS_NAME, 'value-option-confirm')
 
     def __init__(self, marionette):
         Base.__init__(self, marionette)
         Wait(self.marionette).until(
             expected.element_displayed(*self._model_locator))
+
+    @property
+    def screen_element(self):
+        return self.marionette.find_element(*self._page_locator)
 
     @property
     def phone_number(self):
@@ -35,8 +46,35 @@ class DeviceInfo(Base):
         self.marionette.find_element(*self._more_info_button_locator).tap()
         return self.MoreInfo(self.marionette)
 
+    # In order to access UI, the frame needs to be switched to root
+    def tap_update_frequency(self):
+        self.marionette.find_element(*self._update_frequency_locator).tap()
+        self.marionette.switch_to_frame()
+        Wait(self.marionette).until(expected.element_displayed(
+            self.marionette.find_element(*self._update_ok_button_locator)))
+
+    # When leaving the page, return to the saved application frame
+    def exit_update_frequency(self):
+        self.marionette.find_element(*self._update_ok_button_locator).tap()
+        self.apps.switch_to_displayed_app()
+        Wait(self.marionette).until(expected.element_displayed(
+            self.marionette.find_element(*self._page_locator)))
+
+    def tap_reset_phone(self):
+        self.marionette.find_element(*self._reset_button_locator).tap()
+        Wait(self.marionette).until(expected.element_displayed(
+            self.marionette.find_element(*self._reset_confirm_locator)))
+
+    def confirm_reset(self, response=True):
+        if response is True:
+            self.marionette.find_element(*self._reset_confirm_locator).tap()
+        else:
+            self.marionette.find_element(*self._reset_cancel_locator).tap()
+            Wait(self.marionette).until(expected.element_displayed(self.screen_element))
+
     class MoreInfo(Base):
 
+        _more_information_page_locator = (By.ID, 'about-moreInfo')
         _os_version_locator = (By.CSS_SELECTOR, '#about-moreInfo small[data-name="deviceinfo.os"]')
         _hardware_revision_locator = (By.CSS_SELECTOR, '#about-moreInfo small[data-name="deviceinfo.hardware"]')
         _mac_address_locator = (By.CSS_SELECTOR, '#about-moreInfo small[data-name="deviceinfo.mac"]')
@@ -54,6 +92,10 @@ class DeviceInfo(Base):
             Base.__init__(self, marionette)
             Wait(self.marionette).until(
                 expected.element_displayed(*self._os_version_locator))
+
+        @property
+        def screen(self):
+            return self.marionette.find_element(*self._more_information_page_locator)
 
         @property
         def os_version(self):
--- a/tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
+++ b/tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
@@ -17,6 +17,7 @@ from mozlog.structured import get_default_logger
 
 from gaiatest import GaiaTestCase
 from gaiatest.apps.system.app import System
+from gaiatest import DEFAULT_SETTINGS
 
 
 class GaiaImageCompareTestCase(GaiaTestCase):
@@ -27,6 +28,8 @@ class GaiaImageCompareTestCase(GaiaTestCase):
         self.fuzz_factor = kwargs.pop('fuzz_factor')
         self.reference_path = kwargs.pop('reference_path')
         self.screenshots_path = kwargs.pop('screenshots_path')
+        self.mismatch_path = os.path.join(self.screenshots_path, 'mismatches')
+        self.locale = kwargs.pop('locale')
 
         self.logger = get_default_logger()
         self.picture_index = 0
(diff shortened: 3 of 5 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3f8e6a40eb91db16215363d2910c6ff057e1dca1

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3f99ce3d15b3355dcbdff5aa93976ddc75a21854

tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
```diff
@@ -48,17 +50,20 @@ def tearDown(self):
         """At the end of test execution, it checks for the errors"""
         self.assertTrue(self.test_passed, msg=self.failcomment)
 
-    def take_screenshot(self):
-        """invokes screen capture event, crops the status bar, and saves to the file"""
+    # hook to override default locale setting
+    # see https://github.com/mozilla-b2g/gaia/blob/master/tests/python/gaia-ui-tests/gaiatest/gaia_test.py#L1019
+    def modify_settings(self, settings):
+        settings['language.current'] = self.locale
```
geoelectric: Only trick with this is now it'll be significant whether you call modify_settings first or last (one clobbers you w/ its locale, the other does it the other way around). That needs to be communicated somehow.

However, this really misses the point of this method. It's a user hook for a -test- to modify settings, not somewhere for the harness to modify its own settings. It's basically a virtual abstract method.

My fairly strong suggestion is to -not- put the locale override here, instead do it when initial settings are set up (initialize that dict from cmd line instead of just assigning 'en-US') and keep modify_settings() as purely to do per-test overrides.

geoelectric: Second line note, so you get notifications correctly. Actually, I misread this. This is your test override, I thought this was in core harness. My bad on that. This might be OK as-is.

FWIW, I think moving the locale selection from cmdline into the main gaiatest would be entirely reasonable (that's why I thought you were doing so here) in which case my comment would stand.

geoelectric:
Looks great!

tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
npark-mozilla: yeah, thought it might also be useful in main gaiatest, but I wanted to see what other people think about it.  If we do agree to put this in gaiatest, yes then we should set it up initially.

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
```diff
@@ -236,6 +264,10 @@ def open_browsing_privacy_settings(self):
         self._tap_menu_item(self._browsing_privacy_item_locator)
         return BrowsingPrivacy(self.marionette)
 
+    # method to open menu item that does not return object
+    def simple_open_menu(self, menu_locator):
```
mwargers: Why not just rename _tap_menu_item to tap_menu_item and use that?

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
```diff
@@ -276,3 +308,13 @@ def _a11y_click_menu_item(self, menu_item_locator):
...
         checkbox = self.marionette.find_element(by, locator)
         Wait(self.marionette).until(expected.element_enabled(checkbox))
+
+    # this method is a copy of go_back() method in regions/keyboard.py
+    def return_to_main_menu(self, gaia_header=True):
+        # TODO: remove tap with coordinates after Bug 1061698 is fixed
+        if gaia_header:
+            self.marionette.find_element(*self._header_locator).tap(25, 25)
+        else:
+            self.marionette.find_element(*self._non_gaia_header_locator).tap(25, 25)
```
mwargers: If it's a case of _header_locator or _non_gaia_header_locator, you could also make the  _header_locator '.current gaia-header, .current header', but I guess this is fine as well.

tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
mwargers: I think it's a good idea to put in gaiatest.

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
JohanLorenzo: Agreed with @mwargers about the naming. Starting a function with an _ is a convention in Python for saying that a function of an attribute is protected.

That said, this won't solve another issue: If we pass the locator itself, we currently pass protected variables (the locators). To solve this we can either:

* Use the functions `open_*_settings()` already defined
* Or make the locators public.

I'm more in favor of the first option. The fact that `open_*_settings()` return another page object is not problematic, we're not obliged to use them.

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
```diff
@@ -276,3 +308,13 @@ def _a11y_click_menu_item(self, menu_item_locator):
     def _wait_for_toggle_ready(self, by, locator):
         checkbox = self.marionette.find_element(by, locator)
         Wait(self.marionette).until(expected.element_enabled(checkbox))
+
+    # this method is a copy of go_back() method in regions/keyboard.py
+    def return_to_main_menu(self, gaia_header=True):
```
JohanLorenzo: Not sure if we need to specify `gaia_header` in the test. @npark-mozilla, do you think it's a good idea to try to search for the regular gaia_header and if it's not present, find the non gaia one?

tests/python/gaia-ui-tests/gaiatest/tests/graphics/RTL/test_settings_RTL_POC.py
```diff
@@ -0,0 +1,177 @@
...
+    def setUp(self):
+        GaiaImageCompareTestCase.setUp(self)
+
+    def test_settings_app(self):
+
+        settings = Settings(self.marionette)
+        settings.launch()
+
+        #this is located at the end of the page.  If this is shown, everything is rendered.
+        Wait(self.marionette).until(
```
JohanLorenzo: We can put this wait in `settings.launch()` I think. This prevents from duplicating the locator in each test.

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
npark-mozilla: Hmm, but I think not having gaia header in this case actually was a bug (I filed one). Technically it was a workaround to make it work with non-gaia header, so I don't think we should explicitly search for it, since it might delay the tests if we use the timeouts, and it will hide the fact that it does not use the gaia-header.

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
JohanLorenzo: Following up on IRC. I think we can factorize all the `open_*_settings()` like this:

```python
     def open_battery_settings(self):
         return self._open_subpage_settings('Battery')

     def open_wifi_settings(self):
         return self._open_subpage_settings('Wifi')

     def _open_subpage_settings(self, class_name, locator):
         self._tap_menu_item(locator)
         Class = self._get_class_by_name(class_name):
         return Class(self.marionette)

    def _get_class_by_name(self, class_name):
         package_path = 'gaiatest.apps.settings.regions.{}'.format(class_name.lower())
         mod = __import__(package_path, fromlist=[class_name])
         Class = getattr(mod, class_name)
         return Class
```

tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
npark-mozilla: I thought about adding it to the gaiatest too, but since this variable is already in testvars and I'm modifying it to use with imagecompare, I thought it might not be right to integrate it into gaiatest, since in runner/base.py, none of the parameters are about actually modifying a specific testvar value.  Since in most cases  we can easily change the locale by changing the testvar value (in RTL there is a potential need to run the same test in many different locale) perhaps we should see how this works out first.

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
JohanLorenzo: Okay. Let's leave it as is, then

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
npark-mozilla: will incorporate.  only thing is that we need extra parameter for filename and class name, since 'DeviceInfo' class sits in 'device_info.py' file.

Also need to take care of cases where we don't have the classes as well

tests/python/gaia-ui-tests/gaiatest/tests/graphics/RTL/test_settings_RTL_POC.py
npark-mozilla: added

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
npark-mozilla: Ah, but if I do it that way when they have non-gaia header, I won't find it out.  ( I think they have to use gaia header for the settings menu)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f3d17232c39b63b0c63ef3444d183ba601e94473

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
```diff
@@ -144,97 +178,123 @@ def language_menu_item_description(self):
...
+        self._tap_menu_item(locator)
+        class_object = self._get_class_by_name(file_name, class_name)
+        if class_object is not None:
+            return class_object(self.marionette)
+
+    def _get_class_by_name(self, file_name, class_name):
+        if file_name is not None:
+            package_path = 'gaiatest.apps.settings.regions.{}'.format(file_name.lower())
+            mod = __import__(package_path, fromlist = [class_name])
+            class_object = getattr(mod, class_name)
```
JohanLorenzo: I'd rename it to `ClassObject` because we're retrieving the class itself which will be used as a constructor at [line 290](https://github.com/mozilla-b2g/gaia/pull/30847/files#diff-716c59d65bea2e1456305b9fc9d6caacR290).

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
```diff
@@ -144,97 +178,123 @@ def language_menu_item_description(self):
...
+    def open_help_settings(self):
+        return self._open_subpage_settings(self._help_menu_item_locator)
+
+    def _open_subpage_settings(self, locator, file_name=None, class_name=None):
+        self._tap_menu_item(locator)
+        class_object = self._get_class_by_name(file_name, class_name)
+        if class_object is not None:
+            return class_object(self.marionette)
+
+    def _get_class_by_name(self, file_name, class_name):
```
JohanLorenzo: I'd put `file_name=None, class_name=None` here too, to make the intent clearer within this function.

tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
```diff
@@ -75,16 +80,23 @@ def take_screenshot(self):
         self.marionette.set_context(self.marionette.CONTEXT_CONTENT)
 
         # obtain reference filename
```
JohanLorenzo: I'm not sure this comments provides more value to the line below.

tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
```diff
@@ -130,6 +142,12 @@ def image_compare(self, target_img, ref_img, diff_img, fuzz_value):
             err = err.replace('\n', '')
             message = 'WARNING: ' + err + ' pixels mismatched between ' + target_img + ' and ' + ref_img + '\n'
 
+            # move the diff image and the target image to a separate folder
+            if not os.path.exists(self.mismatch_path):
+                os.makedirs(self.mismatch_path)
+            os.rename(target_img, os.path.join(self.mismatch_path, target_img[target_img.rfind("/")+1:]))
```
JohanLorenzo: It might worth to extract `target_img.rfind("/")+1` in a variable. I'm not sure what it should represent.

tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
```diff
@@ -164,6 +182,7 @@ def edge_scroll(self, frame, direction, dist, release=True):
         if release:
             action.release()
         action.perform()
+        time.sleep(1)  # compensate for the time taken for actual action
```
JohanLorenzo: We're waiting for the dynamic scroll to end, is that right?

tests/python/gaia-ui-tests/gaiatest/tests/graphics/RTL/test_settings_RTL_POC.py
```diff
@@ -0,0 +1,173 @@
...
+
+from gaiatest.gaia_graphics_test import GaiaImageCompareTestCase
+from gaiatest.apps.settings.app import Settings
+from marionette_driver import By
+
+
+class TestSettingsRTL(GaiaImageCompareTestCase):
+    _page_locator = (By.ID, 'root')
+
+    def setUp(self):
```
JohanLorenzo: We don't need this setUp anymore

tests/python/gaia-ui-tests/gaiatest/tests/graphics/RTL/test_settings_RTL_POC.py
```diff
@@ -0,0 +1,173 @@
...
+        GaiaImageCompareTestCase.setUp(self)
+
+    def test_settings_app(self):
+
+        settings = Settings(self.marionette)
+        settings.launch()
+
+        self.take_screenshot('main')
+
+        self.scroll(self._page_locator, 'down', 600)
```
JohanLorenzo: I think we can move this method in Settings. This will prevent the locator to be in the test.

tests/python/gaia-ui-tests/gaiatest/tests/graphics/RTL/test_settings_RTL_POC.py
```diff
@@ -0,0 +1,173 @@
...
+        self.take_screenshot('main')
+        self.scroll(self._page_locator, 'down', 600)
+        self.take_screenshot('main')
+        self.scroll(self._page_locator, 'down', 600)
+        self.take_screenshot('main')
+        self.scroll(self._page_locator, 'down', 600)
+        self.take_screenshot('main')
+
+        settings.open_wifi_settings()
```
JohanLorenzo: Let's remove the duplication.
``` python
for subpage in ('wifi', 'sim_manager', ...):
   open_method = settings.__getattribute__('open_{}_settings'.format(subpage))
   open_method()
   self.take_screenshot(subpage)
   settings.return_to_main_menu()
```

Side note: `settings.open_*_settings` sounds redundant. @npark-mozilla how do you feel about just call these method `open_*`? (It might be better to do it in a follow up)

tests/python/gaia-ui-tests/gaiatest/tests/graphics/RTL/test_settings_RTL_POC.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def test_settings_app(self):
+
+        settings = Settings(self.marionette)
+        settings.launch()
+
+        self.take_screenshot('main')
+
+        self.scroll(self._page_locator, 'down', 600)
+        self.take_screenshot('main')
```
JohanLorenzo: I think it's better to remove the duplication here, too:
``` python
for i in range(1, 5):
    # we perform this action 5 times in order to cover the whole length of the settings page
    self.scroll(self._page_locator, 'down', 600)
    self.take_screenshot('main')
```

tests/python/gaia-ui-tests/gaiatest/tests/graphics/RTL/test_settings_RTL_POC.py
```diff
@@ -0,0 +1,173 @@
...
+
+        settings.open_do_not_track_settings()
+        self.take_screenshot('donottrack')
+        settings.return_to_main_menu()
+
+        settings.open_browsing_privacy_settings()
+        self.take_screenshot('browsing_privacy')
+        settings.return_to_main_menu()
+
+        # This is a link to separate app
```
JohanLorenzo: In this case, I don't think we have to cover it in the Settings tests. Let's just put a comment saying why Privacy Control is not here

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
```diff
@@ -144,97 +178,123 @@ def language_menu_item_description(self):
     def bluetooth_menu_item_description(self):
         return self.marionette.find_element(*self._bluetooth_text_locator).text
 
-    def a11y_open_accessibility_settings(self):
-        from gaiatest.apps.settings.regions.accessibility import Accessibility
-        self._a11y_click_menu_item(self._accessibility_menu_item_locator)
-        return Accessibility(self.marionette)
+    def open_wifi_settings(self):
```
JohanLorenzo: That looks :+1: ! We could reduce the amount of functions defined by defining them dynamically, but that would be overkill for now :)

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
npark-mozilla: hmm, in that case i should rename it to class_definition

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
npark-mozilla: fixed

tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
npark-mozilla: true, removed

tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
npark-mozilla: changed the code to make it more readable

tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
npark-mozilla: yes, there is no check that can tell me the scroll action (and the rendering of screen) has completed.

tests/python/gaia-ui-tests/gaiatest/tests/graphics/RTL/test_settings_RTL_POC.py
npark-mozilla: Actually we need to run GaiaImageCompareTestCase.setUp(self), and I'd prefer to run it in the setUp() method since it's not part of the test.

tests/python/gaia-ui-tests/gaiatest/tests/graphics/RTL/test_settings_RTL_POC.py
npark-mozilla: fixed

tests/python/gaia-ui-tests/gaiatest/tests/graphics/RTL/test_settings_RTL_POC.py
npark-mozilla: awesome, keep forgetting that python can do this.  yes, I'll remove _settings suffix as well.

npark-mozilla: wait, if I do this, then opening all submenu methods should also reside in settings.py.  Hmm, then still it's settings task should be okay.

tests/python/gaia-ui-tests/gaiatest/tests/graphics/RTL/test_settings_RTL_POC.py
npark-mozilla: ahh, one small thing with this: if we want to find _screen_locator, then I have to move to the root frame.

npark-mozilla: hmm, for me, the total_height = comes out as 539.33, while the screen_height comes as 569.  The page locator in Settings is (By.ID, 'root')  I don't think it draws the entire menu list and compute its height - i think it draws as you scroll, so it would be difficult to get the total height.

tests/python/gaia-ui-tests/gaiatest/tests/graphics/RTL/test_settings_RTL_POC.py
npark-mozilla: Ah, if we do this way, then it might be difficult to scroll within each of the subpage in the future, because some are scrollable while others are not.  And because of the above issue about being unable to get the total page height, we might have to put if statements within the loop that manually specifies the scroll iterations anyway. So although I much rather prefer to do it your method, I think we have keep it this way for the sake of simplicity.  what do you think?

tests/python/gaia-ui-tests/gaiatest/tests/graphics/RTL/test_settings_RTL_POC.py
npark-mozilla: done

tests/python/gaia-ui-tests/gaiatest/tests/graphics/RTL/test_settings_RTL_POC.py
JohanLorenzo: I agree, `GaiaImageCompareTestCase.setUp()` needs to run. Nevertheless, `TestSettingsRTL` inherits from `GaiaImageCompareTestCase` so, if you don't re-define `setUp()` here, Python knows it has to call `GaiaImageCompareTestCase.setUp()`

tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
JohanLorenzo: Okay, I'd add these details in the comment, for future readers.

tests/python/gaia-ui-tests/gaiatest/tests/graphics/RTL/test_settings_RTL_POC.py
JohanLorenzo: For the longer subpages, I guess we can extract them from the loop, and do the specific job, when we'll come to this issue. Even if only 30% of the subpages were good to be in this loop, that would give us 10 methods saved, that is to say 30 lines not duplicated.

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
```diff
@@ -30,30 +35,59 @@ class Settings(Base):
...
 
     def launch(self):
         Base.launch(self)
         Wait(self.marionette).until(
             expected.element_present(*self._app_loaded_locator))
 
+        # this is located at the end of the page.  If this is shown, everything is rendered.
+        Wait(self.marionette).until(
+            expected.element_present(*self._bluetooth_l10n_locator))
+
```
mwargers: If you do this, the previous Wait for *self._app_loaded_locator is not needed anymore, right?

tests/python/gaia-ui-tests/gaiatest/tests/graphics/RTL/test_settings_RTL_POC.py
JohanLorenzo: Sounds great to me!

tests/python/gaia-ui-tests/gaiatest/tests/graphics/RTL/test_settings_RTL_POC.py
npark-mozilla: ah, didn't know that. fixed.

tests/python/gaia-ui-tests/gaiatest/tests/graphics/RTL/test_settings_RTL_POC.py
npark-mozilla: So in order to extract screenshots from the loop, we have to explicitly mention when to scroll x times.  Then we'll need to put an if statement for each of them as below:
for subpage in ('wifi', 'sim_manager', ...):
   open_method = settings.__getattribute__('open_{}_settings'.format(subpage))
   open_method()
  if subpage == 'wifi':
      self.take_screenshot(subpage)
      self.scroll(subpage)
      self.take_screenshot(subpage)
  if subpage == 'sim_manager':
      self.take_screenshot(subpage)
      self.scroll(subpage)
      self.take_screenshot(subpage)
      self.scroll(subpage)
      self.take_screenshot(subpage)

   settings.return_to_main_menu()

And above is just opening 1 level of submenu.  We also need to open submenus within the submenu (For example, Settings -> Device Information -> More information or Settings -> Accessibility -> Color Filter)  For the person who writes the actual script, wouldn't it be easier to follow the workflow if we don't use loop?

tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
npark-mozilla: done

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=42a750c6d03a9c853800e82f6fbfe120fc8880c3

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=eb49f9d33e50bdfa63049506d5b84ab5704787cb

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a0c6468a775168dc36cb3a292f6d108cfc4759d4

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
npark-mozilla: I think so, the _app_loaded_locator can be visible while the _bluetooth_l10n_locator isn't, but not the other way around.  I'm not sure whether I should change the launch method though

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
```diff
@@ -1,7 +1,6 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
-
```
mwargers: You need to keep this line, no?

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/device_info.py
```diff
@@ -20,6 +29,10 @@ def __init__(self, marionette):
             expected.element_displayed(*self._model_locator))
 
     @property
+    def screen(self):
```
mwargers: Wouldn't it be better to call this screen_element?
I mean, calling it .screen makes it sound like the Screen object or something like that.

tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
```diff
@@ -48,17 +50,20 @@ def tearDown(self):
...
-    def take_screenshot(self):
-        """invokes screen capture event, crops the status bar, and saves to the file"""
+    # hook to override default locale setting
+    # see https://github.com/mozilla-b2g/gaia/blob/master/tests/python/gaia-ui-tests/gaiatest/gaia_test.py#L1019
+    def modify_settings(self, settings):
+        settings['language.current'] = self.locale
+        return settings
 
-        time.sleep(2)  # compensate for the time taken for actual action on previous step
+    def take_screenshot(self,page_name = None):
```
mwargers: Space before page_name

tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
```diff
@@ -48,17 +50,20 @@ def tearDown(self):
         """At the end of test execution, it checks for the errors"""
         self.assertTrue(self.test_passed, msg=self.failcomment)
 
-    def take_screenshot(self):
-        """invokes screen capture event, crops the status bar, and saves to the file"""
+    # hook to override default locale setting
+    # see https://github.com/mozilla-b2g/gaia/blob/master/tests/python/gaia-ui-tests/gaiatest/gaia_test.py#L1019
```
mwargers: This url will be outdated once that file has changed, right?

tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
mwargers: What if there is no locale set? Would that overwrite the language.current setting that might be already there?

tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
```diff
@@ -109,30 +120,40 @@ def take_screenshot(self):
...
+                os.makedirs(self.mismatch_path)
+            target_image_name = target[target.rfind("/")+1:]
+            diff_image_name = diff[diff.rfind("/") + 1:]
+            os.rename(target, os.path.join(self.mismatch_path, target_image_name))
+            os.rename(diff, os.path.join(self.mismatch_path, diff_image_name))
 
         return message
 
-    def edge_scroll(self, frame, direction, dist, release=True):
+    # Make UI action related methods static, so they can be used outside the GaiaImageCompareTestCase object as well.
```
mwargers: I don't understand. Why are these methods in gaia_graphics_test.py then? Why not in gaia_test.py?

tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
mwargers: So page_name is a unique name you can give to your screenshot files? Perhaps add that as a comment as explanation?

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/device_info.py
```diff
@@ -35,8 +48,40 @@ def tap_more_info(self):
...
+
+    def exit_update_frequency(self):
+        self.marionette.find_element(*self._update_ok_button_locator).tap()
+        self.marionette.switch_to_frame(self.device_info_frame)
+        Wait(self.marionette).until(expected.element_displayed(
+            self.marionette.find_element(*self._page_locator)))
+
+    def tap_reset_phone(self):
+        self.marionette.find_element(*self._reset_button_locator).tap()
+        #self.device_info_frame = self.marionette.get_active_frame()
```
JohanLorenzo: Let's remove this commented code.

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/device_info.py
```diff
@@ -35,8 +48,40 @@ def tap_more_info(self):
...
+        Wait(self.marionette).until(expected.element_displayed(
+            self.marionette.find_element(*self._reset_confirm_locator)))
+
+    def confirm_reset(self, response=True):
+        if response is True:
+            self.marionette.find_element(*self._reset_confirm_locator).tap()
+        else:
+            self.marionette.find_element(*self._reset_cancel_locator).tap()
+            Wait(self.marionette).until(expected.element_displayed(self.screen))
+        #self.device_info_frame = self.marionette.get_active_frame()
```
JohanLorenzo: Let's remove this commented code.

tests/python/gaia-ui-tests/gaiatest/tests/graphics/RTL/test_settings_RTL_POC.py
```diff
@@ -0,0 +1,180 @@
...
+        
+        settings.open_media_storage()
+        self.take_screenshot('media_storage')
+        settings.return_to_prev_menu(settings.screen)
+        
+        settings.open_application_storage()
+        self.take_screenshot('application_storage')
+        settings.return_to_prev_menu(settings.screen)
+
+        ### Device Information and its sub pages
```
JohanLorenzo: I don't think this comment adds some values to the lines below.

tests/python/gaia-ui-tests/gaiatest/tests/graphics/RTL/test_settings_RTL_POC.py
```diff
@@ -0,0 +1,180 @@
...
+        device_info_page = settings.open_device_info()
+        self.take_screenshot('dev_info')
+        GaiaImageCompareTestCase.scroll(self.marionette, 'down', device_info_page.screen.size['height'],
+                                        screen=device_info_page.screen)
+        self.take_screenshot('dev_info')
+
+        # Enter more information dialog
```
JohanLorenzo: I don't think this comment adds some values to the lines below.

tests/python/gaia-ui-tests/gaiatest/tests/graphics/RTL/test_settings_RTL_POC.py
```diff
@@ -0,0 +1,180 @@
...
+        moreinfo_page = device_info_page.tap_more_info()
+        self.take_screenshot('dev_info-more_info')
+        GaiaImageCompareTestCase.scroll(self.marionette, 'down', device_info_page.screen.size['height'],
+                                        screen=moreinfo_page.screen)
+        self.take_screenshot('dev_info-more_info')
+        settings.return_to_prev_menu(device_info_page.screen)
+
+        # Enter reset phone dialog
```
JohanLorenzo: I don't think this comment adds some values to the lines below.

tests/python/gaia-ui-tests/gaiatest/tests/graphics/RTL/test_settings_RTL_POC.py
```diff
@@ -0,0 +1,180 @@
...
+        device_info_page.tap_reset_phone()
+        self.take_screenshot('dev_info-reset')
+        device_info_page.confirm_reset(False)
+
+        # Entering system view, need to switch frame
```
JohanLorenzo: I don't think this comment adds some values to the lines below.

tests/python/gaia-ui-tests/gaiatest/tests/graphics/RTL/test_settings_RTL_POC.py
```diff
@@ -0,0 +1,180 @@
...
+        device_info_page.tap_update_frequency()
+        self.take_screenshot('dev_info-update-freq')
+        device_info_page.exit_update_frequency()
+        settings.return_to_prev_menu(settings.screen)
+
+        ### Downloads page
```
JohanLorenzo: I don't think this comment adds some values to the lines below.

tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
npark-mozilla: You're right, if the locale is not set, i should make it so it uses the testvar value.

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
npark-mozilla: thanks!

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/device_info.py
npark-mozilla: you're right, should change that

tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
npark-mozilla: added

tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
npark-mozilla: true, I should remove the reference to the code

tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
npark-mozilla: bit of a historical reason, initially we didn't want to modify the existing code when we were adding imagecompare to gaia.  I might add this to gaia_test.py later, but right now there are a number of marionette bugs associated with these UI actions - for example, pinch method does not work in browser.  So perhaps we should wait until these Actions derived methods become a bit more stable.

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/device_info.py
npark-mozilla: didn't see that, thanks!

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/device_info.py
npark-mozilla: again! sorry about that.

tests/python/gaia-ui-tests/gaiatest/tests/graphics/RTL/test_settings_RTL_POC.py
npark-mozilla: I thought once we populate this script with the sub-page actions, it would become a bit harder to see where each settings menu is tested.  I thought I could make it more discernable by putting a line with #'s after each of them, do you think it's necessary?

tests/python/gaia-ui-tests/gaiatest/tests/graphics/RTL/test_settings_RTL_POC.py
npark-mozilla: true, it should be evident from the code. will remove.

tests/python/gaia-ui-tests/gaiatest/tests/graphics/RTL/test_settings_RTL_POC.py
npark-mozilla: same reason as adding '### Device Information and its sub pages' comment - should we put a clear divider between accessing each menu for readability?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=45c348d6c3ad881fd6ddfaf285b2a1dac3ad0134

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
```diff
@@ -30,30 +36,59 @@ class Settings(Base):
...
+    _achievements_menu_item_locator = (By.ID, "achievements-section")
+
+    _firefox_accounts_menu_item_locator = (By.ID, "menuItem-fxa")
+    _findmydevice_locator = (By.ID, 'menuItem-findmydevice')
+
+    _screen_lock_menu_item_locator = (By.ID, 'menuItem-screenLock')
+    _app_permission_menu_item_locator = (By.ID, 'menuItem-appPermissions')
     _do_not_track_menu_item_locator = (By.ID, 'menuItem-doNotTrack')
+    _browsing_privacy_item_locator = (By.ID, 'menuItem-browsingPrivacy')
+    _privacy_controls_menu_item_locator = (By.ID, 'menuItem-privacyPanel')
```
mwargers: This one is only mentioned once, it's not used currently?

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
```diff
@@ -30,30 +36,59 @@ class Settings(Base):
...
+    _keyboard_menu_item_locator = (By.ID, "menuItem-keyboard")
+    _theme_menu_item_locator = (By.ID, "menuItem-themes")
+    _addon_menu_item_locator = (By.ID, "menuItem-addons")
+    _achievements_menu_item_locator = (By.ID, "achievements-section")
+
+    _firefox_accounts_menu_item_locator = (By.ID, "menuItem-fxa")
+    _findmydevice_locator = (By.ID, 'menuItem-findmydevice')
+
+    _screen_lock_menu_item_locator = (By.ID, 'menuItem-screenLock')
+    _app_permission_menu_item_locator = (By.ID, 'menuItem-appPermissions')
```
mwargers: This one is only mentioned once, it's not used currently?

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
```diff
@@ -144,97 +179,131 @@ def language_menu_item_description(self):
...
+    def open_developer(self):
+        return self._open_subpage(self._developer_menu_item_locator)
+
+    def open_improve(self):
+        return self._open_subpage(self._improve_menu_item_locator)
+
+    def open_help(self):
+        return self._open_subpage(self._help_menu_item_locator)
+
+    def _open_subpage(self, locator, file_name=None, class_name=None):
```
mwargers: I see a relationship between file_name and class_name.
Perhaps we could only use class_name in this function and then use that string, decapitalize it and use '_' where appropriate, so you get the class_name string automatically?

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/device_info.py
```diff
@@ -35,8 +48,36 @@ def tap_more_info(self):
...
+        self.marionette.find_element(*self._update_frequency_locator).tap()
+        self.device_info_frame = self.marionette.get_active_frame()
+        self.marionette.switch_to_frame()
+        Wait(self.marionette).until(expected.element_displayed(
+            self.marionette.find_element(*self._update_ok_button_locator)))
+
+    # When leaving the page, return to the saved application frame
+    def exit_update_frequency(self):
+        self.marionette.find_element(*self._update_ok_button_locator).tap()
+        self.marionette.switch_to_frame(self.device_info_frame)
```
mwargers: I guess this is necessary, but why doesn't self.apps.switch_to_displayed_app() not work here?

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
npark-mozilla: I thought about parsing the file name to get the class name, but since this implicitly assumes that we would always follow this convention, which might not be true.  I didn't want to make an assumption when it involves filesystem names, just to be safe.

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
npark-mozilla: I use it at  def open_app_permissions(self) method

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
npark-mozilla: true, removed.

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/device_info.py
npark-mozilla: Last time I used it it was somewhat unreliable. I'll try it, and if it works, I'll change to that.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6f84a78f8a1aa329d8df37ce2ddfff10c54392c3

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4d4ace3eb8e4dce60fc328d468890ba006bc117d

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
mwargers: I understand.

tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
```diff
@@ -48,17 +51,26 @@ def tearDown(self):
...
-    def take_screenshot(self):
-        """invokes screen capture event, crops the status bar, and saves to the file"""
+    # marionette hook to override default locale setting
+    def modify_settings(self, settings):
 
-        time.sleep(2)  # compensate for the time taken for actual action on previous step
+        if self.locale != 'undefined':
+            settings['language.current'] = self.locale
+        else:
+            settings['language.current'] = DEFAULT_SETTINGS['language.current']
```
mwargers: Ok, that's the place where it goes wrong.
Suppose I have language.current=nl in my testvars.json file, then I think it should be picked up, but because of this, it picks up the default setting instead of the one I've put in my testvars.json file. Perhaps that else: part can be just removed and then it would be working?

tests/python/gaia-ui-tests/gaiatest/mixins/imagecompare.py
```diff
@@ -33,7 +33,12 @@ def __init__(self, **kwargs):
...
         group.add_option('--screenshots-path',
                          default="screenshots",
-                         help="Path for screenshot images, relative to t he current location, "
+                         help="Path for screenshot images, relative to the current location, "
                               "Default folder is %default")
 
+        group.add_option('--locale',
+                         default = "undefined",
+                         help = "locale for the device, "
+                                "When not defined, will use the value from testvar")
```
mwargers: the testvars.json file

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
mwargers: Oh yes, I see it now, thanks

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c796884eabb8444ea4d2fc1e5211d3b4fc4ebfb1

tests/python/gaia-ui-tests/gaiatest/mixins/imagecompare.py
```diff
@@ -33,7 +33,12 @@ def __init__(self, **kwargs):
...
         group.add_option('--screenshots-path',
                          default="screenshots",
-                         help="Path for screenshot images, relative to t he current location, "
+                         help="Path for screenshot images, relative to the current location, "
                               "Default folder is %default")
 
+        group.add_option('--locale',
+                         default = "undefined",
+                         help = "locale for the device, "
+                                "When not defined, will use the value from testvar.json file")
```
mwargers: Nit, it's testvars.json file.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d55b98567126feb1400266020a5b511f8aba659d

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=98ddb1f9e216e9368d8a057d0b4bc48c74f06ec4

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=bba445b2c517b75bde3544c8251543882c0955aa

PR #30850, opened 2015-07-07
Bug 1176527 - Failure in test_bluetooth.py unit test in GaiaDataLayer.enableBluetooth

mwargers:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e8e6e822c3a42b7caefc7d76daaf775c13c43596

tests/atoms/gaia_data_layer.js
```diff
@@ -45,16 +45,17 @@ var GaiaDataLayer = {
 
   disableBluetooth: function() {
     var bluetooth = window.navigator.mozBluetooth;
-    if (bluetooth.enabled) {
+    if (bluetooth.defaultAdapter.state == 'enabled') {
```
JohanLorenzo: I think we need to perform disable() when ['enabling'](https://wiki.mozilla.org/B2G/Bluetooth/WebBluetooth-v2/BluetoothAdapter#BluetoothAdapterState) is happening.

tests/atoms/gaia_data_layer.js
```diff
@@ -45,16 +45,17 @@ var GaiaDataLayer = {
...
       console.log('trying to disable bluetooth');
+      bluetooth.defaultAdapter.disable();
       this.setSetting('bluetooth.enabled', false, false);
       waitFor(
```
JohanLorenzo: We should perform this wait if `disabling` is the current state, too

tests/atoms/gaia_data_layer.js
```diff
@@ -66,16 +67,17 @@ var GaiaDataLayer = {
 
   enableBluetooth: function() {
     var bluetooth = window.navigator.mozBluetooth;
-    if (!bluetooth.enabled) {
+    if (bluetooth.defaultAdapter.state != 'enabled') {
```
JohanLorenzo: `bluetooth.defaultAdapter.state === 'disabled' `+  same comment as above for 'disabling'

tests/atoms/gaia_data_layer.js
```diff
@@ -66,16 +67,17 @@ var GaiaDataLayer = {
...
       console.log('trying to enable bluetooth');
       this.setSetting('bluetooth.enabled', true, false);
+      bluetooth.defaultAdapter.enable();
       waitFor(
```
JohanLorenzo: Same comment as above.

tests/atoms/gaia_data_layer.js
mwargers: Ah, right, so if (bluetooth.defaultAdapter.state != 'disabled') { is probably the solution.

JohanLorenzo: I don't think so. `disabling` also exist and I don't know what could happen if we ask to disable when we're disabling.

tests/atoms/gaia_data_layer.js
mwargers: Yes, which already is the case, no?

tests/atoms/gaia_data_layer.js
mwargers: Why? bluetooth.defaultAdapter.state != 'enabled' does cover all the states we don't want to have, no?

tests/atoms/gaia_data_layer.js
mwargers: I don't understand what you mean.

tests/atoms/gaia_data_layer.js
JohanLorenzo: If the adapter is already in the `disabling` at the time it reaches `if (bluetooth.defaultAdapter.state == 'enabled')`, the wait will never happen

tests/atoms/gaia_data_layer.js
npark-mozilla: So I'm guessing, when bluetooth is being enabled, it goes through the transition of disabled -> enabling -> enabled.  and enabled-> disabling -> disabled the other way.  So if it is in middle of enabling, (but not enabled yet)  we should still wait until it is enabled, then try to disable it in case of line 48.  So if we only check for != 'enabled' state, then it will ignore the disable request of the user.  

I'm wondering whether that actually makes sense or not.  should we ignore the input given during the transition state?  I'm leaning towards no, but you guys might already have reached conclusion since this comment was made 12 hours ago.

ian-liu:
I have some suggestion for the patch.

1. Be careful to access `window.navigator.mozBluetooth.defaultAdapter` because we might get `null` while the platform is creating the instance. In my implementing experience, I sometimes get the `null` value before. It's better to check `window.navigator.mozBluetooth.defaultAdapter`. Then, access `window.navigator.mozBluetooth.defaultAdapter.METHODS`.

2. I find out the script is sill using `window.navigator.mozBluetooth.getDefaultAdapter()`. There is no API `getDefaultAdapter()` in Bluetooth v2. It's changed to `getAdapters()`. And will return array for adapters.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=50ae5f9313033a5f267ec1ba1413eed811563d00

tests/atoms/gaia_data_layer.js
```diff
@@ -43,46 +43,49 @@ var GaiaDataLayer = {
...
+        marionetteScriptFinished(true);
+      },
+      function() {
+        console.log('bluetooth enable status: ' + bluetooth.getAdapters()[0].state);
+
+        // Wait for the adapter state to become 'disabled' or 'enabled'
+        if (adapter.state == 'disabling' || adapter.state == 'enabling')
+          return false;
+
+        var isnotValue = aValue == 'enabled' ? 'disabled' : 'enabled';
```
npark-mozilla: sorry, I'm not familiar with c style conditional operator on python. is this same as below?
if aValue == 'enabled:
   isnotValue = 'disabled'
else
   isnotValue = 'enabled'

if so, can we just check for (adapter.state != aValue) below, instead of checking (adapter.state == isnotValue) ?

mwargers: Ah, you're probably right, I'll do that

JohanLorenzo:
Note: The JSHint job is broken because:
```
tests/atoms/gaia_data_layer.js: line 62, col 84, Line is too long. (ERROR)
tests/atoms/gaia_data_layer.js: line 66, col 11, Expected '{' and instead saw 'return'. (ERROR)
tests/atoms/gaia_data_layer.js: line 72, col 13, Expected '{' and instead saw 'adapter'. (ERROR)
tests/atoms/gaia_data_layer.js: line 74, col 13, Expected '{' and instead saw 'adapter'. (ERROR)
```

tests/atoms/gaia_data_layer.js
```diff
@@ -43,46 +43,49 @@ var GaiaDataLayer = {
...
-          console.log('bluetooth enable status: ' + bluetooth.enabled);
-          return bluetooth.enabled === false;
-        }
-      );
-    }
-    else {
-      console.log('bluetooth already disabled');
+    var adapter = bluetooth.getAdapters()[0];
+
+    if (adapter.state == aValue) {
```
JohanLorenzo: Nit: Use ===, it prevents JavaScript from chaging the type of the values.

tests/atoms/gaia_data_layer.js
```diff
@@ -43,46 +43,49 @@ var GaiaDataLayer = {
...
+    waitFor(
+      function() {
+        console.log('bluetooth ' + aValue);
+        marionetteScriptFinished(true);
+      },
+      function() {
+        console.log('bluetooth enable status: ' + bluetooth.getAdapters()[0].state);
+
+        // Wait for the adapter state to become 'disabled' or 'enabled'
+        if (adapter.state == 'disabling' || adapter.state == 'enabling')
```
JohanLorenzo: Nit: Use ===, it prevents JavaScript from chaging the type of the values.

tests/atoms/gaia_data_layer.js
JohanLorenzo: Nit: Use ===, it prevents JavaScript from chaging the type of the values.

tests/atoms/gaia_data_layer.js
```diff
@@ -43,46 +43,49 @@ var GaiaDataLayer = {
...
+          return false;
+
+        var isnotValue = aValue == 'enabled' ? 'disabled' : 'enabled';
+        if (adapter.state == isnotValue) {
+          console.log('trying to make bluetooth ' + aValue);
+          if (aValue == 'enabled')
```
JohanLorenzo: Nit: Use ===, it prevents JavaScript from chaging the type of the values.

tests/atoms/gaia_data_layer.js
```diff
@@ -43,46 +43,49 @@ var GaiaDataLayer = {
     marionetteScriptFinished(true);
   },
 
-  disableBluetooth: function() {
+  setBluetooth: function(aValue) {
```
JohanLorenzo: `value` is a generic name. Do you think `expectedState` would fit here?

tests/atoms/gaia_data_layer.js
```diff
@@ -43,46 +43,49 @@ var GaiaDataLayer = {
...
+      return;
     }
+
+    waitFor(
+      function() {
+        console.log('bluetooth ' + aValue);
+        marionetteScriptFinished(true);
+      },
+      function() {
+        console.log('bluetooth enable status: ' + bluetooth.getAdapters()[0].state);
```
JohanLorenzo: Why don't we use `adapter` here?

tests/python/gaia-ui-tests/gaiatest/gaia_test.py
```diff
@@ -331,17 +331,17 @@ def bluetooth_disable(self):
 
     @property
     def bluetooth_is_enabled(self):
-        return self.marionette.execute_script("return window.navigator.mozBluetooth.enabled")
+        return self.marionette.execute_script("return window.navigator.mozBluetooth.getAdapters()[0].state === 'enabled'")
```
JohanLorenzo: Maybe we could wrap `window.navigator.mozBluetooth.getAdapters()[0]` in an atom called `getFirstAdapter` and use it in the next 3 following methods. This would avoid the duplication

tests/atoms/gaia_data_layer.js
mwargers: I'll use  `aState` then.  `expectedState` is more for comparing values, while here I'm using an argument.

tests/atoms/gaia_data_layer.js
mwargers: Because I didn't think of doing that, I'll change it.

tests/python/gaia-ui-tests/gaiatest/gaia_test.py
mwargers: Yeah, I could make it  `getDefaultAdapter` and then query for defaultAdapter if it exist and if it doesn't exist, use the first adapter in the getAdapters() list.

tests/atoms/gaia_data_layer.js
JohanLorenzo: I think ` expectedState` everywhere is fine. It's more semantic.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1fd19bfac2aecc4f29f74b5efce22c34c254441e

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a3ad93cc943cef587a97dfcae6cc6e805e3a5e22

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5f20a68e5b63059418265de4b573a602e6dfc854

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1ab112716f81a3ece698d5f784fb87631face7c1

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e9775dba027c5ea6f0c2e782ddeed5c84aa9452c

PR #30853, opened 2015-07-07
WIP - Add code transformer for Raptor tests

snowmantw:
APP=system DEBUGUY=<where the debuguy is> TESTDIR=<where the esp files are> make


autolander:

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=33215ec83edf0a963214d0204bfddc1d6c5d7ce7

build/post-app.js
```diff
@@ -14,6 +14,8 @@ exports.execute = function(options, webapp) {
 
   nodeHelper.require('copy-build-stage-data', options);
 
+  require('./raptor-test-transformer').execute(options, webapp);
```
rickychien: ```nodeHelper.require('multilocale', options);```

rickychien: require('./raptor-test-transformer').execute(options);

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9ee9e22370d6f32095517f3ac1336ddbb9cd6731

build/post-app.js
```diff
@@ -14,6 +14,11 @@ exports.execute = function(options, webapp) {
 
   nodeHelper.require('copy-build-stage-data', options);
 
+  // If we need to transform code for Raptor tests
+  if ('1' === options.RAPTOR_TRANSFORM) {
+    require('./raptor-test-transformer').execute(options, webapp);
```
rickychien: We've already assign webapp to options at https://github.com/snowmantw/gaia/blob/raptoraop/build/post-app.js#L7
so you can access it within the options object

apps/system/lockscreen/raptor/lockscreen.esp.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+module.exports = function(filepath, outpath, requiredir) {
+  var opts = {};
+  if (outpath) {
+    var fs = require('fs');
```
rickychien: Why don't you import module on the top of script? I prefer to do it since it's very clear to see what modules are used in that file.

apps/system/lockscreen/raptor/lockscreen.esp.js
```diff
@@ -0,0 +1,23 @@
...
+    opts.writer = {
+      write: function(advice) {
+        var code = advice.code;
+        fs.writeFile(outpath, code);
+      }
+    };
+  }
+  var Espect = require(requiredir + 'espect.js');
+  var espect = new Espect(opts);
```
rickychien: same here

snowmantw: In this case you need to wait the requiredir pass in, which is not available at the top of file.

apps/system/test/raptor/lockscreen.esp.js
```diff
@@ -0,0 +1,43 @@
...
+  }
+  var Espect = require(requiredir + 'espect.js');
+  var espect = new Espect(opts);
+  espect
+    .select(filepath + ' LockScreen.prototype.unlock')
+    .before(function() {
+      Performance.mark('LockScreen#unlock');
+    })
+    .done()
+  .done();
```
rickychien: I'm just curious about why you have to invoke twice done() here.

apps/system/lockscreen/raptor/lockscreen.esp.js
snowmantw: If you don't need that module in some cases, to require it is only wasting of time. I can take a look to see if the condition is necessary: it's related to the internal logic of Debuguy + Espect in my recollection.

apps/system/test/raptor/lockscreen.esp.js
snowmantw: It's an API from Espect: the first `done` is for the `select`, and the second one is for the whole process constructed here. It's weird only when you select single file to query. If you have multiple selection in one process it would be like:

    espect()
       select()
       ....
       .done()
       .select()
       .done()
    .done()

snowmantw: And the question is just like as to not duplicate two ending angle brackets at HTML. Well, I don't have any better idea now.

package.json
```diff
@@ -4,14 +4,13 @@
...
   "author": "The Gaia Team <dev-gaia@lists.mozilla.org>",
-
   "engineStrict": "true",
   "engines": {
     "node": ">=0.10",
     "npm": ">=2"
   },
-
   "devDependencies": {
+    "SMS": "file:./apps/sms",
```
nullaus: Not sure why you'd want to move this, it was alphabetized.

snowmantw: It's a mistake. Would be fixed.

PR #30864, opened 2015-07-07
Bug 1181136 - [Messages] When sending a MMS in airplane mode, the dis…

julienw:
…play stays in "sending" state after the error r=schung

### DIFF
--- a/apps/sms/views/conversation/js/conversation.js
+++ b/apps/sms/views/conversation/js/conversation.js
@@ -79,6 +79,7 @@ var ConversationView = {
   shouldChangePanelNextEvent: false,
   showErrorInFailedEvent: '',
   previousSegment: 0,
+  buildingMessages: {},
 
   timeouts: {
     update: null,
@@ -856,8 +857,9 @@ var ConversationView = {
     // Update the stored thread data
     Threads.registerMessage(message);
 
-    this.appendMessage(message);
-    TimeHeaders.updateAll('header[data-time-update]');
+    return this.appendMessage(message).then(
+      () => TimeHeaders.updateAll('header[data-time-update]')
+    );
   },
 
   /**
@@ -1768,11 +1770,14 @@ var ConversationView = {
 
     var result;
     if (message.type === 'mms' && !isNotDownloaded && !noAttachment) { // MMS
+      this.buildingMessages[message.id] = messageDOM;
+
       result = SMIL.parse(message).then(
         (slideArray) => this.createMmsContent(slideArray)
-      ).then(
-        (mmsContent) => pElement.appendChild(mmsContent)
-      ).then(
+      ).then((mmsContent) => {
+        pElement.appendChild(mmsContent);
+        delete this.buildingMessages[message.id];
+      }).then(
         () => messageDOM
       );
     } else {
@@ -2452,7 +2457,8 @@ var ConversationView = {
   },
 
   setMessageStatus: function(id, status) {
-    var messageDOM = document.getElementById('message-' + id);
+    var messageDOM =
+      document.getElementById('message-' + id) || this.buildingMessages[id];
 
     if (!messageDOM || messageDOM.classList.contains(status)) {
       return;
--- a/apps/sms/views/conversation/test/unit/conversation_test.js
+++ b/apps/sms/views/conversation/test/unit/conversation_test.js
@@ -1543,183 +1543,213 @@ suite('conversation.js >', function() {
       );
     }
 
-    setup(function(done) {
-      fakeMessage = MockMessages.sms({
+    setup(function() {
+      this.sinon.stub(Navigation, 'isCurrentPanel');
+      this.sinon.spy(ConversationView, 'appendMessage');
+    });
+
+
+    test('onMessageSent for MMS', function(done) {
+      fakeMessage = MockMessages.mms({
         id: 24601,
         delivery: 'sending'
       });
 
-      this.sinon.stub(Navigation, 'isCurrentPanel').
-        withArgs('thread', { id: fakeMessage.threadId }).
-        returns(true);
-
-      this.sinon.spy(ConversationView, 'appendMessage');
+      Navigation.isCurrentPanel.withArgs(
+        'thread', { id: fakeMessage.threadId }
+      ).returns(true);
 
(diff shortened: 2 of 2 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5fb6164a422ed7ba7f896a4b5fb5cb3bb4b197d6

apps/sms/views/conversation/js/conversation.js
```diff
@@ -815,8 +816,9 @@ var ConversationView = {
     // Update the stored thread data
     Threads.registerMessage(message);
 
-    this.appendMessage(message);
-    TimeHeaders.updateAll('header[data-time-update]');
+    return this.appendMessage(message).then(
+      () => TimeHeaders.updateAll('header[data-time-update]')
```
julienw: I'm not sure this call is actually needed... but I'm not really sure where it's done otherwise

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=887b89ceea6cf832eb2c0af1017a22b3e93978a3

apps/sms/views/conversation/test/unit/conversation_test.js
```diff
@@ -1585,183 +1585,213 @@ suite('conversation.js >', function() {
...
+          fakeMessage.deliveryInfo = [
+            {receiver: null, readStatus: 'success'},
+            {receiver: null, readStatus: 'pending'}];
+          MessageManager.on.withArgs('message-read').yield({
+            message: fakeMessage
+          });
 
-        assertMessageStatus('sending');
+          assertMessageStatus('sending');
+        });
```
julienw: mostly an indentation change

apps/sms/views/conversation/test/unit/conversation_test.js
```diff
@@ -1585,183 +1585,213 @@ suite('conversation.js >', function() {
...
-      ConversationView.appendMessage.returnValues[0].then(() => {
+      MessageManager.on.withArgs('message-failed-to-send').yield({
+        message: fakeMessage
+      });
+
+      ConversationView.appendMessage.lastCall.returnValue.then(() => {
         container = document.getElementById('message-' + fakeMessage.id);
+        assertMessageStatus('error');
       }).then(done, done);
     });
```
julienw: this is the real new test

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2aa8311937dd266a3019b26a7fc99b3722759893

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1713,14 +1715,24 @@ var ConversationView = {
...
+      messageReadyPromise = Promise.resolve();
     }
 
-    return Promise.resolve(messageDOM);
+    return messageReadyPromise.catch(
+      (e) => console.error('Error while building the message DOM', e)
+    ).then(
+      () => delete this.buildingMessages[message.id]
+    ).then(
+      () => messageDOM
```
julienw: The idea is to store the currently building message so that we can access it in setMessageStatus even if it's not added to the DOM yet.

steveck-chung: Is it necessary to split `delete this.buildingMessages[message.id]` and `messageDOM` into set later one into next `then` method? It seems could be done in the first promise resolve.

apps/sms/views/conversation/js/conversation.js
```diff
@@ -2403,6 +2415,10 @@ var ConversationView = {
   setMessageStatus: function(id, status) {
     var messageDOM = document.getElementById('message-' + id);
 
+    if (!messageDOM) {
```
steveck-chung: Maybe `var messageDOM = document.getElementById('message-' + id) || this.buildingMessages[id];` ?

apps/sms/views/conversation/js/conversation.js
julienw: I thought it was clearer this way, and I could use only simple arrow functions instead of using {} and return :) but bot work for me

apps/sms/views/conversation/js/conversation.js
julienw: I'm not sure this fits in one line, but maybe on 2 lines it can be more readable indeed

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ab046940ca07977b24a73ec9e3f188e989c696ad

apps/sms/views/conversation/js/conversation.js
steveck-chung: hmm... maybe it's for updating the headers in inbox view. Not sure we really need the updateAll here either but we definitely need to revisit this time update mechanism once we encapsulate the view separately.

apps/sms/views/conversation/js/conversation.js
steveck-chung: ok, it's fine for me

apps/sms/views/conversation/test/unit/conversation_test.js
steveck-chung: Wondering if it's better to assert the this.buildingMessages is added and cleared since we don't want the unexpected binding and cached for DOM element after appended.

apps/sms/views/conversation/js/conversation.js
steveck-chung: BTW I was thinking do we need to return DOM even the mms parsing is failed. The original implementation didn't return it, but returning the DOM here looks better because we can still return the incomplete message DOM with error message.

apps/sms/views/conversation/test/unit/conversation_test.js
julienw: yep good idea

apps/sms/views/conversation/js/conversation.js
julienw: yes I think just like you here :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2460000368b75652ad4fb68b03fc8f6c283372be

PR #30899, opened 2015-07-09
Bug 1181787 - Create nodejs util for parsing and transforming app htm…

lambdabaa:
…l files


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2e6c9bd341a24e031164b5eb78bb2993bc2377c6

tools/html_transformer/html_transformer.js
```diff
@@ -0,0 +1,95 @@
...
+
+exports.load = function(filename) {
+  return new Promise(function(resolve, reject) {
+    jsdom.env(filename, function(errors, window) {
+      if (errors && errors.length) {
+        return reject(errors[0]);
+      }
+
+      exports.window = window;
+      return resolve();
```
eliperelman: I don't know if it's just me, but this feels a little weird. What if instead of resolving with `window` on `exports` asynchronously, you return an instance of an object that refers to the window, etc. Example:

```js
var Transformer = function(window) {
  this.window = window;
};

Transformer.prototype.getSharedDependencies = function() {};

exports.load = function(filename) {
  // blah, blah, blah
  return resolve(new Transformer(window));
};
```

gaye: Yeah that seems fine. In general, I try to only do classes when I know that we'll want to be able to run multiple in the same js context simultaneously since @millermedeiros has convinced me that having your own custom data structures makes it more difficult for others to dive into your codebase

tools/html_transformer/html_transformer.js
```diff
@@ -0,0 +1,95 @@
+var Promise = require('es6-promise').Promise;
+var jsdom = require('jsdom');
+
+exports.window = null;
+
+exports.load = function(filename) {
+  return new Promise(function(resolve, reject) {
+    jsdom.env(filename, function(errors, window) {
+      if (errors && errors.length) {
+        return reject(errors[0]);
```
nullaus: I hope this doesn't mean we can only see one error at a time. This would make fixing the errors a little bit frustrating by having to run N times for N errors present before resolving all of them.

tools/html_transformer/html_transformer.js
```diff
@@ -0,0 +1,95 @@
...
+};
+
+// Neither document.createTreeWalker nor document.createNodeIterator
+// are supported by jsdom so we're going to do this the old fashioned way...
+function getComments(doc) {
+  var results = [];
+  var queue = [doc.body];
+  while (queue.length) {
+    var next = queue.shift();
+    if (next.nodeType === 8 /* COMMENT_NODE */) {
```
nullaus: I believe there's a constant you can access on Node for this instead of hardcoding to '8'. Something like next.nodeType === Node.COMMENT_NODE iirc.

tools/html_transformer/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "html-transformer",
+  "version": "0.0.1",
+  "description": "Figures out app shared deps, localizes your strings, other mischief",
+  "main": "html_transformer.js",
+
+  "dependencies": {
+    "es6-promise": "^2.3.0",
```
rickychien: Node.js already supported native Promise after 0.12.x

tools/html_transformer/test/html_transformer_test.js
```diff
@@ -0,0 +1,27 @@
...
+  test('#getSharedDependencies', function() {
+    expect(htmlTransformer.getSharedDependencies())
+      .to
+      .deep
+      .equal([
+        '/shared/style/toolbars.css',
+        'shared/js/component_utils.js',
+        'shared/elements/gaia_subheader/script.js',
+        'shared/style/tabs.css',
+        '/shared/js/gesture_detector.js'
```
rickychien: For this case of ```href="app://theme.gaiamobile.org/shared/elements/gaia-theme/style.css"``` found it at https://github.com/mozilla-b2g/gaia/blob/167a22b2fcba06f9d09472bbaadd877252ef709a/apps/calendar/index.html#L48

It should be included and parse correctly.

tools/html_transformer/html_transformer.js
rickychien: For maintenance and readability, maybe we can write it like this:
``` javascript
module.exports = {

  window: null,

  load: function() {
   ...
  },

  getSharedDependencies: function() {

  },

  _getSharedElements: function() {
    ...
  }

}
```

tools/html_transformer/test/html_transformer_test.js
gaye: Ah good catch

tools/html_transformer/package.json
gaye: Yeah still not sure about which node we want people to be on

tools/html_transformer/html_transformer.js
gaye: Oh nice

tools/html_transformer/html_transformer.js
gaye: Yeah maybe I'll somehow combine them

tools/html_transformer/html_transformer.js
gaye: I tend to prefer less indentation which is why I usually go for the `exports` style thing rather than `module.exports` when I'm exporting multiple things

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1aa154604ac6cc67134a0033ebb6d4f54e3e90d1

tools/html_transformer/html_transformer.js
gaye: I actually think I won't be able to access that given how jsdom works

gaye: Er scratch that I can do `window.Node.COMMENT_NODE`

gaye: But then I'd have to pass a window into this thing just for that so I might just leave this number magical

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ebacbec5b528a4728c430dbb1b0c3e118a11d8bb

tools/html_transformer/html_transformer.js
```diff
@@ -0,0 +1,105 @@
...
+      if (errors && errors.length) {
+        return reject(errors[0]);
+      }
+
+      return resolve(new HTMLTransformer(window));
+    });
+  });
+};
+
+function HTMLTransformer(window) {
```
snowmantw: I would like to add some comments about what this component is for, and what it would do at the beginning of a "class".

tools/html_transformer/html_transformer.js
```diff
@@ -0,0 +1,105 @@
...
+  this.window = window;
+}
+exports.HTMLTransformer = HTMLTransformer;
```
snowmantw: And for me it would be nice to collect all exports at the bottom of file, not scattering like this.

tools/html_transformer/html_transformer.js
```diff
@@ -0,0 +1,105 @@
...
+  });
+};
+
+// Neither document.createTreeWalker nor document.createNodeIterator
+// are supported by jsdom so we're going to do this the old fashioned way...
+function getComments(doc) {
+  var results = [];
+  var queue = [doc.body];
+  while (queue.length) {
+    var next = queue.shift();
```
snowmantw: I would prevent to modify the queue while iteration is going, which is a basic defensive pattern to prevent errors.

gaye: How would you perform a breadth first search without modifying the queue?

tools/html_transformer/html_transformer.js
gaye: Sure I can go for that

tools/html_transformer/package.json
gaye: Right now everything in gaia runs on 0.10

tools/html_transformer/html_transformer.js
snowmantw: Okay, I've done some children walking like this with for + function calls (not recursion). But I think in this code your expression is clear and small enough, so I think it's fine for me to keep it here.

PR #30928, opened 2015-07-12
Bug 965211 - Render keyboard according to uppercase button status

nblumoe:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b403232739c116d9d6c8584630f0501c07976658

apps/keyboard/js/views/layout_page_view.js
```diff
@@ -137,13 +137,6 @@ LayoutPageView.prototype.render = function render() {
     content.appendChild(kbRow);
   }).bind(this));
 
-  // If this layout does not require different rendering for lowercase state,
-  // we default to uppercase rendering -- this class will tell CSS file to
-  // never toggle button label <span> elements.
-  if (!layout.secondLayout) {
```
timdream: Not sure if we should remove all `secondLayout` property in the layout file. Maybe do it in a follow-up patch when this patch merged and sticks?

PR #30937, opened 2015-07-13
Bug 1183101 - Adds AdvancedTelemetry module to track performance metr…

tamarahills:
…ics and custom app metrics from Gecko.

### DIFF
--- a/apps/sharedtest/test/unit/telemetry_test.js
+++ b/apps/sharedtest/test/unit/telemetry_test.js
@@ -229,4 +229,26 @@ suite('Telemetry:', function() {
       });
     });
   });
+
+  /*
+   * Test that the getSettings() utility function works as expected.
+   * This is required for proper configuration of the module and for
+   * gathering the settings that are sent along with usage data.
+   */
+  suite('getSettings():', function() {
+    suiteSetup(function() {
+      TelemetryRequest.DEBUG = false; // Shut up console output in test logs
+      TelemetryRequest.getSettings;
+
+      var mockSettings = MockNavigatorSettings.mSettings;
+      mockSettings.x = '1';
+      mockSettings.y = '2';
+    });
+
+    test('getSettings()', function(done) {
+      TelemetryRequest.getSettings({x: '3', y: '4', z: '5'}, function(result) {
+        done(assert.deepEqual(result, {x: '1', y: '2', z: '5'}));
+      });
+    });
+  });
 });
--- a/apps/system/js/advanced_telemetry.js
+++ b/apps/system/js/advanced_telemetry.js
@@ -1,0 +1,414 @@
+/**
+ * This module is responsible for managing the AdvancedTelemetry features in
+ * Gaia. Advanced Telemetry includes:
+ *  1) Engineering Performance metrics on a per app basis that are collected
+ *  by the HUD engine in gecko.
+ *  2) Custom App Metrics.  Apps can define their custom metrics and have them
+ *  collected in a histogram as well.  This API is defined via the
+ *  AdvancedTelemetryHelper.
+ *
+ * Metrics are gathered in gecko and requested from AdvancedTelemetry Module
+ * when it is time to ship the metrics to the server depending on the send
+ * interval desired.
+ */
+
+/* global asyncStorage, SettingsListener, uuid, TelemetryRequest */
+(function(exports) {
+  'use strict';
+
+  // This is the asyncStorage key we use to persist our device ID
+  const DEVICE_ID_KEY = 'metrics.advanced_telemetry.deviceID.v4';
+
+  // Constants for events we register/unreg for
+  const IDLE = 'idle';
+  const ACTIVE = 'active';
+  const HUDMETRICS = 'advanced-telemetry-update';
+
+
+  // Constant for debugging lines
+  const AT_DEBUG_PREFIX = '[AdvancedTelemetry]';
+
+  // This is the list of event types we register handlers for
+  const EVENT_TYPES = [
+    IDLE,
+    ACTIVE,
+    HUDMETRICS
+  ];
+
+
+  // This AdvancedTelemetry() constructor is the value we export from
+  // this module. This constructor does no initialization itself: that
+  // is all done in the start() instance method. See bootstrap.js for
+  // the code that actually calls the constructor and the start method.
+  function AdvancedTelemetry() {}
+
+  // We use the acronym AT internally to save typing and make
+  // references to class variables and constants more legible.
+  const AT = AdvancedTelemetry;
+
+  AT.BatchTiming = BatchTiming;
+
+  // This the ping type in the v4 telemetry spec.  Exposed for unit testing.
+  AT.REASON = 'advancedtelemetry';
+
+  // asyncStorage key that stores the BatchTiming start time.
+  AT.BATCH_KEY = 'metrics.advanced_telemetry.data.v4';
+
+  // Set to true to to enable debug output.  Not recommended to turn this
+  // on except while doing deep debugging.
+  AT.DEBUG = false;
+
+  // LOGINFO method allows for debugging of initialization events and
+  // packet sending events and is at a higher level then DEBUG.
+  AT.LOGINFO = false;
+
+  function debug(...args) {
+    if (!AT.DEBUG) {
+      return;
+    }
+
+    var iCount = 0;
+    // Get the total length of all arguments passed
+    args.forEach(function (value) {
+      iCount += value.length;
+    });
+    // if the length is greater than what adb console can handle, chunk it up.
+    if (iCount > 1000) {
+      var longLine = '';
+      args.forEach(function (value) {
+        longLine += value;
+      });
+      // Chop it into chunks of what adb console can handle.
+      var array = longLine.match(/.{1,1000}/g);
+      // Output each chunk to the console
+      array.forEach(function (value) {
+        console.log(AT_DEBUG_PREFIX + value);
+      });
+    } else {
+      // Its < what ADB can handle so just output it normally.
+      args.unshift(AT_DEBUG_PREFIX);
+      console.log.apply(console, args);
+    }
+  }
+
+  // This info level is more of a information level.  It's recommended to
+  // turn this on just to see general events related to initializations,
+  // and sending of payloads, etc.
+  function loginfo(...args) {
+    if (AT.LOGINFO) {
+      args.unshift('[AdvancedTelemetryInfo]');
+      console.log.apply(console, args);
+    }
+  }
+
+  // This value is what turns advanced telemetry on/off
+  AT.TELEMETRY_ENABLED_KEY = 'debug.performance_data.advanced_telemetry';
+
+  // Base URL for sending data reports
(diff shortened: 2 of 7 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=377823f2390f03ad197f52adebfd94c1345c0a14

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b37909dc5bd12b06fe814dee429f8785fe0c108a

apps/system/js/advanced_telemetry.js
```diff
@@ -0,0 +1,426 @@
...
+      args.unshift('[AdvancedTelemetryInfo]');
+      console.log.apply(console, args);
+    }
+  }
+
+  // Debugging function to handle the case where the output length exceeds that
+  // which the adb console can handle.
+  function debuglongLine(longLine) {
+    if (AT.DEBUG) {
+      var array = longLine.match(/.{1,1000}/g);
```
marshall: could the logic of this function be handled internally in debug() ?

apps/system/js/advanced_telemetry.js
```diff
@@ -0,0 +1,426 @@
...
+      return;
+    }
+    this.collecting = true;
+
+    this.metrics = new BatchTiming(true);
+
+    getDeviceID();
+
+    function getDeviceID() {
+      asyncStorage.getItem(DEVICE_ID_KEY, function(value) {
```
marshall: just curious, do you need to reuse the new dogfood ID logic here?

apps/system/js/advanced_telemetry.js
```diff
@@ -0,0 +1,426 @@
...
+        done();
+      }
+    }
+  };
+
+  // This method shuts everything down and is only exposed for unit testing.
+  // Note that this is not the same as the stopCollecting() method which is
+  // used to stop data collection but keep the module running.
+  AT.prototype.stop = function stop() {
+
```
marshall: Should the SettingsListener added in `start()` be removed here?

apps/system/js/advanced_telemetry.js
```diff
@@ -0,0 +1,426 @@
...
+        this.transmit(e.detail);
+        break;
+      default:
+        break;
+    }
+  };
+
+  // Signal gecko to send the histograms with a telemetry message.
+  AT.prototype.getPayload = function getPayload() {
+    console.info('telemetry|MGMT|TIMETOSHIP');
```
marshall: y'all are using the console to communicate to the gecko telemetry backend? :)

apps/system/js/advanced_telemetry.js
```diff
@@ -0,0 +1,426 @@
...
+  };
+
+  // Start a timer to notify when to send the payload.
+  AT.prototype.startBatch = function startBatch() {
+    var self = this;
+    this.timer = setTimeout(function() {
```
marshall: is `self.getPayload()` supposed to be called continuously on an interval or just once? you use `this.interval` as the timeout value, but `self.getPayload()` will only be called once here because you're using `setTimeout()` and not `setInterval()`

apps/system/js/advanced_telemetry.js
```diff
@@ -0,0 +1,426 @@
...
+    // Query the settings db for parameters for the URL
+    AT.getSettings(deviceInfoQuery, function(deviceResponse) {
+      // Note that this wrapper is using the new v4 Unified Telemetry format
+      var wrapper = {
+        type: REASON,
+        id: uuid(),
+        creationDate: new Date().toISOString(),
+        version: AT.TELEMETRY_VERSION,
+        application: {
+          architecture: 'x86',
```
marshall: should this be `arm` (or queried from the OS)?

apps/system/js/advanced_telemetry.js
```diff
@@ -0,0 +1,426 @@
...
+          buildId: deviceResponse['deviceinfo.platform_build_id'],
+          name: AT.TELEMETRY_APP_NAME,
+          version: deviceResponse['deviceinfo.platform_version'],
+          vendor: 'Mozilla',
+          platformVersion: deviceResponse['deviceinfo.platform_version'],
+          xpcomAbi: 'x86-msvc',
```
marshall: I have to assume this is just a placeholder value, since most FxOS devices are using `armeabi` binaries. does the telemetry service care about these values for FxOS?

apps/system/js/advanced_telemetry.js
```diff
@@ -0,0 +1,426 @@
...
+    function send(payload, deviceInfoQuery) {
+      var request = new AdvancedTelemetryPing(payload, deviceInfoQuery,
+        self.deviceID);
+
+      // We don't actually have to do anything if the data is transmitted
+      // successfully. We are already set up to collect the next batch of data.
+      function onload() {
+        loginfo('Transmitted Successfully.');
+        // TODO: Add this line below to AdvancedTelemetryHelper as an API.
+        console.info('telemetry|MGMT|CLEARMETRICS');
```
marshall: is this new console based behavior documented? I'm getting very curious about it now :)

apps/system/js/advanced_telemetry.js
```diff
@@ -0,0 +1,426 @@
...
+      });
+    }
+  };
+
+  function AdvancedTelemetryPing(payload, deviceQuery, did) {
+    if (!payload) {
+      throw new Error('No arguments');
+    }
+    // clone so we don't put data into the object that was given to us
+    this.packet = payload;
```
marshall: This isn't actually a clone, just a reference. If you want to copy the data, the most straight forward way would be:

`this.packet = JSON.parse(JSON.stringify(payload));`

apps/system/js/advanced_telemetry.js
```diff
@@ -0,0 +1,426 @@
...
+  AT.prototype.transmit = function transmit(payload) {
+    var self = this;
+
+    if (!this.collecting) {
+      return;
+    }
+
+    // But assume that it will succeed and start collecting new metrics now
+    this.metrics = new BatchTiming(false);
+    this.startBatch();
```
marshall: it looks like you call `startBatch()` any time the `advanced-telemetry-update` event is sent, but I'm unclear on why you need to send the encoded string `"'telemetry|MGMT|TIMETOSHIP'"` after a timeout from this function. Can you not just do it synchronously?

apps/system/js/advanced_telemetry.js
```diff
@@ -0,0 +1,426 @@
...
+  };
+
+  // Start a timer to notify when to send the payload.
+  AT.prototype.startBatch = function startBatch() {
+    var self = this;
+    this.timer = setTimeout(function() {
+      if (navigator.onLine) {
+        self.getPayload();
+      }
+    }, self.metrics.getInterval());
```
marshall: After reading over this more, It looks like the only reason you're using `BatchInterval` is to hold an interval value to call `setTimeout()` with -- I see no other references to `this.metrics` or `self.metrics` except for initialization. If my other comments about this timeout being unnecessary are true, I think you should be able to get rid of `BatchInterval` all together.

apps/system/js/advanced_telemetry.js
```diff
@@ -0,0 +1,426 @@
...
+  };
+
+  /*
+   * A utility function to get values for all of the specified settings.
+   * settingKeysAndDefaults is an object that maps settings keys to default
+   * values. We query the value of each of those settings and then create an
+   * object that maps keys to values (or to the default values) and pass
+   * that object to the callback function.
+   */
+  AT.getSettings = function getSettings(settingKeysAndDefaults, callback) {
```
marshall: I hate that we have to keep cargo-culting this code, oh well :(

apps/system/js/advanced_telemetry.js
marshall: update: after reading over the proposed gecko patch, I think I understand what you're trying to do here now. I'll be deleting some of my earlier comments :)

apps/system/js/advanced_telemetry.js
tamarahills: Yes, good point.  I will do this.  Thanks.

apps/system/js/advanced_telemetry.js
tamarahills: Yes, but it seems like there may be a change as per discussions in 1171685.  I'm hoping we can just use the same implementations across ftu, aum, and advanced telemetry... So, I'll probably have a follow on for this.

apps/system/js/advanced_telemetry.js
tamarahills: What do you think about putting it in /shared/js/telemetry.js?  That way we would only have one copy of it instead of two.

marshall: Yeah, that is definitely preferable

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e165bd77e98ff1953ae3bf237dce84cfd56c655d

apps/system/js/advanced_telemetry.js
tamarahills: fixed this.

apps/system/js/advanced_telemetry.js
tamarahills: I checked and it doesn't seem like this does anything useful.  I did change it to arm-gcc3 since that is the format for it.

apps/system/js/advanced_telemetry.js
tamarahills: Yes I checked with the telemetry team and we can use the arm for this.  It's one of the dropdowns in the dashboard.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ce812a77da2ea337c647f5053a407af16a552031

apps/sharedtest/test/unit/telemetry_test.js
```diff
@@ -229,4 +229,28 @@ suite('Telemetry:', function() {
...
+   * Test that the getSettings() utility function works as expected.
+   * This is required for proper configuration of the module and for
+   * gathering the settings that are sent along with usage data.
+   */
+  suite('getSettings():', function() {
+    var getSettings;
+
+    suiteSetup(function() {
+      TelemetryRequest.DEBUG = false; // Shut up console output in test logs
+      getSettings = TelemetryRequest.getSettings;
```
marshall: nit: looks like you're saving `getSettings` here, but don't use it below.. I'd say just get rid of it altogether :)

apps/system/js/advanced_telemetry.js
marshall: not sure if you still plan to combine the `debug` and `debuglongLine` functions, but just calling out that it's still here :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=fa736f2961197fdeee41a0c525428019041a8bc5

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=019ca84ff558130586337586059011d7595df43e

PR #30943, opened 2015-07-13
Bug 1183175 - use gmail entry for mozillafoundation.org

asutherland:
Because of ISPDB bug 1177436 we need to copy the gmail.com entry to
mozillafoundation.org like we did for mozilla.com for slightly different
reasons.  (The transition from Zimbra to GMail necessitated it while we
were waiting for DNS changes that would make MX resolution work.)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ea115259ec8e2160bd342226605c51df02d45b85

PR #30952, opened 2015-07-14
Bug 1183566 - [Contacts][NGA] Fix loader.js when path is not specified

jpruden92:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ac79f30ccb0c754fb8e3b4e30b216a352189aeef

apps/communications/contacts/js/loader.js
```diff
@@ -53,7 +54,7 @@
      */
     function doLoad() {
       var name = file.toLowerCase();
-      var path = elementMapping[name][1];
+      var path = elementMapping[name][1] || null;
```
borjasalguero: Check if `elementMapping[name]` exists... I would do something as
`var path;`
`if ( elementMapping[name] && elementMapping[name][1]) {`
`path = elementMapping[name][1];`
`}`

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5573277e69844c01a7d151e6edc7436dd6c291e3

apps/communications/contacts/js/loader.js
```diff
@@ -43,7 +43,8 @@
     search: ['search-view', SHARED_CONTACTS],
     multiple_select: ['multiple-select-view'],
     overlay: ['loading-overlay', SHARED_UTILS],
-    ice: ['ice-view']
+    ice: ['ice-view'],
+    import_sim_contacts: ['', SHARED_UTILS]
```
mancas: Wouldn't be more elegant (and better) to use [null, SHARED_UTILS] instead of an empty string? We can avoid iterate through the DOM looking for a node that doesn't exists. Then in Line 78 we can do something like this:

```javascript
if (elementMapping[name][0]) {
  var node = ...
  toLoad.unshift(node);
}
```

WDYT? @borjasalguero @jpruden92

borjasalguero: I would add `if (elementMapping[name] && elementMapping[name][0]) {`, but looks great!

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=118a95ed829f32de8a774404a7983b71935aea70

PR #30971, opened 2015-07-15
Bug 1182035 - Fixed app updates error in 2G

mcjimenez:
Fixed so when only there are app updates and update.2g enabled is false and device is in 2G, they are not blocked


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=dfca49cd3f06d5f5c69ec458bf17c2547a3c272a

PR #30980, opened 2015-07-15
Bug 942076 - Add events for every possible message state change and use them to update the UI r=julienw

gabrielesvelto:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=351e5671951650f970c0cf9f5218ea09cf81f67b

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e040f9c3b11eb8f1327564e6372ab07bb4a9b03d

apps/wappush/js/messagedb.js
```diff
@@ -76,18 +102,27 @@ var MessageDB = (function() {
...
-        if (event.oldVersion === 0) {
-          var store = db.createObjectStore(MESSAGE_STORE_NAME,
-                                           { keyPath: KEY_PATH });
-          store.createIndex(ID_INDEX, ID_INDEX_KEYPATH, { unique: false });
+        var store = null;
+
+        switch (event.oldVersion) {
+          case 0:
+            store = db.createObjectStore(MESSAGE_STORE_NAME,
+                                         { keyPath: KEY_PATH });
```
julienw: nit: I can see you spent some time in Gecko looking at your indentation style.

Here is what I'm used to, tell me what you think:
```javascript
store = db.createObjectStore(
  MESSAGE_STORE_NAME, { keyPath: KEY_PATH }
);
```
The idea is that I try to never increase indentation by more than 2 spaces.
Especially I don't want to have indentation depend on variable/function names, which is the case here.

apps/wappush/js/messagedb.js
```diff
@@ -76,18 +102,27 @@ var MessageDB = (function() {
...
+            store.createIndex(ID_INDEX, ID_INDEX_KEYPATH, { unique: false });
+            /* falls through */
+          case 1:
+            store = db.createObjectStore(VIEWED_STORE_NAME,
+                                         { keyPath: KEY_PATH });
```
julienw: same here

apps/wappush/js/messagedb.js
```diff
@@ -76,18 +102,27 @@ var MessageDB = (function() {
 
       req.onsuccess = function mdb_opened(event) {
         db = event.target.result;
-        resolve(db);
+        mdb_deleteViewedMessages().then(() => resolve(db));
       };
       req.onerror = function mdb_openError(event) {
         reject(event.target.error);
       };
       req.onupgradeneeded = function mdb_upgradeNeeded(event) {
```
julienw: maybe not in this patch, but as soon as we have more than 3 different versions (0, 1, 2) I think it would be better to move this function to a separate function.

apps/wappush/js/messagedb.js
```diff
@@ -96,35 +131,33 @@ var MessageDB = (function() {
...
     } else {
       store.put(message);
+      mdb_dispatchEvent(updated ? 'update' : 'new', message);
     }
-
-    return status;
   }
 
   /**
    * Adds a messages to the message database according to the out-of-order
```
julienw: nit: "messages" => "message"

apps/wappush/js/messagedb.js
```diff
@@ -96,35 +131,33 @@ var MessageDB = (function() {
...
-   * delivery rules specified in WAP-167 6.2. Returns a promise that resolves
-   * to a string describing the status of the message: 'new' if the message was
-   * new, 'updated' if the message updated an existing message or 'discarded'
-   * if the message was discarded.
+   * delivery rules specified in WAP-167 6.2. Returns a promise that is
+   * resolved when the message has been stored. This will dispatch all the
+   * appropriate events.
```
julienw: nit: a good place to use the jsdoc `@event` and `@fires` :) http://usejsdoc.org/tags-event.html http://usejsdoc.org/tags-fires.html

apps/wappush/js/messagedb.js
```diff
@@ -230,7 +264,8 @@ var MessageDB = (function() {
   function mdb_retrieve(timestamp) {
     return mdb_open().then(function mdb_openResolved(db) {
       var state = { result: null };
-      var transaction = db.transaction(MESSAGE_STORE_NAME, 'readwrite');
+      var transaction =
+        db.transaction([ MESSAGE_STORE_NAME, VIEWED_STORE_NAME ], 'readwrite');
       var promise = mdb_transactionPromise(transaction, state);
```
julienw: if we want to completely kill that `state` parameter (this is the last place where it's used), you can replace `return promise;` at the end of the function by `return promise.then(() => state);`

apps/wappush/js/wappush.js
```diff
@@ -282,49 +266,127 @@
...
+            displayedSiMessage = null;
+            SiSlScreenHelper.hide();
+            CpScreenHelper.showConfirmInstallationDialog(message);
+            return Promise.resolve();
         }
       }
     ).catch(function(error) {
-      console.error('Could not retrieve the message: ' + error.name + '\n');
+      var message = error.message || 'Unknown error';
+      console.error('Could not retrieve the message: ' + message + '\n');
```
julienw: nit: you don't need '\n', and thus you can just separate the prefix from the message with a comma

apps/wappush/test/unit/messagedb_test.js
```diff
@@ -102,196 +106,217 @@ suite('MessageDB', function() {
...
-          assert.equal(message.text, messages.current.text);
-        });
-      }, done);
+    setup(function() {
+      MessageDB.on('new', onNewStub = this.sinon.stub());
     });
 
-    test('message is removed after retrieval', function(done) {
+    teardown(function() {
+      MessageDB.offAll('new');
```
julienw: I'd use `MessageDB.offAll()` in a topmost teardown function

apps/wappush/test/unit/wappush_test.js
```diff
@@ -793,6 +825,10 @@ suite('WAP Push', function() {
       this.sinon.spy(window, 'close');
     });
 
+    teardown(function() {
+      clock.restore();
+    });
```
julienw: nit: you don't need this because we use  `this.sinon.useFakeTimers()` -- it's automatically restored at teardown.

apps/wappush/js/wappush.js
```diff
@@ -115,8 +119,8 @@
     }).catch(function(error) {
       // If we encountered an error don't process messages
       wapPushEnabled = false;
-      error = error || 'Unknown error';
-      console.error('Could not initialize: ', error);
+      var message = error.message || 'Unknown error';
+      console.error('Could not initialize the WAP push manager: ', message);
       return error;
```
julienw: looking at this, I wonder if we should not rethrow the error (so that the promise is still rejected) ?

apps/wappush/js/wappush.js
```diff
@@ -282,49 +266,127 @@
...
+   *
+   * @param {Object} obj The new message in untyped form (i.e. JSON)
+   */
+  function wpm_onUpdate(obj) {
+    DUMP('A message was updated in the DB');
+
+    var message = new ParsedMessage(obj);
+
+    // Update the message if we're currently displaying it
+    if (message.getUniqueId() === displayedSiMessage.getUniqueId()) {
```
julienw: shouldn't you do like in wpm_onDelete and check if `displayedSiMessage` is not null ?

apps/wappush/js/wappush.js
```diff
@@ -282,49 +266,127 @@
...
+   * @param {Object} obj The deleted message in untyped form (i.e. JSON)
+   */
+  function wpm_onDelete(obj) {
+    DUMP('A message was deleted');
+
+    var message = new ParsedMessage(obj);
+
+    // Close the application if we're displaying the deleted message
+    if (displayedSiMessage &&
+      displayedSiMessage.getUniqueId() === message.getUniqueId()) {
```
julienw: nit: I'd indent this by 2 more spaces (just below the previous line's `displayedSiMessage`) so that it's crystal-clear that it's part of the condition and not a line code

apps/wappush/js/wappush.js
```diff
@@ -240,29 +244,9 @@
...
-          wpm_finish();
-        });
-      }
-    }).catch(function wpm_saveRejected(error) {
-      console.log('Could not add a message to the database: ' + error + '\n');
+    return message.save().catch(error => {
+      var message = error.message || 'Unknown error';
+      console.log('Could not add a message to the database: ' + message + '\n');
       wpm_finish();
       return error;
```
julienw: should we rethrow the error here ?

apps/wappush/js/wappush.js
```diff
@@ -282,49 +266,127 @@
...
 
     // Clear the close timer as the application will soon become visible
     app.launch();
     window.clearTimeout(closeTimeout);
     closeTimeout = null;
 
-    return ParsedMessage.load(timestamp).then(
+    return ParsedMessage.load(id).then(
       function wpm_loadResolved(message) {
-        if (message) {
```
julienw: why do you remove the condition ? I'm fine removing the condition if you make it a rejection in ParsedMessage.load or MessageDB.retrieve when we have no message.

apps/wappush/js/wappush.js
```diff
@@ -343,7 +405,8 @@
         }
       },
       function onError(error) {
-        console.error('Notification.get() promise error: ' + error.name);
+        var message = error.message || 'Unknown error';
+        console.error('Could not get notification: ' + message);
         return error;
```
julienw: should we re throw the error ?

apps/wappush/js/parsed_message.js
```diff
@@ -96,7 +93,16 @@
...
+      return (this.expires && (this.expires < Date.now())) || false;
+    },
+
+    /**
+     * Gets the message unique identifier
+     *
+     * @returns {String} An identifier for this message
+     */
+    getUniqueId: function pm_getUniqueId() {
+      return this.timestamp;
```
julienw: I'm still somewhat nervous with using the timestamp as unique id :) but maybe it's good enough

apps/wappush/js/messagedb.js
```diff
@@ -243,7 +278,11 @@ var MessageDB = (function() {
           state.result = message;
```
julienw: should we check whether the message is in the viewed store as well ?

julienw: or is it necessary that we can retrieve a message even if it's in the viewed store so that we can hide it ?

apps/wappush/js/messagedb.js
```diff
@@ -96,35 +131,33 @@ var MessageDB = (function() {
...
     if (message.action === 'delete') {
       mdb_deleteById(store, message.id);
-      status = 'discarded';
+      mdb_dispatchEvent('discard', message);
     } else {
       store.put(message);
+      mdb_dispatchEvent(updated ? 'update' : 'new', message);
     }
-
-    return status;
```
julienw: I don't really like that processMessage dispaches the event. I think it's not its role.
Also the "updated" boolean is quite bad.

How about just returning a boolean (or a string) telling whether the message has been deleted ? And then the caller can do whatever needed to have the proper status.

apps/wappush/js/messagedb.js
gabrielesvelto: Good point, it's awkward for re-factoring and if you have long function names. And yes I did some gecko work lately ;-)

apps/wappush/js/messagedb.js
gabrielesvelto: Yeah, that would be clearer.

apps/wappush/js/messagedb.js
gabrielesvelto: I think this is the easiest solution since it doesn't have any downsides and it doesn't require extra code. Besides if/when WAP Push becomes a component of the SMS app all this thingamagig to delete stuff at the right time because the application is auto-closed will go away.

apps/wappush/js/parsed_message.js
gabrielesvelto: Yeah, it's not pretty and I even thought about changing it but that would require changes to the database and I didn't want to pile even more changes on top of the already existing ones.

apps/wappush/js/wappush.js
gabrielesvelto: Good point.

apps/wappush/js/wappush.js
gabrielesvelto: Yep, same thing

apps/wappush/js/wappush.js
gabrielesvelto: Yes, nice catch.

apps/wappush/js/wappush.js
gabrielesvelto: Yep, will do.

apps/wappush/js/wappush.js
gabrielesvelto: The real reason why I removed it is that we can't have notifications for deleted messages anymore so it shouldn't be possible to call MessageDB.retrieve() on a deleted message anymore. I'll make mdb_retrieve() throw if the message is not found so that it will be harder to misuse it in the future by trying to retrieve messages that don't exist.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=680907158ed24057e5126b84797c2bb9c9e7b670

apps/wappush/js/messagedb.js
gabrielesvelto: Agreed, otherwise it'll get messy.

gabrielesvelto:
Good point, I'll modify ParsedMessage.load() accordingly.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3484014cc88e4fe34bcff9ac02adffaea130c3fd

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9d85fd15a3c52e34977ae7222ee14acb0a54d58f

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4225b1e35c30f325ed00b3f7aed2d74498b24497

PR #30984, opened 2015-07-15
Bug 1183711 - Use the same uuid for ftu ping and AppUsage ping. r=mar…

tamarahills:
…shall.

### DIFF
--- a/apps/system/js/app_usage_metrics.js
+++ b/apps/system/js/app_usage_metrics.js
@@ -54,9 +54,9 @@
   // that it survives across device restarts.
   const PERSISTENCE_KEY = 'metrics.app_usage.data.v2';
 
-  // This is the asyncStorage key we use to persist our device ID
-  // v1 of this ID used a randomly generated String, while v2 uses a UUID
-  const DEVICE_ID_KEY = 'metrics.app_usage.deviceID.v2';
+  // We use the same asyncStorage key as what is used for ftuPing so one
+  // random ID is shared between the two.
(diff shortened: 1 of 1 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c5c013ee32a1d40bb61deabee2a0241bc3e9aac1

apps/system/js/app_usage_metrics.js
```diff
@@ -54,9 +54,9 @@
   // that it survives across device restarts.
   const PERSISTENCE_KEY = 'metrics.app_usage.data.v2';
 
-  // This is the asyncStorage key we use to persist our device ID
-  // v1 of this ID used a randomly generated String, while v2 uses a UUID
-  const DEVICE_ID_KEY = 'metrics.app_usage.deviceID.v2';
+  // We use the same asyncStorage key as what is used for ftuPing so one
+  // random ID is shared between the two.
+  const DEVICE_ID_KEY = 'ftu.pingID';
```
marshall: I don't think this will work exactly the way you want it to. In the AUM service, we delete the generated device ID and metrics when the user opts out of collection -- see https://github.com/tamarahills/gaia/blob/bugfix/1183711-use-same-random-id/apps/system/js/app_usage_metrics.js#L402.

If we just change the source of the generated ID, that would mean the FTU activation would happen again whenever the user opts-out of App Usage Metrics. Maybe we need to add some kind of special logic to account for this?

PR #30991, opened 2015-07-15
Bug 1184246 - Properly set the offset for the IFD data.

hfiguiere:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ad585fb6fab52867f0b95f9432d56b9fea90b9aa

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=dbc2af3491a10a51cb018242e04a4dc29232581d

hfiguiere:
This PR will be deprecated soon.

shared/js/media/jpeg_metadata_parser.js
```diff
@@ -305,7 +312,7 @@ function parseJPEGMetadata(file, metadataCallback, metadataError) {
 
     var total = count * typesize[type];
     var valueOffset = total <= 4 ? offset + 8 :
-      data.getUint32(offset + 8, byteorder);
+        data.getUint32(offset + 8, byteorder) + startOffset;
```
davidflanagan: If I understand correctly, startOffset will always be 10, right?  The code is already full of "+ 10" constants. So I'd rather fix this by just adding a literal 10 here without using a variable.  That seems much simpler than declaring a variable, then later setting its value then still later using the value.

If I'm wrong, and it is possible that sometimes startOffset will be 0 or will be 20 or something, then this code looks good to me.

hfiguiere: Reading another patch I found the + 10 and thought the same. 

The reason I want to keep the +10 here is that the spec is vague about +10. It says "This tags records the offset from the start of the TIFF header to the position where the value itself is recorded." This is mostly "10" when it is an Exif block in a JFIF APP1 segment. So, short term it doesn't matter. Longer term, if we ever have a hardware that support raw files like some Android or Lumia do we'll be happy to support plain TIFF file (includes DNG which is TIFF/EP)

I'll update the PR to take that into account.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=49af8b0d4a6d4e84e6a9e8445d3c4939c6709c05

shared/js/media/jpeg_metadata_parser.js
```diff
@@ -179,6 +184,9 @@ function parseJPEGMetadata(file, metadataCallback, metadataError) {
       throw Error('invalid byteorder in EXIF segment');
     }
 
+    // The TIFF magic started at +10.
+    startOffset += 10;
```
davidflanagan: if parseEXIFData() ever gets called more than once, then startOffset will have a different value each time, which seems wrong.

shared/js/media/jpeg_metadata_parser.js
```diff
@@ -117,7 +121,8 @@ function parseJPEGMetadata(file, metadataCallback, metadataError) {
       var exif = parseEXIFData(data);
 
       if (exif.THUMBNAIL && exif.THUMBNAILLENGTH) {
-        var start = data.sliceOffset + data.viewOffset + 10 + exif.THUMBNAIL;
+        var start = data.sliceOffset + data.viewOffset + startOffset +
```
davidflanagan: From a quick reading, it is easy to think that startOffset is still 0 here and has not been set to 10.  The code is correct, but confusing.  I'd prefer it if you just set startOffset to 10 up above and remove the increment call below.

shared/js/media/jpeg_metadata_parser.js
```diff
@@ -20,6 +20,10 @@ function parseJPEGMetadata(file, metadataCallback, metadataError) {
   // This is the object we'll pass to metadataCallback
   var metadata = {};
 
+  // this is the start offset for the data in exif. Should be
+  // where to find the TIFF magic header.
+  var startOffset = 0;
```
davidflanagan: If this symbol is going to be used throughout, I'd like it to have a more specific name. startOffset is so generic. How about TIFFHeader or TIFFHeaderOffset.

Consider making it a const instead of a variable.  And please just initialize it to 10 instead of starting at 0 and incrementing later.

shared/js/media/jpeg_metadata_parser.js
hfiguiere: Good point. I'll just set it to 10.

shared/js/media/jpeg_metadata_parser.js
hfiguiere: TIFFHeaderOffset it is. I don't agree with the const. For the same argument as "+ 10".

shared/js/media/jpeg_metadata_parser.js
hfiguiere: parseEXIFData() sets it 3 lines above.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4e840b55667087542572e38c8f00d02d7f773bd1

PR #31003, opened 2015-07-16
Bug 1180591 - [Messages][NG] mozSettings shim Implementation, r=julienw, azasypkin

steveck-chung:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d3d980e828d5c4fad984c491d309a2ee04e25eac

apps/sms/services/js/moz_settings/moz_settings_shim.js
```diff
@@ -0,0 +1,74 @@
...
+  },
+
+  /**
+   * Add the Observer and return the changed value via service broadcast.
+   * @param {String} key Key name for Settings.
+   */
+  addObserver(key) {
+    mozSettings.addObserver(key, (e) => {
+      service.broadcast(key, e.settingValue);
+    });
```
julienw: I don't think we need the observer now; I think we just need to get values when we need them.

If we want an addObserver implementation this will need to be more complicated: add only one real observer, and be able to remove observers; because this file may be loaded only once for several clients.

Not sure we really need this complexity, what do you think ?

apps/sms/services/js/moz_settings/moz_settings_shim.js
```diff
@@ -0,0 +1,74 @@
...
+
+var service = null;
+
+var MozSettingsShim = {
+  init(settings) {
+    if (!settings) {
+      return;
+    }
+
+    mozSettings = settings;
```
azasypkin: optional nit: I'd use ```BridgeServiceMixin``` here as well, I've migrated activity shim to this approach, and quite happy so far :)

apps/sms/services/js/moz_settings/moz_settings_shim.js
azasypkin: > I don't think we need the observer now; I think we just need to get values when we need them.

Works for me unless it's hard to migrate to "just need to get values when we need them" :)

> If we want an addObserver implementation this will need to be more complicated: add only one real observer, and be able to remove observers; because this file may be loaded only once for several clients.

Agree, we either make it fully functional (that will be more complex) or skip it for now.

apps/sms/services/js/moz_settings/moz_settings_shim.js
```diff
@@ -0,0 +1,74 @@
...
+  get(key) {
+    return mozSettings.createLock().get(key);
+  },
+
+  /**
+   * Shim for mozSettings.createLock().set API.
+   * @param {Object} setting Settings key and value for .
+   * @return {Promise} Promise that return saved result.
+   */
+  set(setting) {
```
azasypkin: nit: I'd name it ```settings``` to be in line with mozSettingsAPI and say that we can save several settings in one go.

apps/sms/services/js/moz_settings/moz_settings_shim.js
```diff
@@ -0,0 +1,74 @@
...
+   * @return {Promise} Promise that return the settings value from settings DB.
+   */
+  get(key) {
+    return mozSettings.createLock().get(key);
+  },
+
+  /**
+   * Shim for mozSettings.createLock().set API.
+   * @param {Object} setting Settings key and value for .
+   * @return {Promise} Promise that return saved result.
```
azasypkin: typo: same here

apps/sms/services/js/moz_settings/moz_settings_shim.js
```diff
@@ -0,0 +1,74 @@
...
+      service.method(shimMethod, this[shimMethod]);
+    });
+  },
+
+  /* Methods */
+
+  /**
+   * Shim for mozSettings.createLock().get API.
+   * @param {String} key Key name for Settings.
+   * @return {Promise} Promise that return the settings value from settings DB.
```
azasypkin: typo: ```that returns```

apps/sms/services/js/moz_settings/moz_settings_shim.js
steveck-chung: Yeah I considered this change for message shim before. I understand it will work since the worker only API(`importScript`) won't be triggered because libs should be loaded already for the iframe. But should we add some error handling if we do face some problem if the libs could not be loaded correctly for shim(iframe) case in `BridgeServiceMixin`?

apps/sms/services/js/moz_settings/moz_settings_shim.js
steveck-chung: ok, will remove then. It might not be necessary for caching the property.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a5e936ecf71491141ae3cc3fa3542c0f280efe2a

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2386f005dcad201ad0804c479f38fbd933dbfe55

apps/sms/services/js/moz_settings/moz_settings_shim.js
```diff
@@ -0,0 +1,65 @@
...
+   *  several key/value pairs inside.
+   * @returns {Promise} Promise that return saved result.
+   */
+  set(settings) {
+    return mozSettings.createLock().set(settings);
+  }
+};
+
+exports.MozSettingsShim = Object.seal(
+  BridgeServiceMixin.mixin(
```
steveck-chung: Move the question in https://github.com/mozilla-b2g/gaia/pull/31003#discussion_r34971061 to here: Should we do so error handling in BridgeServiceMixin if we want to utilize it for shims.

azasypkin: Could you please clarify a bit, what do you mean by "error handling"?

apps/sms/services/js/moz_settings/moz_settings_shim.js
```diff
@@ -0,0 +1,65 @@
...
+  },
+
+  /* Methods */
+
+  /**
+   * Shim for mozSettings.createLock().get API.
+   * @param {String} key Key name for Settings.
+   * @returns {Promise} Promise that return the settings value from settings DB.
+   */
+  get(key) {
```
julienw: I think `get` should also be able to take an array (maybe handle both cases: array and not array). Notify.js gets 2 settings at once. This is probably not a big deal though (we can do 2 requests with Promise.all) as this case is not performance-sensitive. So as you wish

apps/sms/services/js/moz_settings/moz_settings_shim.js
```diff
@@ -0,0 +1,65 @@
+/* -*- Mode: js; tab-width: 2; indent-tabs-mode: nil; c-basic-offset: 2 -*- /
+/* vim: set shiftwidth=2 tabstop=2 autoindent cindent expandtab: */
```
julienw: we don't need the vim mode

apps/sms/services/js/moz_settings/moz_settings_shim.js
```diff
@@ -0,0 +1,65 @@
...
+ */
+const METHODS = Object.freeze(['get', 'set']);
+
+var mozSettings = null;
+
+var MozSettingsShim = {
+  init(settings) {
+    if (!settings) {
+      return;
+    }
```
julienw: should we return a rejected promise from get/set when we don't have the mozSettings API ?

julienw: or maybe just do as if everything works fine as this will be in desktop mode

apps/sms/services/js/moz_settings/moz_settings_shim.js
```diff
@@ -0,0 +1,65 @@
...
+ * Array of method names that need to be exposed for API shim.
+ * @type {Array.<string>}
+ */
+const METHODS = Object.freeze(['get', 'set']);
+
+var mozSettings = null;
+
+var MozSettingsShim = {
+  init(settings) {
+    if (!settings) {
```
julienw: maybe console.error ?

apps/sms/services/js/moz_settings/moz_settings_shim.js
azasypkin: Honestly I'd not care about this at all, I'd either remove this check or just throw and that's it - it should not happen in real life (on device) and in desktop (we have mock there). Why do you guys think we need this?

apps/sms/services/js/moz_settings/moz_settings_shim.js
```diff
@@ -0,0 +1,65 @@
...
+   * @returns {Promise} Promise that return saved result.
+   */
+  set(settings) {
+    return mozSettings.createLock().set(settings);
+  }
+};
+
+exports.MozSettingsShim = Object.seal(
+  BridgeServiceMixin.mixin(
+    MozSettingsShim,
```
azasypkin: optional nit: looks lit it fits into single line :)

apps/sms/services/test/unit/moz_settings/moz_settings_shim_test.js
```diff
@@ -0,0 +1,108 @@
+/*global MockNavigatorSettings,
```
azasypkin: nit: please alphabetize

apps/sms/services/test/unit/moz_settings/moz_settings_shim_test.js
```diff
@@ -0,0 +1,108 @@
...
+      });
+      this.sinon.stub(MockNavigatorSettings, 'createLock').returns(lockStub);
+      MozSettingsShim.init(MockNavigatorSettings);
+    });
+
+    suite('get >', function() {
+      setup(function() {
+        MozSettingsShim.init(MockNavigatorSettings);
+      });
+      test('returns value successfully', function(done) {
```
azasypkin: nit: new line

apps/sms/services/js/moz_settings/moz_settings_shim.js
steveck-chung: Maybe iframe contains shims forgot to load the bridge or stream, and it will cause shims to load the scripts by importScripts API that not existed in window context. But you still still see the error and figure out the problem anyway, so it's really minor.

apps/sms/services/js/moz_settings/moz_settings_shim.js
steveck-chung: I'm fine to remove it since it won't happen if we initialize currently.

apps/sms/services/js/moz_settings/moz_settings_shim.js
steveck-chung: If it's really not a performance-sensitive case, maybe we can wrap this in settings service or other client side since it's simply API shim here.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=eed29189082a5ed5700519879305aa5ee3cabbb6

apps/sms/services/js/moz_settings/moz_settings_shim.js
azasypkin: > If it's really not a performance-sensitive case, maybe we can wrap this in settings service or other client side since it's simply API shim here.

I agree, it's more client or service (that's one more abstraction level) privilege to enhance API :) I'd do this later only if we see performance penalty for multiple requests.

apps/sms/services/js/moz_settings/moz_settings_shim.js
julienw: ah yeah right we have a mock for mozSettings, I forgot  about it (some time ago we used to work simply by ignoring mozSettings)

apps/sms/services/js/moz_settings/moz_settings_shim.js
azasypkin: > Maybe iframe contains shims forgot to load the bridge or stream, and it will cause shims to load the scripts by importScripts API that not existed in window context. But you still still see the error and figure out the problem anyway, so it's really minor.

Ah, yeah, I don't really care, as you said we'll notice error in console anyway (as I did when was playing with shim_host.html and ConversationService :))

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0fecfab804f9f573b9e2f75272921a3e3c79fb64

apps/sms/services/js/moz_settings/moz_settings_shim.js
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Array of method names that need to be exposed for API shim.
+ * @type {Array.<string>}
+ */
+const METHODS = Object.freeze(['get', 'set']);
+
+var mozSettings = null;
+
+var MozSettingsShim = {
+  init(settings) {
```
julienw: we'll init in a separate startup.js ?

azasypkin: I was thinking about smth like this: https://github.com/mozilla-b2g/gaia/pull/31051/files#diff-0d83de4eba65c6e6dbdf3183d4fcc9f0R6

julienw: yep why not

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2e659fb778291bb629c3c08c19d71ba99669c792

apps/sms/services/js/moz_settings/moz_settings_shim.js
steveck-chung: Yeah I did the same in message shim :)

PR #31037, opened 2015-07-20
Bug 1182071 - Allow System app to fetch Sync keys. r=kgrandon

ferjm:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=716f33f11ea652e26e8b5ec16b3f36a9747ba507

apps/system/js/fx_accounts_client.js
```diff
@@ -82,6 +82,12 @@ var FxAccountsClient = function FxAccountsClient() {
     }, successCb, errorCb);
   };
 
+  var getKeys = function getKeys(successCb, errorCb) {
```
KevinGrandon: I would prefer to make this a promise instead, is that possible?

KevinGrandon: Well after looking again it appears this entire file is callback based, so probably fine for now. Would be nice in the future though.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e33b0745e27635003f277cc494add7986feebb75

PR #31060, opened 2015-07-21
Bug 1186100 - Making 'ls.lock-immediately' lock immediately

cr:
This is a proposed fix for [bug 1186100](http://bugzil.la/1186100) where 'lockscreen.lock-immediately'
as sent by _FindMyDevice_ doesn't lock the device immediately with a pass code
when a pass code timeout has been set.

On unlock, `LockScreen.checkPassCodeTimeout()` only asks for the pass code
when either `.fetchLockedInterval()` or `.fetchUnlockedInterval()` exceed
the pass code timeout.

Setting both `_lastLockedTimeStamp` and `_lastUnlockedTimeStamp` to
something deep in the past will enforce pass code entry on unlock and
avoid a potential race condition.

### DIFF
--- a/apps/system/lockscreen/js/lockscreen.js
+++ b/apps/system/lockscreen/js/lockscreen.js
@@ -354,11 +354,9 @@
       window.SettingsListener.observe('lockscreen.lock-immediately',
         false, (function ls_observeImmediately(value) {
         if (value === true) {
-          this.lock(true);
-          if (this.passCodeEnabled) {
-            // Enforce immediate pass code unlock
-            this.resetTimeoutForcibly();
-          }
+          // Enforce immediate pass code unlock
+          this.resetTimeoutForcibly();
+          window.dispatchEvent(new CustomEvent('lockscreen-request-lock'));
 		}
       }).bind(this));
 
--- a/apps/system/lockscreen/js/lockscreen.js
+++ b/apps/system/lockscreen/js/lockscreen.js
@@ -351,13 +351,15 @@
       // bug 992277. We don't need to use SettingsListener because
       // we're only interested in changes to the setting, and don't
       // keep track of its value.
-      navigator.mozSettings.addObserver('lockscreen.lock-immediately',
-        (function(event) {
-        if (event.settingValue === true) {
+      window.SettingsListener.observe('lockscreen.lock-immediately',
+        false, (function ls_observeImmediately(value) {
+        if (value === true) {
           this.lock(true);
-          // Enforce immediate pass code unlock
-          this.resetTimeoutForcibly();
-        }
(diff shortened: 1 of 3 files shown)


autolander:

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8874b731f2bd0fadc775842c4e1c986b392cfe50

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=54816bd906e8bf9851c437ed8dbf1d24003983d2

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2b533e66a2a148a08faf4038d49156945c41d2f5

apps/system/lockscreen/js/lockscreen.js
```diff
@@ -434,10 +434,14 @@
     // bug 992277. We don't need to use SettingsListener because
     // we're only interested in changes to the setting, and don't
     // keep track of its value.
-    navigator.mozSettings.addObserver('lockscreen.lock-immediately',
-      (function(event) {
-      if (event.settingValue === true) {
+    window.SettingsListener.observe('lockscreen.lock-immediately',
```
lissyx: Please make sure that this is executed only once. This codes lives in the system app, so adding mozSettings observers continuously will leak.

cr: LockScreen registers its settings observers as part of LockScreen.init() which is only ever called once when System starts here:

http://mxr.mozilla.org/gaia/source/apps/system/lockscreen/js/lockscreen_bootstrap.js#27

Do you think this will require deeper verification?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ca266291a40dc1ee4b31fe537a845ec7204242da

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7e82fcb816aaf908a319ed9c182610e8078dc346

apps/system/lockscreen/js/lockscreen.js
```diff
@@ -351,9 +351,13 @@
       // bug 992277. We don't need to use SettingsListener because
```
snowmantw: Since now you use the SettingsListener, you may fix the comment, too.

cr: I don't know who wrote this comment, but it is still valid and should stay. The point it makes is that we should not need to misuse a settings listener for this event-style signalling. I didn't touch that part of the code since I wasn't involved in that design.

PR #31061, opened 2015-07-21
Bug 1186124 - Add "Power button locks immediately" option

cr:
As outlined in [bug 1186124](http://bugzil.la/1186124), this adds a prototype of an
option that immediately requires a pass code after pushing the power button regardless
of the pass code timeout setting.

This is a prototype meant for starting the discussion. No tests, no review, lack of
translations, and perhaps not even the best way to go about this.

### DIFF
--- a/apps/settings/elements/screen_lock.html
+++ b/apps/settings/elements/screen_lock.html
@@ -33,6 +33,19 @@
         <li class="passcode-enabled">
           <button class="passcode-edit" data-l10n-id="change-passcode"></button>
         </li>
+        <li class="passcode-enabled">
+          <p data-l10n-id="powerbutton-behavior">Power button behavior</p>
+          <span class="button icon icon-dialog">
+            <select name="lockscreen.passcode-lock.powerbutton-behavior">
+              <option value="screen-lock" data-l10n-id="powerbutton-screen-lock" selected>
+                Screen lock
+              </option>
+              <option value="passcode-lock" data-l10n-id="powerbutton-passcode-lock">
+                Immediate pass code lock
+              </option>
+            </select>
+          </span>
+        </li>
       </ul>
     </div>
 
--- a/apps/system/js/lock_screen_window_manager.js
+++ b/apps/system/js/lock_screen_window_manager.js
@@ -240,7 +240,15 @@
       // Should be replaced by a proper IAC API in bug 992277.
       var lockListener = (event) => {
         if (true === event.settingValue) {
-          this.openApp();
+          // Ignore when lockscreen is disabled.
+          // Prevents LSWM ending up in a broken, permanently-locked
+          // state when lock-immediately is sent to disabled lockscreen.
+          if (this.states.enabled) {
+            this.openApp();
+          } else {
+            console.warn('LockScreenWindowManager: Ignoring ' +
+              '\'lock-immediately\' on disabled lock screen.');
+          }
         }
       };
 
--- a/apps/system/lockscreen/js/lockscreen_powerbutton_manager.js
+++ b/apps/system/lockscreen/js/lockscreen_powerbutton_manager.js
@@ -1,0 +1,116 @@
+/**
+ Copyright 2015, Mozilla Foundation
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+'use strict';
+
+/**
+ * LockScreen's handling of the power button is implemented here.
+ */
+(function(exports) {
+
+  var LockScreenPowerbuttonManager = function() {
+    /**
+     * Boolean value to track state of locksckscreen.enabled setting.
+     * We must default to false here to ensure that a potential race condition
+     * with the settings observer never results in a permanently locked device.
+     */
+    this.lockScreenEnabled = false;
+    /** String value defining power button behavior. Currently implemented:
+     * - 'screen-lock': default behavior of just locking the screen
+     * - 'passcode-lock': locks and immediately requires pass code on unlock
+     * Only settings listener should change this value to sync with data
+     * in Settings API.
+     */
+    this.buttonBehavior = 'screen-lock';
+  };
+
+  LockScreenPowerbuttonManager.prototype.start =
+    function lspm_start() {
+      window.SettingsListener.observe(
+        'lockscreen.passcode-lock.powerbutton-behavior',
+        'screen-lock', (function(value) {
+          this.setPowerbuttonLockBehavior(value);
+        }).bind(this));
+      window.SettingsListener.observe(
+        'lockscreen.enabled',
+        'screen-lock', (function(value) {
+          this.setLockScreenEnabled(value);
+        }).bind(this));
+      window.addEventListener('sleep', this);
+      return this;
+    };
+
+  /**
+   * The sleep handler hooks the 'sleep' event which is sent by the
+   * key manager when the power button is pressed, and it extends it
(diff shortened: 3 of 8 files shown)


autolander:

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=aeae4393b562c6dce1c868791618377d71a04f48

autolander:

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6552975298afeccb3f590eec2a673b391b69ed77

apps/system/js/screen_manager.js
```diff
@@ -206,6 +206,14 @@
 
         case 'sleep':
           this.turnScreenOff(true, 'powerkey');
+          var lock_enabled = window.lockScreen.enabled;
```
snowmantw: Please note the convention in this file is camel cases. See `this._deviceLightEnabled` above.

apps/system/js/screen_manager.js
```diff
@@ -206,6 +206,14 @@
...
+          var passcode_enabled = window.lockScreen.passCodeEnabled;
```
snowmantw: And it's improper to read another component's status through no the `request` method of System/Service. This would create a tight coupling that hinders us to modify components individually in the future.

snowmantw: You can rather listen the 'sleep' SCM event in `lockscreen.js` to access these properties and do the changes you need in that file. Both `request` and events are for communications among different components, and to decouple them by reducing the direct method calls.

apps/system/lockscreen/js/lockscreen.js
```diff
@@ -754,6 +777,12 @@
     }
   };
 
+  LockScreen.prototype.forcePassCodeUnlock =
```
snowmantw: As I said in Bug 1186100, this should be renamed and changed for a new implementation.

apps/system/lockscreen/js/lockscreen.js
```diff
@@ -64,6 +64,15 @@
...
     passCodeEnabled: false,
 
+	/*
+    * Boolean return whether pushing the power button locks screen
+    * immediately without waiting for lock timeout to pass.
+    * Must not multate directly - use setPassCodeEnabled(val)
+    * Only Settings Listener should change this value to sync with data
+    * in Settings API.
+    */
+    powerButtonLocksEnabled: false,
```
snowmantw: I think it would be better to put your logic into a new LockScreen component under `system/lockscreen/js` and name it as `lockscreen_powerbutton_manager.js` to deal with all these. Although it seems to be a small component, to avoid growing this legacy LockScreen and put each functions in individual components is important. The new component, as a temporary workaround before we could really break all functions into widgets/components of screen lock, can still hold the LockScreen as it's one argument at starting up to get all resources directly. Note this is different from the opinion I left at `screen_manager.js`, which suggest you not access the LockScreen directly, because ScreenManager is totally a different component with different (but coherent) functions from LockScreen, but these sub-components are all related to LockScreen itself.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9d2714e86428151b06283121c547568498aba610

cr:
@snowmantw, your feedback was extremely helpful, and all oddities of the previous solution (which bothered me, too) clicked nicely into a consistent place.The latest push to this PR should come close to what you outlined.

cr:
I will back out the *'lock-immediately'* patch from this PR once it has landed on master, which is a requirement for this feature moving forward. Until then I think it makes more sense to leave it in while review and discussion are still ongoing.

apps/settings/elements/screen_lock.html
```diff
@@ -31,6 +31,15 @@ <h1 data-l10n-id="screenLock-header"></h1>
           </span>
         </li>
         <li class="passcode-enabled">
+          <p data-l10n-id="powerbutton-behavior">Powerbutton behavior</p>
```
timdream: Remove or correct the English string in HTML, as it will be overwritten by l10n property anyway.

cr: I would keep the string, conforming to the convention in this file. Will correct spelling and indent the HTML to avoid long lines.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d5356db16a2a65328f969a3ab8a0aea8b9a7fd90

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=074fcce5a82915f572cb7707d7b380026cacd665

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=932fbefaecc2c208ee73cfb420704cf19a9309b9

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=38addeaccee42b784e1dc3181d994095fbd17756

apps/system/lockscreen/js/lockscreen_powerbutton_manager.js
```diff
@@ -0,0 +1,116 @@
...
+     * - 'passcode-lock': locks and immediately requires pass code on unlock
+     * Only settings listener should change this value to sync with data
+     * in Settings API.
+     */
+    this.buttonBehavior = 'screen-lock';
+  };
+
+  LockScreenPowerbuttonManager.prototype.start =
+    function lspm_start() {
+      window.SettingsListener.observe(
```
snowmantw: Personally I support what Tim said at

https://groups.google.com/forum/#!topic/mozilla.dev.b2g/ZPiHWOSWQR8

So unless I'm moving old code around I would avoid adding new lines using SettingListeners. If it's not necessary, could you change this and to use plain Settings API

apps/system/lockscreen/js/lockscreen_powerbutton_manager.js
```diff
@@ -0,0 +1,116 @@
...
+          }
+      }
+    };
+
+  /**
+   * Value setter for the behavior settings observer. It interprets
+   * everything that is not 'passcode-lock' as the default behavior
+   * 'screen-lock'.
+   */
+  LockScreenPowerbuttonManager.prototype.setPowerbuttonLockBehavior =
```
snowmantw: nit: It's a convention to name a function to handle events/changes as `on-` or `handle-`. There are still some differences between a simple setter and a handler.

apps/system/lockscreen/js/lockscreen_powerbutton_manager.js
```diff
@@ -0,0 +1,116 @@
...
+          this.buttonBehavior = 'screen-lock';
+      }
+    };
+
+  /**
+   * Value setter for the lockscreen.enabled settings observer. It
+   * interprets everything that is not true as its default false.
+   */
+  LockScreenPowerbuttonManager.prototype.setLockScreenEnabled =
+    function lspm_setLockScreenEnabled(value) {
```
snowmantw: I've confirmed that the non-boolean value of this setting value is from the very old days, when the mozSettings still only accept serialized value in strings. So if that's the only concerns I think we can save the code and to treat it as boolean directly.

snowmantw: And there is the naming issue as above.

apps/system/js/lock_screen_window_manager.js
```diff
@@ -240,7 +240,15 @@
       // Should be replaced by a proper IAC API in bug 992277.
       var lockListener = (event) => {
         if (true === event.settingValue) {
-          this.openApp();
+          // Ignore when lockscreen is disabled.
+          // Prevents LSWM ending up in a broken, permanently-locked
+          // state when lock-immediately is sent to disabled lockscreen.
+          if (this.states.enabled) {
```
snowmantw: I think for FMD it need to lock it forcibly no matter whether the enable is true? Or a thief can easily bypass that when the settings is still showing on the screen. And the issue of unable to unlock the phone again after exiting FMD mode as you describe here should be fixed in another bug if we didn't fire one.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f88b0621f58cd63a1f6ad6f511c02a778e0e67c2

PR #31070, opened 2015-07-22
Bug 1044984 - Start seperate worker to run each of the worker.js tests

timdream:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=fa9d4c439260066dad6b01b6e80d7dc66576b4de

apps/keyboard/test/unit/imes/latin/worker_test.js
```diff
@@ -451,22 +448,13 @@ suite('Latin worker', function() {
...
-
-        occ = worker.onmessage.callCount;
-
-        if (filter !== undefined && ev.data.cmd !== filter) {
-          return;
-        }
-
-        clearInterval(iv);
-        callback(ev.data);
+    worker.onmessage = function handleWorkerMessage(evt) {
```
julienw: what do you think of using `addEventListener('message', ...)` instead of `onmessage` ?

apps/keyboard/test/unit/imes/latin/worker_test.js
```diff
@@ -451,22 +448,13 @@ suite('Latin worker', function() {
...
+      if (filter !== undefined && evt.data.cmd !== filter) {
+        return;
       }
-    }, 20);
+
```
julienw: Also, what do you think about unregistering the handler when we receive the right message here ? `worker.removeEventListener('message', handleWorkerMessage)` if you follow my suggestion of using `addEventListener`, or `worker.onmessage = null` otherwise.

julienw: it all depends whether you want to receive the message only once for tests; it seems that's what we do here :)

apps/keyboard/test/unit/imes/latin/worker_test.js
```diff
@@ -451,22 +448,13 @@ suite('Latin worker', function() {
...
+      callback(evt.data);
```
julienw: last but not least: what do you think of `onWorkerMessage` returning a promise instead of using a callback ? Maybe not in this patch though :)

apps/keyboard/test/unit/imes/latin/worker_test.js
```diff
@@ -527,14 +515,23 @@ suite('Latin worker', function() {
...
         }
       }
     };
   }
 
+  test('Worker should throw if no dictData is passed', function(done) {
+    worker.postMessage({ cmd: 'setLanguage', args: ['en_us', null] });
+    onWorkerMessage('error', function(data) {
+      assert.equal(data.cmd, 'error');
+      done();
```
julienw: nit: use `done(() => assert.equal(data.cmd, 'error'))` so that if there is an error it is properly caught instead of having a timeout

julienw: it's not standard mocha, it's an extension we have for a long time in the test-agent (and that mocha should probably adopt :) I need to take the time to upstream this eventually)

julienw: it would also be easier if we are Promise based:
```javascript
onWorkerMessage('error').then(
  () => assert.equal(data.cmd, 'error')
).then(done, done);
```

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=93fce727be2819aa31be9a178e03f36f1c15b27f

PR #31073, opened 2015-07-22
Bug 1177849 - [Gallery] Editing a photo, Exposure slider is an improper graphic

punamdahiya:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d0ea5a92be7b238449b0c679f0d63b8e39012968

pivanov:
r+

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d8fa22811d39e8cfac2ae1b1492f04efea7373c9

apps/gallery/style/gallery.css
```diff
@@ -851,6 +851,7 @@ a.filter-faded {
   height: 2.6rem;
   -moz-margin-start: calc(3.9rem - 1.8rem);
   background: url(images/slider-scrubber.png) no-repeat center center;
+  background-size: 100% auto;
```
davidflanagan: I had to look this up to see why you needed two values.  According to MDN, you can just use the 100% and the second value will be auto by default.  Consider putting the size specification in the background attribute for simplicity.  I think that might be "center/100% center", but I'm not positive. Maybe simpler to keep background-size separate.

Do you understand why the lack of background-size was causing it to come out the wrong size? Is this a bug that needs to be fixed elsewhere in our code?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=51ebcbdfd130ae208b46023003a54ec4c5f1ae67

apps/gallery/style/gallery.css
punamdahiya: Updated patch to include backgound-size in the same line by specifying 100%.  Lack of 100% width dimension to background-size was causing big background image to get truncated in the containing box. 
I will file a bug to look into images picked at 2.25 device pixel ratio for Aries where Aries window.devicepixelratio is  2.

PR #31078, opened 2015-07-23
Bug 1160049 - [Messages] Attach menu should not dismiss when user cancel the replace attachment request, r=azasypkin

steveck-chung:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8f2f7632586ab5618c4602817230163db439fe27

apps/sms/views/conversation/js/compose.js
```diff
@@ -733,7 +733,14 @@ var Compose = (function() {
           dom.message.removeChild(currentAttachmentDOM);
           disposeAttachmentNode(currentAttachmentDOM);
           onContentChanged(newAttachment);
-        }, this._onAttachmentRequestError);
+        }, (err) => {
+          var errId = err instanceof DOMError ? err.name : err.message;
```
azasypkin: Well, honestly I don't really like this solution due to following reasons:

1. When user taps on Cancel it's quite noticeable that menu was hidden and re-appeared that looks a bit clumsy;
2. It doesn't handle the case when activity was cancelled from the activity app. In this case ```pick cancelled``` error is returned (it would be great to confirm this case with UX as well, but not sure whom we can ask about it :( Is Morpheus our new UX?);
3. And it just looks too hackish to manually call ```onAttachmentClick``` in error handler.

Maybe we can modify OptionMenu instead? I was thinking about ```hideOnAction``` option (let's say for the entire OptionMenu at first, but can probably be on item level as well), which default value is ```true``` (to not introduce breaking change). In our case we just pass this pass ```options.hideOnAction: false``` and hide menu manually in all these cases. It would be easier if ```hideOnAction``` was on item option level, but I'm not sure if it's too much.

What do you think? Maybe better ideas exist as well.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d74bf5c3c5fc3a57639ee5241cc1d3ae8281bb1b

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c7665d87ffea45c5625dbe3cbf41014070bf2be8

apps/sms/views/conversation/js/compose.js
steveck-chung: I changed the implementation to set the hideOnAction per item, that menu will not hide after method triggered when hideOnAction is true.

apps/sms/views/conversation/js/compose.js
```diff
@@ -656,9 +656,12 @@ var Compose = (function() {
...
 
-      //'pick cancelled' is the error thrown when the pick activity app is
-      // canceled normally
-      } else if (errId !== 'ActivityCanceled' && errId !== 'pick cancelled') {
+      // Some possible 'activity cancel' error name:
+      // 'pick cancelled': error thrown when the pick activity canceled in most
+      //   of the apps.
+      // 'cancelled': error thrown when the pick activity canceled in contacts.
+      // 'ActivityCanceled': error thrown when pick activity canceled from menu.
+      } else if (!/cancel/i.test(errId)) {
```
steveck-chung: There's something really messy about our cancel :/  The error message is defined by pick apps but not sync with the unique 'ActivityCanceled', and I've seen pick cancelled(most of the apps)/pick 'canceled' (wallpaper and some other apps)/cancelled (contacts only). Maybe we should file a bug to unify the naming of the activity canceled.

shared/js/option_menu.js
```diff
@@ -169,10 +169,13 @@ var OptionMenu = function(options) {
...
       method = action.method || function() {};
 
-      method.apply(null, action.params || []);
+      method(...(action.params || []));
 
-      // Hide action menu when click is received
-      this.hide();
+      // Hide action menu automatically when hideOnAction does not set to true.
+      if (!action.hideOnAction) {
+        console.log('@@@ hide');
```
azasypkin: __reminder:__ don't forget to remove this ```console.log``` before landing.

__question:__ mm the name ```hideOnAction``` and what we do in reality is a bit confusing :) Should not it be like this?:
```
{ 
	l10nId: `replace-attachment-${fileType}`,
	method: onReplace,
	hideOnAction: false
}
.....

var hideOnAction = typeof action.hideOnAction === 'boolean' ? action.hideOnAction : true;
if (hideOnAction) {
  hide();
}
```

azasypkin: Or maybe just give ```hideOnAction``` oposite name that reflects ```please _don't_ hide dialog on dialog action``` :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b5f13de39c2e404cc26bc37423f7dfc7f23c8e87

apps/sms/views/conversation/js/compose.js
azasypkin: Yeah, I remember we were discussing in some bug - ```activity.cancel```, maybe we also need ```activityRequest.cancel``` in addition to ```activityRequest.postError```. For now and for us it's fine like you did I think.

apps/sms/views/conversation/js/compose.js
azasypkin: __question:__ I'm wondering if we should hide replace menu in case of error at all? I see previously if you e.g. choose video and then will try to replace it with too-big one - you'll see error dialog and menu will stay open. It's edgy case, but I'm not sure what is better from UX perspective, could you please confirm with Morpheus if you have him handy? :) But anyway, I won't block on this.

PR #31082, opened 2015-07-23
Bug 1186893 - keyboard.is_keyboard_displayed is always returning true in current builds

mwargers:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=702a9bd9c0b5248f653ccf7fe1643dadee2fa74d

tests/python/gaia-ui-tests/gaiatest/apps/keyboard/app.py
```diff
@@ -156,18 +156,10 @@ def _current_input_mode(self):
...
-        expected_y = -System(self.marionette).software_buttons_height
-
-        Wait(self.marionette).until(
-            lambda m: 'active' in input_window.get_attribute('class') and input_window.location['y'] == expected_y,
-            message='Keyboard inputWindow not interpreted as displayed. Debug is_displayed(): %s, class: %s.'
-            % (input_window.is_displayed(), input_window.get_attribute('class')))
+        Wait(self.marionette).until(lambda m: self.is_keyboard_displayed is True)
 
         keybframe = self.marionette.find_element(*self._keyboard_active_frame_locator)
-        return self.marionette.switch_to_frame(keybframe, focus)
```
npark-mozilla: so no one uses this return object i suppose?

tests/python/gaia-ui-tests/gaiatest/apps/keyboard/app.py
```diff
@@ -335,15 +327,7 @@ def dismiss(self):
...
-        input_window = self.marionette.find_element(*self._input_window_locator)
-
-        # if we have software buttons, keyboar's y will not be 0 but the minus height of the button container.
-        expected_y = int(input_window.size['height']) - System(self.marionette).software_buttons_height
-
-        Wait(self.marionette).until(
-            lambda m: 'inactive' in input_window.get_attribute('class') and
-            not input_window.is_displayed() and
-            (int(input_window.location['y']) == expected_y))
+        Wait(self.marionette).until(lambda m: self.is_keyboard_displayed is False)
```
npark-mozilla: So this would return false when only one of the conditions is true, I guess that would be sufficient to consider the keyboard is not visible from the screen?

tests/python/gaia-ui-tests/gaiatest/apps/keyboard/app.py
mwargers: No, see: http://mxr.mozilla.org/gaia/search?string=switch_to_keyboard
Also see: http://mxr.mozilla.org/gaia/search?string=switch_to_frame&find=&findi=&filter=^[^\0]*%24&hitlimit=&tree=gaia
I believe switch_to_frame doesn't return anything at all.

tests/python/gaia-ui-tests/gaiatest/apps/keyboard/app.py
mwargers: In my opition input_window.is_displayed() should be enough in is_keyboard_displayed to consider the keyboard not visible from the screen.
I don't really why that other condition is necessary.

tests/python/gaia-ui-tests/gaiatest/apps/keyboard/app.py
npark-mozilla: ah I see, weird they did that before then.

tests/python/gaia-ui-tests/gaiatest/apps/keyboard/app.py
npark-mozilla: Hmm, if our assumption is wrong, we'll find out since we check for both. should be okay.

PR #31094, opened 2015-07-24
Bug 1184865 - [Messages][NG] Replace some methods in MessageManager with messaging service in conversation view, r=julienw,azasypkin

steveck-chung:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=79c63447702b6e702af8128efcdaf2b4e74251ca

apps/sms/lib/sww.js
```diff
@@ -25,16 +25,19 @@ Router.prototype.methods = ['get', 'post', 'put', 'delete', 'head',
  * @param handler (Function) payload to be executed if url matches.
```
azasypkin: I suppose this file won't be included in the patch, but leave it here just as a reminder :)

apps/sms/services/js/messaging/messaging_service.js
```diff
@@ -12,7 +13,7 @@ if (!('BridgeServiceMixin' in self)) {
 (function(exports) {
 
 const SERVICE_NAME = 'messaging-service';
-const MOBILE_MESSAGE_CLIENT_NAME = 'mozMobileMessageShim';
+const MOBILE_MESSAGE_CLIENT_NAME = 'moz-mobile-message-shim';
```
azasypkin: __note:__ it doesn't really matter, but I don't feel confident about dash-based shim name, I'd keep it camelCase as we do for other services/shims. I suspect it was done to reuse it for the ```BroadcastChannel``` name, but we'll likely have more complex channel name like ```appInstanceId +'moz-mobile-message-shim'``` to avoid collisions between several app instances.... And with current approach it will look a bit monstrous :) - ```appInstanceId + MOBILE_MESSAGE_CLIENT_NAME + '-channel'```. But it's up to you in the end.

apps/sms/services/js/moz_mobile_message/moz_mobile_message_shim.js
```diff
@@ -16,90 +16,85 @@
...
-
-    service.plugin(streamService);
+    this.initService(endPoint);
 
-    EVENTS.forEach((event) => {
+    Object.keys(EVENTS).forEach((event) => {
       mozMobileMessage.addEventListener(
         event,
-        this[Utils.camelCase(`on-${event}`)]
+        this[Utils.camelCase(`on-${event}`)].bind(this)
```
azasypkin: __optional nit:__ Btw, when I was creating shim_host.html I found out that this file depends on ~760LoC Utils (should be less when minified though) file just for the ```camelCase``` function :) So I'd rather avoid using Utils here unless it's totally required, but up to you.

apps/sms/services/js/bridge_service_mixin.js
```diff
@@ -26,7 +26,7 @@
    * EVENTS, METHODS and STREAMS are all optional.
    */
   var serviceMixin = {
-    initService() {
+    initService(channel) {
```
azasypkin: __recommendation:__ I'd rather name it ```endpoints``` and support array since service can listen on multiple endpoints, especially if shims hosted in single Window - want to talk to each other without using ```BroadcastChannel```. So we can have smth like this:
```
(endpoints || [self]).forEach((endpoint) => service.listen(endpoint));
```
or if we want to always listen for default endpoint:
```
// IIRC bridge internally uses Set to store endpoints, so we should not
// have duplicates if "endpoints" already included "self".
service.listen();

endpoints && endpoints.forEach((endpoint) => service.listen(endpoint));
```

apps/sms/views/conversation/js/conversation.js
```diff
@@ -265,6 +266,8 @@ var ConversationView = {
 
     // Bound methods to be detachables
     this.onMessageTypeChange = this.onMessageTypeChange.bind(this);
+
+    MessagingClient.init();
```
azasypkin: __nit:__ maybe we should lazy load ```MessagingClient``` in both current monolithic and split views and initialize in shared/js/startup.js and views/conversation/startup.js instead? Looks like MessagingService doesn't include anything that maybe needed before visually-loaded for both setups...

apps/sms/services/js/messaging/messaging_service.js
```diff
@@ -182,4 +186,6 @@ exports.MessagingService = BridgeServiceMixin.mixin(
   { methods: METHODS }
 );
 
+MessagingService.init();
```
azasypkin: __note:__ that's why I've added that ```if ('importScripts' in self)``` in my patch :) So that ```init``` is called on-demand in tests and all internal things can be stubbed in advance, on contrary in this case ```init``` will be called once we load file in the test suite.

apps/sms/services/js/messaging/messaging_client.js
```diff
@@ -0,0 +1,50 @@
+/* global bridge
```
azasypkin: __nit:__ ```/* global bridge */```

apps/sms/services/js/messaging/messaging_client.js
```diff
@@ -0,0 +1,50 @@
...
+*/
+
+(function(exports) {
+  'use strict';
+
+  /**
+   * Name of the service that is responsible for messaging service.
```
azasypkin: __typo:__  ```Name of the service that is responsible for messaging actions.``` or something similar? :)

apps/sms/services/js/moz_mobile_message/moz_mobile_message_shim.js
```diff
@@ -2,8 +2,8 @@
 /* vim: set shiftwidth=2 tabstop=2 autoindent cindent expandtab: */
```
azasypkin: __nit:__ it would be great if you can update mozMobileMessage.retrieveMMS in desktop mock so that it returns Promise....

apps/sms/services/js/bridge_service_mixin.js
julienw: I agree with oleg :)
+ add jsdoc !

apps/sms/services/js/messaging/messaging_client.js
```diff
@@ -0,0 +1,50 @@
...
+   */
+  var client;
+
+  var MessagingClient = {
+    init() {
+      var serviceEndpoint = new SharedWorker(
+        '/services/js/messaging/messaging_service.js'
+      );
+
+      serviceEndpoint.port.start();
```
julienw: **note**: I see on MDN (https://developer.mozilla.org/en-US/docs/Web/API/MessagePort/start) that MessagePort are queuing the messages. Good for us, no race is possible !

apps/sms/services/js/messaging/messaging_client.js
```diff
@@ -0,0 +1,50 @@
...
+      return client.method('sendMMS', options);
+    },
+
+    resendMessage(message) {
+      return client.method('resendMessage', message);
+    },
+
+    retrieveMMS(id) {
+      return client.method('retrieveMMS', id);
+    }
```
julienw: wondering if we should make this more automatic; I realize some client methods could be more complex, but we'll have a lot of simple methods like those. Either a forEach, or using a ES6 proxy ? Not urgent and not necessary in this patch, just thinking out loud and asking what you think about this.

apps/sms/index.html
```diff
@@ -128,6 +129,7 @@
   </head>
 
   <body role="application" class="skin-comms theme-communications">
+    <iframe src="/views/shared/shim_host.html" class="shim-host" height="0" width="0"></iframe>
```
julienw: can you please measure with raptor if adding the iframe changes the launch time ? Actually I'm sure it does and I'm also sure it's better to have it in the markup so that we have the shims ready at startup time, but it's good to know how much we'll regress because of this.

julienw: Can you check it still works if we hide the iframe ? (display: hidden for the class, or hidden attribute)

apps/sms/views/shared/js/startup.js
```diff
@@ -47,7 +47,8 @@ var Startup = {
     '/lib/bridge/bridge.js',
     '/services/js/bridge_service_mixin.js',
     '/services/js/activity/activity_shim.js',
-    '/services/js/activity/activity_client.js'
+    '/services/js/activity/activity_client.js',
+    '/services/js/messaging/messaging_client.js'
```
julienw: I think we need this in the startup for conversation view too ?

apps/sms/views/conversation/js/conversation.js
```diff
@@ -2480,26 +2483,23 @@ var ConversationView = {
     // force a number
     var id = +messageDOM.dataset.messageId;
     var iccId = messageDOM.dataset.iccId;
-
-    var request = MessageManager.retrieveMMS(id);
```
julienw: Just throwing some thoughts I had here, but I now think it's better like this.
I was wondering if we could do the call in MessageManager instead. Maybe having 2 different "MessageManager" implementations, switchable using a setting. OR MessageManager itself could decide to use the client or mozMobileMessage directly.

But now I think this be the role of MessagingClient, deciding to either use a worker or to load the worker script inside the window for a fast path.

apps/sms/services/js/messaging/messaging_client.js
azasypkin: Heh, good to know!

apps/sms/services/js/messaging/messaging_client.js
azasypkin: It's would be interesting to leverage Proxies some day - they look so promising to me :)

apps/sms/views/conversation/js/conversation.js
azasypkin: > But now I think this be the role of MessagingClient, deciding to either use a worker or to load the worker script inside the window for a fast path.

Yep, I like this solution more - it gives us more granular control on what we want to replace if needed on contrary to the monolithic catch-all MessageManager.

apps/sms/lib/sww.js
steveck-chung: Right, we don't have to update sww right in the patch.

apps/sms/services/js/messaging/messaging_service.js
steveck-chung: No matter which style we apply it must be horrible considering the channel and possibly app instance here :) Just want to make sure we have the same naming convention here.

apps/sms/services/js/moz_mobile_message/moz_mobile_message_shim.js
steveck-chung: Yeah I'll try to make the desktop mozMobileMessage.retrieveMMS returns Promise in this patch. For the integration test, we'll need separate bugs for it since we don't even have test case for MMS retrieval in integration test now.

apps/sms/services/js/messaging/messaging_client.js
steveck-chung: yeah we can try some new pattern with ES6 features for sure.

apps/sms/index.html
steveck-chung: I'm thinking the same thing about hiding the iframe, will give it a try for all the scenario by raptor.

apps/sms/views/shared/js/startup.js
steveck-chung: I load the client in the conversation view markup, but maybe we can lazyload messageing client since we don't really need it while view startup.

apps/sms/views/conversation/js/conversation.js
steveck-chung: Yeah you are right, MessagingClient/service are not necessary for startup path (and even not necessary for conversation view startup)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=058ee8ef68562d5281e3946b4a661fde676fa9c7

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8c8e3c5be592f97e4b86ae3b9ebb74ab0087c39b

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3ca540285bdbdd27aa2be6c48bc7c3d223d5ed1f

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ba2dc28ad706d27ed6d2a47db947459a07635fb9

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6f820ee6adf3ac771cb62591e07fabd2283f200a

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=05f1aba20a791fe0780609fa28c65e3d4592579c

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=76c1cf40ddca66fbc3f126bc6df88bc4370780e5

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=69cf75f58be6026a90ee2f8ee5e220dcfc245ba0

apps/sms/desktop-mock/navigator_moz_mobilemessage.js
```diff
@@ -3,6 +3,18 @@
 
 'use strict';
 
+// Copied from Utils.Promise 
+function defer() {
```
steveck-chung: Copy the promise defer from Utils. We can still use Utils.Promise.defer but it's the only thing we need in this mock. I can reuse Utils.Promise.defer if you prefer.

apps/sms/index.html
```diff
@@ -127,6 +128,7 @@
   </head>
 
   <body role="application" class="skin-comms theme-communications">
+    <iframe src="/views/shared/shim_host.html" class="shim-host" hidden></iframe>
```
steveck-chung: Set the iframe hidden seems a little better based on profiling result: https://docs.google.com/a/mozilla.com/spreadsheets/d/1u8u7OswOxHD8t4WgjaOrfzdMYx6bkOz3CahxHUYX7Gw/edit?usp=sharing

apps/sms/services/js/messaging/messaging_client.js
```diff
@@ -0,0 +1,54 @@
...
+
+(function(exports) {
+  'use strict';
+
+  /**
+   * Name of the service that is responsible for messaging functionality.
+   * @type {string}
+   */
+  const SERVICE_NAME = 'messaging-service';
+  const TIMEOUT = 5000;
```
steveck-chung: Now we can set the timeout by ourself instead of default 1 sec since it's not enough for retrieving the mms, so I ask Wilson to have additional parameter for timeout. Maybe the best way for messaging service is to set no timeout since it's highly depends on network status and gecko api also have it's timout mechanism.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=064a2f8d206221705dd55d08d9f0debe49f74d37

apps/sms/services/js/messaging/messaging_client.js
```diff
@@ -0,0 +1,54 @@
...
+
+  var serviceEndpoint = null;
+
+  var MessagingClient = {
+    init() {
+      serviceEndpoint = new SharedWorker(
+        '/services/js/messaging/messaging_service.js'
+      );
+
+      client = bridge.client({
```
steveck-chung: Per Willson's cpmment client should be able to accept both original way of parameters and client option below. I think it might be more clear to use options since people will know the parameter immediately.

julienw:
I think you forgot to add the mock messaging client to the patch

apps/sms/desktop-mock/navigator_moz_mobilemessage.js
julienw: if we include utils.js anyway, maybe it's ok to use it ?

apps/sms/desktop-mock/navigator_moz_mobilemessage.js
```diff
@@ -13,6 +25,13 @@
     return;
   }
 
+	// Workaround to force the mozMobileMessage in shim context use window one.
+  // Need to be removed once all the event listeners are ready for shim. 
+  if (location.pathname.indexOf('shim_host') > -1) {
+    navigator.mozMobileMessage = parent.navigator.mozMobileMessage;
+    return;
+  }
```
julienw: maybe easier to have this in a separate JS file that we'd load only in the shim html file ?

apps/sms/desktop-mock/navigator_moz_mobilemessage.js
```diff
@@ -1538,9 +1557,7 @@
   };
 
   MockNavigatormozMobileMessage.retrieveMMS = function(id) {
-    var request = {
-      error: null
-    };
+    var deferred = defer();
     var msgs = messagesDb.messages;
     var idx = 0, len = msgs.length;
     setTimeout(function() {
```
julienw: we don't need the setTimeout if we use promises -- it was here to make the callback call asynchronous but with promises we have this for free

apps/sms/services/js/messaging/messaging_client.js
julienw: when you say "network" you mean the communication between threads I guess ?

apps/sms/services/js/messaging/messaging_client.js
julienw: yeah I like it like this as well

apps/sms/services/js/messaging/messaging_service.js
julienw: we don't have that many objects and we control the names so I'm not too concerned here.

apps/sms/services/js/messaging/messaging_service.js
```diff
@@ -182,4 +188,8 @@ exports.MessagingService = BridgeServiceMixin.mixin(
   { methods: METHODS }
 );
 
+if ('importScripts' in self) {
+  MessagingService.init();  
+}
```
julienw: If you want to know whether you're in a worker, we found that `if (!self.document)` is likely better

apps/sms/services/js/moz_mobile_message/moz_mobile_message_shim.js
```diff
@@ -16,90 +11,89 @@
...
+    var endPoint = new BroadcastChannel(SERVICE_NAME + '-channel');
+    mozMobileMessage = mobileMessage;
+    this.initService(endPoint);
 
-    EVENTS.forEach((event) => {
+    Object.keys(EVENTS).forEach((event) => {
       mozMobileMessage.addEventListener(
         event,
-        this[Utils.camelCase(`on-${event}`)]
+        this['on' + capitalize(event)].bind(this)
```
julienw: just wondering why you removed camelCase here ?

apps/sms/services/js/moz_mobile_message/moz_mobile_message_shim.js
```diff
@@ -111,8 +105,17 @@ var MozMobileMessageShim = {
...
+  /**
+   * retrieveMMS.
+   * @param {Number} id MMS message id.
+   * @returns {Promise.<Boolean|Error>} return true if MMS is retrieved
+   *  successfully or error while failed. 
+   */
   retrieveMMS(id) {
-    return mozMobileMessage.retrieveMMS(id);
+    return mozMobileMessage.retrieveMMS(id).then((message) => {
+      // Retrun true instead of message to avoid clone error issue.
```
julienw: nit: "retrun"

apps/sms/tools/update_nga_libs.sh
```diff
@@ -12,7 +12,7 @@ bower_path="bower_components"
 
 threads_repo="gaia-components/threads#master"
 threads_index="$bower_path/threads/index.js"
-threads_plugins="$bower_path/threads/lib/plugins"
+threads_plugins="$bower_path/threads/src/plugins"
```
julienw: wondering why we move this to src ?

apps/sms/views/conversation/test/unit/conversation_test.js
```diff
@@ -2887,13 +2889,18 @@ suite('conversation.js >', function() {
...
         setup(function() {
           if (!('mozSettings' in navigator)) {
            navigator.mozSettings = null;
           }
 
+          retrievePromise = new Promise((resolve, reject) => {
+            onResolve = resolve;
+            onReject = reject;
+          });
+          MessagingClient.retrieveMMS.returns(retrievePromise);
```
julienw: interesting idea; I never thought of doing it like this (usually I prefer to run `MessagingClient.retrieveMMS.returns(...)` in the test itself so that I can change the return value -- but then I'd need to call handleMessageClick in the test itself as well.

So this is an interesting idea here :)

I don't really like the names `onResolve` and `onReject`; maybe `doResolve` and `doReject` feel better. Also can you nullify these variable in a `teardown` function ? Just a cleanup :)

apps/sms/services/js/messaging/messaging_client.js
julienw: ah no, I understand now, it's the "real" network.

so 5sec is not enough, I just got the issue with a small MMS. It think we should not have a timeout here.

Another possibility is to just ignore the return for the calls that might take a long time and rely on events only. But this would be for a future patch if we want to do this.

Let's just disable the timeout for now.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=abb4051eaf6bd2afefd4b230485002e66516a366

apps/sms/services/js/moz_mobile_message/moz_mobile_message_shim.js
steveck-chung: Ah, it's for remove the dependency for Utils since it's possible that we don't need any utils for shim host(and yes I forgot to remove the Utils loading in shim host...)

apps/sms/tools/update_nga_libs.sh
steveck-chung: The plugins folder direction changed in threads

apps/sms/services/js/moz_mobile_message/moz_mobile_message_shim.js
julienw: as of now I think we maybe don't need to focus on such optimizations; if we feel we need it later on, we can look again. What do you think ?

apps/sms/tools/update_nga_libs.sh
julienw: ah yeah ok, got it

apps/sms/services/js/messaging/messaging_client.js
steveck-chung: Disable timeout patch: https://github.com/gaia-components/threads/pull/75

apps/sms/services/js/moz_mobile_message/moz_mobile_message_shim.js
steveck-chung: But I still think we hardly need Utils for the shims in general. Maybe we can add it back If we really need it for other shims in the future?

julienw: really I don't have a strong opinion; let's see what @azasypkin thinks

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=af8d7488a2781e9b9e3c8d7a2187210874bbb7f9

apps/sms/services/js/moz_mobile_message/moz_mobile_message_shim.js
azasypkin: Personally I'd rather avoid loading Utils.js unless we really need it. On a related note, honestly I'd split utils to view-utils and utils, but it's another story :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ff6afb5b7243cec34cfb14057b2adcff89a4d039

apps/sms/lib/bridge/plugins/contracts/service.js
```diff
@@ -0,0 +1,9 @@
+!((define)=>{define((require,exports,module)=>{
```
azasypkin: I've manually removed this file from my PR, since we don't use it, but it's not a big deal anyway, just wanted to note :)

apps/sms/desktop-mock/navigator_moz_mobilemessage.js
```diff
@@ -1538,9 +1537,7 @@
   };
 
   MockNavigatormozMobileMessage.retrieveMMS = function(id) {
-    var request = {
-      error: null
-    };
+    var deferred = Utils.Promise.defer();
```
azasypkin: __nit:__ shim-host.html includes this file, but doesn't include ```utils.js``` - so maybe we can just use ```new Promise``` here (e.g. instead of ```setTimeout``` below since we don't need timeout now anyway).

apps/sms/desktop-mock/navigator_moz_mobilemessage.js
```diff
@@ -1571,11 +1565,10 @@
           return;
         }
       }
-      if (typeof request.onerror === 'function') {
-        request.onerror.call(request);
-      }
+
```
azasypkin: __nit:__ trailing spaces

apps/sms/services/js/messaging/messaging_client.js
```diff
@@ -0,0 +1,54 @@
...
+
+(function(exports) {
+  'use strict';
+
+  /**
+   * Name of the service that is responsible for messaging functionality.
+   * @type {string}
+   */
+  const SERVICE_NAME = 'messaging-service';
+  const TIMEOUT = 0;
```
azasypkin: __optional nit:__ maybe comment here would be useful (why we need it and what 0 means)?

apps/sms/services/js/messaging/messaging_service.js
```diff
@@ -61,7 +63,11 @@ var MessagingService = {
 
     this.initService();
 
-    this.mozMobileMessageClient = bridge.client(MOBILE_MESSAGE_CLIENT_NAME);
+    this.mozMobileMessageClient = bridge.client({
+      service: MOBILE_MESSAGE_CLIENT_NAME,
+      endpoint: new BroadcastChannel(MOBILE_MESSAGE_CLIENT_NAME + '-channel'),
```
azasypkin: __question:__ Not sure if we need leave comment\todo here that it won't work for the second app instance in case it has been run when _first_ instance was alive, but then was killed/closed?

apps/sms/services/js/moz_mobile_message/moz_mobile_message_shim.js
```diff
@@ -111,8 +105,17 @@ var MozMobileMessageShim = {
     return mozMobileMessage.getMessage(id);
   },
 
+  /**
+   * retrieveMMS.
```
azasypkin: __nit:__ Maybe a bit more wordy description? :)

apps/sms/desktop-mock/navigator_moz_mobilemessage.js
steveck-chung: shim-host.html includes `navigator_moz_mobilemessage_shim.js` actually per julien's comment ;)

apps/sms/views/conversation/index.html
```diff
@@ -116,10 +116,12 @@
   <script defer src="/views/shared/js/contact_renderer.js"></script>
   <script defer src="/views/shared/js/localization_helper.js"></script>
   <script defer src="/views/conversation/js/information.js"></script>
+  <script defer src="/services/js/messaging/messaging_client.js"></script>
   -->
 </head>
 
 <body role="application" class="skin-comms theme-communications">
+  <iframe src="/views/shared/shim_host.html" class="shim-host" hidden></iframe>
```
julienw: maybe add it in /views/inbox already, even if we don't use it here yet ?

apps/sms/services/js/messaging/messaging_service.js
julienw: yes I think we need a bug and a TODO with the bug number here. This is not that important for retrieveMMS but will be for other cases :/

apps/sms/services/test/unit/moz_mobile_message/moz_mobile_message_shim_test.js
```diff
@@ -164,17 +131,39 @@ suite('MozMobileMessageShim >', function() {
...
-        ...args
-      );
+      setup(function() {
+        args = [1];
+      });
+
+      test('success', function() {
+        MockNavigatormozMobileMessage.retrieveMMS.returns(Promise.resolve({
+          id: 'fake message'
+        }));
```
julienw: this is a really a nit, but I'd write this like this:
 ```javascript
MockNavigatormozMobileMessage.retrieveMMS.returns(
  Promise.resolve({ id: 'fake message' })
);
```

I think it's al ittle more readable; what do you think ?

File: apps/sms/services/test/unit/moz_mobile_message/moz_mobile_message_shim_test.js [hunk 34]
```diff
@@ -164,17 +131,39 @@ suite('MozMobileMessageShim >', function() {
...
+            ...args
+          );
+        });
+      });
+
+      test('error', function() {
+        var error = 'fakeError';
+        MockNavigatormozMobileMessage.retrieveMMS.returns(Promise.reject(
+          error
+        ));
```
julienw: same here:
```javascript
MockNavigatormozMobileMessage.retrieveMMS.returns(
  Promise.reject(error)
);
```

apps/sms/desktop-mock/navigator_moz_mobilemessage.js
azasypkin: Mmmm, what is ```navigator_moz_mobilemessage_shim.js```? I don't see this file anywhere? I believe we'll have 404 in browser....

apps/sms/services/js/moz_mobile_message/moz_mobile_message_shim.js
```diff
@@ -111,8 +105,17 @@ var MozMobileMessageShim = {
...
+   * @param {Number} id MMS message id.
+   * @returns {Promise.<Boolean|Error>} return true if MMS is retrieved
+   *  successfully or error while failed. 
+   */
   retrieveMMS(id) {
-    return mozMobileMessage.retrieveMMS(id);
+    return mozMobileMessage.retrieveMMS(id).then((message) => {
```
azasypkin: Not sure if ```true``` makes sense here since we can't have ```false```. Maybe just ```Promise.<void>```?

apps/sms/services/js/bridge_service_mixin.js
```diff
@@ -26,7 +26,12 @@
    * EVENTS, METHODS and STREAMS are all optional.
    */
   var serviceMixin = {
-    initService() {
+    /**
+     * Initialize service to listen on the default window and given endpoint.
```
azasypkin: __typo:__ ```default window``` --> ```default context```/ ```default scope```/```global context``` since it's not window in worker context.

apps/sms/services/test/unit/messaging/messaging_client_test.js
```diff
@@ -0,0 +1,90 @@
...
+  setup(function() {
+    clientStub = sinon.stub({
+      method: () => {}
+    });
+
+    this.sinon.stub(bridge, 'client').returns(clientStub);
+  });
+
+  suite('init', function() {
+    test('client are initialized', function() {
```
azasypkin: __typo:__ ```client is initialized```

apps/sms/services/test/unit/messaging/messaging_client_test.js
```diff
@@ -0,0 +1,90 @@
...
+          assert.equal(successResult, result);
+        }).then(done, done);
+      });
+
+      test('onerror', function(done) {
+        clientStub.method.withArgs(method, ...params).returns(
+          Promise.reject(errorResult)
+        );
+
+        MessagingClient[method](...params).catch((error) => {
```
azasypkin: __nit:__ probably to be on the safe side we need smth like:
```
MessagingClient[method](...params).then(() => { 
 	throw new Error('Promise should not be resolved!');
}, (error) => {
  	sinon.assert.calledWithExactly(clientStub.method, method, ...params);
  	assert.deepEqual(errorResult, error);
}).then(done, done);
```

apps/sms/services/test/unit/messaging/messaging_service_test.js
```diff
@@ -51,7 +52,11 @@ suite('Messaging service >', function() {
   suite('init', function() {
     test('service mixing and client are initialized', function() {
       sinon.assert.calledWith(bridge.service, 'messaging-service');
-      sinon.assert.calledWith(bridge.client, 'mozMobileMessageShim');
+      sinon.assert.calledWith(bridge.client, {
+        service: 'moz-mobile-message-shim',
+        endpoint: sinon.match.instanceOf(BroadcastChannel),
```
azasypkin: __optional nit:__ in my patch I've used slightly modified BroadcastChannelMock (mock remebers name that is passed into its constructor)with the following assert, to make sure we use channel with the right name:
```
endpoint: sinon.match.instanceOf(BroadcastChannel).and(sinon.match.has('name', shimChannel))
```

But up to you.

apps/sms/services/test/unit/messaging/mock_messaging_client.js
```diff
@@ -0,0 +1,10 @@
+/* exported MockMessagingClient */
+'use strict';
+
+var MockMessagingClient = {
+  init: () => {},
+  sendSMS: () => {},
+  sendMMS: () => {},
+  resendMessage: () => {},
+  retrieveMMS: () => {}
+};
```
azasypkin: __nit:__ new line

apps/sms/services/test/unit/moz_mobile_message/moz_mobile_message_shim_test.js
```diff
@@ -50,39 +43,13 @@ suite('MozMobileMessageShim >', function() {
...
-      sinon.assert.calledWith(
-        serviceStub.stream,
-        shimStream,
-        MozMobileMessageShim[shimStream]
-      );
-    });
+    sinon.assert.calledTwice(serviceStub.listen);
+    sinon.assert.calledWith(
+      serviceStub.listen,
+      sinon.match.instanceOf(BroadcastChannel)
```
azasypkin: __optional nit:__ same advice with ```BroadcastChannel``` name here as well, but I'll modify it in my patch anyway, so you can ignore :)

apps/sms/services/test/unit/moz_mobile_message/moz_mobile_message_shim_test.js
azasypkin: __nit:__ also should not we just use ```var messageId = 'fake_message_id'``` instead of these ```...args```? Looks a bit over-engineered and confusing (and then we don't need ```setup``` at all).

apps/sms/views/conversation/js/conversation.js
```diff
@@ -2504,26 +2505,21 @@ var ConversationView = {
...
-
     var button = messageDOM.querySelector('button');
 
     this.setMessageStatus(id, 'pending');
     button.setAttribute('data-l10n-id', 'downloading-attachment');
 
-    request.onsuccess = (function retrieveMMSSuccess() {
+    MessagingClient.retrieveMMS(id).then((message) => {
```
azasypkin: __nit:__ we don't have ```message``` as return result, so better to remove to avoid confusion.

apps/sms/views/conversation/js/conversation.js
```diff
@@ -2504,26 +2505,21 @@ var ConversationView = {
...
-
-    request.onerror = (function retrieveMMSError() {
+    }, (error) => {
       this.setMessageStatus(id, 'error');
       button.setAttribute('data-l10n-id', 'download-attachment');
 
       // Show NonActiveSimCard/Other error dialog while retrieving MMS
-      var errorCode = (request.error && request.error.name) ?
-        request.error.name : null;
+      var errorCode = (error && error.name) ?
```
azasypkin: __optional nit:__ it can be one-liner now! :)

apps/sms/views/conversation/test/unit/conversation_test.js
```diff
@@ -3128,11 +3160,12 @@ suite('conversation.js >', function() {
           assert.isTrue(element.classList.contains('pending'));
         });
         test('click calls retrieveMMS', function() {
-          assert.isTrue(MessageManager.retrieveMMS.calledWith(message.id));
+          sinon.assert.calledWith(MessagingClient.retrieveMMS ,message.id);
```
azasypkin: __nit:__ ```(MessagingClient.retrieveMMS, message.id)``` (space after comma)

apps/sms/views/conversation/test/unit/conversation_test.js
```diff
@@ -2920,11 +2933,12 @@ suite('conversation.js >', function() {
...
         });
         test('click calls retrieveMMS', function() {
-          assert.isTrue(MessageManager.retrieveMMS.calledWith(message.id));
+          sinon.assert.calledWith(MessagingClient.retrieveMMS, message.id);
         });
         suite('response error', function() {
-          setup(function() {
-            MessageManager.retrieveMMS.returnValues[0].onerror();
+          setup(function(done) {
+            retrievePromise.catch(() => done());
```
azasypkin: __optional nit:__ I'd place statement with ```done``` as the last statement in ```setup``` here and in tests below, but up to you.

apps/sms/views/shared/shim_host.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Messages Shim Host</title>
+    <!-- Firefox Desktop shims -->
+    <script defer src="/desktop-mock/navigator_moz_mobilemessage_shim.js"></script>
```
azasypkin: __nit:__ no need in ```defer```;

__question:__ as I've asked above, what is ```/desktop-mock/navigator_moz_mobilemessage_shim.js```?

apps/sms/services/test/unit/moz_mobile_message/moz_mobile_message_shim_test.js
azasypkin: > I think it's al ittle more readable; what do you think ?

Yep, I agree.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2c13417a678fc0fb8b844f1ae7c570928c810732

apps/sms/desktop-mock/navigator_moz_mobilemessage.js
steveck-chung: Sorry my bad :/ , added in the latest commit and you can check if it works for you @azasypkin

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9b976345cf46810d1878ef6a8b9857ab22706db8

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c3b31e8535f7ae972649c6c60f412bc7bb83042b

apps/sms/desktop-mock/navigator_moz_mobilemessage.js
azasypkin: Yeah, looks good!

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c1e8a086db920e703500c6b887f7fac2f0475ed8

PR #31100, opened 2015-07-24
Homescreen stackmanager prototype

sfoster:
WIP prototype to add the homescreen at the bottom of the stack and navigable to via edge gesture and task manager.

### DIFF
--- a/apps/system/js/app_window.js
+++ b/apps/system/js/app_window.js
@@ -1247,7 +1247,19 @@
    * once we need to display the image,
    * and revoke right away after we finish rendering the image.
    */
-  AppWindow.prototype._screenshotBlob = undefined;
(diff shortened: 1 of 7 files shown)


autolander:

apps/system/js/app_window.js
```diff
@@ -1247,7 +1247,19 @@
...
+    get: function() {
+      return this.__screenshotBlob;
+    },
+    set: function(blob) {
+      this.__screenshotBlob = blob;
+      if (blob) {
+        this.publish('screenshotready');
+      }
+    }
+  });
```
etiennesegonzac: can we move this to the browser_mixin.js method?
(looks like the `renewCachedScreenshotBlob` method is never used)

PR #31107, opened 2015-07-26
Bug 1187668 - update settings l10n api uses

zbraniecki:
https://bugzilla.mozilla.org/show_bug.cgi?id=1187668


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9ee188c0f70abc92a8280253791795b12bdd05dd

autolander:

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=bfe73a074b2a07f1f983db8019c85231437df5e0

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7fb323a1f2790fabc3a9f28fa14031c8a5b5ee88

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7c013aa29a3512d4e7a99b0da0dba3b416a7618f

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a0f005d6fbe1a834c82d7273cafa7c101c9cce94

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=92471a2a673535faf91ec4cda4b146526dca1bf1

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3b31a3947518637483a0c07450d8074418bb46c9

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d2c170d990d3626db323fdf30ca831d5d734c9d4

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=223ecbc033119afd587b2a997428b461aefbe81a

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f9887d1679ec5f705a0ae4337edf2057f4fa46e8

apps/settings/elements/firefox_accounts.html
```diff
@@ -29,14 +29,14 @@ <h2 data-l10n-id="fxa-confirm-account">Confirm Your Account</h2>
       </header>
       <ul>
         <li class="description">
-          <p data-l10n-id="fxa-verification-email-sent-msg" id="fxa-unverified-text">
```
gasolin: does it intended to remove the l10n here?

apps/settings/elements/firefox_accounts.html
```diff
@@ -29,14 +29,14 @@ <h2 data-l10n-id="fxa-confirm-account">Confirm Your Account</h2>
...
+          <p id="fxa-unverified-text">
             We will send an email to: foo@bar.com
           </p>
           <p data-l10n-id="fxa-please-verify">
             Click the verification link in your email to confirm your account.
           </p>
-          <p id="fxa-resend-email">
-            Don't see an email? <a href="#">Resend</a>.
+          <p id="fxa-resend-email" data-l10n-id="fxa-dont-see-email-resend">
+            Don't see an email? <a href="#" id="fxa-resend">Resend</a>.
```
gasolin: seems not the same as string defined in `fxa-dont-see-email-resend`. Could we just remove all strings inside of tags?

gasolin:
treeherder fail, please fix the tests

apps/settings/elements/firefox_accounts.html
zbraniecki: yes. The entity requires an attribute (email address), so it will not work without it, so there's no reason to keep it in HTML since it has to be defined from JS as l10n-id/l10n-args couple and it is in `apps/settings/js/firefox_accounts/panel.js#showUnverifiedPanel`

apps/settings/elements/firefox_accounts.html
zbraniecki: Absolutely. It will never show up because we pretranslate to default language during build time, so any strings defined in HTML will be overwrote. We can as well remove them :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f9887d1679ec5f705a0ae4337edf2057f4fa46e8

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7a2f21102b623fe045e39d888f5c91ffcba3c870

zbraniecki:
it was just a treeherder failure over the weekend. The tests are green now (except of some unrelated privacy-panel bug that is on all treeherder runs)

apps/settings/elements/firefox_accounts.html
evelynhung: remove+1

apps/settings/js/bluetooth.js
```diff
@@ -664,13 +666,14 @@ navigator.mozL10n.once(function bluetoothSettings() {
...
       } else {
         // show pair process fail.
-        var msg = _('error-pair-title');
+        var l10nId = 'error-pair-title';
         if (errorMessage === 'Repeated Attempts') {
-          msg = msg + '\n' + _('error-pair-toofast');
+          l10nId = 'error-pair-toofast';
         } else if (errorMessage === 'Authentication Failed') {
-          msg = msg + '\n' + _('error-pair-pincode');
+          l10nId = 'error-pair-pincode';
```
evelynhung: The logic here is changed. It was an error "title + body", but it now only remains body.

evelynhung: So I think you need to update string as you did for `unpair-confirm`. It's easier. :)

apps/settings/js/bluetooth.js
```diff
@@ -685,18 +688,25 @@ navigator.mozL10n.once(function bluetoothSettings() {
...
       if (device.address === connectedAddress) {
-        var msg = _('unpair-title') + '\n' + _('unpair-msg');
-        if (!window.confirm(msg)) {
-          return;
-        }
-        connectedAddress = null;
+        navigator.mozL10n.formatValue('unpair-confirm').then(msg => {
+          if (!window.confirm(msg)) {
+            return;
+          }
```
evelynhung: I'm not pretty sure if it would cause any side effect of being asynchronized. Imagine at the moment user clicking 'unpair' and the UI wasn't frozen by this confirm window *immediately* so the user can click others.. and then at some point the window pops up which may be unexpected. Do you think the promise will be resovled very fast and almost immediately?

apps/settings/js/bluetooth.js
zbraniecki: The reason why I decided to not modify the ID was because previously it would build message like this:
`Unable to pair devices\nUnable to pair with the device. Check that the PIN is correct.`

which I changed to either first line or the latter. Since all messages have the same "Unable to pair with the device" I don't feel like removing the first line changes anything.

Can you look at settings.en-US.properties and let me know what you think?
If you'll still believe I should, I'll change the IDs :)

apps/settings/js/bluetooth.js
zbraniecki: good point. It will be fast, but the idea behind it is that it should have to be to work properly. I'll test it.

apps/settings/js/bluetooth.js
evelynhung: Yep, I noticed there are repeated words but it was UX design of "title" and "body", so we can't remove one of them, otherwise the UI will be a bit weird.

apps/settings/js/bluetooth.js
evelynhung: Thanks. I also have the same worry on all similar changes (delayed pop-up windows), but I guess we just need to test some of them.

apps/settings/elements/firefox_accounts.html
evelynhung: I didn't quite get it when replying the comments above. So correct me if I'm wrong:
1. the new `fxa-dont-see-email-resend` string has a `<a>` tag so it will be applied by following the same DOM structure we defined here as
`Don't see an email? <a href="#" id="fxa-resend">Resend</a>` without removing `href` and `id` attributes, right? (I remember it's an awesome feature of l20n.)
2. what @gasolin suggested was to remove `Don't see an email?` and `Resend` text in HTML because it will be overwritten at build time anyway? If so, I prefer to keep it because it's not purely a string but includes a HTML tag. Keeping the content is more readable to me.

apps/settings/js/icc.js
```diff
@@ -61,6 +59,18 @@ require([
...
+      if (typeof(l10n) === 'string') {
+        element.setAttribute('data-l10n-id', l10n);
+      } else if (l10n.raw) {
+        element.removeAttribute('data-l10n-id');
+        element.textContent = l10n.raw;
+      } else {
+        navigator.mozL10n.setAttributes(element, l10n.id, l10n.attrs);
+      }
+    }
+
```
evelynhung: Sorry I didn't follow the evolution of l10n framework too much. Could you explain a bit the logic here? I don't understand why we need this function and what `l10n.raw` is. Thanks.

evelynhung: Oh, I guess I got it. Using different way to assign either pure text or a l10n-id

evelynhung: The usage of this function and the whole l10n/string processing in this file is not intuitive to me. Could you add some comments to make it more clear? Thanks.

apps/settings/js/modules/date_time.js
```diff
@@ -193,9 +193,7 @@ define(function(require) {
   DateTime.prototype.handleEvent = function(evt) {
     switch (evt.type) {
       case 'moztimechange':
-        navigator.mozL10n.ready((function _updateTime() {
-          this._autoUpdateDateTime();
-        }).bind(this));
+        this._autoUpdateDateTime();
```
evelynhung: I'm curious why you removed `navigator.mozL10n.ready` and `navigator.mozL10n.once` everywhere. :-/

apps/settings/js/modules/date_time.js
```diff
@@ -290,13 +288,15 @@ define(function(require) {
...
       if (iso) {
         return d.toLocaleFormat('%Y-%m-%d');
       } else {
-        var f = new navigator.mozL10n.DateTimeFormat();
-        return f.localeFormat(d, _('shortDateFormat'));
+        return d.toLocaleString(navigator.languages, {
+          year: 'numeric',
+          month: 'long',
+          day: '2-digit',
+        });
```
evelynhung: It's great to see we use native API of Date object instead of DIY!

apps/settings/js/modules/dialog_manager.js
```diff
@@ -169,10 +155,6 @@ define(function(require) {
         return self._loadPanel(dialog.panel.id);
       })
       .then(function() {
-        // 2: l10n is ready
-        return self._initializedL10n();
-      })
-      .then(function() {
         // 3: Get that panel
```
evelynhung: nit: `2: Get that panel`

apps/settings/js/modules/panel_utils.js
```diff
@@ -270,15 +270,13 @@ define(function(require) {
...
-            var _ = navigator.mozL10n.get;
             switch (key) {
               //XXX bug 816899 will also provide 'deviceinfo.software' from
               // Gecko which is {os name + os version}
               case 'deviceinfo.software':
                 navigator.mozL10n.setAttributes(spanFields[i],
                   'deviceInfo_software',
-                  { brandShortName: _('brandShortName'),
-                    os: result['deviceinfo.os'] });
+                  { os: result['deviceinfo.os'] });
```
evelynhung: Do you intend to remove `brandShortName` ?

apps/settings/js/panels/hotspot/panel.js
```diff
@@ -98,8 +98,10 @@ define(function(require) {
       },
 
       _updateHotspotSecurity: function(newValue) {
-        elements.hotspotSecurityType.setAttribute('data-l10n-id',
-          'hotspot-' + newValue);
+        if (newValue) {
+          elements.hotspotSecurityType.setAttribute('data-l10n-id',
+            'hotspot-' + newValue);
+        }
```
evelynhung: nice catch!

apps/settings/js/panels/languages/languages.js
```diff
@@ -30,12 +30,28 @@ define(function(require) {
...
+          month: 'long',
+          year: 'numeric'
+        });
+        var timeString = d.toLocaleString(navigator.languages, {
+          hour12: navigator.mozHour12,
+          hour: 'numeric',
+          minute: 'numeric'
+        });
+
+        if (navigator.language in navigator.mozL10n.qps) {
```
evelynhung: Could you explain what `navigator.mozL10n.qps` is? Thanks.

apps/settings/elements/firefox_accounts.html
zbraniecki: That's what we do in other places, just sth like this:

```
foo=Hello world <a>Click me<a>
```

```
<p data-l10n-id="foo">
  <a href="http://www.mozilla.org" class="highlight"></a>
</p>
```

If you prefer to leave it, I'm ok with that as well. It doesn't affect the result. We're just trying to remove en-US content from HTML files to remove the confusion that happens when we update strings and the content in HTML gets out of sync

apps/settings/js/bluetooth.js
zbraniecki: Hmm, I actually think that the UI would be less weird :)

Would you be ok if I took some screenshots with this change (and without) and NI'ed UX in the bug?

apps/settings/js/icc.js
zbraniecki: We have this pattern documented here: https://developer.mozilla.org/en-US/Firefox_OS/Developing_Gaia/localization_code_best_practices#Writing_APIs_that_operate_on_L10nIDs

I'll add comments!

apps/settings/js/modules/date_time.js
zbraniecki: Because you don't need them anymore! The only reason we had them was for synchronous `mozL10n.get` that we're obsoleting and I'm removing all calls to it in this patch.

The `mozL10n.ready` serves one more purpose - it refires on language change, but in this case it was a bug because it was firing new `mozL10n.ready` on every moztimechange, so after a few, the callback was fired multiple times on languagechange.

With this particular removal, the autoUpdateDateTime will happen only when moztimechange happens.

apps/settings/js/modules/date_time.js
zbraniecki: Yeah! :)

apps/settings/js/modules/panel_utils.js
zbraniecki: Hahah, no! Sorry, the patch makes it unclear what happens here.

So, basically right now in .properties we have the dual-nature of placeables.

```
foo = {{ brandShortName }} loves you
```
will take brandShortName from a variable (l10n-args) or if there will not be such a key it will take another entity's value and in this case, we have an entity
```
brandShortName=Firefox OS
```

so it'll just call it in `deviceInfo_software` entity.

So basically this argument was not needed, because localization knows you want the value of an entity `brandShortName`. :)

apps/settings/js/panels/languages/languages.js
zbraniecki: Yes!

So, we introduced a concept of pseudo-locales that are meant to help developers work with testing their code in different languages before localizers provide translations. Historically we had 4 languages (ar, fr, zh-CN and en-US) in our source code, but that didn't work well since the other locales were outdated and localizers don't translate at the time when developers work on code.

So pseudo-locales just take DEFAULT_LOCALE (usually en_US) and morph it to something that looks similar but different.
* qps-ploc - Is Accented English which looks like en-US but is longer (so you can see how your UI behaves in languages which take more space), and is different so you can see if you didn't hardcode any en-US string accidentally
* qps-plocm - Mirrored English. This pseudo-locale mimics how RTL language like Hebrew or Arabic might behave. You can test how your UI behaves in RTL scenarios.

Unfortunately, for this screen we relied in our tests on verifying that the time and date strings use pseudo-locale. With a switch from our DYI solution to Intl, we lost this ability because Intl doesn't localize the date or time string into accented or mirrored.

I had a choice of either removing the test, or manually running those two strings to morph them into a pseudo locale.
I chose the latter because I think that anyone who will want to switch to pseudo locale in Settings panel will want to see how it looks like on the date/time example.

In the future we will modify the code here to show 'ar' date/time for RTL pseudolocale to make it look even closer.

`navigator.mozL10n.qps` is just an Object that stores current pseudolocales, so I use it here to test if the language currently used is a pseudolocale.
It will not affect any production language behavior.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=bb96f354a1e1749df0279d03118646f82d8537d4

apps/settings/locales/settings.en-US.properties
```diff
@@ -303,7 +303,7 @@ message-read-reports-details = Request read report for each message sent
 # force the proper directionality of parentheses at the end of the string
 # regardless of the direction of the surrounding layout. Remove them only
 # if the translation puts them between other text.
-mmsTitle.innerHTML= &#x202a;Multimedia Messaging (MMS)&#x202c;
+mmsTitle= &#x202a;Multimedia Messaging (MMS)&#x202c;
```
zbraniecki: @stasm - in bug 1139031 you recommended this solution. If I'm correct that is either not needed (because we do inject it in l10n.js for fallback languages) or it should be affecting all <h2> elements and should use CSS dir:auto instead.

Am I correct? I'd like to remove those characters from localization files

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=feb6314f8c047a24c6fedfd355d73cbc47167d8c

apps/settings/js/bluetooth.js
zbraniecki: ok, I don't even know how to get this code to run :(

The only code piece related to the string `error-pair-title` I can run is using this code: https://github.com/mozilla-b2g/gaia/blob/master/apps/settings/js/panels/bluetooth/panel.js#L593-L599 and it's already properly separating the entities.

The only other place where this l10nId is used is here and I don't know how to trigger this code path.

Can you help me? Can we move it to use `DialogService.alert`?

apps/settings/js/bluetooth.js
zbraniecki: Similarly as above, I can't seem to be able to trigger this codepath. All I'm getting is using this: https://github.com/mozilla-b2g/gaia/blob/master/apps/settings/js/panels/bluetooth/panel.js#L517-L522

Can you give me STR to launch this codepath?

apps/settings/js/panels/about/update_check.js
```diff
@@ -150,7 +154,8 @@ define(function(require) {
      */
     _checkForUpdates: function uc__checkForUpdates() {
       if (!navigator.onLine) {
-        alert(this._('no-network-when-update'));
+        navigator.mozL10n.formatValue('no-network-when-update').then(
+          msg => alert(msg));
```
zbraniecki: so, trying to see what will happen if formatValue takes time, I delayed it here for 5s.
The result was that the UI was a bit confusing since nothing has happened.

I can add `disabled` state on the button and remove it when formatValue resolves. It could be also a good idea to add it for when the update is being performed and remove when the checking is complete.

Would that work for you @evelynhung ?

apps/settings/js/panels/hotspot_wifi_settings/panel.js
```diff
@@ -135,7 +134,8 @@ define(function(require) {
 
         // ensure SSID is set
         if (/^\s*$/.test(elements.tethering_ssid.value)) {
-          alert(_('SSIDCannotBeEmpty'));
+          navigator.mozL10n.formatValue('SSIDCannotBeEmpty').then(
+            msg => alert(msg));
```
zbraniecki: Here, with the delay of 5s, the panel returns to the main HotSpot settings and then shows the error dialog.

I can either mingle with the code to make it not return to the previous panel until the alert is shown or I could mingle with the code to make the "OK" button be disabled when the input is empty.
Would that work for you @evelynhung ?

apps/settings/js/panels/simcard_manager/simcard_manager.js
```diff
@@ -152,15 +151,16 @@ define(function(require) {
...
-          } else {
-            this.selectedIndex = prevCardIndex;
-          }
+            if (wantToChange) {
+              SimSettingsHelper.setServiceOnCard('outgoingData',
+                newCardIndex);
+            } else {
+              this.selectedIndex = prevCardIndex;
+            }
+          });
```
zbraniecki: Here it seems to work well, because the event is asynchronous anyway (reacts to blur).

apps/settings/js/panels/sound/tone_manager.js
```diff
@@ -287,7 +289,8 @@ define(function(require) {
       activity.onerror = function() {
         console.log(this.error);
         if (this.error.name === 'NO_PROVIDER') {
-          alert(navigator.mozL10n.get('no-ringtone-app'));
+          navigator.mozL10n.formatValue('no-ringtone-app').then(msg =>
+            alert(msg));
```
zbraniecki: here it just happens later, no confusion beyond 5s lag.

apps/settings/js/panels/wifi/wifi_wps.js
```diff
@@ -72,7 +71,9 @@ define(function(require) {
 
         req.onsuccess = function() {
           if (method === 'myPin') {
-            alert(_('wpsPinInput', { pin: req.result }));
+            navigator.mozL10n.formatValue('wpsPinInput', {
+              pin: req.result
+            }).then(msg => alert(msg));
```
zbraniecki: here it makes sense to separate codepaths - if not myPin then set `inProgress = true;` and fire `onSuccess();`, otherwise do those two in a callback to formatValue after alert.

Does it sound good to you @evelynhung ?

apps/settings/js/modules/date_time.js
evelynhung: Got it, thanks for the explanation! :)

apps/settings/elements/firefox_accounts.html
evelynhung: Okay, let's keep it this time, and file another bug to remove all content in HTML. ;-)

apps/settings/js/modules/panel_utils.js
evelynhung: Cool!

apps/settings/js/panels/about/update_check.js
evelynhung: Yep, it's a good idea to disable the button before formatValue resolves, I assume usually users won't notice this disabled state though. Would it be that much a 5s delay?

apps/settings/js/panels/hotspot_wifi_settings/panel.js
evelynhung: Disabling the OK button is a better solution. I'm surprised the design didn't go to this way. :-/

apps/settings/js/panels/languages/languages.js
evelynhung: Okay, totally understand. Thanks for being thoughtful. :)

apps/settings/js/panels/about/update_check.js
zbraniecki: absolutely not. Actually, at this moment there is no way for it to take any time because all this code will be fired after we already translated DOM and displayed the UI (which means that we have all resources in memory).

In the future we would like to consider switching to runtime language fallbacking which means that in error scenarios (when for some reason the string cannot be retrieved/resolved in user selected language) we will load fallback locale and use the string from there, this can take ~100-200ms.

But I don't ever expect a scenario when retrieving a string takes longer than than, and in all non-error scenarios it is literally below 5ms (timeout of a Promise).

So, while I like to think of formatValue as "it should work even if it would take 5s", I don't believe it's a realistic worry.

apps/settings/js/panels/simpin/simpin.js
```diff
@@ -72,8 +72,7 @@ define(function(require) {
         simPinHTMLs.push(
           this.simPinView({
             'simIndex': index.toString(),
-            'simName': _('simPinWithIndex', { 'index': simPinIndex }),
-            'changeSimLabel': _('changeSimPin')
+            'simPinArgs': JSON.stringify({ 'index': simPinIndex }),
```
evelynhung: Please remove the trailing comma.

apps/settings/js/panels/wifi/wifi_wps.js
evelynhung: Yes, we should do in this way.

apps/settings/js/bluetooth.js
evelynhung: Oh! You remind me that here is a legacy code for Bluetooth API version 1. Sorry I didn't think of it. I asked Ian who wrote the newer version (bluetooth/panel.js#L593-L599) and he said UX reviewed that part. So feel free to ignore my previous comments.

apps/settings/js/bluetooth.js
evelynhung: Same here! The code will be run on v2.2, not on master anymore. (We should remove this file!)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=505ed23a6b746fc4c53552c1e5f3f1115639f9fe

apps/settings/locales/settings.en-US.properties
stasm: AFAIR, this was an edge case in which the translation of this string was missing in Arabic and on buildtime we would fall back to English.  The parenthesis would take the directionality of the surrounding Arabic text.

Now that I think about this, it almost seems like the best solution would be to have L20n's HTML bindings do something like this:

```javascript
const sheet = doc.head.appendChild(document.createElement('style')).sheet;
sheet.insertRule('[data-l10n-id] { unicode-bidi: isolate; }', 0);
```

zbraniecki: so, I tried setting `unicode-bidi: isolate` on `<h2>` it and it didn't work for some reason. I decided to instead of that use `<bdi>` tag inside `<h2>` and that works well.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ba9582259d2eeca5b468ee85d5313f9a0d3ed194

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a558fb756789e05b930294d8cdb0e13e0d4b432d

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=aa2cb1c376c37a7e8d698acb4eab0b5dae1861e5

apps/settings/locales/settings.en-US.properties
```diff
@@ -546,31 +540,8 @@ hotspotSettings-header={{hotspotSettings}}
...
 
-# LOCALIZATION NOTE
-#
-# [For localizers]
-# We can't render these strings correctly for RTL locales, so we have to put LRE (&#x202a;)
-# and PDF (&#x202c;) entities to these translations to force the correct direction.
-#
-# RTL locales should have a translation similar to this:
-# hotspot-wpa-psk.innerHTML= &#x202a;WPA (TKIP)&#x202c;
-# hotspot-wpa2-psk.innerHTML= &#x202a;WPA2 (AES)&#x202c;
```
stasm: Why are you removing this comment?  Most likely, RTL languages won't translate these acronyms, and without the unicode marks, the parens will break.  Should we add `unicode-bidi: isolate` to CSS to prevent this?

zbraniecki: So, once again, I tried `unicode-bidi:isolate` for `mmsTitle` and it did not fix it there. Can I just put it in `<bdi>`?

zbraniecki: Yeah, I don't get it: http://jsfiddle.net/7qopv17s/

zbraniecki: I readded the comment. `unicode-bidi: isolate` doesn't seem to work at all on anything, and also putting `<bdi>` inside the `<option>` doesn't seem to work so I for this case we're stuck with `&#x202a/c`.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a39c7a54649ccf52ce08cdc52f9b17040b28ad50

PR #31109, opened 2015-07-26
Bug 1181571 - Fix lockscreen delay issue

cr:
This is a manually verified fix and prototype tests.

I need some help with the tests. I have trouble testing code paths that resolve asynchronously after method calls have returned, for example in setTimeout() or promises. Is there a test pattern for this case?

### DIFF
--- a/apps/system/lockscreen/js/lockscreen_inputpad.js
+++ b/apps/system/lockscreen/js/lockscreen_inputpad.js
@@ -117,8 +117,8 @@
         }
         // Cancel the default action of <a>
         evt.preventDefault();
+        // handlePassCodeInput triggers updatePassCode
         this.handlePassCodeInput(key);
-        // .handlePassCodeInput() triggers .updatePassCode()
         break;
     }
   };
--- a/apps/system/lockscreen/js/lockscreen_inputpad.js
+++ b/apps/system/lockscreen/js/lockscreen_inputpad.js
@@ -90,12 +90,19 @@
 
   LockScreenInputpad.prototype.handleEvent = function(evt) {
     switch (evt.type) {
+      // The event flow from lockscreen.js is:
+      // on pass code fail:
+      //   - 'validationfailed' -> (validation timeout) -> 'validationreset'
+      // on pass code success:
+      //   - 'validationsuccess'
       case 'lockscreen-notify-passcode-validationfailed':
         this.states.passCodeErrorTimeoutPending = true;
         this.updatePassCodeUI();
         break;
       case 'lockscreen-notify-passcode-validationreset':
       case 'lockscreen-notify-passcode-validationsuccess':
+        // Currently both validationreset and validationsuccess
+        // just need to reset Inputpad's internal state.
         this.states.passCodeEntered = '';
         this.states.passCodeErrorTimeoutPending = false;
         this.updatePassCodeUI();
--- a/apps/system/lockscreen/js/lockscreen_inputpad.js
+++ b/apps/system/lockscreen/js/lockscreen_inputpad.js
@@ -28,22 +28,21 @@
   var LockScreenInputpad = function(lockScreen) {
     this.lockScreen = lockScreen;
     this.configs = {
-      passcodeDecoratingTimeout: 500,
-      padVibrationDuration: 50,
+      padVibrationDuration: 50
     };
     this.states = {
       // Keep in sync with Dialer and Keyboard vibration
       padVibrationEnabled: false,
       passCodeEntered: '',
-      passcodeErrorCounter: 0
+      passCodeErrorTimeoutPending: false
     };
   };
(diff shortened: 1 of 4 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=fbe8085116264cb3dd9d28e159eac98b687d74dd

apps/system/lockscreen/js/lockscreen_inputpad.js
```diff
@@ -117,11 +115,11 @@
         if (!key) {
           break;
         }
-
         // Cancel the default action of <a>
         evt.preventDefault();
         this.handlePassCodeInput(key);
-      break;
+        // .handlePassCodeInput() triggers .updatePassCode()
```
snowmantw: Remove the commented code.

cr: It's not code, just a comment, and I think it's necessary to follow the logic without looking any further.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=35fb616d74fa4bfea1846d6e3274ba2fb6389116

apps/system/lockscreen/js/lockscreen_inputpad.js
```diff
@@ -92,20 +91,19 @@
...
       case 'lockscreen-notify-passcode-validationfailed':
-        return this.decorateErrorPasscodeUI()
-          .then(() => {
-            this.removeErrorPasscodeUI();
-            this.updatePassCodeUI();
-          });
+        this.states.passCodeErrorTimeoutPending = true;
+        this.updatePassCodeUI();
+        break;
+      case 'lockscreen-notify-passcode-validationreset':
```
snowmantw: Although I can understand the possible reason why we need to have two events for the same effects, I think it is better to leave some comments about that. Since according the event flow it's quite strange without knowing the current status of LockScreen:

  validationfailed ----> validationreset  === validationsuccess

So I think we should add few lines here, to explain when the reset would be fired. That is, after the failure and we need to reset the UI.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f51d134acad733df034e2ff268bf8f7b61fc3c31

apps/system/test/unit/lockscreen/lockscreen_test.js
```diff
@@ -297,6 +297,75 @@ suite('system/LockScreen >', function() {
...
+            return e.type ===
+              'lockscreen-notify-passcode-validationfailed';
+          })),
+        'validation fail does not trigger validationfailed event');
+      stubDispatch.restore();
+    });
+    //TODO: test validationreset event triggered async after timeout
+
+    /* TODO: promise resolves after checkPassCode() returns.
+     * TODO: how can we check this?
```
snowmantw: You can use `test('...', function(done) {...})` and then put the `done` in the promise's next step. Namely:

    test('...', function(done) {
        function() {
            return Promise.resolve(true).then(done).catch(done);
            // The first `done` ends this  asynchronous test.
            // The second `done` catch the error and make the test failed to present that
        }
    } );

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ac16d3cf4b0d510b257571eb5405c694260b6f02

cr:
Rebased to today's master.

apps/system/test/unit/lockscreen/lockscreen_test.js
cr: This results in Error: done() invoked with non-Error: true.

cr: When I change from ```.then(done)``` to ```.then(function(){done()})```,  the result is *Error: done() called multiple times* although I can prove by adding a ```console.log()```-call that done is called exactly once, but after the test reported having failed. Looks like there's an implicit ```done()``` somewhere after the test returns.

I suspect that the suggested pattern may not be the solution for this particular case.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=bed700e1d592917e18b41f43746df8780a4e8d99

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=09263c146580291e539ca7ccbe420e95face6080

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=785b58fd5d5b08208f52a648f5d59ba9204eddc1

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=28b85c0b4bcd8f3b8144121393ab24685d9272c1

PR #31114, opened 2015-07-27
Bug 1187945 - Intermittent failure in test_dialer_add_contact.py in self.wait_for_done(return_contacts), NoSuchWindowException

mwargers:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=91be4c1d359bef162ea636e881ba92abc632b18e

mwargers:
How do you use self.apps.displayed_app.origin? Then Contact class would need an origin property too, no?
I'd rather do this later, because I don't know currently how to solve this.

PR #31115, opened 2015-07-27
Bug 1187979 - Improve Inputpad's touch and visual feeback

cr:
Inputpad looks and feels wonky compared to the slick look and feel of
the keyboard panel used while setting the pass code in Settings. Much of
the reason comes from that it only vibrates on key release, the other
that it does not behave well when the user is sliding a finger over the
keys. There is also a considerable delay between touching a key and a
visual reaction to the extent that it's skipped when typing mildly fast.

This prototype patch addresses all of the above:
- make InputPad vibrate instantly on touch
- make InputPad respect user's touchmove
- visualize key presses instantly

### DIFF
--- a/apps/system/lockscreen/js/lockscreen_inputpad.js
+++ b/apps/system/lockscreen/js/lockscreen_inputpad.js
@@ -117,6 +117,7 @@
         this.updatePassCodeUI();
         break;
       case 'touchstart':
+        // TODO: ignore while on error timeout
         touch = evt.changedTouches[0];
         target = document.elementFromPoint(touch.clientX, touch.clientY);
         key = this._keyForTarget(target);
@@ -133,6 +134,7 @@
         }
         break;
       case 'touchmove':
+        // TODO: ignore while on error timeout
         touch = evt.changedTouches[0];
         target = document.elementFromPoint(touch.clientX, touch.clientY);
         key = this._keyForTarget(target);
@@ -150,6 +152,7 @@
         break;
       case 'touchend':
       case 'click':
+        // TODO: ignore while on error timeout
         if (evt.type == 'touchend') {
           evt.preventDefault();  // prevent the 'click'
           touch = evt.changedTouches[0];
@@ -170,49 +173,51 @@
     }
   };
 
+  LockScreenInputpad.prototype._anchorForTarget = function(target) {
+    // Find ancestorial anchor that has info on the key
+    // Current structure: a > div > span
+    try {
+      if (target.tagName !== 'A') {
+        target = target.parentNode;
+      }
+      if (target.tagName !== 'A') {
+        target = target.parentNode;
+      }
+      if (target.tagName == 'A') {
+        return target;
+      }
+      return null;
+    }
+    catch (e) {
+      return null;
+    }
+  };
+
   LockScreenInputpad.prototype._keyForTarget = function(target) {
-    // find ancestorial a
-    // current structure: a > div > span
-    if (target.tagName !== 'A') {
-      target = target.parentNode;
+    var anchor = this._anchorForTarget(target);
+    if (anchor) {
+      return anchor.dataset.key;
+    } else {
+      return null;
     }
-    if (target.tagName !== 'A') {
-      target = target.parentNode;
-    }
-    if (target.tagName !== 'A') {
-      return undefined;
-    }
-    return target.dataset.key;
   };
 
   LockScreenInputpad.prototype._makeKeyActive = function(target) {
-    // find ancestorial a
-    // current structure: a > div > span
-    if (target.tagName !== 'A') {
-      target = target.parentNode;
+    var anchor = this._anchorForTarget(target);
+    if (anchor) {
+      anchor.classList.add('active-key');
+    } else {
+      throw Error('lsip_makeKeyActive called with non-key node');
     }
-    if (target.tagName !== 'A') {
-      target = target.parentNode;
-    }
-    if (target.tagName !== 'A') {
-      return;
-    }
-    target.classList.add('active-key');
   };
 
   LockScreenInputpad.prototype._makeKeyInactive = function(target) {
-    // find ancestorial a
-    // current structure: a > div > span
-    if (target.tagName !== 'A') {
-      target = target.parentNode;
+    var anchor = this._anchorForTarget(target);
+    if (anchor) {
+      anchor.classList.remove('active-key');
+    } else {
+      throw Error('lsip_makeKeyInactive called with non-key node');
     }
-    if (target.tagName !== 'A') {
-      target = target.parentNode;
-    }
-    if (target.tagName !== 'A') {
-      return;
-    }
-    target.classList.remove('active-key');
   };
 
   LockScreenInputpad.prototype.dispatchEvent = function(evt) {
--- a/apps/system/lockscreen/js/lockscreen_inputpad.js
+++ b/apps/system/lockscreen/js/lockscreen_inputpad.js
@@ -99,13 +99,13 @@
   };
 
   LockScreenInputpad.prototype.handleEvent = function(evt) {
-    var touch, target, key;
+    /* The event flow from lockscreen.js is:
+     * on pass code fail:
+     *   - 'validationfailed' -> (validation timeout) -> 'validationreset'
+     * on pass code success:
+     *   - 'validationsuccess'
+     */
     switch (evt.type) {
-      // The event flow from lockscreen.js is:
-      // on pass code fail:
-      //   - 'validationfailed' -> (validation timeout) -> 'validationreset'
-      // on pass code success:
-      //   - 'validationsuccess'
       case 'lockscreen-notify-passcode-validationfailed':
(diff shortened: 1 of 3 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=24fd1bf25e3deb6e204e3fe2ce4d69daa8a4e5d6

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d03ea06eac1cb6cc4bda67d59c10025715ce200e

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=074b856985795c6516479d6901bfd7c27d9570ff

cr:
Rebased to today's master, local tests passing.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=04db3f4b74ae472f906fb500aa45f2fd5f3d22be

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=226afb0d2a85f2751d4d8c7f269950277711f36a

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c885d13f60dd2cea471a47045ae226a25bf18fbc

apps/system/lockscreen/js/lockscreen_inputpad.js
```diff
@@ -111,25 +116,111 @@
       case 'lockscreen-inputappclosed':
         this.updatePassCodeUI();
         break;
+      case 'touchstart':
+        // TODO: ignore while on error timeout
+        touch = evt.changedTouches[0];
+        target = document.elementFromPoint(touch.clientX, touch.clientY);
```
snowmantw: Is it possible to use `evt.target` instead of querying element from the coordination?

https://developer.mozilla.org/en-US/docs/Web/Events/touchstart

apps/system/lockscreen/js/lockscreen_inputpad.js
```diff
@@ -111,25 +116,111 @@
...
+        }
+        this._makeKeyActive(target);
+        this.lastTouched = target;
+        if (this.states.padVibrationEnabled &&
+          !this.states.passCodeErrorTimeoutPending) {
+          navigator.vibrate(this.configs.padVibrationDuration);
+        }
+        break;
+      case 'touchmove':
+        // TODO: ignore while on error timeout
```
snowmantw: Is this still a TODO or just a comment need to be cleared?

apps/system/lockscreen/js/lockscreen_inputpad.js
```diff
@@ -111,25 +116,111 @@
...
+        touch = evt.changedTouches[0];
+        target = document.elementFromPoint(touch.clientX, touch.clientY);
```
snowmantw: `evt.target` ?

apps/system/lockscreen/js/lockscreen_inputpad.js
```diff
@@ -111,25 +116,111 @@
...
+        key = this._keyForTarget(target);
+        if (target.textContent !== this.lastTouched.textContent) {
```
snowmantw: Could you add some comments to make the intention here more clear?

snowmantw: I mean, an overall comment to describe what you're doing here, with these conditional branches.

apps/system/lockscreen/js/lockscreen_inputpad.js
```diff
@@ -47,8 +47,12 @@
     this.passcodeCode = document.getElementById('lockscreen-passcode-code');
     this.passcodePad = document.getElementById('lockscreen-passcode-pad');
     this.emergencyCallBtn = this.passcodePad.querySelector('a[data-key=e]');
+    this.lastTouched = null;
```
snowmantw: May need to describe what the variable is for in brevity.

apps/system/lockscreen/js/lockscreen_inputpad.js
```diff
@@ -111,25 +116,111 @@
...
+      case 'touchend':
       case 'click':
-        var key = evt.target.dataset.key;
-        if (!key &&
-            ('div' === evt.target.tagName.toLowerCase() &&
-             'a' === evt.target.parentNode.tagName.toLowerCase())
-           ) {
-          key = evt.target.parentNode.dataset.key;
+        // TODO: ignore while on error timeout
+        if (evt.type == 'touchend') {
```
snowmantw: No `==` in Gaia is a hinting requirement.

apps/system/lockscreen/js/lockscreen_inputpad.js
```diff
@@ -111,25 +116,111 @@
...
+          evt.preventDefault();  // prevent the 'click'
```
snowmantw: Instead of calling `preventDefault` just for distinguishing these two events, to deal with these two events by two `switch` branch is better to me.

apps/system/lockscreen/js/lockscreen_inputpad.js
```diff
@@ -111,25 +116,111 @@
...
+          touch = evt.changedTouches[0];
+          target = document.elementFromPoint(touch.clientX, touch.clientY);
```
snowmantw: `evt.target` ?

apps/system/lockscreen/js/lockscreen_inputpad.js
snowmantw: And in such case if you concern the duplicated part, make them as a function and call it twice. Or, with such tiny duplicated code is okay I think.

apps/system/lockscreen/js/lockscreen_inputpad.js
```diff
@@ -111,25 +116,111 @@
...
+        } else {  // 'click'
+          target = evt.target;
+        }
+        key = this._keyForTarget(target);
+        if (this.lastTouched) {
```
snowmantw: I still feel it is unclear and need a comment to describe what's the relation between a `lastTouched` and key active/inactive. Maybe re-naming can solve that, but a comment may be better.

apps/system/lockscreen/js/lockscreen_inputpad.js
cr: I started out with .target, but unlike changedTouches it does not stay in sync with where the finger is touching on touch moves. IIRC, .target remains where the touch first went down. Hence I went for the alternative that always works.

apps/system/lockscreen/js/lockscreen_inputpad.js
cr: On touchmove, .target fails to report the current position under the finger. See comment above.

apps/system/lockscreen/js/lockscreen_inputpad.js
cr: Conscious design. See comment above.

apps/system/lockscreen/js/lockscreen_inputpad.js
cr: Added this:
```
    /* lastTouched
     * Keeps track of the key that was under the finger during
     * the previous 'touchmove' event. UI must be redrawn only if
     * the finger moved to a different key.
     */
    this.lastTouched = null;
```

apps/system/lockscreen/js/lockscreen_inputpad.js
cr: Added comments throughout the code.

apps/system/lockscreen/js/lockscreen_inputpad.js
cr: Code was deleted.

apps/system/lockscreen/js/lockscreen_inputpad.js
cr: I chose to go for the minor duplication, because else I feel that I should be writing tests for the new function. A future iteration may optimize the code in this respect.

apps/system/lockscreen/js/lockscreen_inputpad.js
cr: Added comments throughout the code. Refactored the variable to .lastTouchedKey.

In short: it's used to determine whether the finger has moved to a new key since the last event. If and only if that is the case, lastTouchedKey must be made inactive, and the new key must be made active.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b7752fc791d9b342c4c7ce4eeca8b4e9b942c31a

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=33e44b57a00a0c843ead79b5abd8a32b54600cbe

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=aca1184c937b994d959d1effbfa0fb3ec9884ec6

PR #31133, opened 2015-07-28
Bug 1175744 - Intermittent create_event_test.js

millermedeiros:
not sure if that will fix or not since I wasn't able to run the tests locally but giving it a shot.. if it doesn't solve the problem I think we should disable the tests.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=030d838b84200b3d9b7c87d46bd4dd5751d401e3

millermedeiros:
not really the cause of failure.

PR #31163, opened 2015-07-29
Bug 816783 - [email] Autocomplete known contacts when user manually types

jrburke:
This is an extraction of the autocomplete from the front end's conversation branch that works on current master.

It only uses the Contacts API for matches. Bug 1188710 tracks expanding that data source over time to include email addresses that the email app knows about.
- autocomplete.js is the main new file.
- compose.html did not change that much, using ?w=1 to view the diff on github would help show the difference: a new div was added to wrap all the envelope lines, the autocomplete tag was added as a sibling to that div, and a couple elements got some new CSS classes.
- element.js and config.js changed to allow custom elements to be loaded from different directories instead of as only direct children of the cards/ repo.

Did a quick test of RTL, using the mirrored locale with arabic keyboard and entry, and worked as expected, with content aligned right.

I will follow up on any screen reader changes in bug 1189092.

### DIFF
--- a/apps/email/build/email.build.js
+++ b/apps/email/build/email.build.js
@@ -47,6 +47,10 @@
         'iframe_shims',
         'cards/editor_mixins',
 
+        // gesture_detector used by both compose and message_reader layer, so
+        // just include it in the base layer.
+        'shared/js/gesture_detector',
+
         // Bundle most likely card
         'cards/message_list'
       ]
--- a/apps/email/js/cards/cmp/autocomplete.js
+++ b/apps/email/js/cards/cmp/autocomplete.js
@@ -1,0 +1,548 @@
+'use strict';
+define(function(require) {
+
+var GestureDetector = require('shared/js/gesture_detector'),
+    itemTemplate = require('tmpl!./autocomplete_item.html'),
+    mozL10n = require('l10n!'),
+    regExpEscape = require('regExpEscape'),
+    transitionEnd = require('transition_end'),
+    validTypes = {
+      email: true,
+      text: true
+    };
+
+/**
+ * Converts a text string into a span and some text nodes that are placed
+ * as the children of node. The span contains the match for the queryRegExp.
+ * @param  {RegExp} queryRegExp The regexp whose match is highlighted via a
+ * span in the target text.
+ * @param  {String} text        The target text to find the match.
+ * @param  {Element} node       The element that will display the matched text.
+ */
+function formatMatchString(queryRegExp, text, node) {
+  // Just highlight the first match.
+  var match = queryRegExp.exec(text);
+
+  if (match) {
+    var index = match.index,
+        endIndex = index + queryRegExp.source.length;
+
+    var startString = text.substring(0, index);
+    if (startString) {
+      node.appendChild(document.createTextNode(startString));
+    }
+
+    var matchString = text.substring(index, endIndex);
+    if (matchString) {
+      var span = document.createElement('span');
+      span.classList.add('highlight');
+      span.textContent = matchString;
+      node.appendChild(span);
+    }
+
+    var endString = text.substring(endIndex);
+    if (endString) {
+      node.appendChild(document.createTextNode(endString));
+    }
+  } else {
+    node.textContent = text;
+  }
+}
+
+/**
+ * Given a node, if it or an ancestor has the CSS className, return the node
+ * with that className. Returns null if not a match.
+ * @param  {Node} node
+ * @return {Node}
+ */
+function findAncestorWithClass(node, className) {
+  for (; node; node = node.parentNode) {
+    if (node.classList && node.classList.contains(className)) {
+      return node;
+    }
+  }
+
+  return null;
+}
+
+/**
+ * Given a node, if it or an ancestor is an autocomplete item in the
+ * autocomplete list, return the autcomplete item node. Returns null if not a
+ * match.
+ * @param  {Node} node
+ * @return {Node}
+ */
+function getAutocompleteItem(node) {
+  return findAncestorWithClass(node, 'cmp-autocomplete-item');
+}
+
+/**
+ * Autocomplete custom element that contains a list of elements that show the
+ * autocomplete matches.
+ *
+ * This custom element relies on a CSS class-based API to bind to nodes outside
+ * of this autocomplete. The CSS/HTML structure expected by this autocomplete:
+ *
+ * - An element with the CSS class of "cmp-autocomplete-origin" that is the the
+ * master scroll area for the view. It should be an ancestor node to the
+ * autocomplete and also contain the 'cmp-autocomplete-input-list' element that
+ * has all the input elements. This element is used as the "origin" for the
+ * style.top placement of the autocomplete list.
+ *
+ * - An element that holds all the input fields, with the CSS class of
+ * "cmp-autocomplete-input-list". That inputArea will be translateY'd to make
+ * sure the current input focus is visible at the top of the view.
+ *
+ * - Individual rows that hold an autocomplete must specify a
+ * "cmp-autocomplete-input-label" CSS class, so that for a given input, that
+ * parent can be found to be marked as "active". This allows just that row to
+ * change its alignment so its label and + icon are aligned at the bottom with
+ * the text input.
+ *
+ * The autocomplete custom element also expects a "data-source" HTML attribute
+ * to be set on it, which indicates the module ID to use as the data source for
+ * the matches.
+ *
+ * The owner of this custom element instance (typically a view card), should
+ * override the getExistingEntries() method on the instance to return the
+ * current entries, to avoid dupes showing up in subsequent autocomplete
+ * matches.
+ */
+return [
+  require('../base')(),
+  {
+    createdCallback: function() {
+      //TODO: master hack only. In email-model-instance branch, this is not
+      //needed. In the branch, there is a separate 'base' that elements can
+      //mix in that is not assumed to be a card. For now, this is an equivalent
+      //result.
+      this.classList.remove('card');
+
+      this.lastTranslateY = 0;
+      this.inputAreaTranslated = Promise.resolve();
+
+      this.runQuery = this.runQuery.bind(this);
+      this.onFocus = this.onFocus.bind(this);
+      this.onDocumentEvent = this.onDocumentEvent.bind(this);
+      this.onInput = this.onInput.bind(this);
+
+      // Find (when cached HTML) or create list to hold the autocomplete matches
+      // Put the list inside the autocomplete element instead of off the
+      // document, as it should be within a given UI card, in case in-app
+      // overlay UIs like confirm dialogs need to be shown.
+      this.list = this.querySelector('.cmp-autocomplete-list');
+      if (!this.list) {
+        this.list = document.createElement('ul');
+        this.list.classList.add('cmp-autocomplete-list');
+        this.appendChild(this.list);
+      }
+      this.list.addEventListener('keydown', this.onMatchKeyDown.bind(this));
+      this.list.addEventListener('click', this.onMatchClick.bind(this));
+      this.list.innerHTML = '';
+      this.hideList();
+
+      // Find parent that is scrollable and is the origin for measurements.
+      var parentNode = this;
+      while((parentNode = parentNode.parentNode)) {
+        if (parentNode.classList.contains('cmp-autocomplete-origin')) {
+          this.originNode = parentNode;
+          break;
+        }
+      }
+
+      // Hold on to the element that holds all the input fields, to use it for
+      // sizing and setting scroll behavior amongst the inputs while the
+      // autocomplete is active.
+      this.inputArea = this.originNode
+                       .querySelector('.cmp-autocomplete-input-list');
+
+      // Listen for transitionend events on the inputArea, since translateY with
+      // a transition is used to move the input to the top of the view. Want to
+      // wait for the transition to end before showing autocomplete results.
+      // This allows the user to better track where the addresses have gone,
+      // and feels smoother.
+      transitionEnd(this.inputArea, this._translateYTransitionEnd.bind(this));
+
+      // Bind to events triggered by input field children. Need to mark the
+      // elements with an autocomplete class, and do not want to target just
+      // any input in the inputArea, so using explicit per-element bindings
+      // instead of event delegation.
+      var nodes = this.inputArea.querySelectorAll('input');
+      Array.from(nodes, function(inputNode) {
+        if (!validTypes.hasOwnProperty(inputNode.type)) {
+          return;
+        }
+
+        // Add a class to the input to mark it as participating in autocomplete.
+        // This is used to know if taps in them should be disregarded for
+        // onDocumentEvent purposes.
+        inputNode.classList.add('cmp-autocomplete-input');
+
+        inputNode.addEventListener('focus', this.onFocus);
+        inputNode.addEventListener('input', this.onInput);
+      }, this);
+
+      // Use a GestureDetector to know when user tries to scroll the inputArea
+      // element to see addresses that have been scrolled out of the way for
+      // autocomplete maximum real estate. No need to clean this up later, the
+      // destruction of the card using the autocomplete is enough: the watched
+      // element is destroyed along with the autocomplete.
+      this.detector = new GestureDetector(this.inputArea);
+      this.inputArea.addEventListener('pan', (event) => {
+        this.hide();
+      });
+      this.detector.startDetecting();
+
+      // Set up data source.
+      var dataSourceId = this.dataset.source;
+      if (dataSourceId) {
+        this.bindDataSource(dataSourceId);
+      }
+    },
+
+    /**
+     * Custom element callback when element is placed in the document.
+     */
+    attachedCallback: function() {
+      this.onResize = this.onResize.bind(this);
+      window.addEventListener('resize', this.onResize, false);
+
+      // Explicitly want to listen during the capture phase, since most code
+      // listens in bubble phase, and may stopPropagation. For autocomplete
+      // hiding though we always want it to disappear if event is outside of
+      // the interesting areas.
+      document.addEventListener('click', this.onDocumentEvent, true);
+      document.addEventListener('focus', this.onDocumentEvent, true);
+    },
+
+    /**
+     * Custom element callback when element is removed from the document.
+     */
+    detachedCallback: function() {
+      window.removeEventListener('resize', this.onResize, false);
+      document.removeEventListener('click', this.onDocumentClick, true);
+      document.addEventListener('focus', this.onDocumentEvent, true);
+    },
+
+    /**
+     * Triggered by input focus events.
+     */
+    onFocus: function(event) {
+      this.inputNode = event.target;
+
+      // If existing label node, be sure to remove the active class from it. In
+      // cases where the focus goes from one input to the other, hide() does not
+      // run to clear this state.
+      if (this.inputLabelNode) {
+        this.inputLabelNode.classList.remove('cmp-autocomplete-active');
+      }
+
+      this.inputLabelNode = findAncestorWithClass(this.inputNode,
+                                                'cmp-autocomplete-input-label');
+      this.inputLabelNode.classList.add('cmp-autocomplete-active');
+
+      this.query = this.inputNode.value.trim();
+      this.runThrottledQuery();
+    },
+
+    /**
+     * Triggered by any click or focus in the document. Used to determine if the
+     * autocomplete should be hidden.
+     */
+    onDocumentEvent: function(event) {
+      var target = event.target;
+
+      // If no classList, then it is not an element, and should be discarded.
+      if (!target.classList) {
+        this.hide();
+        return;
+      }
+
+      // If an autocomplete input, do not do anything, handled by other code in
+      // this module.
+      if (target.classList.contains('cmp-autocomplete-input')) {
+        return;
+      }
+
+      // Also ignore if it is part of an autocomplete-item.
+      if (getAutocompleteItem(target)) {
+        return;
+      }
+      this.hide();
+    },
+
+    /**
+     * Triggered by input element input events.
+     */
+    onInput: function(event) {
+      this.query = this.inputNode.value.trim();
+      this.runThrottledQuery();
+    },
+
+    /**
+     * Handles window resizes. Could be triggered by keyboard size changes.
+     */
+    onResize: function(event) {
+      if (this.inputNode && !this.isHidden) {
+        this.positionElements();
+      }
+    },
+
+    onMatchKeyDown: function(event) {
+      if (event.key === 'Enter' && !this.isHidden) {
+        this.onMatchClick(event);
+      }
+    },
+
+    /**
+     * Handles taps inside the autocomplete. The tap could occur outside a match
+     * element, in which case means the autocomplete should just close.
+     */
+    onMatchClick: function(event) {
+      var node = getAutocompleteItem(event.originalTarget);
+
+      if (node && node.match) {
+        event.preventDefault();
+        this.dispatchEvent(new CustomEvent('autocompleteSelected', {
+          detail: {
+            match: node.match,
+            inputNode: this.inputNode
+          }
+        }));
+      }
+    },
+
+    /**
+     * Loads the dataSource module.
+     * @param  {String} dataSourceId The module ID for the data source.
+     */
+    bindDataSource: function(dataSourceId) {
+      require([dataSourceId], (dataSource) => {
+        this.dataSource = dataSource;
+        if (this.query) {
+          this.runThrottledQuery();
+        }
+      });
+    },
+
+    /**
+     * Called to trigger finding matches from the dataSource. Set on a timeout
+     * to avoid multiple fast keystrokes from triggering too many rapid queries.
+     */
+    runThrottledQuery: function() {
+      if (!this.query || !this.dataSource) {
+        this.list.innerHTML = '';
+        this.positionAndHide();
+        return;
+      }
+
+      if (!this.timedQueryId) {
+        // The choice of delay tries to balance sending too many queries to the
+        // contacts API vs giving a feeling of responsiveness for the reaction
+        // of the autocomplete. This balance may be affected by the total number
+        // of contacts.
+        this.timedQueryId = setTimeout(this.runQuery, 350);
+      }
+    },
+
+    /**
+     * The owner of this autocomplete instance should set this value if they
+     * have extra spacing around the input that should be maintained for better
+     * visual display, and to avoid small scroll janks on initial positioning
+     * of the input element if it needs to be scrolled to the top of the view.
+     */
+    verticalSpace: 0,
+
+    /**
+     * This method can be overriden by the owner of the autocomplete to return
+     * existing entries already attached to the email. The result should be an
+     * array of objects that have "name" and "address" properties.
+     * @return {Array}   Array of existing entries.
+     */
+    getExistingEntries: function() {
+      return [];
+    },
+
+    /**
+     * Runs the query to find matches from the dataSource.
+     */
+    runQuery: function() {
+      this.timedQueryId = 0;
+
+      this.dataSource(this.query).then((result) => {
+        // Protect against old or out of date queries
+        if (result.query !== this.query) {
+          return;
+        }
+
+        var existingEntries = this.getExistingEntries();
+
+        var contacts = result.contacts;
+
+        this.list.innerHTML = '';
+
+        var queryRegExp = new RegExp(regExpEscape(this.query), 'i');
+
+        contacts.forEach((contact) => {
+          // Only care about contacts with an email address(es).
+          var emails = contact.email;
+          if (!emails || !emails.length) {
+            return;
+          }
+
+          // Contact name could be an array, just use the first one.
+          var name = contact.name;
+          if (Array.isArray(name)) {
+            name = name[0];
+          }
+
+          // Emails are an array, each one should get its own entry in the
+          // autocomplete.
+          emails.forEach((email) => {
+            var address = email.value;
+
+            // Do not add it if the email address is already an entry.
+            var hasExisting = existingEntries.some(function(entry) {
+              return entry.address === address;
+            });
+            if (hasExisting) {
+              return;
+            }
+
+            var node = itemTemplate.cloneNode(true),
+                detailsNode = node
+                               .querySelector('.cmp-autocomplete-item-details'),
+                nameNode = node.querySelector('.cmp-autocomplete-name'),
+                emailNode = node.querySelector('.cmp-autocomplete-email');
+
+            // Set the ariaLabel for the match, for screen reader use only.
+            mozL10n.setAttributes(detailsNode, 'compose-autocomplete-match',
+                                  { name: name, address: address });
+
+            formatMatchString(queryRegExp,
+                              name || address,
+                              nameNode);
+            formatMatchString(queryRegExp, address|| '', emailNode);
+
+            // Store the match data on the node for ease of access when items
+            // are tapped.
+            var match = {
+              name: name,
+              address: address
+            };
+            node.match = match;
+
+            this.list.appendChild(node);
+          });
+        });
+
+        // Always reshow on a query, since each keystroke has the possibility of
+        // breaking the input field to a new line.
+        if (this.list.children.length) {
+          this.show();
+        } else {
+          this.positionAndHide();
+        }
+      });
+    },
+
+    /**
+     * Positions the input area to the top of the view and sizes the list to
+     * take up as much view space as possible.
+     */
+    positionElements: function() {
+      // Always calculate the originTop since the whole card that contains the
+      // origin could be scrolled for some reason, even by an accident.
+      var originTop = this.originNode.getBoundingClientRect().top;
+
+      // Want to use the input element, and not a parent of it, for placement of
+      // the autocomplete, since the input element can move to different lines
+      // as typing occurs, where the parent's origin would not change, and would
+      // be a bad reference point to use.
+      var inputRect = this.inputNode.getBoundingClientRect();
+
+      // Scroll the area so that the input item is now at the top of the
+      // address area.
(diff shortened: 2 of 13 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=411a960402c113922a2b7e1143c0b0aa5ca0078d

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=414e9cb1222ae88470c2acf8d400421416fcd6b3

apps/email/js/cards/cmp/autocomplete.js
```diff
@@ -0,0 +1,476 @@
...
+  } else {
+    node.textContent = text;
+  }
+}
+
+/**
+ * Autocomplete custom element that contains a list of elements that show the
+ * autocomplete matches.
+ *
+ * This custom element relies on a CSS class-based API to relate nodes outside
```
asutherland: I'm confused by the use of "relate" in this comment line.

apps/email/js/cards/cmp/autocomplete_source.js
```diff
@@ -0,0 +1,48 @@
...
+ * @return {Promise}
+ */
+return function match(query) {
+  return new Promise(function(resolve, reject) {
+    if (!contacts || !query) {
+      return resolve(makeResult(query, []));
+    }
+
+    var request = contacts.find({
+      filterBy: ['name', 'email'],
```
asutherland: instead of 'name', I think we want to use 'givenName' and 'familyName'.  There are indices over these too, and "name" appears to be a composite created by the contacts app by taking the 0th entries of the givenName and familyName entries, trim()ing them, then smooshing them together with a space between them (and trimming again in case one is missing).

(So from our perspective, 'name' is really just a duplicate of 'givenName' if there is a givenName, or it's a duplicate of familyName if not.  The downside of favoring both of these is that if searching for "John Doe", if the user decides to type the whole name for some reason, once they get to "John D", 'name' would have kept matching, but familyName/givenName will refuse to give any results at that point.)

asutherland:
I'm loving the thorough comments on everything.

asutherland:
I'm going to step away from reviewing this for a bit, so I want to note some weird behaviour I'm experiencing for me to investigate more when I resume reviewing, or for you to fix if you get to fixin' before I resume:

On b2g-desktop, if I'm doing things on the cc/bcc lines so that magic upward displacement happens, when the autocomplete results list goes away, the displacement is still there.  This goes for both me selecting an autocomplete entry and it being displayed (by mouse clicking), as well as backspacing so there are no results.  cc or bcc will be at the top of the gray box, and the gray box still retains its prior dimensions.  The problem fixes itself when the keyboard goes away and a resize triggers.

apps/email/js/cards/cmp/autocomplete.js
```diff
@@ -0,0 +1,476 @@
...
+      this.runQuery = this.runQuery.bind(this);
+      this.onFocus = this.onFocus.bind(this);
+      this.onBlur = this.onBlur.bind(this);
+      this.onInput = this.onInput.bind(this);
+
+      // Bind to events triggered by input field children. Need to explicitly
+      // bind to the elements instead of bubble since focus/blur do not
+      // bubble.
+      var nodes = this.parentNode.querySelectorAll('input');
+      slice.call(nodes).forEach((inputNode) => {
```
asutherland: fyi-you-are-probably-aware-of: Array.from(nodes) is the new hotness over this dance (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from) that we can safely use and is backport safe through gecko 32 but of course is potentially disaster if we don't assume modern-ish Firefox/Chrome.

apps/email/js/cards/cmp/autocomplete.js
```diff
@@ -0,0 +1,476 @@
...
+      this.match();
+    },
+
+    /**
+     * Triggered by input blur events.
+     */
+    onBlur: function(event) {
+      this._blurRequest = Date.now();
+      // Taps on the autocomplete list can trigger a blur, so give those taps
+      // a chance to emit events before hiding the list.
```
asutherland: I don't think we need this guard or the ramifications of the 200ms timeout on testing.  Specifically, the UI Events Specification (https://w3c.github.io/uievents/) says that the default action for click is to focus the element.  Which means the onMatchClick event below will fire synchronously, as will its synchronous call to dispatchEvent.  Only after onMatchClick returns without preventDefault()ing would the blur/focus series of events occur.

apps/email/js/cards/cmp/autocomplete.js
```diff
@@ -0,0 +1,476 @@
...
+          this.match();
+        }
+      });
+    },
+
+    /**
+     * Called to trigger finding matches from the dataSource. Set on a timeout
+     * to avoid multiple fast keystrokes from triggering too many rapid queries.
+     */
+    match: function() {
```
asutherland: naming nit: I'm not crazy about having `match` be a "go do it" function on the custom element and at the same time a result value object of the form { name, address } annotated onto the result-holding inert child elements.  Note that I'm more concerned about the noun/verb dual usage as opposed to the potential for type confusion between the parent and the children.

asutherland: maybe throttledRunQuery or debouncedRunQuery or something that conveys it's really just us limiting the rate at which we invoke runQuery

apps/email/js/cards/cmp/autocomplete.js
```diff
@@ -0,0 +1,476 @@
...
+        }
+
+        var existingEntries = this.getExistingEntries(this.inputNode);
+
+        var contacts = result.contacts;
+
+        this.list.innerHTML = '';
+        this.topMatch = null;
+
+        var queryRegExp = new RegExp(this.query, 'i');
```
asutherland: We don't need the RegExp.escape polyfill (https://github.com/benjamingr/RegExp.escape/blob/master/polyfill.js), but we should have an escaping function and that seems like a reasonable thing to import as a module, etc.

apps/email/js/cards/cmp/autocomplete.js
```diff
@@ -0,0 +1,476 @@
...
+          // autocomplete.
+          emails.forEach((email) => {
+            var node = itemTemplate.cloneNode(true),
+                nameNode = node.querySelector('.cmp-autocomplete-name'),
+                emailNode = node.querySelector('.cmp-autocomplete-email'),
+                address = email.value;
+
+            // Do not add it if the email address is already an entry.
+            if (existingEntries.some(function(entry) {
+              return entry.address === address;
```
asutherland: Indentation/style nit: my brain parsed this wrong

apps/email/js/cards/compose.js
```diff
@@ -451,6 +467,37 @@ return [
       return bubble;
     },
 
+    getExistingEntriesForAutocomplete: function(inputNode) {
```
asutherland: Should this function just return the union of all the recipients lists instead of scoping by the list?  This implementation seems to assume that I'd want to, for example, include you on both the to and the cc lists, but I don't see why I'd want to do that.  (And both bcc-ing someone and including them on a non-bcc list also seems weird.)

apps/email/js/cards/cmp/autocomplete_source.js
```diff
@@ -0,0 +1,48 @@
...
+      filterOp: 'startsWith',
+      filterValue: query
+    });
```
asutherland: We probably also want to specify filterLimit with a value high enough that we'd expect the user to give up scrolling before they got there and instead type more, but not so high that we waste a lot of disk I/O time, IPC overhead, and then build a crazy big DOM.  Like 30 or 50 or something.  (What's the worst-case count on number of people someone would know named John/James/etc.? :)

apps/email/style/compose_cards.css
```diff
@@ -88,6 +88,58 @@ input.cmp-subject-text {
...
+}
+
+/* When the autocomplete is active, change the flex alignment so that the label
+and + buttons are visible when typing */
+.cmp-autocomplete-noscroll .cmp-envelope-line {
+  align-items: flex-end;
+}
+
+.cmp-autocomplete-spacer {
+  background-color: #f2f2f2;
```
asutherland: going forward discussion: would it make sense to start using CSS variables for things like this where the color really wants to be updated in lock-step with its other uses?

apps/email/js/cards/compose.js
```diff
@@ -551,13 +599,13 @@ return [
...
           break;
       }
+
       if (makeBubble) {
-        // TODO: Need to match the email with contact name.
-        node.style.width = '0.5rem';
-        var mailbox = model.api.parseMailbox(node.value);
-        this.insertBubble(node, mailbox.name, mailbox.address);
-        node.value = '';
+        entryMatch = this.autocomplete.topMatch ||
```
asutherland: The automatic use of topMatch here works out, I think, but the emergent control flow is a little weird, and I think it'd be good to document the intentional behaviour in a comment or markdown file or on the wiki somewhere.  (At your discretion.  Mainly there are situations where someone will file a bug and at least my exact recall of what led to that situation has faded somewhat and I don't want to do the bugzilla legwork to find out.  I had started https://wiki.mozilla.org/Gaia/Email/UX/Decisions at one point, that was mainly for cases like the sub-folder icon asymmetry that I thought was a weird choice and wanted to be able to point people.  Or maybe it's not worth the effort.  Your call!)

Specifically, we're causing command and semicolon act like "tab" might in a lot of autocomplete boxes, with it triggering the selection and clearing the input.  This works great.  But the control flow is also considering doing this when the user types the space character.  We don't make a bubble because the user hasn't typed an "@".  I think this will continue to be fine, but it's awkward, so as alluded to above, I think some type of comment acknowledging this might be in order if we don't want to clean up the logic.  (Not sure it's worth cleaning up the logic.)

apps/email/js/cards/cmp/autocomplete.js
```diff
@@ -0,0 +1,476 @@
...
+      this.parentNode.classList.remove('cmp-autocomplete-noscroll');
+      this.originNode.classList.remove('cmp-autocomplete-noscroll');
+
+      this.spacer.style.minHeight = '';
+
+      this._matchClickedTime = 0;
+    },
+
+    hideList: function() {
+      this.list.classList.add('collapsed');
```
asutherland: This also needs to do `this.parentNode.scrollTop = 0;` in order to address the problem I reported.  This addresses both the "backspace" scenario and the "autocomplete picked" for the cc case.  I believe in both those cases hideList() just ends up getting called.

(This also seems like one of those subtle/complex bugs where a react.js style re-render everything based on current state strategies is desirable.)

asutherland: Uh, noting that I didn't see if this has other horrible side-effects.  Maybe more things should just be calling hide() instead or what not.

apps/email/js/cards/cmp/autocomplete.js
jrburke: Will do a s/relate/bind to/

apps/email/js/cards/cmp/autocomplete.js
jrburke: Ah, thanks for calling this out. Old habit, nice to break it now.

apps/email/js/cards/cmp/autocomplete.js
jrburke: Will use "runThrottledQuery".

apps/email/js/cards/cmp/autocomplete.js
jrburke: Thanks for the catch, I had a mental note to look into it, but forgot to write it down. I created a regExpEscape module that just returns a function, using the polyfill code:

```javascript
'use strict';
/**
 * Taken from the ECMAScript proposal polyfill in the repo:
 * https://github.com/benjamingr/RegExp.escape
 * git commit: de7b32f9f72d15f9d580119963139d1027c46bbe
 * Copyright CC0 1.0 Universal Benjamin Gruenbaum 2015
 */

define(function() {
  return function regExpEscape(str) {
    return str.replace(/[\\^$*+?.()|[\]{}]/g, '\\$&');
  };
});
```

apps/email/js/cards/cmp/autocomplete.js
jrburke: I was too fancy here, extracted the some() call to a separate hasExisting var, then just that var in the if().

apps/email/js/cards/compose.js
jrburke: This was just me being careful and not knowing the expected rules for email addressing. Changed to include all entries attached to the email, not just the current input node. Interface changed to not pass the inputNode.

apps/email/js/cards/cmp/autocomplete_source.js
jrburke: Good catch! I will use 30, as I expect it is easier for the user to keep typing to narrow matches than to scroll through the limited height area for the matches.

apps/email/style/compose_cards.css
jrburke: That sounds good. We pull in shared/elements/gaia-theme/gaia-theme.css because of a shared dependency. That has some vars defined, but it does not include this particular color. I expect if we revisit the UI for making things more common across apps or a general visual style change we will do that then.

apps/email/js/cards/cmp/autocomplete_source.js
jrburke: I went with adding all three of the name options: `filterBy: ['email', 'name', 'givenName', 'familyName']` -- that seems to allow matching on last name only, or if i do first name and a space matches on the whole name.

apps/email/js/cards/cmp/autocomplete.js
jrburke: When I change onBlur to just do this.hide(), then what I log in the console is the blur occurs before the tap on an autocomplete list item, and I do not even get the click event in that case. I can see where the blur emitting needs to happen first.

I could maybe switching to touchend, but I liked the simplicity of just using 'click' event binding. On the other hand, I am always suspicious of setTimeouts.

jrburke:
> On b2g-desktop, if I'm doing things on the cc/bcc lines so that magic upward displacement happens, when the autocomplete results list goes away, the displacement is still there

I did this on purpose to avoid the UI jumping around as new things are entered, or if text box is temporarily cleared probably for new quick typing to follow. So, stability of the UI.

Maybe what I could do is to just leave up an empty white autocomplete region while the focus is still in a text box, so that the scrolled state of the gray envelope area is not shown? Although, it requires an extra tap to then remove the white area and to see the next place where the user might want to tap (like to go to bcc or to subject or start typing text).

If you think that is still confusing or not optimal, I can give it a shot with the jumping around of scroll as the text is cleared/items are tapped in the autocomplete.

asutherland:
The thing that seemed buggy to me was that the "To" region was no longer visible and I could not physically scroll to get it back and there was no obvious UI indicator that I was in a different effective state than when I started.

I think any of the following would work:
* Make it so the user can scroll the 'to' back into place (possibly with a resistance factor?)
* Have the autocompletion mode cause one of those (x) to clear things.  Tapping that could clear the autocomplete mode.  Maybe the "+" could morph into an "x" for screen real-estate/stability reasons?  (Alternately, the whole thing could be rejiggered to be a more obvious modal UI, but I like how smoothly this implementation seems to fit in.)
* Tap-on-empty/invalid autocomplete option to close.  (Or maybe a "close autocomplete" explicit string?)

I agree your goal of avoiding jumpiness is very important.  That's one of the things that struck me about the excellence of this implementation and why I thought we'd have to go more explicit modal to otherwise accomplish the effect.

apps/email/js/cards/cmp/autocomplete.js
asutherland: Oy.  I guess I'm not surprised that it's necessary then.  I was hoping you were just being superstitious (I assumed this because I've literally written logic exactly like this just because I didn't want to think about/debug the problem.)

A zero setTimeout would be fine-ish with me with a little more comment along the lines of "the blur event is dispatched before the click event.  These are dispatched consecutively, so we want to run immediately after any click event would have been dispatched. We can't use a Promise because the micro-task queue will be drained before the click event is processed."

asutherland: Also, if you happen to understand why things are working like they're working, I'd love if you could briefly detail it here.  I had checked quirksmode and some other searching briefly, hoping something would clearly explain what's happening, but my search-fu was weak or what not.

asutherland: oh, I wonder if some of it is the custom logic that B2G injects into content processes to magic up the keyboard.  I could see them monkeying with various events.

apps/email/js/cards/compose.js
jrburke: On further reflection, I just removed topMatch. Favoring it in this keyboard character case could actually lead to the user not being able to type in an address if what they wanted to type and send to was a substring match to an existing contact. So if they had amy@a.com in the contacts but wanted to send to amy@a.co, this logic was a problem.

Plus, given that the ; and , are on secondary keyboard screens, the user is more likely to just tap the autocomplete selection than navigate the keyboard to end selection via the keyboard.

I want to add in keyboard support for pressing down/up arrows to select an autocomplete entry, to allow easy selection of autocomplete. I might do that when working on the accessibility ticket for the autocomplete. For that, I'll need to compensate more for blurs occurring but not collapsing the autocomplete list even when there is not a tap triggering a match click. I might need to do that anyway for the screen reader.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e6ab0da333cc98bfc9aeaf08d6fad3954b4b033a

apps/email/js/cards/cmp/autocomplete.js
jrburke: Just making sure I follow up with all the comments: in a comment I added yesterday, the goal was to keep the input box UI stable at the top, so the hideList is working as desired here without the scrollTop or the additional hide() work.

I will be changing the overall implementation so that if the user tries to scroll up to see the already entered addresses that will work, but via a 'pan' gesture detection and the use of an overflowY. This should address the overall concern about the UI seeming stuck or out of alignment when trying to manipulate/scroll it manually.

jrburke:
> Make it so the user can scroll the 'to' back into place (possibly with a resistance factor?)

I have something working locally which will allow the user to just scroll up to see the 'to' addresses, and feels pretty natural. I stole the technique from the SMS app, which was using a translateY approach (vs the .scrollTop on inputArea that I was using), with a GestureDetector for 'pan' that restores the translateY, and with a transform added for a nice scroll effect.

End result unfortunately is that the implementation has changed enough that I will likely ask for another review pass once I have it ready. Apologies for generating extra review work, but also grateful for the review dialogue that will lead to an even slicker feel for the autocomplete.

apps/email/js/cards/cmp/autocomplete.js
jrburke: Went with this comment (with the name of _matchClickedTime changing to _inProgressActionTime because with the new transition translateY that needs to wait for a transitionend, input focus events as well as clicks are in play):

```javascript
      // Taps on the autocomplete list or to a new input field will trigger the
      // blur event listeners before the focus or click listeners, so use a
      // setTimeout to get an entry in the event loop after those potential
      // listeners can fire. If they did fire (indicated by setting
      // _inProgressActionTime), then do not do the hide() work. Using a
      // setTimeout because a Promise microqueue would end up with this function
      // getting called before the other event listeners are notified.
```

My experience before is that focus and click event listeners fired after blur listeners, since conceptually the UI would need to blur the existing focused element before then focusing on the next element. Otherwise, asking for the focused element could result in two elements being marked as "focused". The same conceptual idea for click.

I expect if we listened for touchstart, that would fire before the blur, but since with the new translateY thing we also want focus events to ignore the hiding in a blur, the setTimeout works generically in that case. I also updated the setTimeout to just use 0 for the interval.

asutherland:
That sounds great.  No worries on the extra review pass.  I trust you to ask for additional reviews when needed, which is one of the reasons I am so free with the r+'s when reviewing your patches!

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=aa683e06e3b4e71fe4736c5afd2eae7ee2847ae2

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f0ee34d2fb9c37d0bec6772f5638ff1d481c1e51

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0418b70f14e466422e3de8692231aee60bdb82d4

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f9324fe76183cf67ada48b1b9cb3288b383090d8

jrburke:
Asking for review again after some significant changes based on review feedback:

* The autocomplete now sits as a tag at the bottom of compose's scrollregion-below-header area. This allows it to properly sit on top of the other content in compose, particularly since the cmp-autocomplete-input-list now uses a transform for scrolling, and that seemed to create a separate layer so that if the autocomplete was in that area, it was getting clipped/hidden by the divs below cmp-autocomplete-input-list.
* cmp-autocomplete-input-list is now scrolled using an translateY transform with a CSS transition to move the input elements to the top of the view. gesture_detector was used to detect 'pan' finger drags to restore the cmp-autocomplete-input-list to translateY(0). This gives it a nice drag feeling when wanting to "scroll up" to see addresses already entered.
* Since a transition is used to move the input box around, the user has a better sense of place, and the autocomplete list is first shown after the transition finishes, to help the user track their sense of place. This meant doing some transitionend work, and the positionElements method (was named sizeList before) has a promise jump to finish its work.
* The onBlur approach with its setTimeout hide was removed, and instead event listeners on the document for click or focus events will hide the autocomplete. I liked the onBlur approach to try to keep the autocomplete encapsulated within itself, but since the autocomplete is already reaching outside of itself for things like window resize events, this seemed fine to do. It removes the setTimeout, and also works better for screen reader use as the matches stay up for selection. So the screen reader situation is much better.
* Due to the better screen reader support, the autocomplete_item.html changed its structure. It was a balancing act to get selectable elements in the screen reader as well as proper full bleed background active/focus coloring that did not affect the margins around the text and bottom border line for each item.
* Support for pressing the down arrow in a full keyboard was added, pressing it while in an input field jumps the focus down to the first item in the match list. Using Tab from then on to select other elements work. The Down arrow in the match list scrolls the match list, just the default browser navigation actions there, nothing added explicitly by this change.
* I did some small HTML/CSS adjustments for RTL in autocomplete_item.html for mixed content, just using the standard dir="auto" and explicit text-align approach we have used elsewhere.

apps/email/build/email.build.js
```diff
@@ -47,12 +47,17 @@
...
+        // just include it in the base layer.
+        'shared/js/gesture_detector',
+
         // Bundle most likely card
         'cards/message_list'
       ]
     },
     {
       name: 'cards/compose',
+      // Exclude gesture_detector
```
asutherland: It seems like this comment is flotsam or needs more comment

apps/email/js/cards/cmp/autocomplete.js
```diff
@@ -0,0 +1,565 @@
...
+}
+
+/**
+ * Given a node, if it or an ancestor has the CSS className, return the node
+ * with that className. Returns null if not a match.
+ * @param  {Node} node
+ * @return {Node}
+ */
+function findAncestorWithClass(node, className) {
+  if (node.classList && node.classList.contains(className)) {
```
asutherland: This method is fine and good with its specialization, but I went spelunking for new DOM methods I wasn't aware of, and found https://developer.mozilla.org/en-US/docs/Web/API/Element/matches which could be used to make this be a more generic firstAncestorQuerySelector method or something.  (And very nitty, but a for loop could eliminate the duplicated logic.)

apps/email/js/cards/cmp/autocomplete.js
```diff
@@ -0,0 +1,565 @@
...
+}
+
+/**
+ * Autocomplete custom element that contains a list of elements that show the
+ * autocomplete matches.
+ *
+ * This custom element relies on a CSS class-based API to bind to nodes outside
+ * of this autocomplete. The CSS/HTML structure expected by this autocomplete:
+ *
+ * - An element with the CSS class of cmp-autocomplete-origin" that is the the
```
asutherland: nit: unpaired double quote

apps/email/js/cards/cmp/autocomplete.js
```diff
@@ -0,0 +1,565 @@
...
+      //mix in that is not assumed to be a card. For now, this is an equivalent
+      //result.
+      this.classList.remove('card');
+
+      this.lastTranslateY = 0;
+      this.inputAreaTranslated = Promise.resolve();
+
+      this.runQuery = this.runQuery.bind(this);
+      this.onFocus = this.onFocus.bind(this);
+      this.onDocumentEvent = this.onDocumentEvent.bind(this);
```
asutherland: General thought not for this specific patch: I wonder if it would be reasonable to just auto-bind() all of the methods on our custom elements.  That's what React.createElement does (the old, pre-"hey just use ES2015 classes" idiom, which I still use), and it's handy since it avoids the boilerplate and the foot-gun of failing to do this.

apps/email/js/cards/cmp/autocomplete.js
```diff
@@ -0,0 +1,565 @@
...
+    onResize: function(event) {
+      if (this.inputNode && !this.isHidden) {
+        this.positionElements();
+      }
+    },
+
+    onMatchKeyDown: function(event) {
+      if (event.key === 'Enter' && !this.isHidden) {
+        this.onMatchClick(event);
+      }
```
asutherland: I feel like this needs ArrowDown and ArrowUp logic of its own or disclaimers.  It's super weird to be able to hit arrow down to focus a match and then seem like you're trapped there.

I realize you did mention that tab works in your bullet change list, but it seems odd to me to support down once and then have it stop working.  This is verging on feature creep quite a bit, of course.  I'd be fine if you have the onKeyDown support explicitly disclaim keyboard navigation as a feature.  Maybe we should just mess with the tabstop ordering so only tab ever works?  (On b2g-desktop I end up tabbing myself into the bcc box from the cc box.)

So I did try tab based on your guidance, but that ended up extra weird.  When my tab focus left the autocomplete items, they didn't go away.  I guess I ended up in the bcc line and then the subject and then the body.  And then, somewhat hilariously, I guess I managed to tab into the "search upsell" box on the message list screen, because all of a sudden the message list search card transitioned onto the screen.  (I'm guessing whatever selector magic we're trying to use to stop text input from happening on other cards is not enough to negate its tab order?)

asutherland:
Something freaky is happening with the autocomplete DOM node I think, but because I'm using b2g-desktop and I don't really know how to be able to remote debug it, it's proving harder to investigate than I was hoping.  (I have hunger brain right now, presumably whatever is going on would be more obvious to a full stomach.)

But in a nutshell, everything I type into the composer area's input area gets duplicated into the autocomplete DOM node that lives just off screen.  If I hit the down arrow twice in the input area, the first down arrow gets eaten to do the focus, but then the second down arrow triggers native scrolling and brings the autocomplete DOM node into view.  This also coincidentally breaks the positioning displacement magic.

apps/email/js/cards/cmp/autocomplete.js
```diff
@@ -0,0 +1,565 @@
...
+
+      // Want to use the input element, and not a parent of it, for placement of
+      // the autocomplete, since the input element can move to different lines
+      // as typing occurs, where the parent's origin would not change, and would
+      // be a bad reference point to use.
+      var inputRect = this.inputNode.getBoundingClientRect();
+
+      // Scroll the area so that the input item is now at the top of the
+      // address area.
+      var translateY = inputRect.top - this.originTop +
```
asutherland: Given what I saw with the down arrow breakage, maybe it would pay to put scrollTop in here too as defensive thing.  We normally should always be scrollTop=0, but when we're not, it's weird.

asutherland:
One b2g-desktop thing that I want to note but I don't think is worth fixing is that the logic gets very faked out in a non-consistent manner by X-focus behavior where a blur happens because my focus has left the b2g-desktop window and then comes back when I re-enter the window.  This might be due to a combination of the delayed resize events from the keyboard coming and going plus our explicit focus/blur-awareness.

apps/email/js/cards/cmp/autocomplete.js
```diff
@@ -0,0 +1,565 @@
...
+                      this.lastTranslateY -
+                      this.verticalSpace;
+
+      if (translateY !== this.lastTranslateY) {
+        this.lastTranslateY = translateY;
+
+        // Create a new promise to track the end of the inputArea translation.
+        this.inputAreaTranslated = new Promise((resolve, reject) => {
```
asutherland: nit: no need to specify reject if we're not going to use it.

apps/email/build/email.build.js
jrburke: Removed.

apps/email/js/cards/cmp/autocomplete.js
jrburke: Good to know, will remember that if we see a lot of this sort of ancestor work going on. Since autocomplete is using a CSS class based API to find things, I like the specialized behavior here, although I have updated it locally to use a for loop. Thanks for pointing that out, good to remove the redundancy.

apps/email/js/cards/cmp/autocomplete.js
jrburke: Fixed and added a note about envelope rows specifying a "cmp-autocomplete-input-label" CSS class.

apps/email/js/cards/cmp/autocomplete.js
jrburke: I was very close to create a helper that accepted a list of method names and it would do this work, but I did not want to disentangle it from this version of base.js from the one on the convoy branch.

On first thought, I'm lukewarm on auto-binding for all methods, seems excessive. I suppose we can talk more about it in convoy.

I was hoping via other APIs to avoid the need for these kinds of explicit .bind() calling. evt now takes object listeners on the convoy branch, so this binding is no longer necessary for those cases, and the data-event binding takes care of it in its cases. Dojo's event module also could handle DOM addEventListener registration in a similar way to the evt object listening, so that might be another option to address this particular case, where it is about DOM event listening vs the non-DOM evt listening.

This case is a bit weird because it is reaching outside of itself to listen to things.

apps/email/js/cards/cmp/autocomplete.js
jrburke: Agreed it is weird, and feels like for it all to work properly, I need to hide more of the the rest of the DOM, either via tabindex=-1 or some other trickery for tab to just work. Probably need to explicitly set tab order so the autocomplete matches appear between valid inputs: if in "to:", tab should be from "to:", all autocomplete entries, then to "cc:".

I do not like the idea of all of that manual setting and unsetting though. For now I just removed this arrowdown support. If you would like to see it with this bunch, then I'll do a separate commit pass in this pull request for it.

apps/email/js/cards/cmp/autocomplete.js
jrburke: Removed.

jrburke:
> But in a nutshell, everything I type into the composer area's input area gets duplicated into the autocomplete DOM node that lives just off screen.

Looking more into this, it is caused by the compose card using that [div to get the size of the text](https://github.com/mozilla-b2g/gaia/blob/474784fc494519a0fab866407c0215aacf984ed9/apps/email/js/cards/compose.js#L561).

So basically the keyboard navigation via tab or arrows is messed up. Given that we have not really tested that before, not too surprising. I removed the down arrow code in the autocomplete for now. I want to do a final pass over the remaining comments, one more test run with the screen reader, then I'll ask for re-review.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8725be4c868a340cc14dd990947593d143fe9175

asutherland:
Aha!  Good find on that sizing logic.  Now I don't feel so bad about being confused about how that was happening, since it was a preexisting problem rather than something I'd find in the diff.  (Although I am of course mortified that I let the bug come into existence in the first place since I think all that code pre-dates you quite extensively.)

Removing the arrow code support works for me.  As a lazy person who tests via b2g-desktop, I do appreciate the keyboard assist, and maybe removing is overkill, except I know we'd inevitably have seen at least a few bug reports about that.

apps/email/js/cards/cmp/autocomplete.js
jrburke: The only scrolling element now is the originNode, the scrollContainer, and that is set to zero on every show() and show is called for every query that has a result. The trick with the down arrow breakage seems to be that it was possible to tab out of the scrollContainer, with the viewport following that, causing a bunch of havoc.

I am able to fix that case by always calculating the originTop. I did a quick performance.now() check, and its cost is negligible to do every time (in the right of the decimal for performance.now() diffs).

I was able to test this failure case, and the fix, in mulet by tapping on the Compose orange header, then pressing arrow down to get to the screenContainer.

jrburke:
> One b2g-desktop thing that I want to note but I don't think is worth fixing is that the logic gets very faked out in a non-consistent manner by X-focus behavior where a blur happens because my focus has left the b2g-desktop window and then comes back when I re-enter the window.

I was thinking this was maybe from using capturing events on document clicks/focus, but changing them to bubbles did not help. In mulet, if I click out of the responsive view of the app, the keyboard goes away. Then when tapping back, it is like the document is tapped, so the autocomplete resets, then a focus is done again by the browser in the input field.

So, going to leave as is as a quirk of the focusing in and out on desktop.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1da43c919b5ce2ec9c9c7787ec9eed33b27d8d85

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a088c7020146a5288b99d2903371f6e9b986106a

PR #31190, opened 2015-07-31
Bug 952456 - Add "Copy Image" option

BorisChiou:
Provide "copy image" option on context menu

### DIFF
--- a/apps/system/js/browser_context_menu.js
+++ b/apps/system/js/browser_context_menu.js
@@ -70,16 +70,29 @@
       var items = [];
 
       // contextmenu.items are specified by the web content via html5
-      // context menu api
+      // context menu api and a specific system command, copy-image.
       if (detail.contextmenu && detail.contextmenu.items.length) {
         detail.contextmenu.items.forEach(function(choice, index) {
-          items.push({
-            label: choice.label,
-            icon: choice.icon,
-            callback: function() {
-              detail.contextMenuItemSelected(choice.id);
-            }
-          });
+          var itemObj = null;
+          switch (choice.id) {
+            case 'copy-image':
+              itemObj = {
+                id: choice.id,
+                label: _(choice.id),
+              };
+              break;
+
+            default:
+              // Customized menu items
+              itemObj = {
+                label: choice.label,
+                icon: choice.icon,
+              };
+          }
+          itemObj.callback = () => {
+            detail.contextMenuItemSelected(choice.id);
+          };
+          items.push(itemObj);
         }, this);
       }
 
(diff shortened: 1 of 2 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3c4a7380ba927d764c2db21e2087a09806f79735

apps/system/js/browser_context_menu.js
```diff
@@ -73,13 +73,22 @@
...
-            icon: choice.icon,
-            callback: function() {
-              detail.contextMenuItemSelected(choice.id);
-            }
-          });
+          var itemObj = null;
+          if (choice.id == 'copy-image') {
+            itemObj = {
+              id: choice.id,
+              label: _(choice.id),
```
timdream: What makes `copy-image` special? If it's just about missing graphics, simply commit a temporary one is better than mess with the code like this....

apps/system/locales/system.en-US.properties
```diff
@@ -600,6 +600,7 @@ save-link=Save linked file
 share-image=Share Image
 share-video=Share Video
 share-audio=Share Audio
+copy-image=Copy Image
```
timdream: Agree we need a UI label regardless.

apps/system/js/browser_context_menu.js
timdream: Never mind, I've figure out why it's needed here...

apps/system/js/browser_context_menu.js
```diff
@@ -73,13 +73,22 @@
...
         detail.contextmenu.items.forEach(function(choice, index) {
-          items.push({
-            label: choice.label,
-            icon: choice.icon,
-            callback: function() {
-              detail.contextMenuItemSelected(choice.id);
-            }
-          });
+          var itemObj = null;
+          if (choice.id == 'copy-image') {
```
timdream: `switch ... case` is better than `if .. else` here IMHO.

apps/system/js/browser_context_menu.js
BorisChiou: OK, I could use a temporary one for copy-image case. Thanks.

apps/system/js/browser_context_menu.js
BorisChiou: Sure, I will upload new patch to use switch case.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8034f561e5500471cc764c1fef7c286cfef98d65

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=31fbbc090a8f57c277999dd6cd3c659396e4cc2d

BorisChiou:
Fix 1 jshint error.

PR #31200, opened 2015-07-31
Bug 1180168 - Fix x-axis scrolling while swiping up on card

sfoster:
Dispatch an event when swiping up (cross-slide) so the task-manager can set overflow-x on the container. This seems to work pretty well but I've not been able to profile yet to see if it introduces new jank

### DIFF
--- a/apps/system/js/card.js
+++ b/apps/system/js/card.js
@@ -386,6 +386,7 @@
             if (verticalY > Math.abs(this.deltaX) &&
                 verticalY > this.SWIPE_WOBBLE_THRESHOLD) {
               this._dragPhase = 'cross-slide';
+              this.publish('crossslidestart');
               // dont try and transition while dragging
               this.element.style.transition = 'transform 0s linear';
               this.onCrossSlide(evt);
@@ -419,6 +420,10 @@
         } else {
           // return it to vertical center
           this._resetY();
+        }
+        if (this._dragPhase === 'cross-slide') {
+          this.publish('crossslideend');
+          this._dragPhase = '';
         }
         break;
     }
--- a/apps/system/js/task_manager.js
+++ b/apps/system/js/task_manager.js
@@ -306,7 +306,7 @@
     window.addEventListener('appterminated', this);
     window.addEventListener('wheel', this);
     window.addEventListener('resize', this);
-
+    window.addEventListener('card-crossslidestart', this);
     this.element.addEventListener('click', this);
   };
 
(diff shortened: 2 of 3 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=296ae6699448a7d58405afa35c401d62bee24bef

apps/system/test/unit/task_manager_test.js
```diff
@@ -471,7 +471,7 @@ suite('system/TaskManager >', function() {
       }
     });
 
-    suite('display cardsview >', function() {
+    suite.only('display cardsview >', function() {
```
etiennesegonzac: /!\ let's not forget the `only`

apps/system/js/card.js
```diff
@@ -386,6 +386,7 @@
             if (verticalY > Math.abs(this.deltaX) &&
                 verticalY > this.SWIPE_WOBBLE_THRESHOLD) {
               this._dragPhase = 'cross-slide';
+              this.publish('crossslidestart');
               // dont try and transition while dragging
               this.element.style.transition = 'transform 0s linear';
```
etiennesegonzac: let's do a
`this.element.style.willChange = 'transform';` too

apps/system/js/task_manager.js
```diff
@@ -745,7 +746,25 @@
...
     }
+    this._handlingCrossSlide = true;
+    var finish = (e) => {
+      if (this.element) {
+        this.element.style.overflowX = '';
+      }
+      this._handlingCrossSlide = false;
+    };
+    this.element.style.overflowX = 'hidden';
+    eventSafety(window, 'card-crossslideend', finish, 1000);
```
etiennesegonzac: let's work through the edge case and handle the `card-crossslideend` properly (ie. like the start event and without event safety).
it's pretty easy to shout ourselves in the foot with this version

etiennesegonzac:
In `  Card.prototype.onCrossSlide = function(evt) {`, I feel like it actually feels better if we don't remove the THRESHOLD from the delta.
It's a small threshold easy forgettable if the card is under the finger where it's supposed too instead of making the threshold "permanent".

apps/system/js/card.js
etiennesegonzac: without forgetting to remove it :)

etiennesegonzac:
let's use the opportunity of this patch to switch resetY to something like
```
  Card.prototype._resetY = function(evt) {
    this.element.style.transition = 'transform 0.15s linear';
    this.element.style.transform = 'translateY(0)';

    var finish = () => {
      this.element.style.removeProperty('transition');
      this.element.style.removeProperty('willChange');
    };
    eventSafety(this.element, 'transitionend', finish, 300);
  };

```

etiennesegonzac:
And maybe to the same for `killApp` :) (letting the card transition for a bit)

PR #31210, opened 2015-08-03
Bug 1190276 - [Messages] In a "new" activity, don't redirect to a thread if there is a specified body r=azasypkin

julienw:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=cd68a8fe8a708f42f9fec1a4855cddda6fb9a288

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=dec879917b7120ef90abb62e259ee273ee930257

apps/sms/views/shared/js/activity_handler.js
```diff
@@ -71,12 +71,17 @@ var ActivityHandler = {
...
+      if (!viewInfo.body) {
+        // try to get a thread from number
+        // if no thread, promise is rejected and we try to find a contact
+        threadPromise = MessageManager.findThreadFromNumber(viewInfo.number)
+          .then((threadId) => viewInfo.threadId = threadId)
+          // case no contact and no thread id: gobble the error
+          .catch(() => {});
+      }
+    }
+    if (viewInfo.number && !viewInfo.body) {
```
azasypkin: __nit:__ likely forgot to remove ```if (viewInfo.number && !viewInfo.body) {```?

apps/sms/views/shared/js/activity_handler.js
```diff
@@ -71,12 +71,17 @@ var ActivityHandler = {
...
       // number or even contact to fill recipients input
       focusComposer = true;
-      // try to get a thread from number
-      // if no thread, promise is rejected and we try to find a contact
-      threadPromise = MessageManager.findThreadFromNumber(viewInfo.number)
-        .then((threadId) => viewInfo.threadId = threadId)
-        // case no contact and no thread id: gobble the error
-        .catch(() => {});
+
+      if (!viewInfo.body) {
```
azasypkin: __nit:__ maybe add comment explaining why we decided to not search for conversation if body is presented? I think it's easy to forget, okay at least for me :)

apps/sms/test/marionette/new_activity_test.js
```diff
@@ -231,5 +231,31 @@ marionette('Messages as "new" activity target', function() {
         messagesApp.Composer.messageInput, 'Message input should be focused'
       );
     });
+
+    test('Activity with thread "number" and "body"', function() {
```
azasypkin: Yay! I like integration tests, thanks for this one! Basically it makes me more confident to introduce risky changes when we have good coverage :)

apps/sms/views/shared/test/unit/activity_handler_test.js
```diff
@@ -864,9 +859,11 @@ suite('ActivityHandler', function() {
...
     test('Should move to the composer and set activity', function(done) {
-      ActivityClient.on.withArgs('new-activity-request').yield(
-        numberActivityData
-      );
+      ActivityClient.on.withArgs('new-activity-request').yield({
+        number: '123',
+        body: 'foo'
+      });
+
       threadDeferred.reject(new Error('No thread for this test'));
```
azasypkin: __nit:__ mm, maybe ```sinon.assert.notCalled(MessageManager.findThreadFromNumber)``` instead of ```threadDeferred.reject(new Error('No thread for this test'));```?

apps/sms/views/shared/test/unit/activity_handler_test.js
```diff
@@ -943,6 +940,24 @@ suite('ActivityHandler', function() {
         );
       }).then(done,done);
     });
+
+    test('existing thread with specified body, navigate to composer',
```
azasypkin: __question:__ what is the value in this test if we already have very similar one ```'Should move to the composer and set activity``` (with added ```sinon.assert.notCalled(MessageManager.findThreadFromNumber)``` in the test above it should not matter whether thread exists or not)?

julienw: you're not wrong, but I'm not wrong either :)

from the unit test writer perspective, we don't know how the code works. So I think the tests as they are now reflect better the "black box" we're testing.

Using a `sinon.assert.notCalled` means we know what happens inside the code. In the end, we don't really care whether it's called or not, as long as we move to the right panel. What's important is our 2 cases: there is a known conversation, there is not a known conversation => what do we do ? Currently we do the same, that's right :) but the 2 cases are not less different.

What do you think ?

azasypkin: Agree, it depends on the side we look at it :) My main points were:
* Make sure that we redirect immediately to view without waiting for the thread search operation to complete (should be noticeable by user anyway though);
* Make test a bit less confusing since we reject/resolve something that doesn't have any effect.

Both are minor points, and "black box" thing you're talking about totally makes sense to me as well - so I'm okay with your intention :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=36df05d8310bbaf7dd61e262d7e5f8a1b640b1f2

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=64f8ca6135d42afcbecbf5332bf0a1b33405a2c0

PR #31228, opened 2015-08-04
Bug 1188926 - Make all the apps have an origin as well as a name and check for that, part 1

mwargers:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1abf0f6d39f42f5f71c40d9b7191bf00c30bd101

tests/python/gaia-ui-tests/gaiatest/apps/phone/app.py
```diff
@@ -10,6 +10,7 @@
 class Phone(Base):
 
     name = "Phone"
```
npark-mozilla: double quote here, perhaps we should make it consistent?  (not sure whether there is a difference though)

JohanLorenzo: There's no difference in python, it's just a matter of style and to be able to include quotes in strings. For instance:
``` python
string = 'That is a "quoted" string'
string2 = "That's a 'quoted' string"
```

tests/python/gaia-ui-tests/gaiatest/apps/calendar/app.py
```diff
@@ -15,6 +15,7 @@
 class Calendar(Base):
 
     name = 'Calendar'
+    origin = 'app://calendar.gaiamobile.org'
```
JohanLorenzo: I think we can factorize the generation of the origin, as it's often the same pattern (`app://appname.gaiamobile.org`). In base.py:
``` python
    @property
    def name(self):
        raise NotImplementedError("{} should implement the class attribute called 'name'".format(self.__class__)  )

    @property
    def origin(self):
        return 'app://{}.gaiamobile.org'.format(self.name.lower())
```

And then we don't have to defined `origin`, except in Dialer and Contacts.

mwargers: How would this work? It's self.apps.displayed_app.name that changes its name depening on the locale that is used.

PR #31243, opened 2015-08-04
Bug 1185274 - Creating Front Camera Case with Timer

PeterBylenga:
Camera/app.py
Creating needed locators and definitions

Functional/test_front_camera_take_photo_with_timer.py
Created test case

### DIFF
--- a/tests/python/gaia-ui-tests/gaiatest/apps/camera/app.py
+++ b/tests/python/gaia-ui-tests/gaiatest/apps/camera/app.py
@@ -3,6 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import time
+import pdb
 
 from marionette_driver import expected, By, Wait
 from marionette_driver.marionette import Actions
@@ -31,9 +32,30 @@ class Camera(Base):
     _hud_locator = (By.CSS_SELECTOR, '.hud')
     _loading_screen_locator = (By.CSS_SELECTOR, '.loading-screen')
     _toggle_flash_button_locator = (By.CSS_SELECTOR, '.test-flash-button')
+    ## These are named equal to current view
+    _camera_switch_to_rear_locator = (By.CSS_SELECTOR, '[data-l10n-id="toggle-camera-front-button"]')
+    _camera_switch_to_front_locator = (By.CSS_SELECTOR, '[data-l10n-id="toggle-camera-rear-button"]')
 
+    _menu_button_locator = (By.CSS_SELECTOR, '[data-l10n-id="menu-button"]')
+    
     _viewfinder_video_locator = (By.CLASS_NAME, 'viewfinder-video')
-
+    
+    # Menu
+    _hdr_enable_locator = (By.CSS_SELECTOR, '[data-l10n-id="settings-option-hdr"]')
+    _self_timer_enable_locator = (By.CSS_SELECTOR, '[data-l10n-id="setting-option-self-timer"]')
+    _5_second_timer_locator = (By.CSS_SELECTOR, '[data-key="secs5"]')
+    _grid_lines_enable_locator = (By.CSS_SELECTOR, '[data-l10n-id="setting-option-grid"]')
+    _close_menu_locator = (By.CSS_SELECTOR, '[data-l10n-id="close-menu-button"]')
+    
+    # Indicator Icons
+    _timer_enabled_locator = (By.CSS_SELECTOR, '[data-l10n-id="self-timer-indicator"]')
+    _hdr_enabled_locator = (By.CSS_SELECTOR, '[data-l10n-id="hdr-indicator"]')
+    _geolocation_enabled_locator = (By.CSS_SELECTOR, '[data-l10n-id="location-indicator"]')
+    _battery_low_indicator = (By.CSS_SELECTOR, '[data-l10n-id="battery-low-indicator"]')
+ 
+    # Timer Animation
+    _timer_animation_locator = (By.CSS_SELECTOR, '[data-timer="active"]')
+    
     # ConfirmDialog
     _select_button_locator = (By.CSS_SELECTOR, '.test-confirm-select')
 
@@ -59,6 +81,21 @@ class Camera(Base):
         # Wait for thumbnail to appear
         self.wait_for_thumbnail_visible()
 
+    def take_photo_with_timer(self):
+        # Wait for camera to be ready to take a picture
+        controls = self.marionette.find_element(*self._controls_locator)
+        Wait(self.marionette, timeout=20).until(
+            lambda m: controls.get_attribute('data-enabled') == 'true')
+         
+        self.tap_capture()
+        time.sleep(1)
+        # Wait for Timer to animate
+        timer_animating = self.marionette.find_element(*self._timer_animation_locator)
+        Wait(self.marionette).until(expected.element_enabled(timer_animating))    
+
+        # Wait for thumbnail to appear
+        self.wait_for_thumbnail_visible()        
+        
     def record_video(self, duration):
         # Start recording
         self.tap_capture()
@@ -115,6 +152,39 @@ class Camera(Base):
         Wait(self.marionette).until(lambda m: not current_camera_mode == self.camera_mode)
         self.wait_for_capture_ready()
 
+    def tap_switch_to_front_camera(self):
+	self.marionette.find_element(*self._camera_switch_to_front_locator).tap()
+	switch_to_rear_button = Wait(self.marionette).until(expected.element_present(*self._camera_switch_to_rear_locator))
+        Wait(self.marionette).until(expected.element_displayed(switch_to_rear_button))
+        
+    def tap_switch_to_rear_camera(self):
+	self.marionette.find_element(*self._camera_switch_to_rear_locator).tap()
+	switch_to_front_button = Wait(self.marionette).until(expected.element_present(*self._camera_switch_to_front_locator))
+        Wait(self.marionette).until(expected.element_displayed(switch_to_front_button))
+		
+    def tap_menu_button(self):
+	self.marionette.find_element(*self._menu_button_locator).tap()
+	menu_opened = Wait(self.marionette).until(expected.element_present(*self._close_menu_locator))
+	Wait(self.marionette).until(expected.element_displayed(menu_opened))
+	
+    def tap_enable_timer(self):
+	self.marionette.find_element(*self._self_timer_enable_locator).tap()
+	option_available = Wait(self.marionette).until(expected.element_present(*self._5_second_timer_locator))
+	Wait(self.marionette).until(expected.element_displayed(option_available))
+	# select 5 seconds, expected to return to viewfinder
+	self.marionette.find_element(*self._5_second_timer_locator).tap()
+	returned_to_viewfinder = Wait(self.marionette).until(expected.element_present(*self._menu_button_locator))
+	Wait(self.marionette).until(expected.element_displayed(returned_to_viewfinder))
+	# Make sure Timer icon is active
+	timer_enabled = Wait(self.marionette).until(expected.element_present(*self._timer_enabled_locator))
+	Wait(self.marionette).until(expected.element_displayed(timer_enabled))
+	
+    def tap_close_menu(self):
+ 	self.marionette.find_element(*self._close_menu_locator).tap()
+	returned_to_viewfinder = Wait(self.marionette).until(expected.element_present(*self._menu_button_locator))
+	Wait(self.marionette).until(expected.element_displayed(returned_to_viewfinder))     
+	
+	
     def tap_toggle_flash_button(self):
         initial_flash_mode = self.current_flash_mode
         toggle = self.marionette.find_element(*self._toggle_flash_button_locator)
@@ -195,6 +265,7 @@ class ImagePreview(Base):
     _gallery_button_locator = (By.CSS_SELECTOR, 'button[data-l10n-id="open-gallery"]')
     _confirm_delete_button = (By.ID, 'dialog-yes')
     _option_menu_locator = (By.CLASS_NAME, 'js-menu')
+    _not_video_preview_locator = (By.CSS_SELECTOR, '[style="display: none;"]')
 
     # for Gallery app switch
     _progress_bar_locator = (By.ID, 'progress')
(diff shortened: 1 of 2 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f5b42e5defbf0d2046c1f90463443378ff5ef875

tests/python/gaia-ui-tests/gaiatest/apps/camera/app.py
```diff
@@ -3,6 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import time
+import pdb
```
npark-mozilla: remove pdb for production checkin

tests/python/gaia-ui-tests/gaiatest/apps/camera/app.py
```diff
@@ -59,6 +81,21 @@ def take_photo(self):
...
         self.wait_for_thumbnail_visible()
 
+    def take_photo_with_timer(self):
+        # Wait for camera to be ready to take a picture
+        controls = self.marionette.find_element(*self._controls_locator)
+        Wait(self.marionette, timeout=20).until(
+            lambda m: controls.get_attribute('data-enabled') == 'true')
+         
+        self.tap_capture()
+        time.sleep(1)
```
npark-mozilla: I know you added time.sleep() for timing issues you saw, but we should try to figure out wait statements that we can rely on to avoid sleep() calls

tests/python/gaia-ui-tests/gaiatest/apps/camera/app.py
```diff
@@ -115,6 +152,39 @@ def tap_switch_source(self):
         Wait(self.marionette).until(lambda m: not current_camera_mode == self.camera_mode)
         self.wait_for_capture_ready()
 
+    def tap_switch_to_front_camera(self):
```
npark-mozilla: the function definitions don't seem to be indented properly.  Perhaps running pep8 tool would help?

tests/python/gaia-ui-tests/gaiatest/apps/camera/app.py
```diff
@@ -195,6 +265,7 @@ class ImagePreview(Base):
     _gallery_button_locator = (By.CSS_SELECTOR, 'button[data-l10n-id="open-gallery"]')
     _confirm_delete_button = (By.ID, 'dialog-yes')
     _option_menu_locator = (By.CLASS_NAME, 'js-menu')
+    _not_video_preview_locator = (By.CSS_SELECTOR, '[style="display: none;"]')
```
npark-mozilla: I think those locators should be defined up top

tests/python/gaia-ui-tests/gaiatest/tests/functional/camera/test_front_camera_take_photo_with_timer.py
```diff
@@ -0,0 +1,49 @@
...
+        self.assertTrue(self.camera.is_thumbnail_visible)
+
+        # Check that picture saved to SD card
+        self.wait_for_condition(lambda m: len(self.data_layer.picture_files) == self.previous_number_of_pictures + 1, 10)
+        self.assertEqual(len(self.data_layer.picture_files), self.previous_number_of_pictures + 1)
+        
+        # Tap on picture preview
+        self.camera.wait_for_thumbnail_visible()
+        self.camera.tap_thumbnail()
+        #self.apps.switch_to_displayed_app()
```
npark-mozilla: are they commented out because it has problems tapping image preview?

tests/python/gaia-ui-tests/gaiatest/apps/camera/app.py
```diff
@@ -115,6 +152,39 @@ def tap_switch_source(self):
...
+	self.marionette.find_element(*self._camera_switch_to_rear_locator).tap()
+	switch_to_front_button = Wait(self.marionette).until(expected.element_present(*self._camera_switch_to_front_locator))
+        Wait(self.marionette).until(expected.element_displayed(switch_to_front_button))
+		
+    def tap_menu_button(self):
+	self.marionette.find_element(*self._menu_button_locator).tap()
+	menu_opened = Wait(self.marionette).until(expected.element_present(*self._close_menu_locator))
+	Wait(self.marionette).until(expected.element_displayed(menu_opened))
+	
+    def tap_enable_timer(self):
```
npark-mozilla: perhaps this can take an argument so we know how long the countdown time is?

tests/python/gaia-ui-tests/gaiatest/apps/camera/app.py
```diff
@@ -212,6 +283,10 @@ def is_video_play_button_visible(self):
     def is_progress_bar_showing(self):
         return self.is_element_displayed(*self._video_progress_bar_locator)
 
+    @property
+    def not_video_preview(self):
+	return self.is_element_displayed(*self._not_video_preview_locator)
+
```
npark-mozilla: I ran this, and this won't work to simply detect the image preview.  If you view the app via WebIDE, in case of video preview, following is displayed: <img class="videoPoster" style="display: block...
But in case of image preview, following is displayed: <img class="image-view" src="blob:app://camera.gaiamoble.org...
So your method should check for the visibility of those classes to see whether it's an image view or video view.

tests/python/gaia-ui-tests/gaiatest/apps/camera/app.py
```diff
@@ -59,6 +81,21 @@ def take_photo(self):
...
+    def take_photo_with_timer(self):
+        # Wait for camera to be ready to take a picture
+        controls = self.marionette.find_element(*self._controls_locator)
+        Wait(self.marionette, timeout=20).until(
+            lambda m: controls.get_attribute('data-enabled') == 'true')
+         
+        self.tap_capture()
+        time.sleep(1)
+        # Wait for Timer to animate
+        timer_animating = self.marionette.find_element(*self._timer_animation_locator)
```
npark-mozilla: when I run the script, it fails here, and it looks like it's because the tap_capture() didn't run properly.  (it didn't even tap the capture button)  I think we need more checks inside tap_capture() method.  How about we add a check there to see whether the capture button is visible?  I think in some cases marionette does not complain when you tap an invisible element.

tests/python/gaia-ui-tests/gaiatest/apps/camera/app.py
mwargers: find_element will wait by itself until the element is found or it times, out, see:
http://mxr.mozilla.org/mozilla-central/source/testing/marionette/driver/marionette_driver/marionette.py#1574
So this sleep() call is kind of useless then anyway.

mwargers: If this is a problem here, it should also be a problem with take_photo method, actually: 
http://mxr.mozilla.org/gaia/source/tests/python/gaia-ui-tests/gaiatest/apps/camera/app.py#51

tests/python/gaia-ui-tests/gaiatest/apps/camera/app.py
mwargers: This selector looks far too generic to me. It will select anything with this style.

tests/python/gaia-ui-tests/gaiatest/apps/camera/app.py
JohanLorenzo: Yeah, that's a good point. If we want more contributors on the tests, we need to have PEP8 running. I filed [bug 1186388](https://bugzilla.mozilla.org/show_bug.cgi?id=1186388) for that.

tests/python/gaia-ui-tests/gaiatest/apps/camera/app.py
mwargers: In fact, this method looks very similar to the take_photo method. Why not use the existing take_photo method and add an extra argument with_timer that defaults to False, but in case you want to use it is doing the checks that you added here?

tests/python/gaia-ui-tests/gaiatest/tests/functional/camera/test_front_camera_take_photo_with_timer.py
```diff
@@ -0,0 +1,49 @@
...
+        self.apps.set_permission('Camera', 'geolocation', 'deny')
+
+    def test_capture_a_photo(self):
+        """https://moztrap.mozilla.org/manage/case/15800/"""
+        self.previous_number_of_pictures = len(self.data_layer.picture_files)
+
+        self.camera = Camera(self.marionette)
+        self.camera.launch()
+
+        # Open Camera Settings and turn on timer
```
JohanLorenzo: Usually, if you have to put a comment within a test, that's a sign that the code is not correctly written. The main reason in this case is:  Usually, the comment just rephrases what the following line does.
For example:
``` python
# Open the XYZ panel
my_app.open_xyz_panel()
```
Having this kind of comments creates duplication, and hence, doubles the maintenance we have to do. For instance, if one day I change the lines to the following and I forget to update the comment:
``` python
# Open the XYZ panel
my_app.open_abc_panel()
```
The comment will lie to the future readers, who'll be confused. ("Where is the truth? Which panel should it open?")

So, to prevent that, let's just rely on the code.

tests/python/gaia-ui-tests/gaiatest/tests/functional/camera/test_front_camera_take_photo_with_timer.py
```diff
@@ -0,0 +1,49 @@
...
+        self.camera.tap_menu_button()
```
JohanLorenzo: Regarding the comment above, the name of this function can be improved I think. How about: 
```python
camera.tap_settings_button()
```

tests/python/gaia-ui-tests/gaiatest/tests/functional/camera/test_front_camera_take_photo_with_timer.py
```diff
@@ -0,0 +1,49 @@
...
+        GaiaTestCase.setUp(self)
+
+        # Turn off geolocation prompt
+        self.apps.set_permission('Camera', 'geolocation', 'deny')
+
+    def test_capture_a_photo(self):
+        """https://moztrap.mozilla.org/manage/case/15800/"""
+        self.previous_number_of_pictures = len(self.data_layer.picture_files)
+
+        self.camera = Camera(self.marionette)
```
JohanLorenzo: I see we have the same issue in the other camera tests. We don't need to use `self` here. Attaching a variable to `self` here, is only useful if you do something with that variable in `setUp()` or `teardown()`. This is not the case here, so we can just have:
```python
camera = Camera(self.marionette)
```

tests/python/gaia-ui-tests/gaiatest/tests/functional/camera/test_front_camera_take_photo_with_timer.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def setUp(self):
+        GaiaTestCase.setUp(self)
+
+        # Turn off geolocation prompt
+        self.apps.set_permission('Camera', 'geolocation', 'deny')
+
+    def test_capture_a_photo(self):
+        """https://moztrap.mozilla.org/manage/case/15800/"""
+        self.previous_number_of_pictures = len(self.data_layer.picture_files)
```
JohanLorenzo: Same comment as [here](https://github.com/mozilla-b2g/gaia/pull/31243/files#r36295876)

tests/python/gaia-ui-tests/gaiatest/apps/camera/app.py
```diff
@@ -59,6 +81,21 @@ def take_photo(self):
         # Wait for thumbnail to appear
         self.wait_for_thumbnail_visible()
 
+    def take_photo_with_timer(self):
+        # Wait for camera to be ready to take a picture
+        controls = self.marionette.find_element(*self._controls_locator)
+        Wait(self.marionette, timeout=20).until(
```
JohanLorenzo: What's the advantage of setting the timeout to 20 seconds here?

tests/python/gaia-ui-tests/gaiatest/apps/camera/app.py
```diff
@@ -115,6 +152,39 @@ def tap_switch_source(self):
         Wait(self.marionette).until(lambda m: not current_camera_mode == self.camera_mode)
         self.wait_for_capture_ready()
 
+    def tap_switch_to_front_camera(self):
+	self.marionette.find_element(*self._camera_switch_to_front_locator).tap()
+	switch_to_rear_button = Wait(self.marionette).until(expected.element_present(*self._camera_switch_to_rear_locator))
+        Wait(self.marionette).until(expected.element_displayed(switch_to_rear_button))
+        
+    def tap_switch_to_rear_camera(self):
```
JohanLorenzo: The following methods have the same logic. I think we can extract this logic in a function:
```python
def tap_switch_to_rear_camera(self):
    self._tap_element_and_wait_for_other_to_be_displayed(self._camera_switch_to_rear_locator, self._camera_switch_to_front_locator)

# And so on, for the other methods

def _tap_element_and_wait_for_other_to_be_displayed(self, selector_to_tap, selector_to_wait):
    self.marionette.find_element(*selector_to_tap).tap()
    Wait(self.marionette).until(expected.element_displayed(
        Wait(self.marionette).until(expected.element_present(*selector_to_wait))))
```

tests/python/gaia-ui-tests/gaiatest/apps/camera/app.py
```diff
@@ -230,6 +305,10 @@ def tap_options(self):
         Wait(self.marionette).until(expected.element_displayed(
             Wait(self.marionette).until(expected.element_present(*self._option_menu_locator))))
 
+    def tap_picture_preview(self):
+        self.marionette.find_element(*self._thumbnail_button_locator).tap()
+	Wait(self.marionette).until(lambda m: self.is_image_preview_visible is True)
```
JohanLorenzo: Nit: No need for `is True`, here. Methods like `self.is_something` imply that a boolean is returned

tests/python/gaia-ui-tests/gaiatest/apps/camera/app.py
mwargers: Yeah, that seems to fix this intermittent failure that I also was seeing.

tests/python/gaia-ui-tests/gaiatest/apps/camera/app.py
JohanLorenzo: Agreed. Also this function does more than it's name implies. I would split it into:
```python
def tap_enable_timer_for(self, length_in_seconds=5):
    self._tap_enable_timer()
    self._choose_length(length_in_seconds)
    self._wait_for_timer_icon_in_view_finder()
```

PR #31268, opened 2015-08-06
Bug 1191863 - Fix fm radio transition.

autra:
FM radio never includes the gaia-theme stylesheet, so some default
variable were not set, especially --transition-duration.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=afbf94f12a37027ed8e77b445de162d1abacacb8

autra:
@vingtetun alternative would be to use a @import directive in `fm.css`. What do you prefer?

PR #31269, opened 2015-08-06
Bug 1191858 - remove the incorrect promise in the test

snowmantw:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6c3a4903c3222580b1ba735c619bacb12c7b10c9

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=08fdd5a903341c0cadf746aecc5ffcc3f08ea5ba

apps/system/test/unit/lockscreen/lockscreen_state_manager_test.js
```diff
@@ -155,7 +153,11 @@ suite('system/LockScreenStateManager', function() {
         return true;
       });
       subject.transfer({}).then(() => {
-        assert.isTrue(stubTransferOut.called, 'the method isn\'t called');
+        assert.isTrue(stubTransferTo.called,
+          'the |transferTo| method isn\'t called');
+        assert.isTrue(stubTransferOut.called,
+          'the |transferOut| method isn\'t called');
+        done();
       }).catch(done);
```
julienw: nits: this is how I write unit tests that have promises:
```javascript
functionReturningPromise().then(() => {
  // assertions
}).then(done, done);
```

Basically the same than what you wrote but I think it's more readable.

As another nit, you can use `sinon.assert.called(spy)` instead of `assert.isTrue(spy.called)`. This is a lot easier to debug when this fails, because sinon gives you some more information. Especially if you use `sinon.assert.calledWith` it will give you how the function was called.

snowmantw: 1. I prefer to express error catching with an explicit `catch`, which for me is a bit more clear than put it as the second parameter of `then`. Although people use Promise should all know the meaning of second parameter.

Anyway I will modify that as you said.

2. Thanks for your reminder. I'll use `called` instead of `isTrue`.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f22eee384c5f87b70d257e7db88085c9a70a66be

PR #31270, opened 2015-08-06
Bug 1176992 - Support linear, count, and exponential histograms in cu…

tamarahills:
…stom metrics.

### DIFF
--- a/apps/sharedtest/test/unit/advanced_telemetry_helper_test.js
+++ b/apps/sharedtest/test/unit/advanced_telemetry_helper_test.js
@@ -1,0 +1,371 @@
+'use strict';
+
+/* global AdvancedTelemetryHelper, MockNavigatorSettings  */
+
+require('/shared/js/advanced_telemetry_helper.js');
+require('/shared/js/settings_listener.js');
+requireApp('system/shared/test/unit/mocks/mock_navigator_moz_settings.js');
+
+/*
+ * This test suite has several sub-suites that verify that:
+ * 1) The metric is formatted correctly for a counter type of histogram
+ * 2) The metric is formatted correctly for a linear type of histogram
+ * 3) The metric is formatted correctly for an exponential type of histogram
+ * 4) The metric is formatted correctly for logging a value for all types
+ * 5) Disabling and Enabling the telemetry flag setting works correctly.
+ */
+suite('AdvancedTelemetryHelper:', function() {
+  var realMozSettings;
+
+  suiteSetup(function() {
+    realMozSettings = navigator.mozSettings;
+    navigator.mozSettings = MockNavigatorSettings;
+  });
+
+  suiteTeardown(function() {
+    navigator.mozSettings = realMozSettings;
+  });
+
+  /*
+   * This suite makes sure that the metrics are formatted properly.
+   */
+  suite('counter metric formatting:', function() {
+    var ath;
+
+    setup(function(done) {
+      this.sinon.stub(console, 'info');
+      // Suppress the error message
+      this.sinon.stub(console, 'warn');
+
+      MockNavigatorSettings.mTriggerObservers(
+        AdvancedTelemetryHelper.TELEMETRY_ENABLED_KEY, { settingValue: true });
+      done();
+    });
+
+    teardown(function() {
+    });
+
+    test('should call console info once', function(done) {
+      ath = new AdvancedTelemetryHelper(AdvancedTelemetryHelper.HISTOGRAM_COUNT,
+        'mymetric');
+      sinon.assert.calledOnce(console.info);
+      done();
+    });
+
+    test('should format the console message correctly', function(done) {
+      ath = new AdvancedTelemetryHelper(AdvancedTelemetryHelper.HISTOGRAM_COUNT,
+        'mymetric');
+      sinon.assert.calledWith(console.info, 'telemetry|mymetric|4|1|0|0');
+      done();
+    });
+
+    test('should not write format the message if missing param',
+    function(done) {
+      ath = new AdvancedTelemetryHelper(
+        AdvancedTelemetryHelper.HISTOGRAM_COUNT);
+      sinon.assert.notCalled(console.info);
+      // Should call a warning message.
+      sinon.assert.calledOnce(console.warn);
+      done();
+    });
+  });
+
+  suite('linear metric formatting:', function() {
+    var ath;
+
+    setup(function(done) {
+      this.sinon.stub(console, 'info');
+      // Suppress the error message
+      this.sinon.stub(console, 'warn');
+
+      MockNavigatorSettings.mTriggerObservers(
+        AdvancedTelemetryHelper.TELEMETRY_ENABLED_KEY, { settingValue: true });
+      done();
+    });
+
+    teardown(function() {
+    });
+
+    test('should call console info once', function(done) {
+      ath = new AdvancedTelemetryHelper(
+        AdvancedTelemetryHelper.HISTOGRAM_LINEAR, 'mymetric', 10, 1, 10000, 5);
+      sinon.assert.calledOnce(console.info);
+      done();
+    });
+
+    test('should format the message correctly with full params',
+    function(done) {
+      ath = new AdvancedTelemetryHelper(
+        AdvancedTelemetryHelper.HISTOGRAM_LINEAR, 'mymetric', 1, 10000, 5);
+      sinon.assert.calledWith(console.info, 'telemetry|mymetric|1|1|10000|5');
+      done();
+    });
+
+    test('should not write format message if missing param', function(done) {
+      ath = new AdvancedTelemetryHelper();
+      sinon.assert.notCalled(console.info);
+      // Should call a warning message.
+      sinon.assert.calledOnce(console.warn);
+      done();
+    });
+
+    test('should not write format message if 3 params sent', function(done) {
+      ath = new AdvancedTelemetryHelper('mymetric', 10, 1);
+      sinon.assert.notCalled(console.info);
+      // Should call a warning message.
+      sinon.assert.calledOnce(console.warn);
+      done();
+    });
+
+    test('should not write format message if 2 params sent', function(done) {
+      ath = new AdvancedTelemetryHelper('mymetric', 10);
+      sinon.assert.notCalled(console.info);
+      // Should call a warning message.
+      sinon.assert.calledOnce(console.warn);
+      done();
+    });
+
+    test('should handle 0 min value and correct to 1', function(done) {
+      ath = new AdvancedTelemetryHelper(
+        AdvancedTelemetryHelper.HISTOGRAM_LINEAR, 'mymetric', 0, 1000, 10);
+      sinon.assert.calledWith(console.info, 'telemetry|mymetric|1|1|1000|10');
+      done();
+    });
+
+    test('should not allow underscores in metricname', function(done) {
+      ath = new AdvancedTelemetryHelper(
+        AdvancedTelemetryHelper.HISTOGRAM_LINEAR, 'my_metric', 0, 1000, 10);
+      sinon.assert.calledWith(console.info, 'telemetry|my-metric|1|1|1000|10');
+      done();
+    });
+  });
+
+  suite('exponential metric formatting:', function() {
+    var ath;
+
+    setup(function(done) {
+      this.sinon.stub(console, 'info');
+      // Suppress the error message
+      this.sinon.stub(console, 'warn');
+
+      MockNavigatorSettings.mTriggerObservers(
+        AdvancedTelemetryHelper.TELEMETRY_ENABLED_KEY, { settingValue: true });
+      done();
+    });
+
+    teardown(function() {
+    });
+
+    test('should call console info once', function(done) {
+      ath = new AdvancedTelemetryHelper(
+        AdvancedTelemetryHelper.HISTOGRAM_EXP, 'mymetric', 0, 10000, 5);
+      sinon.assert.calledOnce(console.info);
+      done();
+    });
+
+    test('should format the message correctly with full params',
+    function(done) {
+      ath = new AdvancedTelemetryHelper(
+        AdvancedTelemetryHelper.HISTOGRAM_EXP, 'mymetric', 1, 10000, 5);
+      sinon.assert.calledWith(console.info,
+        'telemetry|mymetric|0|1|10000|5');
+      done();
+    });
+
+    test('should not write format the message if missing param',
+    function(done) {
+      ath = new AdvancedTelemetryHelper();
+      sinon.assert.notCalled(console.info);
+      // Should call a warning message.
+      sinon.assert.calledOnce(console.warn);
+      done();
+    });
+
+    test('should not write format message if 3 params sent', function(done) {
+      ath = new AdvancedTelemetryHelper('mymetric', 10, 1);
+      sinon.assert.notCalled(console.info);
+      // Should call a warning message.
+      sinon.assert.calledOnce(console.warn);
+      done();
+    });
+
+    test('should not write format message if 2 params sent', function(done) {
+      ath = new AdvancedTelemetryHelper('mymetric', 10);
+      sinon.assert.notCalled(console.info);
+      // Should call a warning message.
+      sinon.assert.calledOnce(console.warn);
+      done();
+    });
+
+    test('should handle 0 min value and correct to 1', function(done) {
+      ath = new AdvancedTelemetryHelper(
+        AdvancedTelemetryHelper.HISTOGRAM_EXP, 'mymetric', 0, 1000, 10);
+      sinon.assert.calledWith(console.info, 'telemetry|mymetric|0|1|1000|10');
+      done();
+    });
+
+    test('should not allow underscores in metricname', function(done) {
+      ath = new AdvancedTelemetryHelper(
+        AdvancedTelemetryHelper.HISTOGRAM_EXP, 'my_metric', 0, 1000, 10);
+      sinon.assert.calledWith(console.info, 'telemetry|my-metric|0|1|1000|10');
+      done();
+    });
+  });
+
+  suite('log metric formatting:', function() {
+    var ath;
+
+    setup(function(done) {
+      this.sinon.stub(console, 'info');
+      // Suppress the error message
+      this.sinon.stub(console, 'warn');
+
+      MockNavigatorSettings.mTriggerObservers(
+        AdvancedTelemetryHelper.TELEMETRY_ENABLED_KEY, { settingValue: true });
+      done();
+    });
+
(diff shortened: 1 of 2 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e15dfe7c27a3ba35f1ac2b33d2d31fbd967efb5f

shared/js/advanced_telemetry_helper.js
```diff
@@ -0,0 +1,154 @@
...
+    var message;
+ 
+    if (name) {
+      message = APP_TELEMETRY_LOG_PREFIX;
+
+      message += '|' + name;
+      // For a counter histogram, the value makes no difference, just adding
+      // 1 here to keep the parsing consistent.  The last parameter is the
+      // constant Histogram type -- in this case HISTOGRAM_COUNT = 4
+      message += '|1|4';
```
russnicoletti: I'm curious why the histogram type is added here but is not added for 'linear' or 'exp' histograms.

shared/js/advanced_telemetry_helper.js
```diff
@@ -0,0 +1,154 @@
...
+      return;
+    }
+
+    var message;
+
+    if (name && value) {
+      message = APP_TELEMETRY_LOG_PREFIX;
+
+      message += '|' + name;
+      message += '|' + value + '|1|';
```
russnicoletti: Using a variable to represent the histogram type (I assume that's what '1' represents here) would make the code more readable.

shared/js/advanced_telemetry_helper.js
```diff
@@ -0,0 +1,154 @@
...
+                                       typeof buckets !== 'undefined') {
+        message += min + '|' + max + '|' + buckets;
+      }
+      console.info(message);
+    }
+    else {
+      console.warn('A telemetry log entry requires two or five parameters');
+    }
+  };
+
```
russnicoletti: The 'exp' function seems identical to the 'linear' function expect for the histogram type. Perhaps it's worth refactoring to remove the duplicate code.

shared/js/advanced_telemetry_helper.js
```diff
@@ -0,0 +1,154 @@
...
+      return;
+    }
+
+    var message;
+
+    if (name && value) {
+      message = APP_TELEMETRY_LOG_PREFIX;
+
+      message += '|' + name;
+      message += '|' + value + '|0|';
```
russnicoletti: Using a variable to represent the histogram type (I assume that's what '0' represents here) would make the code more readable.

shared/js/advanced_telemetry_helper.js
```diff
@@ -0,0 +1,154 @@
...
+ *  var ath = new AdvancedTelemetryHelper();
+ *  // DO SOMETHING
+ *  // This creates and increments the counter on this metric called
+ *  // 'mycountmetric'
+ *  ath.count('mycountmetric'); // Count will be 1
+ *  ath.count('mycountmetric'); // Count is now 2
+ *  // DO SOMETHING
+ *  // This creates and logs a linear histogram called 'mylinearmetric' with
+ *  // value 100, minimum of 0, max of 1000, with 10 buckets.
+ *  ath.linear('mylinearmetric', 100, 0, 1000, 10);
```
russnicoletti: I'm wondering if it's cleaner to either 'initialize' the histogram (specify name, min, max, number of buckets) or to add a value, but not both. In other words, if the user has to explicitly initialize the histogram once, also adding a value at the same time might not be that helpful and may make the API less straightforward.

apps/sharedtest/test/unit/advanced_telemetry_helper_test.js
```diff
@@ -0,0 +1,225 @@
...
+    teardown(function() {
+    });
+
+    test('should not create a message if telemetry disabled', function(done) {
+      MockNavigatorSettings.mTriggerObservers(
+        AdvancedTelemetryHelper.TELEMETRY_ENABLED_KEY, { settingValue: false });
+
+      ath.exp('mymetric', 10, 0, 10000, 5);
+      ath.count('mymetric');
+      ath.count('linear');
```
russnicoletti: Nit: having a 'count' metric called 'linear' is confusing.

apps/sharedtest/test/unit/advanced_telemetry_helper_test.js
```diff
@@ -0,0 +1,225 @@
...
+    test('should format the message correctly without full params',
+    function(done) {
+      ath.linear('mymetric', 10);
+      sinon.assert.calledWith(console.info, 'telemetry|mymetric|10|1|');
+      done();
+    });
+
+    test('should not write format message if missing param', function(done) {
+      ath.linear();
+      sinon.assert.notCalled(console.info);
```
russnicoletti: It may be worth having additional negative test cases where, for example, there are three and four arguments specified.

apps/sharedtest/test/unit/advanced_telemetry_helper_test.js
```diff
@@ -0,0 +1,225 @@
...
+
+    test('should format the message correctly without full params',
+    function(done) {
+      ath.linear('mymetric', 10);
+      sinon.assert.calledWith(console.info, 'telemetry|mymetric|10|1|');
+      done();
+    });
+
+    test('should not write format message if missing param', function(done) {
+      ath.linear();
```
russnicoletti: It may be worth having additional negative test cases where, for example, there are three and four arguments specified.

apps/sharedtest/test/unit/advanced_telemetry_helper_test.js
```diff
@@ -0,0 +1,225 @@
...
+      ath.exp('mymetric', 10);
+      sinon.assert.calledWith(console.info,
+        'telemetry|mymetric|10|0|');
+      done();
+    });
+
+    test('should not write format the message if missing param',
+    function(done) {
+      ath.exp();
+      sinon.assert.notCalled(console.info);
```
russnicoletti: I think it's worth having additional negative test cases where, for example, there are three and four arguments specified.

apps/sharedtest/test/unit/advanced_telemetry_helper_test.js
```diff
@@ -0,0 +1,225 @@
...
+    function(done) {
+      ath.exp('mymetric', 10);
+      sinon.assert.calledWith(console.info,
+        'telemetry|mymetric|10|0|');
+      done();
+    });
+
+    test('should not write format the message if missing param',
+    function(done) {
+      ath.exp();
```
russnicoletti: I think it's worth having additional negative test cases where, for example, there are three and four arguments specified.

shared/js/advanced_telemetry_helper.js
tamarahills: added this.

shared/js/advanced_telemetry_helper.js
tamarahills: added.

shared/js/advanced_telemetry_helper.js
tamarahills: Yes, thanks.  I combined these two.

shared/js/advanced_telemetry_helper.js
tamarahills: Added.

shared/js/advanced_telemetry_helper.js
tamarahills: I thought about this and my thoughts were that this will be similar to a logging api for a developer.  Essentially, you don't want to use multiple lines to log.  Then the initialization becomes an extra line of code for the developer.  So, I opted to keep it thin.  Let me know if you want to talk more about this as there may be some points I'm missing.

apps/sharedtest/test/unit/advanced_telemetry_helper_test.js
tamarahills: Agreed.  I changed this.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5e2d0eb4b7fa5bab7349e03d5735014b6e8c6f5d

shared/js/advanced_telemetry_helper.js
```diff
@@ -0,0 +1,158 @@
...
+      if (typeof min !== 'undefined') { varCount++; }
+      if (typeof max !== 'undefined') { varCount++; }
+      if (typeof buckets !== 'undefined') { varCount++; }
+
+      if (varCount > 3 && varCount < 6) {
+        console.warn('telemetry message must have 2 or 5 params');
+        return;
+      }
+      if (varCount === 6) {
+        if (min < 1) {
```
russnicoletti: I can't think of an example, but it seems like zero could be a valid minimum.

apps/sharedtest/test/unit/advanced_telemetry_helper_test.js
```diff
@@ -0,0 +1,255 @@
...
+
+    test('should not write format message if 4 params sent', function(done) {
+      ath.linear('mymetric', 10, 1, 1000);
+      sinon.assert.notCalled(console.info);
+      // Should call a warning message.
+      sinon.assert.calledOnce(console.warn);
+      done();
+    });
+
+    test('should handle 0 min value and correct to 1', function(done) {
```
russnicoletti: I'm thinking it's possible for zero to be a valid minimum value. For example, a metric like 'contact-search-results' that tracks the number of contacts found per search.

shared/js/advanced_telemetry_helper.js
marshall: This is definitely subjective, but I think I agree with Russ. an extra line of code could make the encapsulation of each histogram a little more obvious, i.e:

```
var linear = ath.linear('mylinearmetric', 0, 1000, 10);
linear.log(100);
linear.log(800);

var exp = ath.exp('myexponentialmetric', 0, 10000, 10);
exp.log(10);
exp.log(8000);

```

shared/js/advanced_telemetry_helper.js
```diff
@@ -0,0 +1,158 @@
...
+  const HISTOGRAM_COUNT = 4;
+  const HISTOGRAM_EXP = 0;
+  const HISTOGRAM_LINEAR = 1;
+  // This is exposed for unit testing purposes
+  ATH.TELEMETRY_ENABLED = false;
+  ATH.TELEMETRY_ENABLED_KEY = 'debug.performance_data.advanced_telemetry';
+  ATH.SETTINGS_LISTENER = false;
+
+  function AdvancedTelemetryHelper() {
+    AdvancedTelemetryHelper.prototype.getTelemetrySetting();
```
marshall: any reason you're calling this function using the prototype instead of just `this.getTelemetrySetting()` ?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ded00df8b0e801fe2f1c8b21920e28174da64266

shared/js/advanced_telemetry_helper.js
```diff
@@ -0,0 +1,158 @@
...
+    if (ATH.SETTINGS_LISTENER) {
+      return;
+    }
+
+    this.telemetryEnabledListener = function telemetryEnabledListener(enabled) {
+      if (enabled) {
+        ATH.TELEMETRY_ENABLED = true;
+      } else {
+        ATH.TELEMETRY_ENABLED = false;
+      }
```
marshall: nit: this can be simplified, maybe just use:
`ATH.TELEMETRY_ENABLED = enabled`

shared/js/advanced_telemetry_helper.js
```diff
@@ -0,0 +1,158 @@
...
+  // <max>: the maximum numeric value for this histogram
+  // <buckets>: the total number of buckets for this histogram.
+  // min/max/buckets are required on the first call.  For subsequent calls they
+  // are optional.
+  AdvancedTelemetryHelper.prototype.linear =
+  function(name, value, min, max, buckets) {
+    if (!ATH.TELEMETRY_ENABLED) {
+      return;
+    }
+    ATH.prototype.writeHistogram(name, value, HISTOGRAM_LINEAR,
```
marshall: should be using `this.writeHistogram` ?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=27ddb45fb3dbd65173098f4aedccb89704208b34

shared/js/advanced_telemetry_helper.js
```diff
@@ -0,0 +1,190 @@
...
+  'use strict';
+
+  const APP_TELEMETRY_LOG_PREFIX = 'telemetry';
+  const ATH = AdvancedTelemetryHelper;
+  const HISTOGRAM_COUNT = 4;
+  const HISTOGRAM_EXP = 0;
+  const HISTOGRAM_LINEAR = 1;
+  // This is exposed for unit testing purposes
+  ATH.TELEMETRY_ENABLED_KEY = 'debug.performance_data.advanced_telemetry';
+  ATH.SETTINGS_LISTENER = false;
```
russnicoletti: This variable name could be more descriptive. As it is, it's purpose is not obvious.

shared/js/advanced_telemetry_helper.js
```diff
@@ -0,0 +1,190 @@
...
+ *  Exponential Metric Usage Example:
+ *  var expMetric = new AdvancedTelemetryHelper();
+ *  // This creates and logs a linear histogram called 'myexpmetric' with
+ *  // minimum of 1 (0 not allowed as a minimum), max of 1000, with 10 buckets.
+ *  expMetric.exp('myexpmetric', 1, 1000, 10);
+ *  // DO SOMETHING
+ *  // This adds to the existing histogram a value of 800.
+ *  expMetric.log(800); // The histogram now has a value in the bucket
+ *                      // that holds the value 800
+ *
```
russnicoletti: 'log' is describing how the function is implemented as opposed to what it is doing. Perhaps 'add' would be more intuitive.

russnicoletti: I'm trying to decide whether or not to use the constructor initialize the histogram. E.g.:
`AdvancedTelemetryHelper('count')`
`AdvancedTelemetryHelper('linear', 1, 1000, 10)`
`AdvancedTelemetryHelper('exp', 1, 1000, 10)`

Pros:
* Makes the API more efficient in that only two function calls are required (construct object, add value)
* Makes it more obvious an instance of the ATH is bound to a specific histogram type (one instance cannot be used to create two or more histograms)

Cons
* The API isn't as self-describing without the descriptive function names corresponding to the types of histograms

I'm not going to take a stand one way or another but I'd say it's worth mentioning in the module description how an instance of the object can only be used to create/add values to a single histogram type.

shared/js/advanced_telemetry_helper.js
```diff
@@ -0,0 +1,190 @@
...
+      }
+    }
+  };
+
+  // API for logging against a pre-existing histogram.
+  // Pre-requisite is to call .count(..), .linear(...), or .exp(...) first.
+  // <value>: The value to log against the histogram.
+  //    for a counter histogram, this can be empty or 1 can be passed
+  //    for .exp or .linear a valid value > 0 must be passed.
+  AdvancedTelemetryHelper.prototype.log = function(value) {
```
russnicoletti: I think it's better to encode the purpose for the API function in the name of the function as opposed to how it's being implemented. 

Nit: The comment is also describing the implementation as opposed to the purpose.

shared/js/advanced_telemetry_helper.js
tamarahills: I was thinking about this myself and can't remember why I discarded the idea... It might have been for side effects in the constructor... Anyhow, I think it's worthwhile, but I think what I'm going to try is requiring the user pass in the histogram type.  It's an extra parameter, but I think it will cut down on the extra logic you mention in your "cons".

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8d5aa819137195ef920429ea431e12056f8d6c82

shared/js/advanced_telemetry_helper.js
```diff
@@ -0,0 +1,203 @@
...
+ *  linear.add(800); // The histogram now has two values, one in the 15 bucket
+ *                   // and one in the bucket that holds the value 800
+ *
+ * Exponential Metric Usage Example:
+ *  // This creates an exponential histogram called 'myexp' with minimum
+ *  // of 1 (0 not allowed as a minimum), max of 1000, with 10 buckets.
+ *  var exp = new AdvancedTelemetryHelper(ATH.HISTOGRAM_EXPONENTIAL, 'myexp',
+ *      1, 1000, 10);
+ *  // This adds the value 15 to the histogram 'myexp'
+ *  exp.add('myexp', 15);
```
russnicoletti: It seems 'add' takes only one argument.

shared/js/advanced_telemetry_helper.js
```diff
@@ -0,0 +1,203 @@
...
+ *  // DO SOMETHING
+ *  count.add(); //Increment the count. The count of the histogram is now 2.
+ *
+ * Linear Metric Usage Example:
+ *  // This creates a linear histogram called 'mylinear' with minimum
+ *  // of 1 (0 not allowed as a minimum), max of 1000, with 10 buckets.
+ *  var linear = new AdvancedTelemetryHelper(ATH.HISTOGRAM_LINEAR, 'mylinear',
+ *      1, 1000, 10);
+ *  // This adds the value 15 to the histogram 'mylinear'
+ *  linear.add('mylinear', 15);
```
russnicoletti: It seems 'add' takes only one argument.

shared/js/advanced_telemetry_helper.js
```diff
@@ -0,0 +1,203 @@
...
+  // <buckets>: the total number of buckets for this histogram.
+  function AdvancedTelemetryHelper(type, name, min, max, buckets) {
+    if (!ATH.SETTINGS_INITIALIZED) {
+      ATH.SETTINGS_INITIALIZED = true;
+      ATH.init();
+    }
+
+    if (typeof type !== 'undefined' && typeof name !== 'undefined') {
+      if (type == ATH.HISTOGRAM_COUNT) {
+        this.count(name);
```
russnicoletti: It seems possible to remove the 'count' function and use createHistogram here, passing '0' for min, max, and buckets.

tamarahills: Yep!  Good call.  Thanks.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0e5bc61c02cc819e59a18d7fb912c88daeff506d

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=cb1bb46ac3bddd8d9588c640758ca471f3c12e30

PR #31288, opened 2015-08-07
Bug 1191849 - Add --locale parameter to select phone's locale in gaia ui test 

nojunpark:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3f39c2e86825c19badefafccb457040fee2da47c

tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
```diff
@@ -28,7 +28,7 @@ def __init__(self, *args, **kwargs):
         self.reference_path = kwargs.pop('reference_path')
         self.screenshots_path = kwargs.pop('screenshots_path')
         self.mismatch_path = os.path.join(self.screenshots_path, 'mismatches')
-        self.locale = kwargs.pop('locale')
+
```
mwargers: Stray return

tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
```diff
@@ -50,14 +50,7 @@ def tearDown(self):
...
 
-    # marionette hook to override default locale setting
-    def modify_settings(self, settings):
-
-        if self.locale != 'undefined':
-            settings['language.current'] = self.locale
-        return settings
-
-    def take_screenshot(self, page_name=None, prewait=0):
+    def take_screenshot(self,page_name=None, prewait=0):
```
mwargers: Removed a space that shouldn't be removed.

tests/python/gaia-ui-tests/gaiatest/mixins/gaia.py
```diff
@@ -19,6 +19,11 @@ def __init__(self, **kwargs):
                          default=False,
                          help='restart target instance between tests')
 
+        group.add_option('--locale',
+                         default="undefined",
```
JohanLorenzo: I wonder if we could simplify the code, if we put `default='en-US'`, here. We wouldn't need to do checks like 
``` python 
if self.locale != 'undefined'
```
Also, what happens if a user doesn't put that value into the testvars.json? The help message doesn't make it clear. Something like this, looks a bit clearer, I think:
```python
       group.add_option('--locale',
                        default = "en-US",
                        help = "locale for the device, "
                               "This value can be overridden with testvars.json")
```

tests/python/gaia-ui-tests/gaiatest/mixins/gaia.py
```diff
@@ -19,6 +19,11 @@ def __init__(self, **kwargs):
...
+                         help="locale for the device, "
+                                "When not defined, will use the value from testvars.json file")
```
JohanLorenzo: You need add the key in `testvars_template.json`

npark-mozilla: added

tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
npark-mozilla: fixed

tests/python/gaia-ui-tests/gaiatest/gaia_graphics_test.py
npark-mozilla: fixed

tests/python/gaia-ui-tests/gaiatest/mixins/gaia.py
npark-mozilla: So I did put default='en-US' initially, but then it would default to English each time the automation test starts, and this might not be ideal.  Currently it is not mandatory to have language.locale value set in the testvars.json file, and the current default behavior is to go with the device's last language configuration, which is more intuitive

I will clarify the help comment, but it should be "This value overrides the value in testvars.json" since this parameter processing happens after getting the default values from testvars.  I'll change the comment to reflect that.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=555b90f0b4d52073b838792207b7a1f9e9443f89

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7142c23061bbfe06549fd7477aa0133521d5d6bd

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e0626768d0bef00653ccb30919c10b7452d47f9a

tests/python/gaia-ui-tests/gaiatest/mixins/gaia.py
mwargers: I don't think we should put a default for language.current in there. I think the default of the pone is without anything set in language.current and it defaults to a language just fine (en-US).

npark-mozilla: Yes, the default language of the phone immediately after the flash of a new build is en-US, but after that the phone boots up with whatever language that was previously set. IF we put a default language value, it might be quite a change to non-english testers.

JohanLorenzo: Okay, let's show the in the comment that the default language is the one set on the device, then.

npark-mozilla: good idea, will do.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4ba06a9a4b5d077e9beee1ece963967777b62f8c

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5c265e5737f7d4bf559da8512218232870c2abe1

PR #31292, opened 2015-08-08
Bug 1190548 - Implement explanations inside of gaia-switch

KevinGrandon:
https://bugzilla.mozilla.org/show_bug.cgi?id=1190548

### DIFF
--- a/apps/settings/elements/bluetooth.html
+++ b/apps/settings/elements/bluetooth.html
@@ -7,17 +7,17 @@
 
     <div>
       <ul>
-        <li class="bluetooth-status hint">
+        <li class="bluetooth-status">
           <gaia-switch data-ignore>
             <label data-l10n-id="bluetooth"></label>
+            <details data-l10n-id="bluetooth-enable-msg" id="bluetooth-enable-msg"></details>
           </gaia-switch>
-          <p data-l10n-id="bluetooth-enable-msg" id="bluetooth-enable-msg" class="explanation"></p>
         </li>
-        <li hidden class="device-visible hint">
+        <li hidden class="device-visible">
           <gaia-switch data-ignore>
             <label data-l10n-id="bluetooth-visible-to-all"></label>
+            <details><bdi class="bluetooth-device-name"></bdi></details>
           </gaia-switch>
-          <p class="explanation"><bdi class="bluetooth-device-name"></bdi></p>
         </li>
         <li hidden id="bluetooth-rename">
           <button id="rename-device" data-l10n-id="rename-device" disabled></button>
--- a/shared/elements/gaia_switch/style.css
+++ b/shared/elements/gaia_switch/style.css
@@ -47,7 +47,7 @@ gaia-switch {
   content: '';
   position: absolute;
   right: 0;
-  top: 50%;
+  top: 2.5rem;
   width: 6rem;
   margin: -1.4rem 0 0;
   height: 2.7rem;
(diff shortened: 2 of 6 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e4a613b031a5af90eeb57150854cf4072b001c69

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=daadc4648f810065f876bac09ce2d0fbe76fe53d

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2dc87a3e2e99c7960055bd05b30f6b067623de8c

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8a22f63f365b9f7f664b1475547de94e9ff47816

apps/settings/elements/bluetooth.html
```diff
@@ -7,17 +7,17 @@ <h1 data-l10n-id="bluetooth-header"></h1>
...
             <label data-l10n-id="bluetooth"></label>
+            <details data-l10n-id="bluetooth-enable-msg" id="bluetooth-enable-msg"></details>
           </gaia-switch>
-          <p data-l10n-id="bluetooth-enable-msg" id="bluetooth-enable-msg" class="explanation"></p>
         </li>
-        <li hidden class="device-visible hint">
+        <li hidden class="device-visible">
           <gaia-switch data-ignore>
             <label data-l10n-id="bluetooth-visible-to-all"></label>
+            <details><bdi class="bluetooth-device-name"></bdi></details>
```
gasolin: :+1

shared/elements/gaia_switch/style.css
```diff
@@ -47,7 +47,7 @@ gaia-switch {
   content: '';
   position: absolute;
   right: 0;
-  top: 50%;
+  top: 2.5rem;
```
gasolin: haven't try in real device yet, but does the `title only` style panel's text position be effect by this change?

(The label position for `gaia-switch > label` & `gaia-switch > label + details` are not the same)

KevinGrandon: So this will change the switch to always be near the top of the label. We can follow what we did previously, but it does seem a bit weird/inconsistent to me. Happy to implement how you think is best though, previously we did: 

label only -> switch is at vertical center
label + details -> switch is aligned at top

The problem I see with that is what happens when you have a label that wraps multiple lines as well as a details object. In that case the switch would be aligned near the top, so why not always align at the top? Maybe that's not really a common use-case though.

gasolin: I flash the device and see no obvious visual difference between previous and patched version, so I think the patch is good enough :)

shared/elements/gaia_switch/style.css
```diff
@@ -88,3 +88,22 @@ span[dir="rtl"]#switch input ~ span:after {
   left: 0;
   right: unset;
 }
+
+/******************************************************************************
+ * Helper elements
```
gasolin: Helper elements to explain the functionality of this switch

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f2835618a0f08fc911696eb1dcdfe27dffb45799

PR #31336, opened 2015-08-12
Bug 1193978 - Automated Test [Clock] Use stopwatch, set a few laps, reset.

cuff-links:
…topwatch.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9271e415771e98eba3a78130ce097f5cee310a28

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_laps.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def setUp(self):
+        GaiaTestCase.setUp(self)
+
+        self.clock = Clock(self.marionette)
+        self.clock.launch()
+
+    def test_click_run_stopwatch_laps(self):
+
+        # Switch to the stopwatch view.
```
JohanLorenzo: I think we can get rid of all the comments in this file. One comment is a duplication of the following code line. If one day, the test needed to be changed, we'll have 2 places to maintain for each change. Some people might forget to change the comment => The comment will lie, which will confuse future readers.

To prevent that, the code should be self-explanatory, which is mostly the case here.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_then_reset.py
```diff
@@ -0,0 +1,39 @@
...
+
+class TestClockRunStopWatch(GaiaTestCase):
+
+    def setUp(self):
+        GaiaTestCase.setUp(self)
+
+        self.clock = Clock(self.marionette)
+        self.clock.launch()
+
+    def test_click_run_stopwatch_then_reset(self):
```
JohanLorenzo: Same comment about the unnecessarily comments.

tests/python/gaia-ui-tests/gaiatest/apps/clock/app.py
```diff
@@ -27,9 +27,18 @@ def launch(self):
...
+    def switch_view(self, view_name):
+        Wait(self.marionette).until(
+            expected.element_present(*(By.ID, self._clock_views.get(view_name)))).tap()
+        if view_name == 'stopwatch':
+            from gaiatest.apps.clock.regions.stopwatch import StopWatch
+            return StopWatch(self.marionette)
+        elif view_name == 'timer':
+            from gaiatest.apps.clock.regions.timer import Timer
+            return Timer(self.marionette)
+        else:
```
JohanLorenzo: I'd suggest to fail if the `view_name` is not expected. Otherwise, as a test writer, if I made a typo, I wouldn't immediately understand why I go to the Alarm panel instead.

That is to say:
``` python
else view_name == 'alarm':
    from gaiatest.apps.clock.regions.alarm import Alarm
    return Alarm(self.marionette)
else: 
    raise AttributeError('{} is not view that you can switch to'.format(view_name))
```

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/alarm.py
```diff
@@ -2,77 +2,63 @@
...
 
-class NewAlarm(Clock):
-
-    _alarm_name_locator = (By.ID, 'alarm-name')
-    _repeat_menu_locator = (By.ID, 'repeat-menu')
-    _sound_menu_locator = (By.ID, 'sound-menu')
-    _snooze_menu_locator = (By.ID, 'snooze-menu')
-    _done_locator = (By.ID, 'alarm-done')
-    _time_button_locator = (By.XPATH, "//li[input[@id='time-select']]")
+class Alarm(Clock):
```
JohanLorenzo: I wonder if we can also inherit from `PageRegion`. It looks safer to me.

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/alarm.py
```diff
@@ -2,77 +2,63 @@
...
+    _stopwatch_tab_locator = (By.ID, 'stopwatch-tab')
```
JohanLorenzo: The next 3 locators should belong to the Clock class only. Otherwise, we duplicate the logic into 2 Page classes.

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/alarm.py
```diff
@@ -2,77 +2,63 @@
...
-        self.marionette.find_element(*self._time_button_locator).tap()
-        from gaiatest.apps.system.regions.time_picker import TimePicker
-        return TimePicker(self.marionette)
-
-
-class EditAlarm(NewAlarm):
-
-    def __init__(self, marionette):
-        NewAlarm.__init__(self, marionette)
+    def alarms(self):
```
JohanLorenzo: I'd rebrand that to `items`. So the use would be: `alarm.items` instead of `alarm.alarms`. What do you think?

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/new_alarm.py
```diff
@@ -0,0 +1,78 @@
...
+
+class NewAlarm(Clock):
+
+    _alarm_name_locator = (By.ID, 'alarm-name')
+    _alarm_view_locator = (By.CSS_SELECTOR, 'section.skin-dark.panel.active')
+    _repeat_menu_locator = (By.ID, 'repeat-menu')
+    _sound_menu_locator = (By.ID, 'sound-menu')
+    _snooze_menu_locator = (By.ID, 'snooze-menu')
+    _done_locator = (By.ID, 'alarm-done')
+    _time_button_locator = (By.XPATH, "//li[input[@id='time-select']]")
```
JohanLorenzo: I wonder if we could just the locator to: `(By.ID, 'time-select')`

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
```diff
@@ -0,0 +1,63 @@
...
+    _stopwatch_pause_locator = (By.CSS_SELECTOR, '#stopwatch-controls .stopwatch-pause')
+    _stopwatch_reset_locator = (By.CSS_SELECTOR, '#stopwatch-controls .stopwatch-reset')
+    _stopwatch_resume_locator = (By.CSS_SELECTOR, '#stopwatch-controls .stopwatch-resume')
+    _stopwatch_lap_locator = (By.CSS_SELECTOR, '#stopwatch-controls .stopwatch-lap')
+    _all_laps_locator = (By.CSS_SELECTOR, '.stopwatch-laps li')
+
+    def __init__(self, marionette):
+        Clock.__init__(self, marionette)
+        view = self.marionette.find_element(*self._stopwatch_view_locator)
+        Wait(self.marionette).until(lambda m: view.location['x'] == 0 and view.is_displayed())
```
JohanLorenzo: I wonder if this wait could be factorized in `Clock`. Not sure of that

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
```diff
@@ -0,0 +1,63 @@
...
+    @property
+    def current_time(self):
+        current_time = Wait(self.marionette).until(expected.element_present(*self._stopwatch_time_locator))
+        return current_time.text
+
+    @property
+    def laps(self):
+        return [self.LapItem(self.marionette, lap) for lap in self.marionette.find_elements(*self._all_laps_locator)]
+
+    def start_stopwatch(self):
```
JohanLorenzo: I think we can simplify the name to just start. So a usage of this class can be: `stopwatch.start()` instead of `stopwatch.start_stopwatch()`

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_laps.py
```diff
@@ -0,0 +1,54 @@
...
+        stopwatch_view = self.clock.switch_view("stopwatch")
+
+        # Verify that the stopwatch time is at 0.
+        assert stopwatch_view.current_time == '00:00.00'
```
JohanLorenzo: `GaiaTestCase` inherits from [unittest.TestCase](https://docs.python.org/2/library/unittest.html), so you can use self.assertEqual(stopwatch_view.current_time, '00:00.00'). If the test fails at that place, the error message will be clearer. Python will show the string difference. With the current assert, Python will just say: AssertionError with the stacktrace.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_set_alarm.py
```diff
@@ -14,8 +14,10 @@ def setUp(self):
         self.clock.launch()
 
     def test_clock_set_alarm(self):
+        # Added a view switching method to app.py for getting a Clock object.
+        alarm = self.clock.switch_view('alarm')
```
JohanLorenzo: I think we also miss the change in the "create new alarm" test

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/new_alarm.py
silne30: I could change that. But that was existing functionality.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_laps.py
silne30: Nice. Updating now.

tests/python/gaia-ui-tests/gaiatest/apps/clock/app.py
silne30: I thought of using an Enum but it seems like overkill. Especially since this is the only place it would be used. If we were going to be using it throughout our whole suite, I think an enum would be more appropriate. What say you @JohanLorenzo?

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/alarm.py
silne30: It makes sense but I wanted to use the functionality of the Clock __init__. Also, when I thought about the actual views, they seem to be views derived from the clock view rather than just a region of the Clock view. They are an entire view for themselves. That's why I stuck with them inheriting from Clock. What say you @JohanLorenzo?

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/alarm.py
silne30: Done

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/alarm.py
silne30: Maybe to  be more specific, alarm_items. Items can be a bit ambiguous as was alarms. That's what I did in the Stopwatch app. I did lap_items.

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/new_alarm.py
silne30: Done.

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
silne30: Nah. It should be in the init of this object.

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
silne30: Agreed. Done.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_then_reset.py
silne30: Done

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/alarm.py
JohanLorenzo: Yeah, `alarm_items` is fine.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=63cab88f55bc8621b21ac00cfdfcb89c776b0180

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_set_alarm.py
silne30: Fixed that as well. I had not updated this test.

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
```diff
@@ -0,0 +1,63 @@
+ # This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from marionette_driver import expected, By, Wait
+
+from gaiatest.apps.clock.app import Clock
+from gaiatest.apps.base import PageRegion
+
```
mwargers: One return too many?

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_then_reset.py
```diff
@@ -0,0 +1,32 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from gaiatest import GaiaTestCase
+from gaiatest.apps.clock.app import Clock
+
```
mwargers: Needs one more return

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_laps.py
```diff
@@ -0,0 +1,44 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from gaiatest import GaiaTestCase
+from gaiatest.apps.clock.app import Clock
+
```
mwargers: Need one more return

tests/python/gaia-ui-tests/gaiatest/apps/clock/app.py
```diff
@@ -9,13 +9,11 @@
 
 
 class Clock(Base):
-
     name = 'Clock'
-
-    _alarm_view_locator = (By.ID, 'edit-alarm')
     _alarm_create_new_locator = (By.ID, 'alarm-new')
-    _all_alarms_locator = (By.CSS_SELECTOR, '#alarms li')
+    _clock_view_locator = (By.ID, 'alarm-tab')
```
mwargers: Shouldn't this locator be in alarm.py?

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/alarm.py
```diff
@@ -2,77 +2,61 @@
...
-class NewAlarm(Clock):
-
-    _alarm_name_locator = (By.ID, 'alarm-name')
-    _repeat_menu_locator = (By.ID, 'repeat-menu')
-    _sound_menu_locator = (By.ID, 'sound-menu')
-    _snooze_menu_locator = (By.ID, 'snooze-menu')
-    _done_locator = (By.ID, 'alarm-done')
-    _time_button_locator = (By.XPATH, "//li[input[@id='time-select']]")
+class Alarm(Clock):
+    _edit_alarm_view_locator = (By.ID, 'edit-alarm')
```
mwargers: This locator doesn't seem to be used.

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/new_alarm.py
```diff
@@ -0,0 +1,78 @@
...
+
+from marionette_driver import expected, By, Wait
+
+from gaiatest.apps.clock.app import Clock
+
+
+class NewAlarm(Clock):
+
+    _alarm_name_locator = (By.ID, 'alarm-name')
+    _alarm_view_locator = (By.CSS_SELECTOR, 'section.skin-dark.panel.active')
```
mwargers: I'm not sure this is a great locator to use. It seems rather general (and theme specific?). Perhaps it's better to use section[data-panel-id="alarm_edit"] here?

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/new_alarm.py
```diff
@@ -0,0 +1,78 @@
...
+    def tap_time(self):
+        self.marionette.find_element(*self._time_button_locator).tap()
+        from gaiatest.apps.system.regions.time_picker import TimePicker
+        return TimePicker(self.marionette)
+
+
+class EditAlarm(NewAlarm):
+
+    def __init__(self, marionette):
+        NewAlarm.__init__(self, marionette)
```
mwargers: Is this necessary? Because you're already inheriting form NewAlarm here, NewAlarm.__init__ is already automatically called, no?

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/new_alarm.py
```diff
@@ -0,0 +1,78 @@
...
+        Wait(self.marionette).until(expected.element_displayed(done))
+        done.tap()
+        from gaiatest.apps.clock.regions.alarm import Alarm
+        return Alarm(self.marionette)
+
+    def tap_time(self):
+        self.marionette.find_element(*self._time_button_locator).tap()
+        from gaiatest.apps.system.regions.time_picker import TimePicker
+        return TimePicker(self.marionette)
+
```
mwargers: One return too many?

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/new_alarm.py
mwargers: Actually, EditAlarm is the same as NewAlarm.

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class StopWatch(Clock):
+    _stopwatch_view_locator = (By.CSS_SELECTOR,'#stopwatch-panel.active')
```
mwargers: I don't think you need the .active part here. This locator is used to make sure this panel is visible anyway.

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
```diff
@@ -0,0 +1,63 @@
...
+    @property
+    def current_time(self):
+        current_time = Wait(self.marionette).until(expected.element_present(*self._stopwatch_time_locator))
+        return current_time.text
+
+    @property
+    def laps(self):
+        return [self.LapItem(self.marionette, lap) for lap in self.marionette.find_elements(*self._all_laps_locator)]
+
+    def start(self):
```
mwargers: Perhaps it's better to call it tap_start here, since this is tapping on the start button?

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
```diff
@@ -0,0 +1,63 @@
...
+        Wait(self.marionette).until(expected.element_present(*self._stopwatch_start_locator)).tap()
+
+    def pause(self):
```
mwargers: Same here.

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
```diff
@@ -0,0 +1,63 @@
...
+        Wait(self.marionette).until(expected.element_present(*self._stopwatch_pause_locator)).tap()
+
+    def reset(self):
```
mwargers: Same here?

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
```diff
@@ -0,0 +1,63 @@
...
+        Wait(self.marionette).until(expected.element_present(*self._stopwatch_reset_locator)).tap()
+
+    def resume(self):
```
mwargers: Same here?

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
```diff
@@ -0,0 +1,63 @@
...
+        Wait(self.marionette).until(expected.element_present(*self._stopwatch_resume_locator)).tap()
+
+    def record_lap(self):
```
mwargers: Same here?

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_laps.py
```diff
@@ -0,0 +1,44 @@
...
+class TestClockRunStopWatch(GaiaTestCase):
+
+    def setUp(self):
+        GaiaTestCase.setUp(self)
+
+        self.clock = Clock(self.marionette)
+        self.clock.launch()
+
+    def test_click_run_stopwatch_laps(self):
+
```
mwargers: This is not based on a specific Moztrap test by any chance?

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_laps.py
```diff
@@ -0,0 +1,44 @@
...
+        stopwatch_view.start()
+
+        self.assertNotEqual(stopwatch_view.current_time, '00:00.00')
+
+        stopwatch_view.record_lap()
+
+        self.assertEqual(len(stopwatch_view.laps), 2)
+        self.assertNotEqual(stopwatch_view.laps[0].time, '00:00.00')
+        self.assertNotEqual(stopwatch_view.laps[1].time, '00:00.00')
```
mwargers: Perhaps we could add another assert that checks that laps[0].time is not equal to laps[1].time?

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_then_reset.py
```diff
@@ -0,0 +1,32 @@
...
+class TestClockRunStopWatch(GaiaTestCase):
+
+    def setUp(self):
+        GaiaTestCase.setUp(self)
+
+        self.clock = Clock(self.marionette)
+        self.clock.launch()
+
+    def test_click_run_stopwatch_then_reset(self):
+
```
mwargers: By any chance based on a particular Moztrap case?

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_then_reset.py
```diff
@@ -0,0 +1,32 @@
...
+        self.assertEqual(stopwatch_view.current_time, '00:00.00')
+
+        stopwatch_view.start()
+
+        self.assertNotEqual(stopwatch_view.current_time, '00:00.00')
+
+        stopwatch_view.pause()
+
+        self.assertNotEqual(stopwatch_view.current_time, '00:00.00')
```
mwargers: I think it would be good for a check that the pause button actually pauses too.
So something like that after 1 second or so, stopwatch_view.current_time has stayed the same.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_set_alarm.py
```diff
@@ -14,8 +14,10 @@ def setUp(self):
         self.clock.launch()
 
     def test_clock_set_alarm(self):
+        # Added a view switching method to app.py for getting a Clock object.
```
mwargers: By any chance based on some Moztrap testcase?

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_laps.py
```diff
@@ -0,0 +1,44 @@
...
+
+        stopwatch_view.pause()
+        recorded_time = stopwatch_view.current_time
+
+        stopwatch_view.resume()
+        self.assertNotEqual(stopwatch_view.current_time, recorded_time)
+
+        stopwatch_view.pause()
```
mwargers: Perhaps another check here for: self.assertEqual(len(stopwatch_view.laps), 2) ?
To make sure that the laps haven't been removed by tapping the resume/pause button?

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_laps.py
silne30: No this wasn't. I just though it would be good to have a shorter test of the stopwatch functionality.

tests/python/gaia-ui-tests/gaiatest/apps/clock/app.py
silne30: Fixed. Also renamed it since that was what confused me.

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/alarm.py
silne30: It's used on line 19. view.location['x']

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/new_alarm.py
silne30: Done.

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/new_alarm.py
silne30: That did not work. Changed it back.

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
silne30: Fixed.

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
silne30: Fixed.

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
silne30: fixed.

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
silne30: Fixed

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
silne30: Fixed

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
silne30: Fixed

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
silne30: Fixed

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_laps.py
silne30: Done.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_laps.py
silne30: Done.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_laps.py
silne30: I did the inverse and made sure that the length equaled 0. That way if it incorrectly only removes one item, it doesn't pass unexpectedly. Look at line 41.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_set_alarm.py
silne30: This was an existing test. Not sure what moztrap test case this was referencing. However, I made changes to the structure of the app so I updated the test.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b38645af5ff9074df8241f6fd83790ae00c731ee

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/new_alarm.py
JohanLorenzo: You're right, that should be enough:
``` python
class EditAlarm(NewAlarm):
    pass
```

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
```diff
@@ -0,0 +1,66 @@
...
+    def current_time(self):
+        current_time = Wait(self.marionette).until(expected.element_present(*self._stopwatch_time_locator))
+        return current_time.text
+
+    @property
+    def lap_items(self):
+        return [self.LapItem(self.marionette, lap) for lap in self.marionette.find_elements(*self._all_laps_locator)]
+
+    def tap_start(self):
+        Wait(self.marionette).until(expected.element_present(*self._stopwatch_start_locator)).tap()
```
JohanLorenzo: What's the advantage of this wait compared to: `self.marionette.find_element(*locator).tap()`?

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
```diff
@@ -0,0 +1,66 @@
...
+    def tap_reset(self):
+        Wait(self.marionette).until(expected.element_present(*self._stopwatch_reset_locator)).tap()
+
+    def tap_resume(self):
+        Wait(self.marionette).until(expected.element_present(*self._stopwatch_resume_locator)).tap()
+
+    def tap_lap(self):
+        Wait(self.marionette).until(expected.element_present(*self._stopwatch_lap_locator)).tap()
+
+    def verify_pause(self):
```
JohanLorenzo: To me, verifications should be done in the test. I'd put that function in `test_clock_run_stopwatch_then_reset.py`

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/alarm.py
JohanLorenzo: I don't think `Clock.__init__()` has special. The init method in Clock is the same as [Base](https://github.com/mozilla-b2g/gaia/blob/eb1193cd6d71e260744754c6fda6f67f2f90d90f/tests/python/gaia-ui-tests/gaiatest/apps/clock/app.py#L21). `PageRegion` also uses that same `__init__()`

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
```diff
@@ -0,0 +1,66 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from marionette_driver import expected, By, Wait
+from gaiatest.apps.clock.app import Clock
+from gaiatest.apps.base import PageRegion
+import time
+
+class StopWatch(Clock):
+    _stopwatch_view_locator = (By.CSS_SELECTOR,'#stopwatch-panel')
```
JohanLorenzo: By.ID

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/new_alarm.py
```diff
@@ -0,0 +1,78 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import time
+
+from marionette_driver import expected, By, Wait
+
+from gaiatest.apps.clock.app import Clock
+
+
+class NewAlarm(Clock):
```
JohanLorenzo: Same comment about the `PageRegion`.

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
```diff
@@ -0,0 +1,66 @@
+ # This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from marionette_driver import expected, By, Wait
+from gaiatest.apps.clock.app import Clock
+from gaiatest.apps.base import PageRegion
+import time
+
+class StopWatch(Clock):
```
JohanLorenzo: Same comment about the `PageRegion`.

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/timer.py
```diff
@@ -0,0 +1,12 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from marionette_driver import expected, By, Wait
+
+from gaiatest.apps.clock.app import Clock
+from gaiatest.apps.base import PageRegion
+
+
+class Timer(Clock):
```
JohanLorenzo: Same comment about the `PageRegion`.

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/timer.py
```diff
@@ -0,0 +1,12 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from marionette_driver import expected, By, Wait
```
JohanLorenzo: Unused imports

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_set_alarm.py
JohanLorenzo:
```
# Added a view switching method to app.py for getting a Clock object.
```
I'm not sure this comment has a lot of value. This looks like duplication with the Page Class. If somebody needs to know why the test changed that way. `git log tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_set_alarm.py` can be used, and it'll point to that bug and this PR.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_create_new_alarm.py
```diff
@@ -22,7 +22,7 @@ def test_clock_create_new_alarm(self):
         https://moztrap.mozilla.org/manage/case/1772/
         https://moztrap.mozilla.org/manage/case/1775/
         """
-
+        alarm_view = self.clock.switch_view("alarm")
         # Set the time on the device
```
JohanLorenzo: This comment lies, doesn't it?

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_create_new_alarm.py
```diff
@@ -34,19 +34,19 @@ def test_clock_create_new_alarm(self):
         alarm_label_text = "test4321"
 
         # get the number of alarms set, before adding the new alarm
```
JohanLorenzo: We can apply the [boy scout rule](http://www.informit.com/articles/article.aspx?p=1235624&seqNum=6) here. This comment has no value, as getting the number of things before doing an action is a common testing pattern.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_create_new_alarm.py
```diff
@@ -34,19 +34,19 @@ def test_clock_create_new_alarm(self):
...
         self.data_layer.set_time(_seconds_since_epoch)
 
         # create a new alarm with the default values that are available
-        new_alarm = self.clock.tap_new_alarm()
+        new_alarm = alarm_view.tap_new_alarm()
 
         # Ensure label has the default placeholder and text
         self.assertEquals(new_alarm.alarm_label_placeholder, 'Alarm name')
 
         # set label
```
JohanLorenzo: Boy scout rule

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_set_alarm.py
```diff
@@ -14,8 +14,10 @@ def setUp(self):
...
+        alarm = self.clock.switch_view('alarm')
 
-        new_alarm = self.clock.tap_new_alarm()
+        new_alarm = alarm.tap_new_alarm()
 
         # Ensure repeat has the default value
```
JohanLorenzo: Same thing here about the boy scout rule.

tests/python/gaia-ui-tests/gaiatest/apps/clock/app.py
```diff
@@ -9,13 +9,10 @@
...
 
 class Clock(Base):
-
     name = 'Clock'
-
-    _alarm_view_locator = (By.ID, 'edit-alarm')
     _alarm_create_new_locator = (By.ID, 'alarm-new')
-    _all_alarms_locator = (By.CSS_SELECTOR, '#alarms li')
     _visible_clock_locator = (By.CSS_SELECTOR, '#clock-view .visible')
+    _clock_views = {"stopwatch":"stopwatch-tab", "alarm":"alarm-tab", "timer":"timer-tab"}
```
mwargers: Add a space between : and "?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=82595baaa484c77a1a2444b49b3f1c80b87cdecd

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=14673e75a868ca78e022e97a7530892d59fdb140

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
silne30: Done.

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
silne30: Done.

tests/python/gaia-ui-tests/gaiatest/apps/clock/app.py
```diff
@@ -27,9 +24,20 @@ def launch(self):
...
+            from gaiatest.apps.clock.regions.timer import Timer
+            return Timer(self.marionette)
+        elif view_name == 'alarm':
+            from gaiatest.apps.clock.regions.alarm import Alarm
+            return Alarm(self.marionette)
+        else:
+            raise AttributeError('{} is not a view that you can switch to'.format(view_name))
 
     @property
     def banner_notification(self):
```
JohanLorenzo: After taking a closer look at the app, I think the next 2 functions should belong to the Alarm view. What do you think @silne30 ?

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
silne30: Going to keep this functionality until we figure out what the standard is.

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/new_alarm.py
silne30: Done.

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
silne30: Done.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/manifest.ini
```diff
@@ -10,3 +10,9 @@ smoketest = true
 stable = false
 
 [test_clock_set_alarm.py]
+
+[test_clock_run_stopwatch_laps.py]
+skip-if = device == "desktop"
```
mwargers: This test should be able to run on desktop, the fact that it fails there makes me think there are issues with the test.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_laps.py
```diff
@@ -0,0 +1,43 @@
...
+        self.clock.launch()
+
+    def test_click_run_stopwatch_laps(self):
+
+        stopwatch_view = self.clock.switch_view("stopwatch")
+
+        self.assertEqual(stopwatch_view.current_time, '00:00.00')
+
+        stopwatch_view.tap_start()
+
```
mwargers: I would add a small import time time.sleep(0.1) here. That makes the following assert pass on b2g desktop then.
On the Flame device this wasn't an issue, apparently, because everything is slower there.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_then_reset.py
```diff
@@ -0,0 +1,33 @@
...
+        self.clock.launch()
+
+    def test_click_run_stopwatch_then_reset(self):
+
+        stopwatch_view = self.clock.switch_view("stopwatch")
+
+        self.assertEqual(stopwatch_view.current_time, '00:00.00')
+
+        stopwatch_view.tap_start()
+
```
mwargers: I would add a small import time time.sleep(0.2) here. That makes the following assert pass on b2g desktop then.
On the Flame device this wasn't an issue, apparently, because everything is slower there.

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/timer.py
silne30: Done

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/timer.py
silne30: Done.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_create_new_alarm.py
silne30: Moved to line 39

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_create_new_alarm.py
silne30: Deleted

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_then_reset.py
```diff
@@ -0,0 +1,33 @@
...
+        self.assertNotEqual(stopwatch_view.current_time, '00:00.00')
+
+        stopwatch_view.tap_pause()
+
+        self.assertNotEqual(stopwatch_view.current_time, '00:00.00')
+        self.assertTrue(stopwatch_view.verify_laps_not_same())
```
mwargers: We tapped pause here before, why do you expect the times to not be the same here?
Why do we need a separate verify_laps_not_same method here? You only used it once. Or are we planning on using it more often?

tests/python/gaia-ui-tests/gaiatest/apps/clock/app.py
silne30: Done.

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
```diff
@@ -0,0 +1,66 @@
...
+        Wait(self.marionette).until(expected.element_present(*self._stopwatch_reset_locator)).tap()
+
+    def tap_resume(self):
+        Wait(self.marionette).until(expected.element_present(*self._stopwatch_resume_locator)).tap()
+
+    def tap_lap(self):
+        Wait(self.marionette).until(expected.element_present(*self._stopwatch_lap_locator)).tap()
+
+    def verify_laps_not_same(self):
+        paused_time = Wait(self.marionette).until(expected.element_present(*self._stopwatch_time_locator)).text
```
mwargers: You could use self.current_time here, no?

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
```diff
@@ -0,0 +1,66 @@
...
+        time.sleep(2)
+        second_paused_time = Wait(self.marionette).until(expected.element_present(*self._stopwatch_time_locator)).text
```
mwargers: Same here, you could use self.current_time here, no?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6468e8907163e6e555d002030e1ceef3780c94c0

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_then_reset.py
mwargers: I tried something like this:
        time1 = stopwatch_view.current_time
        time.sleep(2)
        self.assertEqual(time1, stopwatch_view.current_time, "both times not the same")

And that works on desktop and device. I don't think this would warrant a new method.

tests/python/gaia-ui-tests/gaiatest/apps/clock/app.py
```diff
@@ -23,13 +19,21 @@ def launch(self):
...
                 *self._visible_clock_locator))))
-        Wait(self.marionette).until(expected.element_displayed(
-            Wait(self.marionette).until(expected.element_present(
-                *self._alarm_create_new_locator))))
 
-    @property
-    def alarms(self):
-        return [self.Alarm(self.marionette, alarm) for alarm in self.marionette.find_elements(*self._all_alarms_locator)]
+    def switch_view(self, view_name):
+        Wait(self.marionette).until(
```
mwargers: I think you can just as well use find_element here, see info: http://mxr.mozilla.org/mozilla-central/source/testing/marionette/driver/marionette_driver/marionette.py#1574
"
If no element is immediately found, the attempt to locate an element will be repeated for up to the amount of time set by set_search_timeout()
"

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/alarm.py
```diff
@@ -2,77 +2,66 @@
...
 
-class NewAlarm(Clock):
-
-    _alarm_name_locator = (By.ID, 'alarm-name')
-    _repeat_menu_locator = (By.ID, 'repeat-menu')
-    _sound_menu_locator = (By.ID, 'sound-menu')
-    _snooze_menu_locator = (By.ID, 'snooze-menu')
-    _done_locator = (By.ID, 'alarm-done')
-    _time_button_locator = (By.XPATH, "//li[input[@id='time-select']]")
+class Alarm(PageRegion):
```
mwargers: I think you need to add a blank line here.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_create_new_alarm.py
```diff
@@ -75,17 +74,17 @@ def test_clock_create_new_alarm(self):
...
-        self.assertFalse(self.clock.alarms[0].is_alarm_active, 'user should be able to turn on the alarm.')
+        alarm_view.alarm_items[0].tap_checkbox()
+        alarm_view.alarm_items[0].wait_for_checkbox_to_change_state(False)
+        self.assertFalse(alarm_view.alarm_items[0].is_alarm_active, 'user should be able to turn on the alarm.')
 
         # turn on the alarm
-        self.clock.alarms[0].tap_checkbox()
-        self.clock.dismiss_banner()
-        self.assertTrue(self.clock.alarms[0].is_alarm_active, 'user should be able to turn off the alarm.')
+        alarm_view.alarm_items[0].tap_checkbox()
```
mwargers: Shouldn't there be a alarm_view.alarm_items[0].wait_for_checkbox_to_change_state(True) line here?

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/alarm.py
```diff
@@ -2,77 +2,66 @@
...
+        return NewAlarm(self.marionette)
+
+    class AlarmItem(PageRegion):
+
+        _label_locator = (By.CSS_SELECTOR, '.label')
+        _check_box_locator = (By.CSS_SELECTOR, '.input-enable')
+        _enable_button_locator = (By.CSS_SELECTOR, '.alarmList.alarmEnable')
+        _time_locator = (By.CSS_SELECTOR, '.time')
+
+        def time(self):
```
mwargers: This should become a @property.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_create_new_alarm.py
```diff
@@ -54,15 +53,15 @@ def test_clock_create_new_alarm(self):
...
                         'Alarms count did not increment')
 
         # verify the label of alarm
-        alarms = self.clock.alarms
+        alarms = alarm_view.alarm_items
         self.assertEqual(len(alarms), 1)
         self.assertEqual(alarms[0].label, alarm_label_text)
 
-        alarm_time = self.clock.alarms[0].time()
+        alarm_time = alarm_view.alarm_items[0].time()
```
mwargers: If you make time a property, you can get rid of the ()

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/new_alarm.py
silne30: Done

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
```diff
@@ -0,0 +1,66 @@
...
+        return paused_time != second_paused_time
+
+    class LapItem(PageRegion):
+        _lap_name = (By.CSS_SELECTOR, '.lap-name')
+        _lap_time = (By.CSS_SELECTOR, '.lap-duration')
+
+        @property
+        def name(self):
```
mwargers: This one doesn't seem to be used, but I guess it will be in future tests.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/manifest.ini
silne30: This line was added, not because it failed, but because it's a dogfood test and we don't run those on device. I will remove the statement though.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_laps.py
silne30: Added.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_then_reset.py
silne30: I put the functionality in the test instead of the app object. You raised a good point. It would only be used once.

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
silne30: Fixed

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
silne30: Fixed

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d49eff227e21433823e3f95c313bc4768c170557

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
mwargers: I asked this question in https://mail.mozilla.org/private/gaia-ui-automation/2015-August/001848.html
I would prefer to use find_element, I find it easier to read.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_laps.py
```diff
@@ -0,0 +1,45 @@
...
+        GaiaTestCase.setUp(self)
+
+        self.clock = Clock(self.marionette)
+        self.clock.launch()
+
+    def test_click_run_stopwatch_laps(self):
+
+        stopwatch_view = self.clock.switch_view("stopwatch")
+
+        time.sleep(0.2)
```
mwargers: This line needs to be after line 24, stopwatch_view.tap_start(), otherwise it still fails on b2g desktop.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_then_reset.py
```diff
@@ -0,0 +1,42 @@
...
+        stopwatch_view = self.clock.switch_view("stopwatch")
+
+        time.sleep(0.2)
+        self.assertEqual(stopwatch_view.current_time, '00:00.00')
+
+        stopwatch_view.tap_start()
+
+        self.assertNotEqual(stopwatch_view.current_time, '00:00.00')
+
+        first_time = stopwatch_view.current_time
```
mwargers: I think we want this after line 30, stopwatch_view.tap_pause(), we want to check if the pause button works correctly, so that after this time.sleep of 2 seconds, we still have the same current_time on the stopwatch.
We already checked that the stopwatch was working after the tap_start, so we don't need another check for that.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_then_reset.py
```diff
@@ -0,0 +1,42 @@
...
+
+        stopwatch_view.tap_pause()
+
+        self.assertNotEqual(stopwatch_view.current_time, '00:00.00')
+
+        time.sleep(2)
+
+        second_time = stopwatch_view.current_time
+
+        self.assertNotEqual(first_time, second_time)
```
mwargers: This should then become:
self.assertEqual(first_time, stopwatch_view.current_time)

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/new_alarm.py
JohanLorenzo: I meant `NewAlarm` inheriting from `PageRegion` instead of `Clock`

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/alarm.py
silne30: Thanks. I missed that.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=259b997ce6664c1aa00b90a6c380c7ef4bd8d169

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
mwargers: What I understood from Dave Hunt's answers is that you use find_element if you are sure the element is there from the start and Wait element_present, when it's not immediately there and then you also check that the element is displayed, often.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_then_reset.py
```diff
@@ -0,0 +1,41 @@
...
+
+        self.assertNotEqual(stopwatch_view.current_time, '00:00.00')
+
+        time.sleep(0.2)
+
+        self.assertNotEqual(first_time, stopwatch_view.current_time)
+
+        stopwatch_view.tap_reset()
+
+        time.sleep(0.2)
```
mwargers: This time.sleep(0.2) doesn't seem necessary.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_then_reset.py
```diff
@@ -0,0 +1,41 @@
...
+        first_time = stopwatch_view.current_time
+
+        stopwatch_view.tap_pause()
+
+        self.assertNotEqual(stopwatch_view.current_time, '00:00.00')
+
+        time.sleep(0.2)
+
+        self.assertNotEqual(first_time, stopwatch_view.current_time)
+
```
mwargers: I'm sorry, I'm not making myself clear here, I guess.

Instead of this:
first_time = stopwatch_view.current_time
stopwatch_view.tap_pause()
self.assertNotEqual(stopwatch_view.current_time, '00:00.00')
time.sleep(0.2)
self.assertNotEqual(first_time, stopwatch_view.current_time)

I would like to see something like this:
stopwatch_view.tap_pause()
first_time = stopwatch_view.current_time
time.sleep(0.2)
self.assertEqual(first_time, stopwatch_view.current_time)

That would test that the pause button actually works, namely that the stopwatch time actually stops.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/manifest.ini
mwargers: You mean on desktop, I guess. Anyway, the statement is still there, I noticed.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_create_new_alarm.py
mwargers: ?

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_then_reset.py
```diff
@@ -0,0 +1,41 @@
...
+    def setUp(self):
+        GaiaTestCase.setUp(self)
+
+        self.clock = Clock(self.marionette)
+        self.clock.launch()
+
+    def test_click_run_stopwatch_then_reset(self):
+
+        stopwatch_view = self.clock.switch_view("stopwatch")
+
```
JohanLorenzo: Nit: We can remove some blank lines here.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_then_reset.py
JohanLorenzo: +1

tests/python/gaia-ui-tests/gaiatest/apps/clock/regions/stopwatch.py
silne30: It will. Leaving that functionality.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_create_new_alarm.py
silne30: Fixed.

tests/python/gaia-ui-tests/gaiatest/tests/functional/clock/test_clock_run_stopwatch_then_reset.py
silne30: Removed.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6011421228007c03e8ecdd4e029b595ed1ea1188

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9487d0bc9e2a642ad92c9331738e53a0f7f7a722

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d5c2ac08d9cb6e8d8677022bb7c963623e163adf

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3331b7b890a0f9f6bd81cd7d77f670bf9380243d

PR #31338, opened 2015-08-13
Bug 1194042 - Implement a global cut-or-copied state for text selection dialog across apps

aethanyc:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=01fe5984de61e8f46f04f630d0e2d99ce4f7b3f6

timdream:
This patch is pretty sad because we are relying on global hidden state, linked between instances. The only way to avoid this would be to create some kind of controller/manager and have instances explicitly link to it, but that's an overkill so I agree we should not do that. However, we could make our global state more transparent by avoiding getter/setter and give it an explicit instance & naming. I will comment on the lines to suggest how to do that.

apps/system/js/app_text_selection_dialog.js
```diff
@@ -3,10 +3,13 @@
...
   var DEBUG = false;
   var _id = 0;
-  /**
-   * Text Selection Dialog of the AppWindow
-   */
 
+  // Timer id obtained in _launchCutOrCopiedTimer() shared by all apps. If cut
+  // command or copy command has been performed within CUT_OR_COPIED_TIMEOUT_MS,
+  // it will be non-null.
+  var _cutOrCopiedTimeoutId = null;
```
timdream: In this file, let create a `AppTextSelectionDialogGlobalStates` constructor, and a variable `globalStates` in this closure to keep the instance.

apps/system/js/app_text_selection_dialog.js
```diff
@@ -16,13 +19,20 @@
...
     this._injected = false;
-    this._hasCutOrCopied = false;
     this._isCommandSendable = false;
     this._transitionState = 'closed';
     this.textualmenuDetail = null;
     this.bindOnObservePrefChanged = this.onObservePrefChanged.bind(this);
     SettingsListener.observe('copypaste.enabled', true,
                              this.bindOnObservePrefChanged);
+
+    Object.defineProperty(this, '_hasCutOrCopied', {
```
timdream: Here, let's do 

```
this.globalStates = globalStates || new AppTextSelectionDialogGlobalStates()
```

to allow all `AppTextSelectionDialog` getting the same GlobalStates instances.

apps/system/js/app_text_selection_dialog.js
```diff
@@ -16,13 +19,20 @@
...
+      get: function() { return !!_cutOrCopiedTimeoutId; },
```
timdream: The one and only state the GlobalStates instance is `hasCutOrCopied`, which can be found by looking at the timeout ID. Therefore, timeoutId should be kept in the GlobalStates instance.

apps/system/js/app_text_selection_dialog.js
```diff
@@ -355,12 +362,21 @@
...
-      this._resetCutOrCopiedTimeout = window.setTimeout(function() {
-        this._hasCutOrCopied = false;
-      }.bind(this), this.RESET_CUT_OR_PASTE_TIMEOUT);
+  AppTextSelectionDialog.prototype._cancelCutOrCopiedTimer =
+    function tsd_cancelCutOrCopiedTimer() {
+      window.clearTimeout(_cutOrCopiedTimeoutId);
+      _cutOrCopiedTimeoutId = null;
+    };
+
+  AppTextSelectionDialog.prototype._launchCutOrCopiedTimer =
```
timdream: The timer functions here can be then moved to the GlobalStates instance, since their purpose is only about managing the timeout ID (and the `hasCutOrCopied` state)

timdream: GlobalStates instance should also implement a `resetGlobalStates` method for unit tests to reset the state.

apps/system/test/unit/app_text_selection_dialog_test.js
```diff
@@ -234,13 +231,19 @@ suite('system/AppTextSelectionDialog', function() {
     assert.isTrue(stubChangeTransitionState.calledWith('opened'));
   });
 
-  test('_resetCutOrCopiedTimer', function() {
+  test('_launchCutOrCopiedTimer', function() {
```
timdream: These method-level unit tests can be moved accordingly.

We would need another test to ensure two instances are getting the same globalStates instance.

timdream:
https://github.com/aethanyc/gaia/blob/1194042/apps/system/test/unit/app_text_selection_dialog_test.js#L36-L37 I do not know if text agent actually reloads the file in every test. If so we will not be getting the same closure in every test. Need to watch out if there is a problem rewriting the test.

aethanyc:
The unit test agent will reload app_text_selection_dialog_test.js. That is, setup() and teardown() will be called once for every test.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b5a6d372e79a3282b806b02dfe4d142d9ab89439

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=10eb9deac4589e0d5a8e3412bd8a5016246d2d3d

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=518736ab013eb39c0f6a445eaf99d14cebf99d32

PR #31340, opened 2015-08-13
Bug 1183058 - [Contacts]Enable "Order by last name", and then kill th…

arcturus:
…e contact app or reboot device without tapping "Done", and "Order by last name" will not work when it is enable unless you kill contact and relaunch again. r=borjasalguero


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5099c397f31026b51c0b23d430fb4bac26364eac

PR #31355, opened 2015-08-14
Bug 1191471 - Fix failure in test_keyboard.py

mwargers:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=fdfcf56cc98ccfae40a91ac5897ef154e5c58213

tests/python/gaia-ui-tests/gaiatest/apps/contacts/regions/contact_form.py
```diff
@@ -114,8 +117,13 @@ def type_comment(self, value):
         element.send_keys(value)
 
     def tap_comment(self):
-        self.marionette.find_element(*self._add_new_note_locator).tap()
+        element = self.marionette.find_element(*self._add_new_note_locator)
+        element.tap()
+        self.marionette.execute_script(
+            'arguments[0].scrollIntoView(true);', [element])
```
JohanLorenzo: This call is great! [Element.scrollIntoView()](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) could be useful for the Graphics test in Settings. What do you think, @npark-mozilla ?

If so, how about moving that method into `Base.py`? Something like:
``` python
def scroll_to_element(element, should_be_at_top=True):
    self.marionette.execute_script(
        'arguments[0].scrollIntoView(arguments[1]);', [element, should_be_at_top])
```

npark-mozilla: This looks good, but I thought this method is already there somewhere in gaiatest already, because I remember seeing something like this.  If not, we should definitely put it into the base class definition.

mwargers: Afaik, there isn't http://mxr.mozilla.org/gaia/search?string=scrollintoview&find=.py
But I'll file a bug about this.

mwargers: This is now https://bugzilla.mozilla.org/show_bug.cgi?id=1195341

PR #31375, opened 2015-08-17
Bug 994567 - [Messages] The contact name is not looked up for exact m…

anubhav7495:
…atches in the recipient panel r=julienw

### DIFF
--- a/tests/python/gaia-ui-tests/gaiatest/gaia_test.py
+++ b/tests/python/gaia-ui-tests/gaiatest/gaia_test.py
@@ -1163,7 +1163,11 @@ class PasscodeTestCase(GaiaTestCase):
         SET_DIGEST_ALGORITHM = 'lockscreen.passcode-lock.digest.algorithm'
 
         settings = {}
+        # The code for setting the passcode uses ArrayBuffers.
         # ArrayBuffers are represented as objects keys from 0 to n-1.
+        # The settings DB does not support this and sees an array buffer of [3,6,9] objects
+        # of the format {0: 3, 1: 6, 2: 9} (hence objects with keys from 0 to n-1)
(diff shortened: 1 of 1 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c78eb612008d6be80387c85abd45ee7fd9ca1dd1

apps/sms/views/shared/js/contacts.js
```diff
@@ -246,13 +246,13 @@
 
     findExact: function contacts_findExact(filterValue) {
       return this.findBy({
-        filterBy: ['givenName', 'familyName'],
+        filterBy: ['tel', 'givenName', 'familyName'],
         filterOp: 'contains',
```
julienw: so, I checked closed, and for the phone number, we need the "match" operation.

Actually this is already implemented by `Contacts.findByPhoneNumber` and I don't think we can do better than doing 2 calls. And actually it's likely better to call `Contacts.findByAddress` to handle the email case as well.

So here is what I think could work. In https://github.com/mozilla-b2g/gaia/blob/09e233d1476e8745f192f13259d7b16662bb2e7e/apps/sms/views/conversation/js/conversation.js#L2618, replace this line by:
```javascript
  return Promise.all([
    Contacts.findExact(fValue),
    Contacts.findByAddress(fValue)
  ]).then(([ result1, result2 ]) => result1.concat(result2));
```
(thanks to your bug 1153888 we don't need to check that result1 or result2 are not null :) ).

Does it make sense ?

anu7495: Yup, it makes sense. i will try the code u suggested above and will get back to you with results.

PR #31400, opened 2015-08-18
Bug 1193392 unit test the new home screen

gmarty:


autolander:

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=cc6929109aea5519ed248f48de3411eafebc7056

apps/homescreen/test/unit/app_test.js
```diff
@@ -1,36 +1,564 @@
...
+    teardown(() => {
+      addAppIconStub.restore();
+    });
+
+    test('should only accept valid apps', () => {
+      app.addApp({});
+      assert.isFalse(addAppIconStub.called, 'Empty app objects');
+
+      app.addApp({ manifest: { role: 'system' } });
+      assert.isFalse(addAppIconStub.called, 'Excluded role app');
```
Cwiiis: I don't think this test will pass anymore, we have to add hidden role apps because the role can change in an update - instead, we add it but keep it hidden by setting display:none on the icon container.

apps/homescreen/test/unit/app_test.js
```diff
@@ -1,36 +1,564 @@
...
+          application: { manifest: {} }
+        }
+      }));
+      assert.isTrue(addAppStub.calledOnce);
+      addAppStub.restore();
+    });
+  });
+
+  suite('App#addIconContainer()', () => {
+    test('should return a HTML element with a order property', () => {
```
Cwiiis: nit: s/a order/an order/

apps/homescreen/test/unit/app_test.js
```diff
@@ -1,36 +1,564 @@
...
+
+    test('should call snapScrollPositionSub() on first child', done => {
+      var appendChildSpy = sinon.spy(app.icons, 'appendChild');
+      app.startupMetadata = [{ order: 1 }];
+      app.addIconContainer(0);
+
+      setTimeout(() => {
+        assert.isTrue(appendChildSpy.called);
+        var spyCall = appendChildSpy.getCall(0);
+        assert.equal(typeof spyCall.args[1], 'function');
```
Cwiiis: Should this also check the name of the function?

apps/homescreen/test/unit/app_test.js
```diff
@@ -1,36 +1,564 @@
...
+
+  suite('App#storeAppOrder()', () => {
+    setup(() => {
+      app.metadata.mSetup();
+    });
+
+    test('should persist apps in sorted order', () => {
+      app.icons.appendChild(getIcon('abc'));
+      app.icons.appendChild(getIcon('def'));
+      app.icons.appendChild(getIcon('gih'));
```
Cwiiis: nit: why gih instead of ghi?

apps/homescreen/test/unit/app_test.js
```diff
@@ -1,36 +1,564 @@
...
+        };
+      });
+
+      test('should snap to icon height when the screen is small', () => {
+        app.scrollable.clientHeight = 150;
+        app.snapScrollPosition(0);
+        assert.equal(app.scrollable.style.scrollSnapPointsY, 'repeat(100px)');
+      });
+
+      test('should snap to screen height if high enough', () => {
```
Cwiiis: This test isn't right, the code never snaps to screen height - it always snaps to as many icons as can fit whole in the screen (so if the clientHeight here was 390, it would be repeat(300px) and this test would fail).

apps/homescreen/test/unit/app_test.js
```diff
@@ -1,36 +1,564 @@
...
+        getElementsByClassName: () => [],
+        querySelector: () => {
+        },
+        open: () => {
+          done();
+        }
+      };
+
+      app.showActionDialog(dialog, null, []);
+      clock.tick(100);
```
Cwiiis: It might be good to alter app.js to export the const it uses, rather than reproducing the value here. I'm fine either way, your call.

apps/homescreen/test/unit/app_test.js
```diff
@@ -1,36 +1,564 @@
...
+    });
+  });
+
+  suite('App#handleEvent()', () => {
+    suite('contextmenu', () => {
+      test('should open a dialog', () => {
+        var showActionDialogStub = sinon.stub(app, 'showActionDialog');
+        app.handleEvent(new CustomEvent('contextmenu'));
+        assert.isTrue(showActionDialogStub.called);
+        //showActionDialogStub.restore();
```
Cwiiis: nit: should the stub setup be done in setup/teardown? Why is this commented out?

apps/homescreen/test/unit/app_test.js
```diff
@@ -1,36 +1,564 @@
...
+          scrollTop: 50
+        };
+        app.handleEvent(new CustomEvent('scroll'));
+        assert.isTrue(app.scrolled);
+        assert.equal(app.shadow.className, 'visible');
+
+        app.scrollable.scrollTop = 0;
+        app.handleEvent(new CustomEvent('scroll'));
+        assert.isFalse(app.scrolled);
+        assert.equal(app.shadow.className, '');
```
Cwiiis: It might be nicer to use assert.isTrue/False(classList.contains('scroll')) here? I don't imagine we'll be adding other classes to this, but just in case.

apps/homescreen/test/unit/datastore_test.js
```diff
@@ -0,0 +1,370 @@
...
+    var names = [];
+    var onUpgradeNeededEvent = version => {
+      return {
+        target: {
+          result: {
+            createObjectStore: (name, options) => {
+              names.push(name);
+              return {
+                createIndex: (name, keyPath, options) => {
+                }
```
Cwiiis: nit, no need for a new-line I think here, makes it a bit harder to read.

apps/homescreen/test/unit/datastore_test.js
```diff
@@ -0,0 +1,370 @@
...
+                            id: 'abc/',
+                            data: {}
+                          },
+                          continue: () => {
+                          }
+                        }
+                      }
+                    });
+                  }
+                };
```
Cwiiis: nit, these bracket cascades get very hard to read, I think it'd be worth compressing some of these things to one line. e.g.
cb({ target: {
  result: { value: {
    id: 'abc/',
    data: {}
  },
  continue: () => {}
} });

apps/homescreen/test/unit/datastore_test.js
```diff
@@ -0,0 +1,370 @@
...
+          });
+      });
+    });
+
+    suite('Datastore#getAll()', () => {
+      test('should retrieve data from object store', done => {
+        attachDBTransaction(bookmarks.storeName);
+        bookmarks.getAll()
+          .then(results => {
+            assert.equal(results.length, 1);
```
Cwiiis: It feels like for this test we should add more than 1 entry in the store to make sure that it returns all of them.

apps/homescreen/test/unit/datastore_test.js
```diff
@@ -0,0 +1,370 @@
...
+            assert.deepEqual(results[0], { id: 'abc/', data: {} });
+            done();
+          });
+      });
+    });
+  });
+
+  suite('Datastore#updateRevision()', () => {
+    test('should set the datastore revision ID in lastRevision', done => {
```
Cwiiis: I think it would be good to have a test that makes sure the revision also gets updated when it changes.

apps/homescreen/test/unit/metadata_test.js
```diff
@@ -0,0 +1,205 @@
...
+    var onUpgradeNeededEvent = version => {
+      return {
+        target: {
+          result: {
+            createObjectStore: (name, options) => {
+              names.push(name);
+              return {
+                createIndex: (name, keyPath, options) => {
+                }
+              };
```
Cwiiis: Same nit wrt the brackets here - this read read more easily as `return { createIndex: (name, key, options) => {} };`

apps/homescreen/test/unit/metadata_test.js
```diff
@@ -0,0 +1,205 @@
...
+
+                    cb({
+                      target: {
+                        result: {
+                          value: value,
+                          continue: () => {
+                          }
+                        }
+                      }
+                    });
```
Cwiiis: Same here with the brackets.

apps/homescreen/test/unit/metadata_test.js
```diff
@@ -0,0 +1,205 @@
...
+      test('should store apps with order', () => {
+        attachDBTransaction('order');
+        var id = 'abc/', order = 0;
+        setMetadata(id, order, undefined);
+        assert.equal(Object.keys(objectStores).length, 1);
+        assert.equal(objectStores.order.id, id);
+        assert.equal(objectStores.order.order, order);
+      });
+
+      test('should store apps with icon', () => {
```
Cwiiis: This and the test above should probably also check (or there should be a separate test for) that the order and icon remain unchanged when they aren't set, respectively.

apps/homescreen/test/unit/app_test.js
gmarty: Ah I guess at my age that'd be a good thing to learn the alphabet in order ;-P

apps/homescreen/test/unit/datastore_test.js
gmarty: Yeah, sorry, blame the autoformat of my IDE. I'll fix that manually.

apps/homescreen/test/unit/app_test.js
gmarty: This test check that the callback parameter of `this.icons.appendChild` is set as a function. But it is an anonymous function. `this.snapScrollPosition(0)` is not called directly. Is that sufficient?

apps/homescreen/test/unit/app_test.js
gmarty: I'm using a very high value (`clock.tick(10000);`) to avoid exporting the real one here.

autolander:

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=eb31b12676d4feabc66551a8ffe4e6c918edc0d2

apps/homescreen/test/unit/datastore_test.js
gmarty: I added a test just below.

apps/homescreen/test/unit/metadata_test.js
gmarty: Fixed. See below.

apps/homescreen/test/unit/datastore_test.js
Cwiiis: nit: These braces are still a bit crazy, but I can live with it if it's a pain to reformat.

apps/homescreen/test/unit/datastore_test.js
```diff
@@ -0,0 +1,383 @@
...
+        done();
+      });
+    });
+
+    test('should update the datastore revision ID in lastRevision', done => {
+      bookmarks.init().then(() => {
+        bookmarks.updateRevision();
+        assert.equal(bookmarks.lastRevision, bookmarks.datastore.revisionId);
+        bookmarks.datastore.revisionId = 'xxx-zzz';
+        bookmarks.init().then(() => {
```
Cwiiis: I think it'd be better to test this by emitting a change event, rather than calling init again.

apps/homescreen/test/unit/metadata_test.js
```diff
@@ -0,0 +1,245 @@
...
+                              target: {
+                                result: {
+                                  value: value,
+                                  continue: () => {}
+                                }
+                              }
+                            });
+                          }
+                        }
+                      }
```
Cwiiis: nit: These braces too

autolander:

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9facdb6619323719985f473a72094211e70cfd2a

autolander:

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6eb68739be315c87d6a46f53c1237b6466f7f43f

PR #31407, opened 2015-08-19
Bug 1196077 - [email] Refactor message_list into some smaller custom elements

jrburke:
There are three parts to this refactor. The unifying theme is instance based, smaller pieces. These parts were part of the general convoy effort, but landing them now will help reduce the overall convoy review hit and allows for some other front end refactors that do not depend on convoy to occur in the meantime.

Keeping the separate commits to help review the three pieces separately. The last commit is some fixes for the "model now instance-based" first commit. These fixes were found in testing after the message_list refactoring to smaller pieces. Since some of the changes are in these refactored pieces, it was easier to add them to the end of the commit list instead of reordering them to fit with the original commit. The main theme of the fix commit: make sure to pass the model instance when pushing a compose card.

Notes on the commits:
## model now instance-based with a defaultModel instance

The goal of this change is to change cards to receive their data/model as an instance via onArgs. model.js is now model_create, and it allows creating instances. There is a model_create defaultModel that is used as a bootstrap model, one that can be used for any entry point into the app, and for use when cards.pushStartCard() is needed from an activity entry point to go back to the message_list.

It is not very interesting on its own, just sets the stage for individual cards to receive their data via instance creation (onArgs in the cards case) instead of holding on to a global.
## HeaderCursor to instance

Builds on the conceptual change with passing model around to allow creating instances of HeaderCursor in each card. Now message_list and message_list_search will get their own HeaderCursor instances, with allow them separate messages slices, and avoids the weirdness where "searchMode" needed to be checked.

HeaderCursor still tracks "searchMode" for now, but it is just informational now, not used for checks (now that the _Break message_list into smaller pieces_ commit is part of this pull request).

With the convoy changes, HeaderCursor will be replaced with ListCursor instances, and with that, tracking searchMode at all goes completely away. I expect ListCursor to be something that could be considered for the (B) concept talked about in [bug 937959 comment 6](https://bugzilla.mozilla.org/show_bug.cgi?id=937959#c6).

There is a small change in mail_app.js onVisibilityChange function: it was getting called on a notification entry point, where the visibility change call happened before the notification startup completed. onVisibilityChange is there specifically for the sync case, so the check in that function was scoped to just that startup case.

I will be testing v2.2 to be sure that is not an issue there, but I suspect the visibilitychange issue is only an issue for master. It is not horrible if it happens, we just push a message_list card then immediately push a message_reader card, but felt it was best to avoid work if we can.
## Break message_list into smaller pieces

This is the most interesting change. There are a lot of files touched, but they are mostly moving the cards to mix in base_card instead of base. base is now used as the basic custom element baseline for all the email custom elements, and now there are some email custom elements that are base-derived custom elements, where only full fledged card modules mix in base_card.

**message_list** has been refactored to use some email custom elements, with the related HTML and JS for those elements moved out of message_list. Looking at the .html files for these modules gives the best overview of what moved where:
- **lst/edit_header**
- **lst/msg_vscroll**: Handles the VScroll instance and the specific message list logic around the basic VScroll.
- **lst/edit_toolbar**

The move of the HTML and logic were mostly mechanical copy/paste operations.

**message_list_search** is now its own fully realized card, and not a mixin hack of message_list. With that change the search specific HTML/logic has been moved out of message_list. message_list_search uses some of the new smaller email custom elements also used in message_list.

The **mixins/data-event.js** and **mixins/data-prop.js** mixins have fixes to only apply themselves if the element using those attributes is not inside another custom element. This was needed because the custom element approach used in email. Their interior DOM structure is visible for caching purposes and there is no hard concept of a private DOM. Using shadowRoot would give that privacy, but that would not work with the HTML cache that email uses, where the HTML should be complete on its own without needing any backing JS to be loaded for the first render.

One of the reasons to land this pull request is to then look at changing the approach to the HTML cache to allow shadowRoot-based elements. But the changes in data-event/data-prop allow the HTML cache to continue to work with email custom elements made up of other email custom elements.

There is a new mixin, **mixins/dom_evt.js**, and it just allows converting a DOM event inside the custom element to a custom event name that is emitted out of the custom element, and doing so declaratively in the HTML template.

**container_listen.js** was extracted from the old cards base.js bindContainerHandler, as it was a specialized function only needed in some scenarios, not all base_card derivatives.

**html_cache.js** was modified to make sure all email custom elements (which have a .email-ce CSS class on them courtesy of base.js) get data-cached set on them, so that when they are hydrated from the HTML cache, the hydrated HTML is used for the internals instead of creating a new internal DOM structure from the template (template.js does the data-cached checking).

This refactoring to smaller pieces is carried forward in the convoy branch further, where conv_list is introduced.
## model now instance-based - fixes

Just fixes for the first commit where I forgot to pass model in some cases, and there were changes on master that happened after the initial commit was done. Also removes the autocomplete.js hack that was on master that is no longer needed now that base.js does not assume a full card base.

---

Tested activity and notification entry points, and manually testing of features. The cards that are not message_list/message_list_search/message_reader/compose are mostly unaffected by the changes, except for the base->base_card mixin change.

Commits will be squashed to one commit if review is passed.

### DIFF
--- a/apps/email/js/cards/base_card.js
+++ b/apps/email/js/cards/base_card.js
@@ -1,0 +1,68 @@
+'use strict';
+define(function(require) {
+  var mozL10n = require('l10n!'),
+      base = require('./base');
+
+  // Set up the global time updates for all nodes.
+  (function() {
+    var formatter = new mozL10n.DateTimeFormat();
+    var updatePrettyDate = function updatePrettyDate() {
+      var labels = document.querySelectorAll('[data-time]');
+      var i = labels.length;
+      while (i--) {
+        labels[i].textContent = formatter.fromNow(
+          labels[i].dataset.time,
+          // the presence of the attribute is our indicator; not its value
+          'compactFormat' in labels[i].dataset);
+      }
+    };
+    var timer = setInterval(updatePrettyDate, 60 * 1000);
+
+    function updatePrettyDateOnEvent() {
+      clearInterval(timer);
+      updatePrettyDate();
+      timer = setInterval(updatePrettyDate, 60 * 1000);
+    }
+    // When user changes the language, update timestamps.
+    mozL10n.ready(updatePrettyDateOnEvent);
+
+    // On visibility change to not hidden, update timestamps
+    document.addEventListener('visibilitychange', function() {
+      if (document && !document.hidden) {
+        updatePrettyDateOnEvent();
+      } else {
+        // If not visible, clear the interval to be a battery good citizen.
+        clearInterval(timer);
+      }
+    });
+
+  })();
+
+  return function baseCard(templateMixins) {
+    // Set up the base mixin
+    return [
+      base(templateMixins),
+      {
+        createdCallback: function() {
+          // Set up extra classes and other node information that distinguishes
+          // as a card. Doing this here so that by the time the createdCallback
+          // provided by the card so that the DOM at that point can be used for
+          // HTML caching purposes.
+          if (this.extraClasses) {
+            this.classList.add.apply(this.classList,
+                                        this.extraClasses);
+          }
+
+          this.classList.add('card');
+        },
+
+        batchAddClass: function(searchClass, classToAdd) {
+          var nodes = this.getElementsByClassName(searchClass);
+          for (var i = 0; i < nodes.length; i++) {
+            nodes[i].classList.add(classToAdd);
+          }
+        }
+      }
+    ];
+  };
+});
--- a/apps/email/js/cards/lst/default_vscroll_data.js
+++ b/apps/email/js/cards/lst/default_vscroll_data.js
@@ -1,0 +1,33 @@
+// Default data used for the VScroll component, when data is not loaded yet for
+// display in the virtual scroll listing. The u2583 character is a solid bar
+// character, so a string of them gives a "redacted" look to a line. Just a
+// temporary placeholder until the real data comes in.
+define({
+  'isPlaceholderData': true,
+  'id': 'INVALID',
+  'author': {
+    'name': '\u2583\u2583\u2583\u2583\u2583\u2583\u2583\u2583',
+    'address': '',
+    'contactId': null
+  },
+  'to': [
+    {
+      'name': ' ',
+      'address': ' ',
+      'contactId': null
+    }
+  ],
+  'cc': null,
+  'bcc': null,
+  'date': '0',
+  'hasAttachments': false,
+  'snippet': '\u2583\u2583\u2583\u2583\u2583\u2583\u2583\u2583' +
+             '\u2583\u2583\u2583\u2583\u2583\u2583\u2583\u2583' +
+             '\u2583\u2583\u2583\u2583\u2583\u2583\u2583\u2583',
+  'isRead': true,
+  'isStarred': false,
+  'sendStatus': {},
+  'subject': '\u2583\u2583\u2583\u2583\u2583\u2583\u2583\u2583' +
+             '\u2583\u2583\u2583\u2583\u2583\u2583\u2583\u2583' +
+             '\u2583\u2583\u2583\u2583\u2583\u2583\u2583\u2583'
+});
--- a/apps/email/js/cards/lst/edit_controller.js
+++ b/apps/email/js/cards/lst/edit_controller.js
@@ -1,0 +1,190 @@
+'use strict';
+
+define(function(require, exports) {
+  var cards = require('cards'),
+      ConfirmDialog = require('confirm_dialog'),
+      deleteConfirmMsgNode = require('tmpl!../msg/delete_confirm.html'),
+      mozL10n = require('l10n!'),
+      toaster = require('toaster');
+
+  return {
+    createdCallback: function() {
+      this.editMode = false;
+      this.selectedMessages = null;
+      this.editModeEnabled = false;
+    },
+
+    setEditMode: function(editMode) {
+      // Do not bother if edit mode is not enabled yet.
+      if (!this.editModeEnabled) {
+        return;
+      }
+
+      this._setEditMode(editMode);
+    },
+
+    // This function is called from setEditMode() after ensuring that
+    // the backend is in a state where we can safely use edit mode.
+    _setEditMode: function(editMode) {
+      var i;
+
+      this.editMode = editMode;
+
+      if (editMode) {
+        this.classList.add('show-edit');
+
+        this.selectedMessages = [];
+        this.updateDomEditControls();
+      }
+      else {
+        this.classList.remove('show-edit');
+
+        this.selectedMessages = null;
+      }
+
+      // Reset checked mode for all message items.
+      var msgNodes = this.msgVScroll.querySelectorAll('.msg-header-item');
+      for (i = 0; i < msgNodes.length; i++) {
+        this.updateDomMessageChecked(msgNodes[i], false);
+      }
+
+      if (this.editModeChanged) {
+        this.editModeChanged(editMode);
+      }
+    },
+
+    // Event handler wired up in HTML template
+    setEditModeStart: function() {
+      this.setEditMode(true);
+    },
+
+    // Event handler wired up in HTML template
+    editHeaderClose: function() {
+      this.setEditMode(false);
+    },
+
+    /**
+     * Toggles the selection state of the node and therefore the stored
+     * `selectedMessages` state based on the previous state of the selection.
+     * The change in the total number of selections is then reflected in DOM
+     * updates.
+     * @param  {Element} msgNode
+     */
+    toggleSelection: function(msgNode) {
+      var header = msgNode.message;
+      var idx = this.selectedMessages.indexOf(header);
+      if (idx !== -1) {
+        this.selectedMessages.splice(idx, 1);
+      } else {
+        this.selectedMessages.push(header);
+      }
+      this.updateDomMessageChecked(msgNode, idx === -1);
+      this.updateDomEditControls();
+    },
+
+    /**
+     * Set or unset the select state on a node based on the edit mode.
+     */
+    updateDomSelectState: function(msgNode, message) {
+      if (this.editMode) {
+        this.updateDomMessageChecked(msgNode,
+          this.selectedMessages.indexOf(message) !== -1);
+      } else {
+        msgNode.removeAttribute('aria-selected');
+      }
+    },
+
+    /**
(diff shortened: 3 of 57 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=16a158effa9cdbbf65729419e38a35fd8cfb2356

apps/email/js/cards/lst/msg_click.js
```diff
@@ -0,0 +1,162 @@
...
+
+      var model = this.model,
+          headerCursor = this.headerCursor;
+
+      function pushMessageCard() {
+        cards.pushCard(
+          'message_reader', 'animate',
+          {
+            model: model,
+            headerCursor: headerCursor,
```
asutherland: available es2015 magic reminder / stylistic consensus question: we can just write "model, headerCursor, header" here (still on separate lines if desired). I've adopted this in GELAM, although I don't think I care.  I mainly just feel like a sucker for typing things twice when I could only type it once.

Having said that, it also simplifies things now that I'm starting to pass arguments through with destructuring.  Ex:
```js
function bob({ hat, cat, bat }) {
}
function mob() {
  let hat = 1, cat = 2, bat = 3;
  bob({ hat, cat, bat });
}
```

(And I've started to do that for the cases where the arg-list starts having more than 3 items or multiple booleans passed.)

apps/email/test/unit/header_cursor_test.js
```diff
@@ -18,10 +18,11 @@ suite('HeaderCursor', function() {
...
     testConfig({ done: done }, [
       'evt',
-      'header_cursor'
-    ], function(_evt, HeaderCursor) {
+      'header_cursor',
+      'model_create',
+    ], function(_evt, HeaderCursor, mc) {
       evt = _evt;
-      headerCursor = HeaderCursor.cursor;
+      headerCursor = new headerCursor(mc.defaultModel);
```
asutherland: Capitalization error after the new.

apps/email/js/cards/base_card.js
```diff
@@ -0,0 +1,65 @@
...
+        labels[i].textContent = formatter.fromNow(
+          labels[i].dataset.time,
+          // the presence of the attribute is our indicator; not its value
+          'compactFormat' in labels[i].dataset);
+      }
+    };
+    var timer = setInterval(updatePrettyDate, 60 * 1000);
+
+    function updatePrettyDateOnEvent() {
+      clearTimeout(timer);
```
asutherland: (I know this is just moved code, but): We should change this clearTimeout to clearInterval for consistency, even though nsGlobalWindow::ClearTimeout and nsGlobalWindow::ClearInterval call the same underlying helper and the WHATWG spec seems to imply that they must behave the same because there seems to only be a single list of active timers.

apps/email/js/cards/base_card.js
```diff
@@ -0,0 +1,65 @@
...
+      timer = setInterval(updatePrettyDate, 60 * 1000);
+    }
+    // When user changes the language, update timestamps.
+    mozL10n.ready(updatePrettyDateOnEvent);
+
+    // On visibility change to not hidden, update timestamps
+    document.addEventListener('visibilitychange', function() {
+      if (document && !document.hidden) {
+        updatePrettyDateOnEvent();
+      }
```
asutherland: (I know this is just moved code, but): Maybe we should call clearInterval in an else case here too so we can be extra good CPU citizens?

apps/email/js/cards/lst/default_vscroll_data.js
```diff
@@ -0,0 +1,31 @@
+// Default data used for the VScroll component, when data is not
+// loaded yet for display in the virtual scroll listing.
```
asutherland: enhancement request: maybe we could put in the comment an explanation of what an \u2583 is.  Or the actual literal character, depending on how much we were trying to avoid putting high characters in.

apps/email/js/cards/lst/edit_controller.js
```diff
@@ -0,0 +1,180 @@
...
+    },
+
+    // This function is called from setEditMode() after ensuring that
+    // the backend is in a state where we can safely use edit mode.
+    _setEditMode: function(editMode) {
+      var i;
+
+      this.editMode = editMode;
+
+      // XXX the manual DOM play here is now a bit overkill; we should very
```
asutherland: I think you can remove this XXX comment since you removed all of the manual manipulation of "collapsed" across all the buttons.  (kudos!)

apps/email/js/cards/lst/edit_controller.js
```diff
@@ -0,0 +1,180 @@
...
+
+    // Event handler wired up in HTML template
+    editHeaderClose: function() {
+      this.setEditMode(false);
+    },
+
+    /**
+     * Set or unset the select state based on the edit mode.
+     */
+    setSelectState: function(msgNode, message) {
```
asutherland: (I know this just just moved code, but) maybe we can rename this to updateDomSelectedState or reflectDomSelectedState or something like that since this is really about reflecting existing internal state.  Whereas "set" makes it seem like we're manipulating internally tracked state.

Ditto for setMessageChecked below, which it also happens to call.  Especially with the manipulation of the checked state, it makes it seem like we care about the DOM state, but we don't.  The true state is in `selectedMessages`.

apps/email/js/cards/lst/edit_controller.js
```diff
@@ -0,0 +1,180 @@
...
+
+      this.editMode = editMode;
+
+      // XXX the manual DOM play here is now a bit overkill; we should very
+      // probably switch top having the CSS do this for us or at least invest
+      // some time in cleanup.
+      if (editMode) {
+        this.classList.add('show-edit');
+
+        this.selectedMessages = [];
```
asutherland: Er, and note that despite my function name nitpicking in this area, I don't think we actually want to deal with the https://bugzilla.mozilla.org/show_bug.cgi?id=824001 issue yet since I think we do want the MailAPI to host a selection management abstraction that can avoid us having to manually deal with id's, and that can maybe help us with the statistics of how many messages in the set are starred/unread/etc. that we use to decide on UI button states.

(In fact, I'm tempted to implement it as basically a ViewSlice/ListView you can cram messages/conversations into and it keeps the messages in there updated.  This could be used in extensions for maintaining temporary stacks of messages to deal with for to-do lists/etc.)

apps/email/js/cards/lst/msg_click.js
```diff
@@ -0,0 +1,162 @@
...
+
+      var header = messageNode.message;
+
+      // Skip nodes that are default/placeholder ones.
+      if (header && header.isPlaceholderData) {
+        return;
+      }
+
+      if (this.editMode) {
+        var idx = this.selectedMessages.indexOf(header);
```
asutherland: I'm honestly not sure if it will make things appreciably cleaner, but I think it would be better to move the logic inside the conditional here up into the edit controller since that's where all these properties are defined and otherwise manipulated.

asutherland:
oh neat, I can comment on the specific commits and have it show up in here.  I've been doing my commits on the overall one because long ago those comments would have just lived of detached floating through the aether. Go github!

apps/email/js/cards/lst/edit_controller.js
jrburke: Removed in upcoming feedback 1 commit.

apps/email/js/cards/lst/edit_controller.js
jrburke: Agreed on saving 824001 for later, when we have the new conversation stuff available.

apps/email/js/cards/lst/msg_click.js
jrburke: Neat, applied in upcoming feedback 1 commit.

apps/email/test/unit/header_cursor_test.js
jrburke: Fixed in upcoming feedback 1 commit.

apps/email/js/cards/base_card.js
jrburke: Used clearInterval in upcoming feedback 1 commit.

apps/email/js/cards/base_card.js
jrburke: Added else that does clearInterval(timer) in upcoming feedback 1 commit.

apps/email/js/cards/lst/default_vscroll_data.js
jrburke: Added comment in upcoming feedback 1 commit:
```
// The u2583 character is a solid bar
// character, so a string of them gives a "redacted" look to a line. Just a
// temporary placeholder until the real data comes in.
```

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b59cf755ca749ddd22d3a7f3e41ac3dc674f056c

apps/email/js/cards/lst/edit_controller.js
jrburke: In feedback 2 commit, the name changes are:

* setSelectState -> updateDomSelectState
* setMessageChecked -> updateDomMessageChecked
* selectedMessagesUpdated -> updateDomSelectedMessages

jrburke: Actually, went with selectedMessagesUpdated -> updateDomEditControls since it was just about the edit controls and not the individual message selections in the DOM.

apps/email/js/cards/lst/msg_click.js
jrburke: Moved to edit_controller as toggleSelection:

```javascript
    /**
     * Toggles the selection state of the node and therefore the stored
     * `selectedMessages` state based on the previous state of the selection.
     * The change in the total number of selections is then reflected in DOM
     * updates.
     * @param  {Element} msgNode
     */
    toggleSelection: function(msgNode) {}
```

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1a0957c9e9db351ef1f560c604d1266d4921e762

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d4554042468cd6dcbbc1048dd042f4466c491e70

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e4ccafe5bcfe0e7c823e1189d5903451e0faa038

PR #31414, opened 2015-08-19
Bug 1191104 - [Bluetooth]turn default bluetooth_context.enabled to undefined to avoid briefly appear dialog, r=tim

gasolin:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=fff9ecc2fd5a17ad7ab4c9a3c4317abbfa1243fe

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f9cbcf93f528799805867cc8af542d71304ac465

apps/bluetooth/js/modules/transfer_manager.js
```diff
@@ -175,7 +175,8 @@ define(function(require) {
     _onBluetoothEnabledChanged:
     function bttm__onBluetoothEnabledChanged(enabled) {
       Debug('_onBluetoothEnabledChanged(): enabled = ' + enabled);
-      if (enabled) {
+      // make sure the value is defined
```
timdream: Simply `if (enabled === undefined) { return; }` here?

Do we want to run the `} else {` block if the value is `undefined`?

timdream: Maybe need a `console.error()` in that early return block since supposedly the settings should **never** "changed" to `undefined`?

gasolin: The origin intention is to align with other observe function call, but it's fine to just not running the "changed" callback at all initially. So I'll remove it.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8d8021d881fcded96027bd8203ee3a85e74f0952

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9ef62e305a0ad6117f8e794dca1dde3df12c8690

gasolin:
Since `enabled` property is only used in this case, we can just remove it. Will ask for review.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6ee95aa1c22baef3c722a444db530297ec084fe2

gasolin:
After revisit, the `enabled` property is still used internally in bluetooth_context, so will keep as our offline discussion result  `just not running the "changed" callback at all initially` to prevent any side effect

PR #31421, opened 2015-08-19
Bug 1195404 - Fix failure in test_settings_wifi.py

mwargers:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b3f32f8d05e51b29f41fa63cdcd3d934ad9f2b93

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e0a3fb18274742be45514539319a6d1899910209

tests/python/gaia-ui-tests/gaiatest/tests/functional/settings/test_settings_wifi.py
```diff
@@ -20,7 +20,9 @@ def test_connect_to_wifi_via_settings_app(self):
         settings.launch()
         wifi_settings = settings.open_wifi()
 
+        self.assertFalse(wifi_settings.is_wifi_enabled, "WiFi should be disabled")
```
JohanLorenzo: Nit: The error message doesn't add that much value to code. I'd delete it. Same thing for the following assertion.

mwargers: Sorry, I only saw this comment until I checked in.
Why you say this assert doesn't add much value? If wifi is disabled, but in the settings it shows as enabled, then this would catch it.

JohanLorenzo: No problem. I meant: If the test fails at that point, we'll see in the stacktrace: 
`self.assertFalse(wifi_settings.is_wifi_enabled, ...)` => the same information will be displayed. In other words, the comment saying "wifi should be disabled" doesn't add any extra information. Hence, that's a kind of duplication in our code base. That's why I think it's better if we remove them.

mwargers: Ah, I see. That's indeed not great. I haven't paid much attention to that.

PR #31426, opened 2015-08-19
Bug 1169010 - Reusing appWindows when clicking on pinned sites r=benfrancis

albertopq:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=44fda0f3252bce11824614a236488cd513d01350

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=415df07a3e9077eee983904384458406bcf6add6

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=67f2cef5b6ce63170c3cfa451f107844b6836b7b

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4eded56b4eadf73ec2fab7ef6bbb66db69eecd64

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f6172de0f2abb5b17c3cf0d78b26a736af7c745c

apps/system/js/app_window.js
```diff
@@ -223,6 +223,12 @@
     this.getTopMostWindow()._setActive(enable);
   };
 
+  AppWindow.prototype.inScope = function(scope) {
+    var url = new URL(this.url);
```
benfrancis: I think this.url needs to be this.config.url (don't ask me why, but .url doesn't get updated)

benfrancis: Also, I learned that new URL() should be surrounded by a try/catch because it can throw an exception.

apps/system/js/wrapper_factory.js
```diff
@@ -96,37 +96,48 @@
...
-
-        // If we already have a browser and we receive an open request,
-        // display it in the current browser frame.
-        var activeApp = Service.query('AppWindowManager.getActiveWindow');
-        if (activeApp && (activeApp.isBrowser() || activeApp.isSearch())) {
-          activeApp.navigate(url);
-          return;
-        }
+      switch (name) {
+        case 'pinned':
```
benfrancis: I think this string should be prefixed by a "_" as it's a special case like "_blank". I wonder whether "_samescope" might be a better name as that's really what it does? It will open in any window with the same scope. "pinned" is a bit specific to the UI of Firefox OS.

apps/system/js/app_chrome.js
```diff
@@ -830,7 +830,10 @@
           this.expand();
         }
         this.scrollable.scrollTop = 0;
-        this.pinned = false;
+        // TODO: Check if the new location is pinned
+        // Taking it from the initial value for now
+        var chromeConfig = this.app.config && this.app.config.chrome;
+        this.pinned = chromeConfig && chromeConfig.pinned;
```
benfrancis: I think we should probably leave this to bug 1168962 and just always expand browser chrome when the location changes for now. A window should not remain pinned for its entire lifetime.

apps/system/js/app_chrome.js
```diff
@@ -855,6 +858,9 @@
       this.setThemeColor('');
     }
     this.setSiteIcon();
+    if (this.pinned) {
+      this.pin();
```
benfrancis: Same as above, this is causing the browser chrome to expand and then collapse again on navigation. Let's leave this to bug 1168962.

apps/system/js/wrapper_factory.js
```diff
@@ -182,8 +191,10 @@
 
     generateBrowserConfig: function wf_generateBrowserConfig(features) {
       var config = {};
+      config.chrome = {};
       config.title = features.name;
       config.icon = features.icon || '';
+      config.chrome.pinned = features.pinned || false;
```
benfrancis: Is there a particular reason we have to set this? I'm not sure it's a good idea to set it if it never gets updated. A window can transform between pinned and unpinned state during its lifetime, we should track this state on the AppChrome object rather than set it in config for the lifetime of the window.

apps/system/js/wrapper_factory.js
albertopq: Sounds good to me. Actually, didn't add the '_' just because of that. It didn't sound too generic for using it as a keyword... '_samescope' sounds good. Thanks!

apps/system/js/wrapper_factory.js
albertopq: That's not entirely true. This config is the initial configuration for the chrome on a new window. We'll need (somehow) to specifiy to a new created window that the initial state is 'pinned'. Not sure if the features object is the best way, but couldn't think in another one for now. That doesn't mean that the reConfig method in the AppChrome cannot reset the pinned config in that chrome depending on the url that is loaded afterwards.

apps/system/js/app_chrome.js
albertopq: So then if I open a pinned site, and click on any link, it gets expanded? This was implemented in this bug, mainly to make it more testable, so I guess it makes sense to leave it to bug 1168962. I wouldn't mind to leave it and make the 'pinned' feature in bookmarks dependent on the pinning_the_web_enabled setting, though.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f2c219584f66076a79fd4bc73d1d5a0716a2e491

apps/system/js/app_window.js
albertopq: Good catch!

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a6a5ed33f01ebf078dd703073c0515dedeba2900

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=933a3760c582824e08a0d4167c58d3ccc87f058b

apps/system/js/app_chrome.js
```diff
@@ -777,6 +778,8 @@
 
       // Check if this is just a location-change to an anchor tag.
       var anchorChange = false;
+      var fisrstLocationChange = !this._currentURL;
```
benfrancis: typo?

apps/system/js/app_window.js
```diff
@@ -223,6 +223,12 @@
     this.getTopMostWindow()._setActive(enable);
   };
 
+  AppWindow.prototype.inScope = function(scope) {
+    var url = new URL(this.config.url);
+    var path = url.hostname + url.pathname;
+    return path.indexOf(scope) === 0;
```
benfrancis: Maybe this should only match browser windows as being in scope? Currently tapping a bookmark at m.facebook.com will re-use a window of the Facebook app, I'm not sure we want to go that far just yet, I think we need to treat apps separately.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=fee621d3f97f2ab80850dd82ae9a89a09fc360dc

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8713709e10b9915104dd597ddd74084f84d5447f

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9dba3c212b8251a97e618cdf88eb15c62571796d

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9dba3c212b8251a97e618cdf88eb15c62571796d

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=24c4944970a61d20e40a8a4d97c1981a35711477

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=24c4944970a61d20e40a8a4d97c1981a35711477

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f66a27f1bcacf33c27d96c97c987af6e63087eb6

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=efbffaf5e65387f0eb3a7ce528a46d8e7fc07204

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=efbffaf5e65387f0eb3a7ce528a46d8e7fc07204

PR #31439, opened 2015-08-20
Bug 1190791 - Add test that causes Connection lost issues in Marionette

mwargers:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4f4fa4e3d21d9041efef6949fe852a9b5e4faab8

tests/python/gaia-ui-tests/gaiatest/tests/unit/test_wifi_multiple_connect.py
```diff
@@ -0,0 +1,17 @@
...
+
+from marionette.marionette_test import parameterized
+
+from gaiatest import GaiaTestCase
+
+
+class TestWiFiMultipleConnect(GaiaTestCase):
+
+    @parameterized("1st_run")
+    @parameterized("2nd_run")
```
JohanLorenzo: I'd add a comment to explain why we need a second run of this test. Without this information we might end up changing the test to:

``` python
    def test_connect_and_forget_all_networks(self):
        for x in range(0, 60):
             # blah blah
```

mwargers: I'll do that. I would like to know where to put this test in, as you proposed to be part of sanity testing. In which functional area should it be part of? 
It should not be part of unit testing, because it's really not a unit test.

JohanLorenzo: How about leaving it in `gaiatest/functional` ?

PR #31474, opened 2015-08-21
Bug 1197144 - Make test_import_edit_export_contact.py even more robust

mwargers:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=db6bb6fcaf56356e01fbeccaf45c308b3a62848b

tests/python/gaia-ui-tests/gaiatest/tests/functional/contacts/test_import_edit_export_contact.py
```diff
@@ -68,6 +68,8 @@ def test_import_edit_export_contact(self):
         self.assertEqual(self.sim_contact['tel'][0]['value'], self.contact.tel['value'])
 
     def tearDown(self):
-        self.data_layer.delete_sim_contact(self.sim_contact['id'])
+        for contact in self.data_layer.sim_contacts:
+            if contact['name'][0].startswith('Name'):
+                self.data_layer.delete_sim_contact(contact['id'])
```
npark-mozilla: This should be safe, but wouldn't it be better if we create a contact name with 'GAIAUITEST_' prefix?  then it's almost guaranteed that no one would use such prefix, and when the user later encounters this contact on the sim card (by any chance the deletion didn't work) they'll know right away where this contact came from.

mwargers: That is a good idea. The only problem is that we're already piling up contacts with 'Name***' on them in the SIM cards.

JohanLorenzo: :+1: Let's [prefix the names there](https://github.com/mozilla-b2g/gaia/pull/31474/files#diff-087812ad7aa03d9cdb90e5349e7bcd6aR14).

mwargers: What about the existing situation then?

JohanLorenzo: It'd be better if the clean up of the contacts already present, happened in another script, that we run only once (on each device). What do you think?

mwargers: I guess that's possible. But about how many contributors are we talking? And are they really storing contacts that start with 'Name'? To me it sounds like a hypothetical situation here.
And to be honest, it should be expected that data is wiped, running these tests.

JohanLorenzo: I'd prefer to take no risk. A company might start with "name" and be in the SIM contacts.
We also should prepare for having more contributors.

We don't warn the user of his sim card being wiped. If we go on the wipe road, I'd make that clear on MDN and on the harness.

mwargers: "A company might start", you say. So currently, we don't have any company or anyone that's using sim card contacts on their device that starts with 'Name' in it and use it for Gaia UI tests, right?
Can't we just check this in, leave it in for a week, so devices on the lab have these test contacts wiped from their sim cards, then make the adjustment to use 'GAIAUITEST_' afterwards?

JohanLorenzo: I really think we should be careful with the SIM card and reduce the risk to 0. But I'm fine with leaving this for a week.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=254ff9b37adddfc78bd99e8d9cd4e81657f387b2

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=11f8b1522db01b3df20b82ae1cc52199e1a83d47

tests/python/gaia-ui-tests/gaiatest/tests/functional/contacts/test_import_edit_export_contact.py
```diff
@@ -11,7 +11,7 @@ def setUp(self):
         GaiaTestCase.setUp(self)
         curr_time = repr(time.time()).replace('.', '')
 
-        self.contact = MockContact(givenName='Name%s'%curr_time[12:], name='Name%s'%curr_time[12:], familyName='')
+        self.contact = MockContact(givenName='GAIAUITEST_%s'%curr_time[12:], name='GAIAUITEST_%s'%curr_time[12:], familyName='')
```
JohanLorenzo: Nit: We can extract the prefix in a variable, so we'll have 1 place to change, the next time we modify it.

mwargers: MockContact already seems to have some existing prefilled values in there? Why don't we use those?
http://mxr.mozilla.org/gaia/source/tests/python/gaia-ui-tests/gaiatest/mocks/mock_contact.py#29

JohanLorenzo: Fine by me

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=49baf84ce52776de77e711e4a74ba43918083d8e

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=37f4fdb612c34672f3a0951b75c9f31549ddb684

PR #31482, opened 2015-08-23
Bug 1197570 - don't load manifest when no basedir r=asppsa

asppsa:
If there is no locale basedir, getManifestProperties tries to open an
invalid path, triggering an error.  This patch makes
getManifestProperties return null instead.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=df3b85681087c63fd9295373dd9e76542b10caf8

PR #31494, opened 2015-08-24
Bug 1197787 - Use a normal function instead of an arrow function in calendar_service.js

jandem:
We're changing the semantics in bug 889158 to match the spec, so we need to fix this one.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=71ef70c89dab5b42bf904f1709745545615f79b8

PR #31495, opened 2015-08-24
Bug 1197787 - Don't use `arguments` in arrow function in calendar_service.js

jandem:
We're changing the semantics in bug 889158 to match the spec, so we need to fix this one.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=17ffee4a66fcac89243703f4671a676163b325aa

PR #31505, opened 2015-08-24
Bug 1027117 - Remove all .innerHTML attributes from l10n resources

zbraniecki:
https://bugzilla.mozilla.org/show_bug.cgi?id=1027117


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4ea3392d2ba1c3d8e7c8ff1b795f45dc2148d078

apps/system/locales/system.en-US.properties
```diff
@@ -201,7 +201,7 @@ wantToInstallNow=Do you want to install the update now? You will not be able to
 installNow=Install Now
 notNow=Not Now
 # the <span> element will be displayed when the download finishes
-downloadingUpdateMessage.innerHTML=Downloading updates… <span>{{ progress }} downloaded.</span>
+downloadingUpdateMessage=Downloading updates… <span>{{ progress }} downloaded.</span>
```
stasm: How is this `<span>` shown?  Do we need to add it to the source HTML?

apps/system/locales/system.en-US.properties
```diff
@@ -209,12 +209,12 @@ downloadingAppProgress={{ progress }} / {{ max }} downloaded
...
 wantToCancelAll=Are you sure you want to cancel all downloads?
 downloadError=There was an error while downloading the updates.
-updateAvailableInfo.innerHTML        = {[ plural(n) ]}
-updateAvailableInfo.innerHTML[one]   = {{ n }} update available. <span>Tap for more info.</span>
-updateAvailableInfo.innerHTML[two]   = {{ n }} updates available. <span>Tap for more info.</span>
-updateAvailableInfo.innerHTML[few]   = {{ n }} updates available. <span>Tap for more info.</span>
-updateAvailableInfo.innerHTML[many]  = {{ n }} updates available. <span>Tap for more info.</span>
-updateAvailableInfo.innerHTML[other] = {{ n }} updates available. <span>Tap for more info.</span>
+updateAvailableInfo        = {[ plural(n) ]}
+updateAvailableInfo[one]   = {{ n }} update available. <span>Tap for more info.</span>
```
stasm: Same here, how is the tap on this `<span>` handled?  Do we need to add it to the source HTML?

shared/js/l10n.js
```diff
@@ -1940,17 +1940,7 @@
 
     var entity = this.ctx.getEntity(l10n.id, l10n.args);
 
-    var value;
-    if (entity.attrs && entity.attrs.innerHTML) {
-      // XXX innerHTML is treated as value (https://bugzil.la/1142526)
-      value = entity.attrs.innerHTML;
-      console.warn(
-        'L10n Deprecation Warning: using innerHTML in translations is unsafe ' +
```
stasm: Do we really want to remove this warning now?  Or, perhaps turn it into an error?  Possible on parse-time.

apps/system/locales/system.en-US.properties
zbraniecki: I don't know, but I don't think that this patch changes the behavior, right? If lack of `<span>` in the source HTML was a problem it would be visible without the patch as well, right?

shared/js/l10n.js
zbraniecki: Why? It seems to me that we can start treating innerHTML the same way as we'd be treating entity.onclick. Why special case it?

stasm: Ah, good point.  So just ignore it via the DOM overlays white-list.  Cool!

apps/system/locales/system.en-US.properties
stasm: Yes, but since we're switching to DOM overlays, it's an opportunity to clean the code which previously had to handle that `span` every time the language changes.  Or maybe there is no such code and it's a simple `parent.complete span { display: inline; }` or someting.  I don't know, hence the question :)

PR #31509, opened 2015-08-25
Bug 1198132 - Apply gaia-header tweaks to Gallery to avoid reflow at …

janus926:
…startup.

### DIFF
--- a/apps/gallery/index.html
+++ b/apps/gallery/index.html
@@ -76,13 +76,13 @@
     <!-- for all views that display the thumbnails -->
     <section role="region" id="thumbnail-views" class="skin-dark">
       <!-- This header only shown on tablet view  -->
-      <gaia-header id="thumbnails-list-top" class="thumbnails-list">
+      <gaia-header id="thumbnails-list-top" class="thumbnails-list" title-start="0" title-end="0" no-font-fit>
         <h1 id="thumbnail-list-title" data-l10n-id="gallery"></h1>
       </gaia-header>
-      <gaia-header id="selected-header" class="thumbnails-select" action="close">
+      <gaia-header id="selected-header" class="thumbnails-select" action="close" title-start="50" title-end="0" no-font-fit>
         <h1 id="thumbnails-number-selected"></h1>
       </gaia-header>
-      <gaia-header id="pick-header" class="thumbnails-pick" action="close">
+      <gaia-header id="pick-header" class="thumbnails-pick" action="close" title-start="50" title-end="0" no-font-fit>
         <h1 id="pick-header-title" data-l10n-id="pickoneimage2"></h1>
       </gaia-header>
 
@@ -105,7 +105,7 @@
     </section>
 
     <section role="region" id="fullscreen-view" class="skin-dark">
-      <gaia-header id="fullscreen-toolbar-header" action="back" class="large-only">
+      <gaia-header id="fullscreen-toolbar-header" action="back" class="large-only" title-start="50" no-font-fit>
         <h1 data-l10n-id="gallery" id="fullscreen-title"></h1>
         <a id="fullscreen-camera-button-large">
           <span data-icon="camera"></span>
@@ -163,7 +163,7 @@
     </form>
 
     <section role="region" id="crop-view" class="skin-dark">
-      <gaia-header id="crop-top" action="back">
+      <gaia-header id="crop-top" action="back" title-start="50" no-font-fit>
         <h1 id="crop-header" data-l10n-id="cropimage"></h1>
         <button id="crop-done-button" data-l10n-id="done">done</button>
       </gaia-header>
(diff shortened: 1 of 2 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=06a65686a7e51e101183e23967bc6d7d2a602ed5

apps/gallery/index.html
```diff
@@ -105,7 +105,7 @@ <h1 id="pick-header-title" data-l10n-id="pickoneimage2"></h1>
     </section>
 
     <section role="region" id="fullscreen-view" class="skin-dark">
-      <gaia-header id="fullscreen-toolbar-header" action="back" class="large-only">
+      <gaia-header id="fullscreen-toolbar-header" action="back" class="large-only" no-font-fit>
```
davidflanagan: This one has an action attribute, so why doesn't it have title-start="50" like the other headers with action buttons?
The "large-only" class means that this header is only for the no-longer-supported tablet form factor, so we don't really have any way of testing the change here.

Given that this title bar will be hidden, will it still run the font-fit logic on startup? That is, do we need the no-font-fit logic here, or can we leave this change out?

I'd prefer to leave it out (since it is untestable and unsupported) if possible.

apps/gallery/js/gallery.js
```diff
@@ -400,6 +400,9 @@ function setView(view) {
   }
   switch (view) {
     case LAYOUT_MODE.list:
+      if (!isPhone) {
+        $('thumbnails-list-top').removeAttribute('no-font-fit');
+      }
```
davidflanagan: Is this header running the font-fit algorithm even though it is never displayed on the phone? 
In fact, none of these gaia header elements are displayed when the app starts up, and I think this is why I never bothered to use this no-font-fit trick for the app.

apps/gallery/js/gallery.js
```diff
@@ -421,8 +428,15 @@ function setView(view) {
...
     case LAYOUT_MODE.edit:
+      $('edit-header').removeAttribute('no-font-fit');
       NFC.unshare();
       break;
+    case LAYOUT_MODE.pick:
+      $('pick-header').removeAttribute('no-font-fit');
+      break;
+    case LAYOUT_MODE.crop:
+      $('crop-top').removeAttribute('no-font-fit');
+      break;
```
davidflanagan: I think I would slightly prefer the simplicity of calling removeAttribute for all of these headers in startup.js, after the performance.mark('visuallyLoaded') call.

Could you just do something like this in setup.js:

  document.querySelectorAll('gaia-header').forEach(function(header) { header.removeAttribute('no-font-fit'); });

That would be slightly less efficient than this code, but simpler and easier to maintain, and it keeps the font-fit off the startup path.

If you've got good reasons to put this code here in setView(), let's discuss them in the bug. But setView() is already more complicated than I'd like.

apps/gallery/index.html
davidflanagan: Now that I've read the rest of the patch, I suspect that all of the headers are doing font-fits even if they are never displayed like this one. So I suppose you do need to add the no-font-fit here.  Please consider adding title-start as well just to match the other ones.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=834f322d82405505c7344a9893e61de983bf08d0

apps/gallery/index.html
janus926: Yes, font-fit will be done no matter it's displayed or not :(

Added |title-start="50"|.

apps/gallery/js/gallery.js
janus926: Moved to startup.js after performance markers.

apps/gallery/js/startup.js
```diff
@@ -271,6 +271,11 @@
       // and is ready to interact with.
       window.performance.mark('visuallyLoaded');
       window.performance.mark('contentInteractive');
+
+      var headers = document.querySelectorAll('gaia-header');
```
davidflanagan: Please add a comment explaining what this code is doing.

apps/gallery/js/startup.js
```diff
@@ -271,6 +271,11 @@
...
+      [].forEach.call(headers, function(header) {
```
davidflanagan: When I suggested forEach(), I was thinking that querySelectorAll() returned a true array instead of a NodeList. If I was writing it, though, I'd use a for(;;) loop, or a for (h of headers) {} instead of having to do the [].foreach.call hack. But I'm okay with it like this if you want to leave it this way.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=cc1d55ec985005a0daa6ccd7825e6def2d47fa84

apps/gallery/js/startup.js
janus926: Added.

apps/gallery/js/startup.js
janus926: Use for(;;) instead.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=39a2939ff7ff84f64d11fe8cd5272b47f5a76eed

PR #31520, opened 2015-08-25
Bug 1198427 - A way to add extra apps to a target at build time R=ferjm

michielbdejong:
A few open issue:
- Not sure how/where to add tests for this?
- Should update https://developer.mozilla.org/en-US/Firefox_OS/Developing_Gaia/Customizing_build-time_apps if/once this lands
- It's a bit ugly that you have to specify multiple apps with a '\ ', like:

```
EXTRA_APPS := "dev_apps/app1\ dev_apps/app2"
```

in local.mk now, to make sure bash handles that string correctly, any idea how to fix that?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -426,6 +426,7 @@ GAIA_ALLAPPDIRS=$(shell find -L $(GAIA_DIR)$(SEP)apps $(GAIA_DIR)$(SEP)dev_apps 
 
 GAIA_APPDIRS?=$(shell $(call run-js-command,scan-appdir, \
 		GAIA_APP_CONFIG="$(GAIA_APP_CONFIG)" \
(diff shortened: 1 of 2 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=fab80b6220caf8f6adb6f30a160c4ba394f36ade

michielbdejong:
After thinking about it some more, I don't think this is a good idea. For the Synchronizer app, we should:
* add it in dev_apps/ (so that it's not built by targets that include `apps/*`)
* add it to the engineering target (so people can play with it if they set `ENABLE_DATA_SYNC := 1`)
* when the public Syncto servers go into operation, add it to the production target, so foxfooders can play with it when they enable it in the developer menu

PR #31524, opened 2015-08-25
Bug 1162586 - [Calendar] move database interactions off the main thread

millermedeiros:
removed all the DB calls from the main thread. still need to update unit tests to match the new structure... took way longer than expected because of all sorts of regressions and insane amount of coupling between modules.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7972957bcce029619a4528f4ad7a0aeb157d200e

PR #31525, opened 2015-08-25
Bug 1192744 - Use mozHour12 in SMS message report

zbraniecki:
https://bugzilla.mozilla.org/show_bug.cgi?id=1192744

### DIFF
--- a/apps/sms/views/conversation/js/information.js
+++ b/apps/sms/views/conversation/js/information.js
@@ -90,10 +90,11 @@ const reportDateTimeFormatOptions = {
 };
 
 function completeLocaleFormat(timestamp) {
-  return new Date(+timestamp).toLocaleString(
-    navigator.languages,
-    reportDateTimeFormatOptions
-  );
+  var options = Object.assign(
+    { hour12: navigator.mozHour12 },
+    reportDateTimeFormatOptions);
+
+  return new Date(+timestamp).toLocaleString(navigator.languages, options);
 }
 
 function l10nContainsDateSetup(element, timestamp) {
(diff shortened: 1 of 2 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=cda3be2ee2628c8d13efd05a2d4b243ae9cd768b

apps/sms/views/conversation/js/information.js
```diff
@@ -82,6 +82,7 @@ const MESSAGE_EVENTS = [
 ];
 
 const reportDateTimeFormatOptions = {
+  hour12: navigator.mozHour12,
```
julienw: this doesn't work properly, you need to add it dynamically at render time, otherwise the following STR doesn't work:
1. open the SMS app.
2. open the Settings app, go to Date and time panel, and change the Time format
3. return to the SMS app, go to a conversation, long press a message and display the report
=> the format is wrong

It's because the "hour12" property is set when the app is launched here.

However it's properly changed when we change the time format while the report is actually displayed. But that's not enough :D

The fact that no unit test change was necessary should have alerted you ;)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=495cbf0f8fa175a8e9261dfd6c8cfde0b46a7ec5

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4f4498e1a73f83598b1fe57952e2276f38281887

PR #31539, opened 2015-08-26
Bug 1169002 - [User Story] View Pinned Pages. r=albertopq

Cwiiis:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f5f41a08ff85ce84fb4060271f952e42a082c097

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=56577da5009dc4e9951cc125b418b389c2aaa6b1

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=eb5ad8369bcbceaadc8a9905a88d00266f7b861a

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=fa25d4eac3d5332c7ff5bb1a04c04640791243de

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1f3fe84d024bf1d39d7c1b991ce02f197314415a

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ef08caf9e7f3208b6ce7e63a6eb212791495f5b4

apps/homescreen/index.html
```diff
@@ -33,39 +37,59 @@
...
-        <p class="body" data-l10n-id="resume-download-body"></p>
-      </section>
-      <button class="action" data-l10n-id="resume-download-action"></button>
-      <button on-click="close" data-l10n-id="cancel-action"></button>
-    </gaia-dialog>
-    <gaia-dialog id="settings" class="dialog-action">
-      <button class="action" data-l10n-id="homescreen-settings"></button>
-      <button on-click="close" data-l10n-id="cancel-action"></button>
-    </gaia-dialog>
+  <div id="page-indicator"><span class="active"></span><span></span></div>
```
albertopq: What are those spans? Should we add some semantic to make it more understandable (accessible)?

apps/homescreen/index.html
```diff
@@ -33,39 +37,59 @@
...
+    <div id="panels"><!--
```
albertopq: Are those  intentional?

apps/homescreen/index.html
Cwiiis: Good call, probably - the spans are the dots that indicate which page is active - Do you have any suggestions here with regards to accessibility?

apps/homescreen/index.html
Cwiiis: Yes, the white-space ends up inserting actual space in the page layout - I could work around this in CSS, but rather than complicating that, it's easier to just remove the white-space in the document with comments.

apps/homescreen/js/app.js
```diff
@@ -523,6 +527,14 @@ const SETTINGS_VERSION = 0;
 
       // Display the top shadow when scrolling down
       case 'scroll':
+        if (e.target === this.panels) {
```
albertopq: Is this for updating the horizontal pages indicator? In that case, do we have any event when that transition finishes, instead of checking it on every sroll?

apps/homescreen/js/datastore.js
```diff
@@ -198,7 +201,9 @@
           (event) => {
             var cursor = event.target.result;
             if (cursor) {
-              results.push(cursor.value);
+              if (filter(cursor.value)) {
```
albertopq: Is this for searching pinned pages? Shouldn't we create an index and search into the datastore using it instead of getting all and filtering afterwards?

apps/homescreen/js/pages.js
```diff
@@ -0,0 +1,174 @@
...
+        var background = {};
+        if (page.screenshot) {
+          background.src = URL.createObjectURL(page.screenshot);
+        }
+        if (page.themeColor) {
+          background.themeColor = page.themeColor;
+        }
+        card.background = background;
+
+        IconsHelper.getIconBlob(page.url, PAGES_ICON_SIZE, page)
```
albertopq: Shouldn't we use the shared/homescreen/icon here, as we do in app_chrome? That will use the GaiaIconRenderer for applying the same effects in all places.

apps/homescreen/js/app.js
Cwiiis: Yes, and unfortunately not - not until https://bugzilla.mozilla.org/show_bug.cgi?id=1172171 anyway.

apps/homescreen/js/datastore.js
Cwiiis: This is for only storing/displaying pinned pages from the places datastore (which has all history) - as far as I know, there's no way of creating an index/query directly from a DataStore: https://developer.mozilla.org/en-US/docs/Web/API/DataStore (the spec even says to do what we're doing - synchronise and keep your own index for querying purposes)

apps/homescreen/js/pages.js
Cwiiis: I'll take a look - this just sets the icon blob on the gaia-pin-card element, but I assume you're saying that there's some shared code to do that? I think Guillaume copied this code from app_chrome, but I'll have another look.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f10af70097ba30d844db7dcc8432a29176b31c28

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d2831f6ec45e3c2c8b27a2b09a6792911b61142b

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0df623868729704b529fa881c2c9bd034f48beee

apps/homescreen/js/pages.js
Cwiiis: So just checked, and this is basically exactly what system does in app_window.js. Ideally, this logic would be hidden in the card component, but I suppose that can be dealt with later.

apps/homescreen/index.html
```diff
@@ -33,39 +37,59 @@
...
-        <p class="body" data-l10n-id="resume-download-body"></p>
-      </section>
-      <button class="action" data-l10n-id="resume-download-action"></button>
-      <button on-click="close" data-l10n-id="cancel-action"></button>
-    </gaia-dialog>
-    <gaia-dialog id="settings" class="dialog-action">
-      <button class="action" data-l10n-id="homescreen-settings"></button>
-      <button on-click="close" data-l10n-id="cancel-action"></button>
-    </gaia-dialog>
+  <div id="page-indicator"><span></span><span></span></div>
```
albertopq: Mhm, despite it says the comment is outdated, I can't see any change on those spans?

apps/homescreen/js/datastore.js
albertopq: Sorry, I thought we were using indexedDB

apps/homescreen/js/pages.js
albertopq: Yeah, you are right. Forgot that we moved the Icon to the icons_helper. Thanks!

apps/homescreen/index.html
Cwiiis: The a11y support was added in the Javascript code.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8ca5341305f76158fe9ae754ebab0d9e29f4cb76

PR #31573, opened 2015-08-28
Bug 1198824 - use call state to verify isOnCall()

alastor0325:
Don't check the current audio channel in the isOnCall(), because now we can't set the "telephony" type correctly.
It would success after landing the bug1129882. However, at that time, we need to revert this changeset and also need to remove the call state checking. See bug819858.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=bf578094154a893b1d0f8a13cd049f6cbf710aa7

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=513c58d4a4341baa71a8aa28705fe81a5eab5bea

PR #31596, opened 2015-08-31
Bug 1200116 - [Messages][Tests] ThreadGenerator should be able to create threads with MMS. r=julien

azasypkin:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=bc3733e9c678d672829132d96e0bf9f97ec6ce3b

apps/sms/test/marionette/conversation_test.js
```diff
@@ -154,11 +154,11 @@ marionette('Conversation Panel Tests', function() {
           type: 'Mobile'
         }]
       }]);
+
+      messagesApp.Inbox.findConversation(threads[1].id).tap();
```
azasypkin: I want to remove ```firstConversation``` for a long time already (and remove it in my drafts dialog patch), for the following reason: sometimes threads can be generated with the same timestamp (using Date.now) so that the order in which they appear  can change from run to run that leads to intermittents. So here I've just used the chance to do that.

apps/sms/test/marionette/conversion_banner_test.js
```diff
@@ -132,7 +132,11 @@ marionette('Message Type Conversion Banner', function() {
       var mmsThread = ThreadGenerator.generate({
         numberOfMessages: 1,
         messageType: 'mms',
-        participants: ['a@b.c']
+        participants: ['a@b.c'],
+        attachments: [
```
azasypkin: Just for the Demo purpose - to show how to create messages with attachments.

apps/sms/test/marionette/generators/thread.js
```diff
@@ -5,48 +5,116 @@
...
-        delivery: 'sent',
-        body: thread.body + ' (messageId: ' + id + ')',
-        timestamp: Date.now()
-      });
+      var message = null;
+      var messageBody = thread.body + ' (messageId: ' +
+        (ThreadGenerator.uniqueMessageId + 1) + ')';
+      if (parameters.messageType === 'sms') {
+        message = this.generateSMS({
+          threadId: thread.id,
```
azasypkin: It seems here is too much repetition, but I didn't want to mix real message properties with ThreadGenerator parameters and it's for tests only after all :)

apps/sms/test/marionette/generators/thread.js
```diff
@@ -5,48 +5,116 @@
...
+    }
 
+    return _.assign(
+      { id: smsParameters.id || ++ThreadGenerator.uniqueMessageId },
+      DEFAULT_SMS,
+      smsParameters
+    );
+  },
+
+  generateMMS: function(mmsParameters) {
```
azasypkin: Using these public methods we can create any threads we want without pretty dumb ```generate``` function.

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
```diff
@@ -32,6 +32,70 @@ Services.obs.addObserver(function(document) {
...
+        canvas.width = canvas.height = 0;
+        canvas = null;
+
+        resolve(blob);
+      }, type, quality);
+    });
+  }
+
+  function rehydrateAttachments(storage) {
+    var attachmentPromises = [];
```
azasypkin: Somewhat ugly method, haven't found significantly better way to rewrite it yet - so keeping as is....

apps/sms/test/marionette/generators/thread.js
julienw: agreed: for example mixed sms/mms content

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
```diff
@@ -32,6 +32,70 @@ Services.obs.addObserver(function(document) {
...
+    storage.threads.forEach((thread) => {
+      thread.messages.forEach((message) => {
+        if (!message.attachments) {
+          return;
+        }
+
+        message.attachments.forEach((attachment, i) => {
+          var blobPromise;
+
+          if (attachment.type.indexOf('image') >= 0) {
```
julienw: nit: use `if (attachment.type.startsWith('image/'))` ?

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
```diff
@@ -32,6 +32,70 @@ Services.obs.addObserver(function(document) {
...
+          attachmentPromises.push(
+            blobPromise.then((blob) => {
+              message.attachments[i] = Cu.cloneInto({ content: blob }, window);
+            })
+          );
+        });
+      });
+    });
+
+    return Promise.all(attachmentPromises).then(() => storage);
```
julienw: nit: it feels out of place to return "storage" as promise value; I'd rather return a promise with no value. It feels like you return "storage" just becuse it makes things easier to you then :)

But I don't have a strong opinion so you can keep it if you want.

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
```diff
@@ -32,6 +32,70 @@ Services.obs.addObserver(function(document) {
...
+        resolve(blob);
+      }, type, quality);
+    });
+  }
+
+  function rehydrateAttachments(storage) {
+    var attachmentPromises = [];
+
+    storage.threads.forEach((thread) => {
+      thread.messages.forEach((message) => {
```
julienw: nit: maybe guard against an undefined `messages` ?

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
julienw: nested structures are always awkward to iterate with, but I think it's ok here.

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
azasypkin: Yeah, I have a mixed feelings as well, but in the end I left it as is because ```rehydrateAttachments``` basically modifies storage and returns that modified storage, I can return empty promise there instead, and below do smth like this:
```
storagePromise = storagePromise.then((storage) => {
	return rehydrateAttachments(storage || {
	  threads: new Map(),
	  recipientToThreadId: new Map(),
	  uniqueMessageIdCounter: 0
	}).then(() => storage);
});
```
Do you like it better? :)

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
azasypkin: Yep

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
azasypkin: Right!

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
azasypkin: Rewriting using the approach above...

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d3feb3218d8058a8bcfd01663923b9d9ff26aed9

apps/sms/test/marionette/mocks/mock_navigator_moz_mobile_message.js
```diff
@@ -43,12 +103,16 @@ Services.obs.addObserver(function(document) {
...
+        if (!storage) {
+          return {
+            threads: new Map(),
+            recipientToThreadId: new Map(),
+            uniqueMessageIdCounter: 0
+          };
+        }
+
+        return rehydrateAttachments(storage).then(() => storage);
       });
```
julienw: or even:
```javascript

storagePromise = storagePromise.then((storage) => {
   return storage || {
       threads: new Map(),
       recipientToThreadId: new Map(),
       uniqueMessageIdCounter: 0
   };
}).then(rehydrateAttachments).then(() => storage);
```

julienw: also, maybe getStorage could reject if it finds no storage, in that case we could do:

```javascript
storagePromise = storagePromise.catch(() => {
   return {
       threads: new Map(),
       recipientToThreadId: new Map(),
       uniqueMessageIdCounter: 0
   };
}).then(rehydrateAttachments).then(() => storage);
```

julienw: but no big deal, as you wish.

azasypkin: Regarding rejection if no storage found - good idea, why not, I'll consider that in one of the future integration related patches (just don't want to wait for Treeherder once again :p) :)

PR #31611, opened 2015-08-31
Bug 1196717 - Dogfood test - Delete thread in SMS

JohanLorenzo:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=332e7f38af8dcbca0a5e00e6a5fd40beedf55912

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=42ef2623d487bc4a46cafcd97072657913cbaf92

tests/python/gaia-ui-tests/gaiatest/apps/contacts/regions/contact_details.py
```diff
@@ -94,11 +94,9 @@ def a11y_click_edit(self):
...
 
     def tap_back(self):
-        el = self.marionette.find_element(*self._details_header_locator)
-        Wait(self.marionette).until(expected.element_displayed(el))
-        # TODO: remove tap with coordinates after Bug 1061698 is fixed
-        el.tap(25, 25)
-        Wait(self.marionette).until(expected.element_not_displayed(el))
+        back_button = self.marionette.find_element(*self._details_header_locator)
+        self.tap_element_from_system_app(element=back_button)
+        Wait(self.marionette).until(expected.element_not_displayed(back_button))
```
mwargers: tap_element_from_system_app is switching to the system app, so I'm not sure how this works.
In general, I would prefer it we're going away from negative checks like these.
But I guess it works, because otherwise you haven't written it like this.

tests/python/gaia-ui-tests/gaiatest/apps/messages/app.py
```diff
@@ -52,20 +64,57 @@ def tap_first_received_message(self):
...
+    def tap_thread_options(self):
+        self.marionette.find_element(*self._thread_options_button).tap()
+        from gaiatest.apps.messages.regions.options import ThreadOptions
+        return ThreadOptions(self.marionette)
+
+    def enter_select_mode(self):
+        edit_header = self.marionette.find_element(*self._edit_thread_header)
+        thread_options = self.tap_thread_options()
+        thread_options.select_threads()
+        Wait(self.marionette).until(lambda m: edit_header.rect['y'] == 0)
```
mwargers: Could you perhaps also add and edit_header.is_displayed() here? Becuase I think that edit_header.rect['y'] is also 0 when it's not displaying.

tests/python/gaia-ui-tests/gaiatest/apps/messages/regions/options.py
```diff
@@ -0,0 +1,52 @@
...
+
+    _root_locator = (By.CSS_SELECTOR, 'form.visible')
+
+    def __init__(self, marionette):
+        root = marionette.find_element(*self._root_locator)
+        PageRegion.__init__(self, marionette, root)
+        self.wait_for_panel_to_appear()
+
+    def wait_for_panel_to_appear(self):
+        Wait(self.marionette).until(lambda m: self.root_element.rect['y'] == 0)
```
mwargers: Here too, I would prefer it if you would also check for root_element.is_displayed().

tests/python/gaia-ui-tests/gaiatest/apps/messages/regions/options.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def wait_for_panel_to_hide(self):
+        Wait(self.marionette).until(expected.element_not_displayed(self.root_element))
+
+    def _tap_and_wait_until_hidden(self, locator):
+        self.root_element.find_element(*locator).tap()
+        self.wait_for_panel_to_hide()
```
mwargers: You're only using this once. And the method itself consists of only 1 line of code. I'd prefer if you just used that 1 line of code here.

tests/python/gaia-ui-tests/gaiatest/tests/functional/messages/test_delete_thread.py
```diff
@@ -0,0 +1,58 @@
...
+        except KeyError:
+            raise SkipTest('Plivo account details not present in test variables')
+
+        plivo = PlivoUtil(
+            self.testvars['plivo']['auth_id'],
+            self.testvars['plivo']['auth_token'],
+            self.testvars['plivo']['phone_number']
+        )
+
+        plivo.send_sms(to_number=self.environment.phone_numbers[0].replace('+', ''),
```
mwargers: Why not send an sms to the phone number this phone is using?

tests/python/gaia-ui-tests/gaiatest/apps/messages/regions/options.py
mwargers: I guess it's for future use.

tests/python/gaia-ui-tests/gaiatest/tests/functional/messages/test_delete_thread.py
```diff
@@ -0,0 +1,58 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from marionette import SkipTest
+from marionette_driver import Wait
```
mwargers: Is this import necessary?

tests/python/gaia-ui-tests/gaiatest/tests/functional/messages/test_delete_thread.py
```diff
@@ -0,0 +1,58 @@
...
+            self.testvars['plivo']['auth_token'],
+            self.testvars['plivo']['phone_number']
+        )
+
+        plivo.send_sms(to_number=self.environment.phone_numbers[0].replace('+', ''),
+                       message=self._generate_text())
+
+        system = System(self.marionette)
+        system.wait_for_notification_toaster_displayed(timeout=300,
+                                                       message="Notification did not appear. SMS database dump: %s " % self.data_layer.get_all_sms())
```
mwargers: Wasn't the problem with this not that the message is compiled at the time that the wait_for_notification_toaster_displayed function is called, so the SMS database dump is from before the timeout of 300s?

tests/python/gaia-ui-tests/gaiatest/tests/functional/messages/test_delete_thread.py
```diff
@@ -0,0 +1,58 @@
...
+    def test_delete_threads(self):
+        """ https://moztrap.mozilla.org/manage/case/15927/ """
+
+        messages = Messages(self.marionette)
+        messages.launch()
+
+        new_message = messages.create_new_message(recipients=[self.environment.phone_numbers[0]],
+                                                  message=self._generate_text())
+        new_message.save_as_draft()
+
```
mwargers: Perhaps we can have a self.assertEqual(len(messages.threads), 2) here?

tests/python/gaia-ui-tests/gaiatest/apps/contacts/regions/contact_details.py
mwargers: Wait a minute, is this tap_element_from_system_app method really necessary here? 
I mean, we're not closing an app here, do we? We just stay in the contacts app, or not?

tests/python/gaia-ui-tests/gaiatest/tests/functional/messages/test_delete_thread.py
JohanLorenzo: `self.assertGreaterEqual(len(messages.threads), 2)`, in case where we receive an extra SMS from the carrier.

tests/python/gaia-ui-tests/gaiatest/apps/messages/app.py
JohanLorenzo: I'll add it, to make the intent clearer. FYI, the header is at y < 0, when it's hidden

tests/python/gaia-ui-tests/gaiatest/apps/contacts/regions/contact_details.py
JohanLorenzo: We need to tap the close button, which is in an HTML custom element. So Marionette doesn't see the button

tests/python/gaia-ui-tests/gaiatest/tests/functional/messages/test_delete_thread.py
JohanLorenzo: Sorry, I don't understand the question. The SMS is sent to the phone being tested.

tests/python/gaia-ui-tests/gaiatest/tests/functional/messages/test_delete_thread.py
JohanLorenzo: Good catch! I'll change that!

tests/python/gaia-ui-tests/gaiatest/apps/messages/regions/options.py
JohanLorenzo: Yeah, [YAGNI](https://en.wikipedia.org/wiki/YAGNI) could apply here. But for consistency (we already have the "wait for appear" method), I'd rather keep it.

tests/python/gaia-ui-tests/gaiatest/apps/messages/app.py
JohanLorenzo: Done

tests/python/gaia-ui-tests/gaiatest/apps/messages/regions/options.py
JohanLorenzo: Done.

tests/python/gaia-ui-tests/gaiatest/tests/functional/messages/test_delete_thread.py
JohanLorenzo: Good catch. Removed.

tests/python/gaia-ui-tests/gaiatest/tests/functional/messages/test_delete_thread.py
JohanLorenzo: Done.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=402af577ee04b730a3968827b95b964552c02b73

tests/python/gaia-ui-tests/gaiatest/apps/contacts/regions/contact_details.py
mwargers: Yes, but I mean the tap_element_from_system_app call you're making here. Is that really necessary? Because this should only be used when it is really necessary.

tests/python/gaia-ui-tests/gaiatest/tests/functional/messages/test_delete_thread.py
mwargers: Yes, but you could use the phone itself to send an sms to itself, like we do in:
http://mxr.mozilla.org/gaia/source/tests/python/gaia-ui-tests/gaiatest/tests/functional/messages/test_sms_notification.py
That way, we don't need to use external resources like Plivo.

tests/python/gaia-ui-tests/gaiatest/tests/functional/messages/manifest.ini
```diff
@@ -69,3 +69,7 @@ skip-if = device == "desktop"
 skip-if = device == "desktop"
 # Bug 1172460 - Intermittent failure in test_sms_notification_removed_when_sms_deleted.py in wait_for_notification_toaster_displayed
 stable = false
+
+[test_delete_thread.py]
+skip-if = device == "desktop"
+dogfood = true
```
mwargers: If you keep on using Plivo, then I guess you need to add external = true here.

tests/python/gaia-ui-tests/gaiatest/apps/messages/app.py
mwargers: rect calls getElementRect in the marionette listener, I think, which uses getBoundingClientRect()
http://mxr.mozilla.org/mozilla-central/source/testing/marionette/driver/marionette_driver/marionette.py#149
http://mxr.mozilla.org/mozilla-central/source/testing/marionette/listener.js#1569
When an element is display: none, the .x and .y values are 0, I think.

I think when you mean that when header is hidden, it is displayed, but it's not at a place where it can be seen.

tests/python/gaia-ui-tests/gaiatest/apps/contacts/regions/contact_details.py
JohanLorenzo: Isn't that the way we tap on Shadow DOM elements?

tests/python/gaia-ui-tests/gaiatest/apps/messages/app.py
JohanLorenzo: I'll create a function to make the explanation clearer.

What happens here is: The header goes down. Once some pixels of the elements are displayed on the screen, `element.is_displayed()` returns `True`. So the wait will be over, even though the transition of the header is not done yet. Waiting on y == 0, allows us to know when the transition is done.

tests/python/gaia-ui-tests/gaiatest/tests/functional/messages/test_delete_thread.py
JohanLorenzo: I'd rather stop from using the data layer to send SMS. This makes us dependent from the internal APIs. Hence, we can use the data layer in an integration test, but not in an end to end test.

On another note, if we send an SMS to itself, we'd have only 1 thread to delete, instead of 2.

tests/python/gaia-ui-tests/gaiatest/tests/functional/messages/manifest.ini
JohanLorenzo: You're right. I added it.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=469999228c760d0023e3c0919ca35169a9eb9747

tests/python/gaia-ui-tests/gaiatest/tests/functional/messages/test_delete_thread.py
mwargers: I don't see why it's wrong to be dependant from an internal API here.
We use these data_layer things all the time to get into a state from where we want to start test from. We could also do all the manual steps in automation to send the SMS, but that would make the test much slower, to get to the part which we want to test here.
The only thing that we want to test here is the delete thread feature in SMS. We could also create the entries in the SMS database directly, for instance, to work from there to test the delete thread feature.

tests/python/gaia-ui-tests/gaiatest/apps/contacts/regions/contact_details.py
mwargers: No, it's used when the button that's being tapped upon closes the app. When that happens, Marionette is out of whack, because it still thinks the active frame that is being worked on is that app. See https://bugzilla.mozilla.org/show_bug.cgi?id=1164078 for more explanation.
For shadow DOM elements, you just have to use coordinates.

JohanLorenzo: Okay. I'll change it

JohanLorenzo: I rolled back these changes.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3dd2e1873e6c80368d05e76400bab4f65aaee42d

PR #31624, opened 2015-08-31
Bug 1175623 - support variable length PINs in the lockscreen

davidflanagan:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=70b0775becd40c47bbf92d50b8aedc450339b72b

apps/system/lockscreen/js/lockscreen_inputpad.js
```diff
@@ -66,7 +70,7 @@
   LockScreenInputpad.prototype.renderUI = function() {
     return new Promise((resolve, reject) => {
       this.toggleEmergencyButton();
-      this.updatePassCodeUI();
+      this.setPinLength(this.pinLength);
```
snowmantw: I prefer to write it as:

    this.setPinLength(this.pinLength);
    this.updatePassCodeUI();

Since it avoid to do too much things inside one function, which is also because the name `setPinLength` looks like a simple setter, not a function drawing thing.

snowmantw: But since the `setPinLength` conditionally call the `updatePassCodeUI`, the condition may need to be moved to here as well, rather than in the setter.

apps/system/lockscreen/js/lockscreen_inputpad.js
```diff
@@ -144,6 +148,26 @@
...
+      for(var i = 0; i < n; i++) {
+        this.passcodeCode.appendChild(document.createElement('span'));
+      }
+
+      // Allow CSS to display the n digits at the right size.
+      this.passcodeCode.dataset.pinLength = n;
+
+      // Set the right states on all of the spans
+      this.updatePassCodeUI();
+    }
```
snowmantw: I don't think it's good to make a function that may do nothing. Maybe we should move this condition to where it is called, and to call it only when the condition is true.

snowmantw: And I think it's part of the issue from we combined setter and rendering function into one.

PR #31647, opened 2015-09-02
Bug 1200517 - Let text-selection-dialog show on top of ActivityWindow.

chenpighead:
When adding page to home screen in bookmark app, there's an appWindow with class name 'ActivityWindow' is created. Since ActivityWindow is under bookmark app
w/ z-index = 65536, the app-based text-selection-dialog under bookmark app w/ z-index = 2 is covered, which looks like a no show.

In this patch, the z-index change only applies to the text-selection-dialog under each appWindow, so this won't affect the z-index relationships between appWindows and other divs (ex. notifications).

### DIFF
--- a/apps/system/style/textselection_dialog/textselection_dialog.css
+++ b/apps/system/style/textselection_dialog/textselection_dialog.css
@@ -11,7 +11,7 @@
   transition: opacity var(--transition-duration) ease;
 }
 .appWindow > .textselection-dialog {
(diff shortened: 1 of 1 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b3a3fc53eb3078e8d900cc001273816e48f56ccc

PR #31659, opened 2015-09-02
Bug 1199182 - Ensure grid height gets set when necessary. r=albertopq

Cwiiis:


apps/homescreen/js/app.js
```diff
@@ -174,8 +178,7 @@ const SETTINGS_VERSION = 0;
               }
               this.bookmarks.get(id).then((bookmark) => {
                 this.addAppIcon(bookmark.data);
-                this.storeAppOrder();
```
albertopq: Why do we need to remove this?

Cwiiis: It's redundant - Newly added apps will always be at the end of the grid and the order is saved as soon as they get rearranged (app order being saved is backed by a marionette test, I believe).

apps/homescreen/js/app.js
```diff
@@ -436,7 +443,7 @@ const SETTINGS_VERSION = 0;
       this.handleEvent({ type: 'scroll' });
     },
 
-    snapScrollPosition: function(bias) {
+    refreshMetrics: function(snap, bias) {
```
KevinGrandon: I'm not really a fan of this function naming, afterall, naming things is the hardest problem in computer science. When I think metrics, I think statistics collected by some server.

My preference here would be an individual getter for both snap and bias, more clarity that way and avoids the boolean trap you have now.

Cwiiis: Yeah, good call - I think I didn't because the snap part relies on information from the calculation part, but I could easily just return that information. Will do.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d42ae2cedc6ff4940eb99d7546bf6891a41d944b

apps/homescreen/test/unit/app_test.js
```diff
@@ -341,55 +355,126 @@ suite('Homescreen app', () => {
...
+
+    test('should remove overflow and scroll to nearest snap point', () => {
+      app.pendingGridHeight = 500;
+      app.pageHeight = 100;
+      app.scrollable.scrollTop = 10;
+
+      app.snapScrollPosition();
+      assert.equal(app.scrollable.style.overflow, '');
+      assert.isTrue(scrollToSpy.calledWith(
+                      { left: 0, top: 0, behavior: 'smooth' }));
```
KevinGrandon: nit: seems like some funky indenting.

PR #31661, opened 2015-09-02
Bug 1201070 - emptied signals video unload, handle abort

sfoster:
https://bugzilla.mozilla.org/show_bug.cgi?id=1201070

Looks like we can rely on emptied always firing when unloading, so the abort handler can be moved up to onVideoUnloaded where it will always get registered


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=fb82c900d20bfa6522022c2af741d161e2dd4e0c

PR #31666, opened 2015-09-03
Bug 1153781 - Revoke notifications for deleted events

lambdabaa:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=fda34316be238d1e3816129893bf8d01203a041d

apps/calendar/js/notification.js
```diff
@@ -37,6 +37,15 @@ exports.sendNotification = function(title, body, url) {
   });
 };
 
+exports.revokeNotificationsForEvent = function(event) {
+  var title = event.remote.title;
```
jrburke: Using title seems imprecise, particularly since the setting of the title for the sendNotification() seems to involve an l10n.get() call? I am still new to the calendar code, so maybe that is not correct? If so though, either matches could not match, or I could see a case where there are two events with the same title, but for different times?

Since this module has a closeNotifications(url), is it possible to calculate or obtain the related url for this event? If not, and it is better to use event IDs, you can pass data in notifications now, via a `data` property in the notification options. The data should be something that is fairly simple (no functions and such), in this case though id with a string value should be fine. Then in this method, it can look at the data object to get the event ID associated with the notification.

If you want an example of data usage, in email, the sending of `data` in notification options is done here:
https://github.com/mozilla-b2g/gaia/blob/master/apps/email/js/sync.js

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=71256668a9169f86331d294915a1d65868a82277

PR #31677, opened 2015-09-03
Bug 1184012 - Add Utils.throttle method R=steveck

sroucheray:
I implemented Utils.throttle and its unit tests.

### DIFF
--- a/apps/sms/views/conversation/js/conversation.js
+++ b/apps/sms/views/conversation/js/conversation.js
@@ -72,6 +72,12 @@ var ConversationView = {
   // activity after this delay after moving to the thread list.
   LEAVE_ACTIVITY_DELAY: 3000,
 
+  THROTTLE_DELAY: 300,
+  THROTTLE_OPTS: {
+    preventFirstCall: true,
+    preventLastCall: false
+  },
+
   draft: null,
   recipients: null,
   // Set to |true| when in edit mode
@@ -129,15 +135,30 @@ var ConversationView = {
                               this.onVisibilityChange);
 
     this.toField.addEventListener(
-      'keypress', this.toFieldKeypress.bind(this), true
+      'keypress', 
+      Utils.throttle(
+        this.toFieldKeypress.bind(this),
+        this.THROTTLE_DELAY, 
+        this.THROTTLE_OPTS),
+      true
     );
 
     this.toField.addEventListener(
-      'input', this.toFieldInput.bind(this), true
+      'input', 
+      Utils.throttle(
+        this.toFieldInput.bind(this),
+        this.THROTTLE_DELAY, 
+        this.THROTTLE_OPTS),
+      true
     );
 
     this.toField.addEventListener(
-      'focus', this.toFieldInput.bind(this), true
+      'focus', 
+      Utils.throttle(
+        this.toFieldInput.bind(this),
+        this.THROTTLE_DELAY, 
+        this.THROTTLE_OPTS),
+      true
     );
 
     this.sendButton.addEventListener(
--- a/apps/sms/views/conversation/test/unit/conversation_test.js
+++ b/apps/sms/views/conversation/test/unit/conversation_test.js
@@ -371,10 +371,13 @@ suite('conversation.js >', function() {
 
         placeholder.dispatchEvent(new CustomEvent('input', { bubbles: true }));
 
+        // The event is throttled with a delay of 300ms
+        this.sinon.clock.tick(300);
         contactPromise.then(() => done(), done);
       });
 
       test('does display found contacts', function() {
+        sinon.assert.calledOnce(suggestionRenderer.render);
         sinon.assert.calledWithMatch(suggestionRenderer.render, {
           contact: contact,
           input: '999',
@@ -5560,6 +5563,11 @@ suite('conversation.js >', function() {
       });
 
       test('Compose.refresh is called', function() {
+        // The event is throttled with a delay of 300ms 
+        this.sinon.clock.tick(200);
+        sinon.assert.notCalled(Compose.refresh);
+
+        this.sinon.clock.tick(300);
         sinon.assert.calledOnce(Compose.refresh);
       });
     });
--- a/apps/sms/views/shared/js/utils.js
+++ b/apps/sms/views/shared/js/utils.js
@@ -711,6 +711,59 @@
     },
 
     /**
+     * Returns a rate limited function, useful to reduce the number of function
+     * calls when listening to user input (keypress) or browser events.
+     * @params {function} func - The function that should be rate-limited
+     * @params {number=300} delay - Min amount of time (ms) between two calls
+     * @params {Object} options - Parameters for managing edge calls. 
+     * Attention : when both preventFirstCall and preventLastCall are set to 
+     * true, `func` may never be called.
+     * @param {Boolean=false} options.preventFirstCall - Prevent first 
+     * call when true.
+     * @param {Boolean=false} options.preventLastCall - Prevent last trailing
+     * call when true.
+     * @returns {function} - The rate limited function that will call the
+     *  original function
+     */
+    throttle: function(func, delay = 300, options = {}) {
+      var timeout = null;
+      var previous = 0;      
+      var preventFirstCall = !!options.preventFirstCall;
+      var preventLastCall = !!options.preventLastCall;
+
+      if(typeof func !== 'function'){
+        throw new Error('func must be a Function');
+      }
+
+      if(typeof delay !== 'number' || delay < 0){
+        throw new Error('delay must be a positive number');
+      }
+
+      return function(...args) {
+        var now = Date.now();
+        if (!previous && preventFirstCall){
+          previous = now;
+        }
+        var remaining = delay - (now - previous);
+        if (remaining <= 0) {
+          if (timeout) {
+            clearTimeout(timeout);
+            timeout = null;
+          }
+          previous = now;
+          func.apply(this, args);
+        } else if (!timeout && !preventLastCall) {
+          timeout = setTimeout(() => {
+            previous = preventFirstCall ? 0 : Date.now();
+            timeout = null;
+
+            func.apply(this, args);
+          }, remaining);
+        }
+      };
+    },
+
+    /**
      * Shows modal alert dialog with single OK button to dismiss it.
      * @param {string|{ raw: string|Node }|{id: string, args: Object }} message
      * Message displayed in the alert. 1. If "message" is string then it's
--- a/apps/sms/views/shared/test/unit/utils_test.js
+++ b/apps/sms/views/shared/test/unit/utils_test.js
@@ -1153,6 +1153,159 @@ suite('Utils', function() {
     });
   });
 
+  suite('Utils.throttle', function() {
+    setup(function() {
+      // Starting fake timers clock at Unix epoch (timestamp of 0) would 
+      // prevent Utils.throttle to work properly so init it to 1 
+      this.sinon.useFakeTimers(1);
+    });
+
+    test('right typeof parameters', function() {
+        assert.throw(
+          () => Utils.throttle(null), 
+          Error, 
+          'func must be a Function'
+        );
+        
+        assert.throw(
+          () => Utils.throttle(() => {}, null),
+          Error, 
+          'delay must be a positive number'
+        );
+        
+        assert.throw(
+          () => Utils.throttle(() => {}, -1),
+          Error, 
+          'delay must be a positive number'
+        );
+    });
+
+    test('parameters are passed to the throttled func', function() {
+      var dummyObjectA = {};
+      var dummyObjectB = {};
+
+      var func = function() {
+        assert.equal(arguments[0], dummyObjectA);
+        assert.equal(arguments[1], dummyObjectB);
+      };
+
+      var throttlerFunc = Utils.throttle(func, 100);
+      
+      throttlerFunc(dummyObjectA, dummyObjectB);
+    });
+
+    test('testing this in the throttled func', function() {
+      var self = this;
+      var dummyObject = {};
+
+      var throttlerFunc = Utils.throttle(function() {
+        assert.equal(this, dummyObject);
+      }, 100);
+      throttlerFunc.call(dummyObject);
+
+      var throttlerFunc2 = Utils.throttle(()=>{
+        assert.equal(this, self);
+      }, 100);
+      throttlerFunc2.call(dummyObject);
+    });
+
+    test('preventing first call', function() {
+      var delay = 100;
+      var firstCallFuncToExecute = sinon.stub();
+      var firstCallThrottleFuncToExecute = 
+        Utils.throttle(firstCallFuncToExecute, delay);
+      
+      var notFirstCallFuncToExecute = sinon.stub();
+      var notFirstCallThrottleFuncToExecute = 
+        Utils.throttle(
+          notFirstCallFuncToExecute, 
+          delay,
+          {
+            preventFirstCall: true
+          }
+        );
+
+      firstCallThrottleFuncToExecute();
+      this.sinon.clock.tick(delay - 1);
+      sinon.assert.calledOnce(firstCallFuncToExecute);
+
+      notFirstCallThrottleFuncToExecute();
+      this.sinon.clock.tick(delay - 1);
+      sinon.assert.notCalled(notFirstCallFuncToExecute);
+    });
+
+    test('preventing last call', function() {
+      var delay = 100;
+      var lastCallStub = sinon.stub();
+      var notLastCallStub = sinon.stub();
+      var notLastCallFuncToExecute = Utils.throttle(
+        notLastCallStub, 
+        delay, 
+        {
+          preventLastCall: true
+        }
+      );
+      var lastCallFuncToExecute = Utils.throttle(lastCallStub, delay);
+
+      lastCallFuncToExecute();
+      notLastCallFuncToExecute();
+      this.sinon.clock.tick(delay);
+      
+      lastCallFuncToExecute();
+      notLastCallFuncToExecute();
+      this.sinon.clock.tick(delay / 2);
+
+      this.sinon.clock.tick(delay);
+      sinon.assert.calledTwice(lastCallStub);
+      sinon.assert.calledOnce(notLastCallStub);
+    });
+
+    test('multiple calls under delay should limit rate to 1 call', function() {
+      var delay = 100;
+      var stubOnce = sinon.stub();
+      var stubOnceThrottler = Utils.throttle(stubOnce, delay);
+
+      stubOnceThrottler();
+      this.sinon.clock.tick(delay - 1);
+      // Under delay these calls will be filtered 
+      stubOnceThrottler();
+      stubOnceThrottler();
+      stubOnceThrottler();
+      sinon.assert.calledOnce(stubOnce);
+    });
+
+    test('second call after delay should trigger 2 calls', function() {
+      var delay = 100;
+      var stubTwice = sinon.stub();
+      var stubTwiceThrottler = Utils.throttle(stubTwice, delay);
+
+      stubTwiceThrottler();
+      this.sinon.clock.tick(delay);
+      // This one is called after the delay
+      stubTwiceThrottler();
+      this.sinon.clock.tick(delay);
+      // But then these calls are filtered 
+      stubTwiceThrottler();
+      stubTwiceThrottler();
+      sinon.assert.calledTwice(stubTwice);
+    });
+
+    test('3 calls separated by delay should trigger 3 calls', function() {
+      var delay = 200;
+      var stubThrice = sinon.stub();
+      var stubThriceThrottler = Utils.throttle(stubThrice, delay);
+
+      stubThriceThrottler();
+      this.sinon.clock.tick(delay);
+      stubThriceThrottler();
+      this.sinon.clock.tick(delay);
+      stubThriceThrottler();
+      this.sinon.clock.tick(delay);
+      sinon.assert.calledThrice(stubThrice);
+    });
+
+  });
+
   suite('Modal dialogs >', function() {
     var dialogMock;
     setup(function() {
--- a/apps/sms/views/shared/test/unit/mock_utils.js
+++ b/apps/sms/views/shared/test/unit/mock_utils.js
@@ -31,6 +31,7 @@ var MockUtils = {
   closeNotificationsForThread: () => Promise.resolve(),
   imageToDataUrl: Utils.imageToDataUrl,
   debounce: Utils.debounce,
+  throttle: Utils.throttle,
   alert: Utils.alert,
   confirm: Utils.confirm,
   Promise: Utils.Promise,


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=06728cd6e30fbc3ff9ae412511e50188918b0a7d

apps/sms/views/shared/js/utils.js
```diff
@@ -711,6 +711,60 @@
...
+     * calls when listening to user input (keypress) or browser events.
+     * @params {function} func - The function that should be rate-limited
+     * @params {number} delay - Min amount of time (ms) between two calls
+     * @param {Boolean} preventfirstCall - Prevent first synchronous call when
+     * true.
+     * @param {Boolean} preventLastCall - Prevent last trailing call when true.
+     * @returns {function} - The rate limited function that will call the
+     *  original function
+     */
+    throttle: function(func, delay, preventfirstCall, preventLastCall) {
```
steveck-chung: Seems like you can use [default parameter](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/default_parameters) if your delay / preventfirstCall / preventLastCall rarely apply the value other than default 300ms/ true /false

apps/sms/views/shared/js/utils.js
```diff
@@ -711,6 +711,60 @@
...
+        }
+        var remaining = delay - (now - previous);
+        args = arguments;
+        if (remaining <= 0 || remaining > delay) {
+          if (timeout) {
+            clearTimeout(timeout);
+            timeout = null;
+          }
+          previous = now;
+          func.apply(this, args);
```
steveck-chung: You can apply [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_operator) here like `func(...arguments)`

apps/sms/views/shared/js/utils.js
```diff
@@ -711,6 +711,60 @@
...
+      }
+
+      return function() {
+        var now = Date.now();
+        if (!previous && preventfirstCall){
+          previous = now;
+        }
+        var remaining = delay - (now - previous);
+        args = arguments;
+        if (remaining <= 0 || remaining > delay) {
```
steveck-chung: Could you explain when will this `remaining > delay` condition happen ?

apps/sms/views/shared/test/unit/utils_test.js
```diff
@@ -1153,6 +1153,134 @@ suite('Utils', function() {
...
+
+    test('multiple calls under delay should limit rate to one call', function(){
+      var clock = sinon.useFakeTimers(Date.now());
+      var delay = 100;
+      var stubOnce = sinon.stub();
+      var stubOnceThrottler = Utils.throttle(stubOnce, delay);
+
+      stubOnceThrottler();
+      clock.tick(delay - 1);
+      //Under delay these calls will befiltered
```
steveck-chung: typo: `befiltered`

apps/sms/views/shared/test/unit/utils_test.js
```diff
@@ -1153,6 +1153,134 @@ suite('Utils', function() {
     });
   });
 
+  suite('Utils.throttle', function() {
+    setup(function() {
+      //this.sinon.useFakeTimers();
```
steveck-chung: You can remove the whole setup if you don't need it. BTW why not use `this.sinon.useFakeTimers()` here and `this.sinon.clock.tick()` in the test?

apps/sms/views/shared/test/unit/utils_test.js
```diff
@@ -1153,6 +1153,134 @@ suite('Utils', function() {
...
+
+    test('second call after delay should trigger 2 calls', function(){
+      var clock = sinon.useFakeTimers(Date.now());
+      var delay = 100;
+      var stubTwice = sinon.stub();
+      var stubTwiceThrottler = Utils.throttle(stubTwice, delay);
+
+      stubTwiceThrottler();
+      clock.tick(delay);
+      // Over the delay this is call
```
steveck-chung: typo: call => called, or you could comment `This one is called after the delay`

apps/sms/views/shared/js/utils.js
sroucheray: You right I am going to remove this condition.

apps/sms/views/shared/js/utils.js
sroucheray: I could use `func(...arguments)` here but not on line 758 where the apply call is wrapped in a setTimeout + `arguments` is a pseudo array so I would have to use Array.from(arguments) before spreading it.
I tried to use the rest parameter on function in line 737 and then use it along with the spread operator but it seems to break all the unit test (I suspect because it is launch from node).
What do you think ?

apps/sms/views/shared/test/unit/utils_test.js
sroucheray: I don't use `this.sinon.useFakeTimers()` in the setup because it breaks the throttle code in the tests that don't use `this.sinon.clock.tick()` by redefining Date and setTimeout. You are right I am going to remove the all setup function.

apps/sms/views/shared/js/utils.js
sroucheray: My mistake. I was trying the code with both rest parameter and arguments variable was still around. It seems it is not allowed to have both.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7bec2249e0fdb9892b43fdbe9cfc0f2d506e5e02

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a42ab0352aa579403d078095403f56da83190382

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2c168b4d38df9bdfd44b1a4d89d5fa0fb7f90792

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2c168b4d38df9bdfd44b1a4d89d5fa0fb7f90792

apps/sms/views/shared/test/unit/utils_test.js
```diff
@@ -1153,6 +1153,131 @@ suite('Utils', function() {
     });
   });
 
+  suite('Utils.throttle', function() {
+    test('right typeof parameters', function(){
+      assert.throw(() => Utils.throttle(null),
```
steveck-chung: nit: indention and space around the arrow:
```
assert.throw(
  () => Utils.throttle(null), 
  Error, 
  'func must be a Function'
);
```

apps/sms/views/shared/test/unit/utils_test.js
```diff
@@ -1153,6 +1153,131 @@ suite('Utils', function() {
     });
   });
 
+  suite('Utils.throttle', function() {
+    test('right typeof parameters', function(){
+      assert.throw(() => Utils.throttle(null), 
+                   Error, 
+                   'func must be a Function');
+      assert.throw(() => Utils.throttle(()=>{}, null),
```
steveck-chung: nit: indention and space around the arrow:
```
assert.throw(
  () => Utils.throttle(() => {}, null), 
  Error, 
  'delay must be a positive number'
);
```

apps/sms/views/shared/js/utils.js
steveck-chung: Ah, it my fault about the misleading and the pattern you used here is rest parameter. After second thought maybe you can simply use `arguments` directly here.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0d0a7231f39c5e57f5ff992c1cd0720571066a98

apps/sms/views/shared/js/utils.js
```diff
@@ -711,6 +711,57 @@
...
+     * @param {Boolean=false} preventfirstCall - Prevent first synchronous call 
+     * when true.
+     * @param {Boolean=false} preventLastCall - Prevent last trailing call
+     * when true.
+     * @returns {function} - The rate limited function that will call the
+     *  original function
+     */
+    throttle: function(func,
+                       delay = 300, 
+                       preventfirstCall = false,
```
steveck-chung: I was thinking that maybe you can set delay = 300 and preventfirstCall = true to default here and then you don't have to add 300/true as parameter in conversation.js. But if you still want to specify the delay and preventfirstCall in conversations, maybe you can make the delay/preventfirstCall/preventLastCall into an option for readability:
```
const THROTTLE_OPTS = {
  delay: 300,
  preventfirstCall : true,
  ...
}
``` 
And use this default option instead, wdyt?

apps/sms/views/shared/test/unit/utils_test.js
```diff
@@ -1153,6 +1153,131 @@ suite('Utils', function() {
     });
   });
 
+  suite('Utils.throttle', function() {
+    test('right typeof parameters', function(){
```
steveck-chung: nit: add a space before open brace.

apps/sms/views/shared/js/utils.js
sroucheray: I have seen several implementations of throttle and most part of the time preventFirstCall, if present, is opt-in. So I thing it's better to keep its default value to false. Also, I think, while delay as a default of 300ms, it's better to repeat it when used because it increases readability.
I like the idea of default value in a const object because it avoid repeating magic numbers.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2899932f9adcad0726002e9c8877b35873478630

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=be4a3efc028a046c79aaf4d414561c0e1057bece

apps/sms/views/conversation/js/conversation.js
```diff
@@ -94,6 +94,10 @@ var ConversationView = {
 
   multiSimActionButton: null,
   init: function conv_init() {
+    const THROTTLE_OPTS = {
```
steveck-chung: You can put this const at the top of the script, or treat is as conversationView's property like others.

apps/sms/views/shared/js/utils.js
steveck-chung: If you prefer the option here, you can:
```
throttle: function(func, options) {
  var delay = options.delay || 300;
  var preventfirstCall = options.preventfirstCall || false;
  ...
}
```
up to you

apps/sms/views/conversation/js/conversation.js
```diff
@@ -129,15 +133,27 @@ var ConversationView = {
                               this.onVisibilityChange);
 
     this.toField.addEventListener(
-      'keypress', this.toFieldKeypress.bind(this), true
+      'keypress', 
+      Utils.throttle(this.toFieldKeypress.bind(this),
```
steveck-chung: and you can set the option as parameter if you apply https://github.com/mozilla-b2g/gaia/pull/31677/files#r38887892

steveck-chung: And another really minor reminder is the message app's convention: if you want to make the some parameters into another new line, please set all the parameter into new line with 2 space indention.

apps/sms/views/conversation/test/unit/conversation_test.js
```diff
@@ -333,7 +333,9 @@ suite('conversation.js >', function() {
     suite('rendering suggestions list', function() {
       var suggestionRenderer, unknownRenderer;
       var contact, unknown;
+      var clock;
       setup(function(done) {
+        clock = sinon.useFakeTimers(Date.now());
```
steveck-chung: I really don't think you'll need to give the Date.now() here since it's not absolute timestamp sensitive test case. Since the  sinon.useFakeTimers already set at the top of the setup, you can simply use this.sinon.clock.tick(ms) without further setup for your test(and you don't have to restore unless you want to do something specific in the same test)

sroucheray: Good to know. I am going to get rid of Date.now() here.

apps/sms/views/shared/js/utils.js
sroucheray: Having one parameter for all is a better way of handling options but it adds one line per default assignment. This is a trade off. It seems that we can't use yet [Destructuring assignement](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Function_argument_defaults) here because it is only available in 41.
If using options I would prefer to initialize delay this way (0 must be a valid delay but would be treated as false) :
```javascript
var delay = (typeof options.delay !== undefined) ? options.delay : 300;
```

What do you think ?

apps/sms/views/conversation/js/conversation.js
sroucheray: OK, I was not sure it was important enough to be part of the conversationView's property. I'll move it there.

apps/sms/views/shared/js/utils.js
steveck-chung: Thanks for the reminder, we can use destructuring assignement actually since 2.5 must binds with gecko version later then 41(should be 43) so you can apply destructuring assignement here if you want.

apps/sms/views/shared/test/unit/utils_test.js
```diff
@@ -1153,6 +1153,131 @@ suite('Utils', function() {
     });
   });
 
+  suite('Utils.throttle', function() {
+    test('right typeof parameters', function() {
+      assert.throw(() => Utils.throttle(null),
```
steveck-chung: nit: also the minor convention issue: https://github.com/mozilla-b2g/gaia/pull/31677/files#r38888049

apps/sms/views/shared/test/unit/utils_test.js
```diff
@@ -1153,6 +1153,131 @@ suite('Utils', function() {
...
+
+      firstCallThrottleFuncToExecute();
+      sinon.assert.calledOnce(firstCallFuncToExecute);
+
+      notFirstCallThrottleFuncToExecute();
+      sinon.assert.notCalled(notFirstCallFuncToExecute);
+    });
+
+    test('preventing last call', function() {
+      var clock = sinon.useFakeTimers(Date.now());
```
steveck-chung: Same as https://github.com/mozilla-b2g/gaia/pull/31677/files#r38888591 you can setup the useFakeTimers at the top of the Utils.throttle suite and use this.sinon.clock.tick in the test. You can reference the debounce unit test part : https://github.com/mozilla-b2g/gaia/pull/31677/files#diff-52ec1815f0d22fe9a17ef2259891a61dR1127

apps/sms/views/shared/test/unit/utils_test.js
```diff
@@ -1153,6 +1153,131 @@ suite('Utils', function() {
...
+
+    test('preventing first call', function() {
+      var delay = 100;
+      var firstCallFuncToExecute = sinon.stub();
+      var firstCallThrottleFuncToExecute = 
+        Utils.throttle(firstCallFuncToExecute, delay);
+      
+      var notFirstCallFuncToExecute = sinon.stub();
+      var notFirstCallThrottleFuncToExecute = 
+      Utils.throttle(notFirstCallFuncToExecute, delay, true);
```
steveck-chung: nit: indention

apps/sms/views/shared/js/utils.js
sroucheray: Finally I think we can't use destructuring assignement here until bug #932080 is fixed [Support default values in destructuring](https://bugzilla.mozilla.org/show_bug.cgi?id=932080)

apps/sms/views/shared/js/utils.js
```diff
@@ -711,6 +711,57 @@
...
+        if (remaining <= 0) {
+          if (timeout) {
+            clearTimeout(timeout);
+            timeout = null;
+          }
+          previous = now;
+          func.apply(this, args);
+        } else if (!timeout && !preventLastCall) {
+          timeout = setTimeout(() => {
+            previous = preventfirstCall ? 0 : Date.now();
```
steveck-chung: This part looks weird: If `preventfirstCall` is true here, it means we will need to set the previous to the current time to prevent the next closed call triggered, but the condition here seems reserved. Please correct me if I'm wrong.

steveck-chung:
It's kind of silly question, but is it possible that user set both preventFirstCall/preventLastCall to true at thte same time? It seems the call won't be triggered if we called only once in this case. Maybe we should prevent someone set both flag to true?

apps/sms/views/shared/js/utils.js
sroucheray: Sorry but by "reserved", do you mean that the line 742 condition is not reached ?

sroucheray:
If preventFirstCall and preventLastCall are both set to true and we called only once, the call won't be triggered. You are right, but is it not the desired behaviour ? It is an edge case and we might prevent it from happening or just document it. There might be cases where we want this to happen. So I would document it in the function header and let developers take responsability for it.
What do you think ?

apps/sms/views/conversation/test/unit/conversation_test.js
sroucheray: Actually I need to remove all the `clock = sinon.useFakeTimers(Date.now());` as you suggested :)

apps/sms/views/shared/test/unit/utils_test.js
sroucheray: Actually 4 on 8 of the tests do not require fakeTimer to pass so I wondered if that was not a better idea to use fakeTimer only where it's needed and use real (none faked) Date and setTimeout objects when possible (which is closer to real usage).
Tell me if you don't agree and I'll move fakeTimer to the suite setup !

apps/sms/views/shared/test/unit/utils_test.js
sroucheray: Should we enforce spaces around arrow in `.eslintrc` with the [arrow-spacing](http://eslint.org/docs/rules/arrow-spacing) rule ?

apps/sms/views/shared/js/utils.js
sroucheray: I was thinking about this and finally made a change in the signature with 3 parameters. Two of them are not "options" (you need to think about their values to make it works) : func and delay, the last one group the prevent* options in one object (those have good defaults and are useful to tweak edge cases).

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5dcd881cf81b9ae5f502e2dc9136545f1a60bb8b

apps/sms/views/shared/js/utils.js
```diff
@@ -711,6 +711,62 @@
...
+     * @param {Boolean=false} options.preventLastCall - Prevent last trailing
+     * call when true.
+     * @returns {function} - The rate limited function that will call the
+     *  original function
+     */
+    throttle: function(func, delay = 300, options = {}) {
+      var timeout = null;
+      var previous = 0;
+      
+      options = options || {};
```
steveck-chung: I don't think it's necessary? Since default options is {} already.

apps/sms/views/conversation/test/unit/conversation_test.js
```diff
@@ -5532,8 +5535,14 @@ suite('conversation.js >', function() {
...
 
       test('Compose.refresh is called', function() {
+        // The event is throttled with a delay of 300ms 
+        this.sinon.clock.tick(200);
+        sinon.assert.notCalled(Compose.refresh);
+
+        this.sinon.clock.tick(300);
         sinon.assert.calledOnce(Compose.refresh);
       });
+
```
steveck-chung: nit: No need for new line here.

apps/sms/views/shared/test/unit/utils_test.js
```diff
@@ -1153,6 +1153,157 @@ suite('Utils', function() {
     });
   });
 
+  suite('Utils.throttle', function() {
+    setup(function() {
+      this.sinon.useFakeTimers(1);
+    });
+
+    test(
```
steveck-chung: nit: For the unit test here, we don't split the description and function into 2 different unless the description is too long. So you can simply `test('description', function() {` Sorry for about  confusing :/

apps/sms/views/shared/test/unit/utils_test.js
```diff
@@ -1153,6 +1153,157 @@ suite('Utils', function() {
...
+      'right typeof parameters', 
+      function() {
+        assert.throw(
+          () => Utils.throttle(null), 
+          Error, 
+          'func must be a Function');
```
steveck-chung: nit: please put the close parenthesis to new line just like the close bracket here and below.

apps/sms/views/shared/js/utils.js
```diff
@@ -711,6 +711,62 @@
...
+        if (remaining <= 0) {
+          if (timeout) {
+            clearTimeout(timeout);
+            timeout = null;
+          }
+          previous = now;
+          func.apply(this, args);
+        } else if (!timeout && !preventLastCall) {
+          timeout = setTimeout(() => {
+            previous = preventFirstCall ? 0 : Date.now();
```
steveck-chung: Sorry about the previous confusing comment :/ I was thinking why we need to reset previous to 0 if preventFirstCall is true. But now I think we can simply reset the previous for all the cases, wdyt?

apps/sms/views/shared/test/unit/utils_test.js
```diff
@@ -1153,6 +1153,157 @@ suite('Utils', function() {
...
+      var stubThriceThrottler = Utils.throttle(stubThrice, delay);
+
+      stubThriceThrottler();
+      this.sinon.clock.tick(delay);
+      stubThriceThrottler();
+      this.sinon.clock.tick(delay);
+      stubThriceThrottler();
+      this.sinon.clock.tick(delay);
+      sinon.assert.calledThrice(stubThrice);
+      this.sinon.clock.restore();
```
steveck-chung: Seems no need to restore here.

apps/sms/views/shared/test/unit/utils_test.js
```diff
@@ -1153,6 +1153,157 @@ suite('Utils', function() {
     });
   });
 
+  suite('Utils.throttle', function() {
+    setup(function() {
+      this.sinon.useFakeTimers(1);
```
steveck-chung: Not sure why you give 1 as parameter here, is it necessary for your test? It seems not necessary to set the specific timestamp for your test.

apps/sms/views/shared/js/utils.js
```diff
@@ -711,6 +711,62 @@
     },
 
     /**
+     * Returns a rate limited function, useful to reduce the number of function
+     * calls when listening to user input (keypress) or browser events.
```
steveck-chung: nit: About the edge case that preventFisrCall/preventLastCall both be true. I think it's enough to add some comments here. But it's very optional :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e5e345d4981b9d14e0f6388e66d2bb2dcb50d8b8

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=13fdbeaeb54eecc67a523bf9933ce4bc61b4276f

apps/sms/views/shared/test/unit/utils_test.js
sroucheray: It is necessary because by default sinon init its clock to Unix epoch (timestamp of 0) which prevent the function to work as expected.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7db3e8b665fad1ce47682d2febfaa511967d9f42

PR #31679, opened 2015-09-03
Bug 1201567 - [Settings] Convert SimPin panel switch to use gaia-switch

KevinGrandon:
https://bugzilla.mozilla.org/show_bug.cgi?id=1201567

### DIFF
--- a/apps/settings/js/panels/simpin/simpin.js
+++ b/apps/settings/js/panels/simpin/simpin.js
@@ -20,6 +20,8 @@ define(function(require) {
         this.isAirplaneMode = (AirplaneModeHelper.getStatus() === 'enabled');
 
         this._elements.simPinContainer.addEventListener('click', this);
+        this._elements.simPinContainer.addEventListener('change',
+          this.simPinSwitched.bind(this));
         this.addIccDetectedEvent();
         this.addIccUndetectedEvent();
         this.addAirplaneModeChangeEvent();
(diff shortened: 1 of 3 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c598bb2f9587349199b68d4b94b136ae6a1b8d94

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4bd79f5e43f65ec5eaed5fa94bb2aae0fee47c01

apps/settings/js/panels/simpin/simpin.js
```diff
@@ -126,15 +127,23 @@ define(function(require) {
       cardIndex = parseInt(cardIndex, 10);
 
       switch (type) {
-        case 'checkSimPin':
```
KevinGrandon: In the previous PR, you asked:
> 

why not handle with this._elements.simPinContainer.addEventListener('change', this) ? Therefore we can keep this switch condition and remove simPinSwitched method.

The reason is due to the way that gaia-switch fires both click and change events. We can handle it inside of handle event, but then need to invert the clicked state, and that feels wrong to me. (Changed doesn't fire until after the click set is different). I think having a separate method is cleaner for now, I've added a comment.

apps/settings/js/panels/simpin/simpin.js
```diff
@@ -34,12 +36,11 @@ define(function(require) {
...
         <li class="simpin-enabled simpin-enabled-${simIndex}
           simpin-${simIndex}">
-          <label class="pack-switch">
-            <input type="checkbox" data-ignore data-sim-index="${simIndex}"
-              data-type="checkSimPin"/>
-            <span data-l10n-id="simPinWithIndex"
-              data-l10n-args="${simPinArgs}"></span>
-        </label>
+          <gaia-switch data-ignore data-sim-index="${simIndex}"
+              data-type="checkSimPin">
```
gasolin: therefore we don't need `data-type="checkSimPin"` here sinice we don't go through handleEvent

apps/settings/js/panels/simpin/simpin.js
gasolin: ooh I did not notice the difference between 'change' and `click` event, thanks for update.

apps/settings/js/panels/simpin/simpin.js
KevinGrandon: Good call, will address. Thanks!

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=054de1d64925da5da188a115cb1ffde6ead8563d

PR #31680, opened 2015-09-03
Bug 1155099 - Make 'MHz' localizable.

hfiguiere:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a155d9fc8fc257deb6648330693c93fdd69d7381

apps/fm/js/fm.js
```diff
@@ -619,12 +622,16 @@ var favoritesList = {
...
+
+    var subElem = document.createElement('div');
+    subElem.className = 'fav-list-frequency';
+    subElem.textContent = item.frequency.toFixed(1);
+    subElem.dataset.unit = mhzString;
+    elem.appendChild(subElem);
+
+    subElem = document.createElement('div');
+    subElem.className = 'fav-list-remove-button';
+    elem.appendChild(subElem);
```
justindarc: nit: This might be a little easier to read as an ES6 template string

apps/fm/js/fm.js
```diff
@@ -883,6 +890,12 @@ document.addEventListener('DOMLocalized', function onDOMLocalized(e) {
   window.performance.mark('navigationLoaded');
   document.removeEventListener('DOMLocalized', onDOMLocalized);
 
+  // localize MHz
+  document.l10n.format('mhz').then(str => {
```
justindarc: Will this automatically re-localize if the locale is changed after the app is launched?

apps/fm/js/fm.js
justindarc: I'm starting to think we should maybe just add a `<span data-l10n-id="mhz"></span>`

hfiguiere: It is still better than innerHTML :-/

As for adding the span, I thought about it but I'm not sure what the intent was originally to use the :after pseudo-element.

apps/fm/js/fm.js
hfiguiere: Good question. I'm tempte to say I don't think. Hence asking feedback from the localization team.

apps/fm/js/fm.js
anaran: I think a :after pseudo-element would be bad because that is intangible (cannot be selected and copied as text IIRC).

hfiguiere: The span would complexifies a lot the layout.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=13e3d0b6423089b04ade6c623588ac379b1cee06

hfiguiere:
With this the localization is now updated.

apps/fm/js/fm.js
```diff
@@ -619,12 +622,16 @@ var favoritesList = {
...
-    html += '<div class="fav-list-frequency">';
-    html += item.frequency.toFixed(1);
-    html += '</div>';
-    html += '<div class="fav-list-remove-button"></div>';
-    elem.innerHTML = html;
+
+    var subElem = document.createElement('div');
+    subElem.className = 'fav-list-frequency';
+    subElem.textContent = item.frequency.toFixed(1);
+    subElem.dataset.unit = mhzString;
```
zbraniecki: so, if you don't want to show "Mhz-NOTLOCALIZED" in the UI, and you are adding it in DOMLocalized handler anyway, why do you do this here as well?

apps/fm/js/fm.js
```diff
@@ -158,6 +158,9 @@ var mozFMRadio = navigator.mozFM || navigator.mozFMRadio || {
   aGlobal.SpeakerManager = SpeakerManager;
 })(window);
 
+// the mhzString that will be localized.
+var mhzString = 'MHz-NOTLOCALIZED';
```
zbraniecki: Why do you at all initialize this string here? It seems that it should not be used outside of the `formatValue` call in the `DOMLocalized` handler?

apps/fm/js/fm.js
justindarc: It increases visibility to us when a translation is missing.

zbraniecki: When translation is missing we will display string ID (`mhz` in this case), so the only scenario in which it will be displayed is when DOMLocalized is never fired.

One more thing to consider is that means that we always switch from "Mhz-NOTLOCALIZED" to localized Mhz string which may trigger reflow, but I guess that it's not a significant performance cost here at all.

Anyway, I'm not strongly opinionated here :)

hfiguiere: I can change the string. I'll post a followup.

As for creating here, it is because it is called when you create a bookmark. So it gets called in two cases:
1. when the app starts to show the existing bookmarks
2. when the use create a bookmark.
(here bookmark is a radio frequency you want to keep)

In the DOMLocalized event it just relocalize the one currently displayed.

zbraniecki: but in both cases it should be localized first, right?

hfiguiere: Should be. If this is called before DOMLocalized is handled, then it will be localized in the handler.

zbraniecki: ok. It just seems like there is a risk of a FOUC (Flash of Untranslated Content) if we inject the nodes before DOMLocalized and then we update them on DOMLocalized, but if you don't see it, let's leave it the way it is :)

PR #31698, opened 2015-09-04
Bug 1182522 - [Aries] Wait(self.marionette).until(lambda m: grippy.lo…

JohanLorenzo:
…cation['y'] == utility_tray.size['height']) is never satisfied

### DIFF
--- a/tests/python/gaia-ui-tests/gaiatest/apps/system/regions/utility_tray.py
+++ b/tests/python/gaia-ui-tests/gaiatest/apps/system/regions/utility_tray.py
@@ -41,6 +41,10 @@ class UtilityTray(Base):
     def a11y_clear_all_notifications(self):
         self.accessibility.click(self.marionette.find_element(*self._notification_clear_locator))
 
+    @property
+    def cost_control_widget(self):
+        return CostControlWidget(self.marionette)
+
     def tap_settings_button(self):
         self.marionette.find_element(*self._quicksettings_app_locator).tap()
 
@@ -71,3 +75,20 @@ class Notification(PageRegion):
         self.root_element.find_element(*self._title_locator).tap()
         from gaiatest.apps.email.regions.read_email import ReadEmail
         return ReadEmail(self.marionette)
+
+
+class CostControlWidget(Base):
+    _cost_control_widget_locator = (By.CSS_SELECTOR, '#cost-control-widget > iframe')
+    _data_usage_view_locator = (By.ID, 'datausage-limit-view')
+
+    def __init__(self, marionette):
+        Base.__init__(self, marionette)
+        usage_iframe = self.marionette.find_element(*self._cost_control_widget_locator)
+        self.marionette.switch_to_frame(usage_iframe)
+
+    def tap(self):
+        self.marionette.find_element(*self._data_usage_view_locator).tap()
+
+    def wait_for_limit_to_be_reached(self):
+        usage_view = self.marionette.find_element(*self._data_usage_view_locator)
+        Wait(self.marionette, timeout=40).until(lambda m: 'reached-limit' in usage_view.get_attribute('class'))
--- a/tests/python/gaia-ui-tests/gaiatest/tests/functional/cost_control/test_cost_control_data_alert_mobile.py
+++ b/tests/python/gaia-ui-tests/gaiatest/tests/functional/cost_control/test_cost_control_data_alert_mobile.py
@@ -2,7 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from marionette_driver import By, Wait
+from marionette_driver import Wait
 
 from gaiatest import GaiaTestCase
 from gaiatest.apps.search.app import Search
@@ -11,10 +11,6 @@ from gaiatest.apps.system.app import System
 
 
 class TestCostControlDataAlertMobile(GaiaTestCase):
-
-    # notification bar locators
-    _cost_control_widget_locator = (By.CSS_SELECTOR, '#cost-control-widget > iframe')
-    _data_usage_view_locator = (By.ID, 'datausage-limit-view')
 
     def setUp(self):
         GaiaTestCase.setUp(self)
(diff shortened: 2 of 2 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0efe52b578ea4059b5caf06c846dc86afb275b28

tests/python/gaia-ui-tests/gaiatest/visual_components/visual_component.py
```diff
@@ -0,0 +1,7 @@
+from gaiatest.apps.base import PageRegion
+
+
+class VisualComponent(PageRegion):
```
npark-mozilla: So is there a chance for a frame switching to occur within this class?  If so we might run into the problem that we discussed.  (looks like it doesn't happen here though)

tests/python/gaia-ui-tests/gaiatest/visual_components/switch.py
```diff
@@ -0,0 +1,18 @@
+from marionette_driver import Wait
+from gaiatest.visual_components.visual_component import VisualComponent
+
+
+class Switch(VisualComponent):
```
npark-mozilla: Would this support both switch and gaia-switch?  I think gaia-switch uses slightly different method to check its state.  Not sure whether youre is_checked() would work in case of gaia-switch

tests/python/gaia-ui-tests/gaiatest/visual_components/visual_component.py
JohanLorenzo: Not in this case. If that happens, we'll have to change PageRegion a bit more

tests/python/gaia-ui-tests/gaiatest/visual_components/switch.py
JohanLorenzo: Currently, it doesn't. We can do it this way, though:

```python
class Switch(VisualComponent):
    @abstractmethod
    def is_checked(self):
        pass


class GaiaSwitch(Switch):
    def is_checked(self):
        return self.marionette.execute_script("return arguments[0].wrappedJSObject.checked", [self.root_element])

class HtmlSwitch(Switch)
    def is_checked(self):
        return # TBD
```

tests/python/gaia-ui-tests/gaiatest/visual_components/visual_component.py
npark-mozilla: this would be great, btw, is there a reason to use js script rather than something like self.is_custom_element_checked(switch) (in gaia-switch) or self.is_element_checked(switch)?

tests/python/gaia-ui-tests/gaiatest/visual_components/switch.py
npark-mozilla: this would be great, btw, is there a reason to use js script rather than something like self.is_custom_element_checked(switch) (in gaia-switch) or self.is_element_checked(switch)?

JohanLorenzo: The only reason is, if we decide to keep it, this will replace `is_custom_element_checked()` ;)

npark-mozilla: ah, awesome! yeah, we might as well support both, although in current PR you are not actually using gaia-switch, it would be easier for the next person to apply it when gaia-switch needs to be triggered.

tests/python/gaia-ui-tests/gaiatest/apps/system/regions/utility_tray.py
```diff
@@ -71,3 +75,20 @@ def tap_notification(self):
         self.root_element.find_element(*self._title_locator).tap()
         from gaiatest.apps.email.regions.read_email import ReadEmail
         return ReadEmail(self.marionette)
+
```
mwargers: 2 newlines.

tests/python/gaia-ui-tests/gaiatest/visual_components/visual_component.py
mwargers: Wat is a VisualComponent? Isn't everything a VisualComponent? I mean, can we have name that better describes what it is? To me, almost everything can be a visual component.

tests/python/gaia-ui-tests/gaiatest/visual_components/switch.py
mwargers: So Switch/HtmlSwitch is the input switch and GaiaSwitch is gaia-switch?
This isn't part of this pull request, right? Could you file a new bug on this?

mwargers: Afaik, the plan is to replace HTMLSwitch with GaiaSwitch completely.

tests/python/gaia-ui-tests/gaiatest/visual_components/switch.py
```diff
@@ -0,0 +1,18 @@
...
+        self._toggle_and_verify_state(final_state=True)
+
+    def disable(self):
+        self._toggle_and_verify_state(final_state=False)
+
+    @property
+    def is_checked(self):
+        return self.marionette.execute_script("return arguments[0].wrappedJSObject.checked", [self.root_element])
+
+    def _toggle_and_verify_state(self, final_state):
```
mwargers: I would prefer also a check here that the begin state is not already equal to the final_state, so I guess: Wait(self.marionette).until(lambda m: self.is_checked is not final_state)

tests/python/gaia-ui-tests/gaiatest/visual_components/visual_component.py
mwargers: Although I have to say, I don't really know a better name. Perhaps FormControl or something?

tests/python/gaia-ui-tests/gaiatest/tests/functional/cost_control/test_cost_control_data_alert_mobile.py
```diff
@@ -46,29 +41,19 @@ def test_cost_control_data_alert_mobile(self):
...
-        browser.wait_for_page_to_load(180)
-
-        browser.switch_to_content()
-        Wait(self.marionette, timeout=60).until(lambda m: "sample_png_02.png" in m.title)
-        browser.switch_to_chrome()
+        browser.wait_for_page_to_load(timeout=180)
+        browser.wait_for_title_to_load(title='sample_png_02.png')
 
         system = System(self.marionette)
+        system.wait_for_notification_toaster_displayed()
```
mwargers: Isn't the risk here that the notification toaster has already been displayed while the browser is busy loading (during wait_for_page_load)?

tests/python/gaia-ui-tests/gaiatest/tests/functional/cost_control/test_cost_control_data_alert_mobile.py
```diff
@@ -46,29 +41,19 @@ def test_cost_control_data_alert_mobile(self):
...
-        # The timeout is increased, because for some reason, it takes some time
-        # before the limit view is shown (the browser has to finish loading?)
-        usage_view = self.marionette.find_element(*self._data_usage_view_locator)
-        Wait(self.marionette, timeout=40).until(lambda m: 'reached-limit' in usage_view.get_attribute('class'),
-             message='Data usage bar did not breach limit')
-        usage_view.tap()
-
-        self.wait_for_condition(lambda m: self.apps.displayed_app.name == cost_control.name)
+        cost_control_widget = utility_tray.cost_control_widget
+        cost_control_widget.wait_for_limit_to_be_reached()
```
mwargers: It is strange that we have to wait for the limit to be reached here, when the notification toaster about 'limit reached' has already been shown.
But it's better to be safe. Iirc, this test is really fragile.

tests/python/gaia-ui-tests/gaiatest/apps/system/regions/utility_tray.py
JohanLorenzo: That's expected. You put 2 new lines to separates 2 external classes

tests/python/gaia-ui-tests/gaiatest/tests/functional/cost_control/test_cost_control_data_alert_mobile.py
JohanLorenzo: That's a good point. We can remove these lines then:
```python
       browser.wait_for_page_to_load(timeout=180)
       browser.wait_for_title_to_load(title='sample_png_02.png')
```

tests/python/gaia-ui-tests/gaiatest/tests/functional/cost_control/test_cost_control_data_alert_mobile.py
JohanLorenzo: The update of the widget is not done directly (especially on Flame).

tests/python/gaia-ui-tests/gaiatest/visual_components/switch.py
JohanLorenzo: You're right. I'll do it

tests/python/gaia-ui-tests/gaiatest/visual_components/visual_component.py
JohanLorenzo: That's also a good point. At the moment, let's focus on the switches, the checkboxes, the text input, which are all FormControl. I'll use this class name, then

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=97127810c6eaff830b3030acd2adf20d70a6cf18

tests/python/gaia-ui-tests/gaiatest/apps/system/regions/utility_tray.py
mwargers: Oops, you're right.

PR #31699, opened 2015-09-04
Bug 1201951 - upgraded slugid/taskcluster-client

petemoore:
This is to switch to using the new `slugid.nice()` function instead of the `slugid.v4()` function. The `nice()` function guarantees that slugs start with `[A-Za-f]` which is safer, especially as it excludes slugids starting with a `-` which can cause problems when used as command line parameters.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9c7b4998ad433cd4066576c9f500ca2185913643

petemoore:
See https://bugzil.la/1194767 for details about the problem this PR solves. Essentially we are upgrading all systems interfacing with taskcluster (not just gaia) to use the new updated taskcluster libraries. See [slugid npm page](https://www.npmjs.com/package/slugid) and [taskcluster-client npm page](https://www.npmjs.com/package/taskcluster-client#generating-slugids) for more information about the switch from using arbitrary v4 slugs to "nice" v4 slugs.

tests/taskcluster/bin/graph
```diff
@@ -76,7 +76,7 @@ var currentRunId = process.env.RUN_ID;
 //////////////////////
 // Boostrap the graph
 //////////////////////
-taskIdMap['bootstrap'] = slugid.v4();
+taskIdMap['bootstrap'] = slugid.nice();
```
jonasfj: Could also just use: `taskcluster.slugid()`
Then remove `slugid` as a dependency from `package.json`

nullaus: If we can get rid of the dependency for slugid for gaia and instead have that be provided by taskcluster client that would certainly be nice but this also works as is so I will leave it up to the implementer.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=86708c42e9df59fb68616ae4ac73e60998a59cf5

petemoore:
Encorporated feedback... Please review again. =)

jonasfj:
nice, I think this is good...

PR #31702, opened 2015-09-04
Bug 1196239 - SyncEngine, r=ferjm

michielbdejong:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=199f4b26718373d20254239f64dd7ddf8c243549

apps/sync/doc/example-adapter.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+/* exported ExampleAdapter */
+
+var ExampleAdapter = {
+  update: function(kintoCollection) {
+    kintoCollection.list().then(list => {
+      for(var i=0; i<list.data.length; i++) {
```
ferjm: I know this is an example only, but let's keep the same coding style everywhere:
`for (var i = 0; i = list.data.length; i++)`

apps/sync/js/sync-engine/syncengine.js
```diff
@@ -0,0 +1,274 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
ferjm: It would be great if you could add some comments about the overall sync process here.

apps/sync/js/sync-engine/syncengine.js
```diff
@@ -0,0 +1,274 @@
...
+var SyncEngine = (function() {
+  var FxSyncIdSchema = Kinto.createIdSchema({
+    constructor: function(collectionName) {
+      this.collectionName = collectionName;
+    },
+    generate: function() {
+      var bytes = new Uint8Array(9);
+      crypto.getRandomValues(bytes);
+      var binStr = '';
+      for (var i=0; i<9; i++) {
```
ferjm: nit: spaces

`for (var i = 0; i < 9; i++)`

apps/sync/js/sync-engine/syncengine.js
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    this._kB = options.kB;
+    this._collections = {};
+    this._controlCollections = {};
+    this._adapters = {};
+    this._fswc = new FxSyncWebCrypto();
+    this._initKinto({
+       URL: options.URL,
+       assertion: options.assertion,
```
ferjm: Assertions can expire. So we need a way to refresh this value.

apps/sync/js/sync-engine/syncengine.js
```diff
@@ -0,0 +1,274 @@
...
+       xClientState: options.xClientState
+    });
+  };
+
+  SyncEngine.prototype = {
+    _initKinto: function(kintoCredentials) {
+      this._kinto = new Kinto({
```
ferjm: We need to take care of the multiple users case.

If I am not wrong, Kinto creates an IndexedDB under the hood where it stores all its metadata and collections related information. On Firefox OS, the user can log in the device with multiple Firefox Accounts. So if the user logs in with an account `newuser@domain.org`, we need to be sure that we don't keep using the data locally stored for `olduser@domain.org`

apps/sync/js/sync-engine/syncengine.js
```diff
@@ -0,0 +1,274 @@
...
+        bucket: 'syncto',
+        remote: kintoCredentials.URL,
+        headers: {
+          'Authorization': 'BrowserID ' + kintoCredentials.assertion,
```
ferjm: Same thing about expired assertions.

apps/sync/js/sync-engine/syncengine.js
```diff
@@ -0,0 +1,274 @@
...
+        this._controlCollections[collectionName] =
+            this._kinto.collection(collectionName);
+        this._controlCollections[collectionName].use(
+            new ControlCollectionIdSchema(collectionName, keyName));
+      };
+      addControlCollection('meta', 'global');
+      addControlCollection('crypto', 'keys');
+    },
+
+    _getCollection: function(collectionName, isControlCollection=false) {
```
ferjm: nit: spaces around `=`. Here and in the rest of the patch, please.

apps/sync/js/sync-engine/syncengine.js
```diff
@@ -0,0 +1,274 @@
...
+      .then((cryptoKeysRecord) => {
+        var cryptoKeys;
+        try {
+          cryptoKeys = JSON.parse(cryptoKeysRecord.data.payload);
+        } catch (e) {
+          return Promise.reject(new SyncEngine.UnrecoverableError(`Could not pa\
+rse crypto/keys payload as JSON`));
+        }
+        return cryptoKeys;
+      }).then((cryptoKeys) => {
```
ferjm: Follow the same pattern for Promises, please.

    Promise.then(() => {
    }).then(() => {
    });

or

    Promise
    .then(() => {
    })
    .then(() => {
    })

but not a mixture of them.

On gaia I've seen the former most of the times. But up to you :)

apps/sync/js/sync-engine/syncengine.js
```diff
@@ -0,0 +1,274 @@
...
+    },
+
+    _updateCollection: function(collectionName) {
+      return this._syncCollection(collectionName)
+      .then(() => this._adapters[collectionName].update(
+          this._collections[collectionName]))
+      .then(() => this._syncCollection(collectionName));
+    },
+
+    syncNow: function() {
```
ferjm: Note that we won't always be synchronizing all collections. The user can choose not to sync some specific collections via Settings app.

apps/sync/doc/example-adapter.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+/* exported ExampleAdapter */
+
+var ExampleAdapter = {
+  update: function(kintoCollection) {
```
ferjm: What is update supposed to do?

apps/sync/js/sync-engine/syncengine.js
ferjm: Oh well, unless we create a SyncEngine instance every time we want to do a sync.

apps/sync/js/sync-engine/syncengine.js
```diff
@@ -0,0 +1,274 @@
...
+        try {
+          cryptoKeys = JSON.parse(cryptoKeysRecord.data.payload);
+        } catch (e) {
+          return Promise.reject(new SyncEngine.UnrecoverableError(`Could not pa\
+rse crypto/keys payload as JSON`));
+        }
+        return cryptoKeys;
+      }).then((cryptoKeys) => {
+        //Cannot do this from the constructor because we need cryptoKeys first,
+        //and it's not nice to return a promise from a constructor:
```
ferjm: That's true. But we can queue sync requests until we have the crypto keys :)
I'd prefer to avoid the need to do `.connect` if that's possible.

apps/sync/js/sync-engine/syncengine.js
```diff
@@ -0,0 +1,274 @@
...
+        return this._initFxSyncWebCrypto(cryptoKeys);
+      });
+    },
+
+    registerAdapter: function(collectionName, adapter) {
```
ferjm: What's the use case for registering an adapter dynamically? We can leave the option to do it this way, but it would be great if we could allow adapters registration via SyncEngine constructor.

Allowing adapters to be registered as constructor args and queuing sync requests until the crypto keys are retrieved, we could have an API as simple as:

    let syncEngine = new SyncEngine(options);
    syncEngine.syncNow(collections);

apps/sync/js/sync-engine/syncengine.js
ferjm: Also, can't we infer if it's a control collection instead of depending on this flag?

apps/sync/js/sync-engine/syncengine.js
```diff
@@ -0,0 +1,274 @@
...
+      // http://kintojs.readthedocs.org \
+      //     /en/latest/api/#fetching-and-publishing-changes
+
+      return collection.sync()
+      .catch(err => {
+        throw err;
+      }).then(syncResults => {
+        if (syncResults.ok) {
+          return syncResults;
+        } else {
```
ferjm: No need for `else` you are already early returning in the previous line

apps/sync/js/sync-engine/syncengine.js
```diff
@@ -0,0 +1,274 @@
...
+      }).catch(err => {
+        if (err instanceof TypeError) {
+          throw new SyncEngine.UnrecoverableError();
+        } else if (err instanceof Error && typeof err.request === 'object') {
+          if (err.request.status === 401) {
+            throw new SyncEngine.AuthError();
+          } else {
+            throw new SyncEngine.TryLaterError();
+          }
+        } else {
```
ferjm: No need for `else`

apps/sync/js/sync-engine/syncengine.js
```diff
@@ -0,0 +1,274 @@
...
+          payloadObj.storageVersion === 5);
+    },
+
+    _initFxSyncWebCrypto: function(cryptoKeys) {
+      this._fswc = new FxSyncWebCrypto();
+      return this._fswc.setKeys(this._kB, cryptoKeys).then(() => {}, (err) => {
+        if (err === 'SyncKeys hmac could not be verified with current main ' +
+            'key') {
+          throw new SyncEngine.UnrecoverableError();
+        } else {
```
ferjm: ditto

apps/sync/js/sync-engine/syncengine.js
```diff
@@ -0,0 +1,274 @@
...
+      } catch(e) {
+        return false;
+      }
+      return (typeof payloadObj === 'object' &&
+          payloadObj.storageVersion === 5);
+    },
+
+    _initFxSyncWebCrypto: function(cryptoKeys) {
+      this._fswc = new FxSyncWebCrypto();
+      return this._fswc.setKeys(this._kB, cryptoKeys).then(() => {}, (err) => {
```
ferjm: nit: s/(err)/err/

The same applies for the whole patch

ferjm: You can remove `.then` and simply use `.catch`

apps/sync/js/sync-engine/syncengine.js
```diff
@@ -0,0 +1,274 @@
...
+    }
+    if (typeof options.kB !== 'string') {
+      throw new Error('options.kB should be a String');
+    }
+
+    this._kB = options.kB;
+    this._collections = {};
+    this._controlCollections = {};
+    this._adapters = {};
+    this._fswc = new FxSyncWebCrypto();
```
ferjm: Add `this._kinto = null;`, please. This way we won't dynamically modify the object.

apps/sync/js/sync-engine/syncengine.js
michielbdejong: Yes, as discussed, I opened https://bugzilla.mozilla.org/show_bug.cgi?id=1202627 about this as a follow-up bug.

apps/sync/js/sync-engine/syncengine.js
michielbdejong: Yes, kinto.js does not provide a way to update the options of an existing Kinto object, so it will be necessary to construct a new Kinto object. The simplest way to do this is to construct a whole new SyncEngine object. So in the [dbPrefix follow-up bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1202627#c2), I'll optimize the SyncEngine constructor so that it doesn't fetch meta/global and crypto/keys if a local copy of them are already available for the current user.

apps/sync/js/sync-engine/syncengine.js
michielbdejong: If the assertion has expired, a 401 http response will come from the Syncto server, which will result in `SyncEngine#syncNow` rejecting its promise with a `SyncEngine.AuthError`. So then the sync app knows it needs to get a new assertion from SyncCredentials (not part of SyncEngine and not implemented yet).

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b24b69bcbfe710fc0d2b7c878599e8bc9cbb9184

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f0f9a46eb21001dd4e459809735c7ca2d316f8c7

apps/sync/js/sync-engine/syncengine.js
```diff
@@ -0,0 +1,358 @@
...
+ * - Creating kinto.js collections for the 'meta' and 'crypto' collections, with
+ *   the appropriate IdSchemas (the 'meta' collection only contains one record,
+ *   called 'global', and the 'crypto' collection only contains one record,
+ *   called 'keys').
+ * - Setting up the FxSyncWebCrypto instance to be used for cryptographically
+ *   encoding / decoding records, before uploading / after downloading them.
+ * - Registering the DataAdapters to be used on each collection (see below).
+ *
+ * The `syncNow` method takes care of:
+ * - If called for the first time, or if this failed the previous time:
```
ferjm: or if the user changed

apps/sync/js/sync-engine/syncengine.js
```diff
@@ -0,0 +1,358 @@
...
+      });
+    }
+  });
+
+  /**
+    * SyncEngine - Constructor.
+    * @param {Object} options Should contain the following fields:
+    *                         * URL - e.g. 'http://localhost:8000/v1/'
+    *                         * assertion - a BrowserID assertion (Base64)
+    *                         * xClientState - xClientState header value
```
ferjm: If I am not wrong, we cannot generate xClientState on the System app unless we use some of the crypto tools that the Sync app has. So you'll need to generate it here.

apps/sync/js/sync-engine/syncengine.js
```diff
@@ -0,0 +1,358 @@
...
+    },
+    generate: function() {
+      return this.keyName;
+    },
+    validate: function(id) {
+      return (id === this.keyName);
+    }
+  });
+
+  var WebCryptoTransformer = Kinto.createRemoteTransformer({
```
ferjm: I wonder if we need to create this even if there are no changes locally or in the server.

ferjm: Do we need to do this even if there are no changes locally or remotely?

apps/sync/js/sync-engine/syncengine.js
```diff
@@ -0,0 +1,358 @@
...
+  SyncEngine.UnrecoverableError.prototype = Object.create(Error.prototype);
+
+  SyncEngine.TryLaterError = function() {
+    this.message = 'try later';
+  };
+  SyncEngine.TryLaterError.prototype = Object.create(Error.prototype);
+
+  SyncEngine.AuthError = function() {
+    this.message = 'unauthorized';
+  };
```
ferjm: We'll need to review this later when we put all the pieces together. But that can be done in a follow up.

apps/sync/test/unit/sync-engine/adapter-mock.js
```diff
@@ -0,0 +1,24 @@
...
+/* exported
+  AdapterMock
+*/
+
+var AdapterMock = (action, args) => {
+  return {
+    update(kintoColl) {
+      if (action) {
+        return kintoColl[action].apply(kintoColl, args);
+      } else {
```
ferjm: No need for `else`

apps/sync/test/unit/sync-engine/adapter-mock.js
```diff
@@ -0,0 +1,24 @@
...
+
+'use strict';
+
+/* exported
+  AdapterMock
+*/
+
+var AdapterMock = (action, args) => {
+  return {
+    update(kintoColl) {
```
ferjm: is this syntax correct?

michielbdejong: Are you referring to the use of ES6 functions object-member shorthand? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions#Description

apps/sync/js/sync-engine/syncengine.js
michielbdejong: The user will be an option in the SyncEngine constructor (see bug	1202627), so it cannot change for a given SyncEngine instance.

apps/sync/js/sync-engine/syncengine.js
michielbdejong: Yes! Created bug 1204827 for this.

apps/sync/js/sync-engine/syncengine.js
michielbdejong: OK!

apps/sync/test/unit/sync-engine/adapter-mock.js
michielbdejong: Oops, I thought I had fixed all of those, sorry.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5b8f69936359c0d75323c969566fffb4e07a8db0

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=540b632645f7620e793e6a3aa2b31453d75dacf5

michielbdejong:
Rebased and squashed.

PR #31705, opened 2015-09-04
Bug 1196096 - [DataSync] Sync Manager. r=etienne, michielbdjong, seanlee

ferjm:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1147698007b01fd068af3c0489d8a8e9636bdafb

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=21c5312ce37ee489ddf53a5452246f9a6941811d

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8fbfe92c8e1ba34ec6c6c54849fe734a02e41f74

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a08f50e0ab842692cf81c8d2a505650632f6f6c0

apps/system/js/sync_manager.js
```diff
@@ -0,0 +1,460 @@
...
+              return;
+            }
+            Service.request('SyncStateMachine:error', ERROR_SYNC_APP_KILLED);
+          });
+
+        });
+      };
+    },
+
+    /** Settings observers **/
```
michielbdejong: Can these two be done with one piece of code instead of two? Could become a lot of duplicated code once we have 10 or more collections to sync.

apps/system/js/sync_manager.js
```diff
@@ -0,0 +1,460 @@
...
+        });
+      });
+    },
+
+    doSync: function(assertion, keys, collections) {
+      this.debug('Syncing with', collections);
+      this.iacRequest({
+        assertion: assertion,
+        keys: keys,
+        collections: collections
```
michielbdejong: Also need the Syncto server URL here, as well as xClientState. And `keys == (kA, kB)`, right?

apps/system/js/sync_manager.js
```diff
@@ -0,0 +1,460 @@
...
+      }).then(result => {
+        if (result.error) {
+          console.error('Error while trying to sync', result.error);
```
michielbdejong: I was thinking of returning 3 errors from the Sync app, which differ in how the system app should deal with then:

* Auth errors (get a new assertion)
* Try Later errors (e.g. network timeouts, or Syncto server report 503 Temporarily Unavailable)
* Unrecoverable errors (report failure to the user; would only happen as a result of bugs, disk full, etc.)

apps/system/js/sync_manager.js
```diff
@@ -0,0 +1,460 @@
...
+ */
+
+(function() {
+  const FXA_EVENT = 'mozFxAccountsUnsolChromeEvent';
+
+  const SYNC_TASK = 'gaia::system::firefoxsync';
+  const SYNC_REQUEST_IAC_KEYWORD = 'gaia::sync::request';
+
+  const HISTORY = 'history';
+  const PASSWORDS = 'passwords';
```
michielbdejong: Maybe make this an array, so we don't have any code that is particular to a specific collection. In fact, it would be nice if the Systems app would be entirely agnostic to collection names (it is possible to enumerate `sync.collection` from settings, and get the collection names from there?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=59e7a863e27f763795fd8a73fda351a08e5c4a56

apps/system/js/sync_manager.js
```diff
@@ -0,0 +1,471 @@
...
+    // ( No need to be observed )
+    // These settings must only be changed by the System app and read
+    // by the Settings app.
+
+    // Indicates the last error in case that 'sync.state' is 'errored'.
+    // -- 'sync.state.error',
+
+    // Last time we synchronized successfully.
+    // -- 'sync.lastTime'
+  ];
```
weilonge: The three properties ('sync.state', 'sync.state.error', 'sync.lastTime') look like read-only properties for Settings app (or other apps which would like to control fxsync), but these properties could be changed by any apps which have mozSettings permission. The state will probably be ruined by other apps.
Can these state/lastTime information be retrieved via IAC way just like iac-gaia::sync::management?

apps/system/js/sync_manager.js
```diff
@@ -0,0 +1,471 @@
...
+        case 'sync':
+          try {
+            Service.request('SyncStateMachine:' + request.name).catch(e => {
+              console.error(e);
+            });
+          } catch(e) {
+            console.error(e);
+          }
+          break;
+        default:
```
weilonge: Add one case here for apps to retrieve the sync state/lastTime in IAC way.

shared/js/sync/errors.js
```diff
@@ -0,0 +1,31 @@
...
+  var unrecoverableErrors = {
+    ERROR_REQUEST_SYNC_REGISTRATION: 'Cannot register sync request',
+    ERROR_GET_FXA_ASSERTION: 'Cannot get Firefox Accounts assertion',
+    ERROR_SYNC_REQUEST: 'Cannot perform sync request'
+  };
+
+  var recoverableErrors = {
+    ERROR_SYNC_APP_KILLED:
+      'The app was killed while performaning a sync operation',
+    ERROR_SYNC_APP_GENERIC: 'Error while trying to sync'
```
weilonge: We need to support L10N for these error messages as well, and errors.js only defines the L10N index/key.
If settings app will show these error messages, settings.en-US.properties would be like this with 'fxsync' prefix:
fxsync-error-request-sync-registration=Cannot register sync request
fxsync-error-get-fxa-assertion=Cannot get Firefox Accounts assertion
fxsync-error-sync-request=Cannot perform sync request
fxsync-error-sync-APP-killed=The app was killed while performaning a sync operation
fxsync-error-sync-APP-generic=Error while trying to sync

apps/system/js/sync_manager.js
weilonge: Agree with Michiel here :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7bbaad30f778c1a6ac5f6fea76373094748cf3d7

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6e30c8d1a3e2bacaf675b58b44bf8fc6ad3a405e

apps/system/js/sync_manager.js
ferjm: Yes, I can add a common handler helper, but unfortunately the handlers themselves need to be split. That's how BaseModule defines it.

apps/system/js/sync_manager.js
ferjm: I'll add the server URL, but I think xClientState needs to be calculated on the Sync app as we don't have the crypto library at this point.

And yes, `keys == (kA, kB)`

apps/system/js/sync_manager.js
ferjm: Ok, we can add the specific error handlers in a follow up.

apps/system/js/sync_manager.js
ferjm: Unfortunately there is no way to iterate over `sync.collection.*`

apps/system/js/sync_manager.js
ferjm: Good point!

shared/js/sync/errors.js
ferjm: System doesn't need to contain any l10n string. It's up to the Settings app to translate these errors into user understandable errors. All that System will be giving is error codes.

Anyway, I am refactoring this code to remove the strings and convert them into simple error codes.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e67546fc724c445fb36a293b204acaf71d007a15

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f9e99dca11303760e21da41e5d9858fea2485cd7

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a8fc40863763ea28f1c6a4d88eb669a87f555af1

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6f4d0ce7e39656b0abf6486809f3a56a3a5502ec

apps/system/js/sync_manager.js
```diff
@@ -0,0 +1,547 @@
...
+        LazyLoader.load('/shared/js/sync/errors.js', () => {
+          navigator.mozSetMessageHandler('request-sync', event => {
+            this.debug('Sync request');
+            if (event.task != SYNC_TASK) {
+              return;
+            }
+            Service.request('SyncStateMachine:sync');
+          });
+
+          SyncManager.SETTINGS.forEach(setting => {
```
etiennesegonzac: you should not have do to this... https://github.com/mozilla-b2g/gaia/blob/master/apps/system/js/base_module.js#L433-L435

apps/system/js/sync_manager.js
```diff
@@ -0,0 +1,547 @@
...
+          return Promise.resolve();
+        }
+
+        this.debug('Registering sync request', location.href);
+
+        return navigator.sync.register(SYNC_TASK, {
+          // minInterval is in seconds.
+          // XXX 2 minutes is only for testing purposes. Will default to 1
+          //     hour.
+          minInterval: this.settings.get('sync.scheduler.interval') || 120,
```
etiennesegonzac: maybe you can use the same `DEBUG` flag that the `BaseModule#debug` to do this just to make sure we don't forget it later on

apps/system/js/sync_manager.js
```diff
@@ -0,0 +1,547 @@
...
+    // able to handle:
+    //   - enable
+    //   - disable
+    //   - sync
+    //   - getInfo
+    //
+    // SyncStateMachine is responsible of checking if the request is valid in
+    // the current situation and progressing it accordingly by publishing the
+    // corresponding events.
+    '_handle_iac-gaia::sync::management': function(event) {
```
etiennesegonzac: Can we use a simpler channel name? This method name is banana :)
I suggest `sync-management` but if you really want the namespacing `gaia-sync-management` would at least prevent the introduction of a _third_ special character in the method name.

apps/system/js/sync_manager.js
```diff
@@ -0,0 +1,547 @@
...
+            this.store.set(key, value);
+            resolve();
+          });
+        });
+        promises.push(promise);
+      });
+      return Promise.all(promises);
+    },
+
+    saveState: function(state) {
```
etiennesegonzac: nit: I'd rename to `updateState` since the saving is done by the property setter and it fits nicely with the fact that without parameter this will do a refresh.

apps/system/test/unit/sync_manager_test.js
```diff
@@ -0,0 +1,784 @@
...
+        requestStub = this.sinon.stub(MockService, 'request', () => {
+          return Promise.resolve();
+        });
+        syncManager.start().then(done);
+      });
+    });
+
+    teardown(() => {
+      syncManager.stop();
+      saveStateSpy.restore();
```
etiennesegonzac: You get the `restore` for free if you use `this.sinon...`

apps/system/js/sync_manager.js
etiennesegonzac: on that note, we might want to add a small unit test for this behavior (set / refresh)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=dea569400bfa835e8fed0efa8b36e4298f905d4f

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3f6f72ceed6f7d7f2a74e343d47ff020c739a2e3

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2efb922bc2e0cd915f1adfc17022ba8c2d7899e7

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8348caebc2e03ad2a40434554b724a349289ca16

PR #31710, opened 2015-09-06
Bug 1064144 - Improve focus behaviour when opening a draft

autra:


autolander:

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9341322cde15ea07057951581865518784b70c09

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=dc81b208460c25abf55f6d86423c1df3cf101ddb

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7a5dfbea1c0c35283156f3181006f74231fedcb4

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=655155f97e341367982d6e72526a41266e52c862

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=dd85b89e0435d7321853d1f6814f6f1ae3a66206

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b05989d7f794185aec0113470bdf3042b5421a70

autra:
So there are 2 commits, the second one only fixing some trailing whitespaces. I included it because my vim keeps fixing them, so I thought I'd better commit it anyway. I don't know what is the policy for this kind of fixes in gaia...

apps/sms/views/conversation/js/conversation.js
```diff
@@ -769,10 +779,11 @@ var ConversationView = {
         return;
       }
 
+      var promises = [];
       // Recipients will exist for draft messages in threads
       // Otherwise find them from draft recipient numbers
       this.draft.recipients.forEach(function(number) {
-        Contacts.findByAddress(number).then(function(contacts) {
+        promises.push(Contacts.findByAddress(number).then(function(contacts) {
```
autra: this `then` callback was executed after `draftPromise.then` line 648. In other word, now that we check the recipients to decide where to focus, we need to wait for their initialization

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3e5e340b505c3705e423b609a5f29df34be091b2

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=464db92c6678faff65f070017069dd0a94f72702

apps/sms/test/marionette/drafts_test.js
```diff
@@ -283,6 +289,34 @@ marionette('Messages Drafts', function() {
       });
     });
 
+    suite('Thread-less recipient-less drafts', function() {
```
autra: I put this test in its own suite. Didn't find any easy way to include that in `Thread-less drafts` test suite... Do you have a better idea?

apps/sms/views/conversation/js/conversation.js
```diff
@@ -642,10 +642,20 @@ var ConversationView = {
...
-        this.recipients.focus();
+      switch (args.focus) {
+        case 'composer':
+          Compose.focus();
+          break;
+        case 'recipients':
+          this.recipients.focus();
+          break;
+        default:
+          if (this.hasValidRecipients()) {
```
julienw: right now we don't save invalid recipients, but we will eventually. Wondering what we should do if we have only invalid recipients... I'd say... we'll see when it will happen :p

apps/sms/views/conversation/js/conversation.js
julienw: yep :) but you can use "this.draft.recipients.map" (instead of forEach) and you don't need to use the array and push in it.

julienw: and while you're here, maybe convert the functions to arrow functions, and remove the "bind" calls

apps/sms/test/marionette/drafts_test.js
julienw: I think it's ok here.

apps/sms/test/marionette/lib/views/new-message/view.js
```diff
@@ -167,6 +167,12 @@ NewMessageView.prototype = {
     return this.composerAccessors.messageInput.scriptWith(function(el) {
       return document.activeElement === el;
     });
+  },
+
+  isRecipientsInputFocused: function() {
+    return this.accessors.recipientsInput.scriptWith(function(el) {
+      return document.activeElement === el;
+    });
```
julienw: wondering if we should have a more generic "isElementFocused" that we could reuse. I don't know where we can have such a toolset though... @azasypkin what do you think ?

apps/sms/views/conversation/test/unit/conversation_test.js
```diff
@@ -6786,6 +6783,8 @@ suite('conversation.js >', function() {
         ConversationView.initRecipients();
         Navigation.isCurrentPanel.withArgs('composer').returns(true);
         this.sinon.stub(ConversationView.recipients, 'focus');
+        this.sinon.stub(Compose, 'focus');
+        this.sinon.stub(ConversationView, 'hasValidRecipients');
```
julienw: I don't think it's a valid way to stub internal methods. Can you write tests that actually add drafts with/without recipients instead ?

apps/sms/views/shared/js/activity_handler.js
```diff
@@ -211,22 +211,22 @@ var ActivityHandler = {
...
-  toView: function ah_toView(message, focusComposer) {
+  toView: function ah_toView(message, focus) {
     var navigateToView = () => {
       // If we have appropriate thread then let's forward user there, otherwise
       // open new message composer.
       if (message.threadId) {
         return Navigation.toPanel('thread', {
           id: message.threadId,
-          focusComposer: focusComposer
+          focus: focus
```
julienw: nit: use just the shortcut "focus"

apps/sms/views/conversation/js/conversation.js
autra: If I understand correctly, `this.hasValidRecipients` only checks that the input value is not empty and finite right? So actually, it actually checks whether or not we have something.In case of only invalid recipients, maybe we'd want to focus the recipients input instead of the composer... Yep, we'll see what happens :-p

apps/sms/views/conversation/test/unit/conversation_test.js
autra: Could you elaborate why it is not valid? What do you mean by that?

Actually here, as we are unit-testing, isn't it better to treat results of other functions as part of the external state, and therefore to mock them, just like parameters? That's usually how I write my unit test (but are we really unit-testing here?) . WDYT?

apps/sms/test/marionette/lib/views/new-message/view.js
azasypkin: Yeah, would be great to have shared tools for that and we don't have them now :/ What about ```view_tools (or better name).js``` in [1]?

[1] https://github.com/mozilla-b2g/gaia/tree/master/apps/sms/test/marionette/lib/views/shared

apps/sms/test/marionette/drafts_test.js
azasypkin: __nit:__ and please ```Conversation-less drafts``` :) I noticed that new contributors are confused by the fact that we use both ```conversation``` and ```thread``` across our codebase.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=33aaa253a4c93a0dad77b048469c42a8396c891f

apps/sms/test/marionette/drafts_test.js
autra: Oh. So there's a difference between conversation and thread? :-)

azasypkin: I think they are the same, also in inbox-view terms conversation-less draft is "conversation" as well, @julienw please correct me if I'm wrong :)

apps/sms/views/conversation/test/unit/conversation_test.js
julienw: In the term "unit testing", we're testing a "unit of code"; the unit of code is the whole object ConversationView.
So ideally we should not mock other methods of this object while testing a method of this object. We don't need to test how the unit works internally, this is not what's important. What's important is how the unit works with inputs/outputs.

Also this makes it easier to refactor the unit.

if we feel like we need to, it usually means the object holds several responsiblities and we should split it.

In this case, I don't think it's the case though. It's just as easy to add something (or not) to ConversationView.recipients.numbers (I think ConversationView.recipients is a MockRecipients, but please check this).

Hope this is clearer !

apps/sms/test/marionette/drafts_test.js
julienw: we changed our terminology from "thread" to "conversation" as this is more understandable and less overloaded...

In Inbox "conversation-less draft" is an item of the list, don't know if we should call this a "conversation" but at least it has the same info :) maybe in Inbox we should use the more general term "inbox item" :)

apps/sms/views/conversation/test/unit/conversation_test.js
autra: Completely clear and making sense, thanks!

apps/sms/test/marionette/drafts_test.js
azasypkin: > In Inbox "conversation-less draft" is an item of the list, don't know if we should call this a "conversation" but at least it has the same info :) maybe in Inbox we should use the more general term "inbox item" :)

Heh, note that we're going to return both conversations and conversation-less drafts as "conversation/conversation-info" via ```ConversationService.getAllConversations``` :) [1]

[1] https://github.com/mozilla-b2g/gaia/blob/master/apps/sms/services/README.md

julienw: yeah I know; maybe we can find better semantics

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e77d8fb29484f3aa82be2d7e4812e9c44383b43f

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1d1e05128db19b15e1cc56b082c119eafe33ffe5

apps/sms/test/marionette/drafts_test.js
```diff
@@ -200,7 +206,7 @@ marionette('Messages Drafts', function() {
       inboxView = new InboxView(client);
     });
 
-    suite('Thread-less drafts', function() {
+    suite('Conversation-less drafts', function() {
```
autra: So renaming also the other `Thread` of this test :-)

apps/sms/test/marionette/lib/views/conversation/view.js
```diff
@@ -6,6 +6,7 @@ var ConversationAccessor = require('./accessors');
 var MessageAccessor = require('./message_accessors');
 var ComposerAccessor = require('../shared/composer_accessors');
 var MenuAccessor = require('../shared/menu_accessors');
+var Tools = require('../shared/tools.js');
```
autra: Here I'm not really sure I've done things the way you expected. This certainly allows to factorize the `scriptWith...` stuff, but I feel I should've gone farther. 

On the other hand, if I remove or move the `isMessageInputFocused` and `isRecipientInputFocused`, I would need to declare `messageAccessor` and `composerAccessor` in `Tools` and I didn't feel that worthy. WDYT?

apps/sms/test/marionette/lib/views/shared/tools.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+/* global module */
+
+function Tools() {
+}
```
autra: I created a constructor here like in other files, but is it ok to just export a single object containing my method?

apps/sms/test/marionette/lib/views/conversation/view.js
autra: Actually I realised I forgot some isElementFocused in individual tests. Will update.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d1327f79e61b573450773fb421f95f06934264cc

julienw:
> I created a constructor here like in other files, but is it ok to just export a single object containing my method?

I actually prefer it like you did in the last commit :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a9947831a46976d017a7f7ecf2f9c656fcfad28c

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e7e58954b53dd9312034ff129e6a27abe7af47e7

PR #31718, opened 2015-09-07
Bug 1202388 - Factorize the checkboxes and the swiches into 1 class

JohanLorenzo:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8732315d8b2323fc64fd27e5ed1d03ec6ee48fdb

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8f0e32f2ce9da08742a923084219a4571466de0b

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6f54eb1fc1fba971947d3d78f8f61b1a08748e74

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e36bc7cbaab5f175d506ab6dd7239e830beab921

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/wifi.py
```diff
@@ -18,13 +17,10 @@ class Wifi(Base):
...
     def is_wifi_enabled(self):
-        element = self.marionette.find_element(*self._wifi_enabled_checkbox_locator)
-        return self.is_custom_element_checked(element)
+        return GaiaBinaryControl(self.marionette, self._wifi_enabled_checkbox_locator).is_checked
 
     def enable_wifi(self):
-        element = self.marionette.find_element(*self._wifi_enabled_checkbox_locator)
-        element.tap()
-        self.wait_for_custom_element_checked_state(element)
+        return GaiaBinaryControl(self.marionette, self._wifi_enabled_checkbox_locator).enable()
```
mwargers: The return here is not needed.

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/wifi.py
```diff
@@ -4,9 +4,8 @@
 
 from marionette_driver import expected, By, Wait
 from marionette_driver.errors import StaleElementException
-
 from gaiatest.apps.base import Base
-
+from gaiatest.form_controls.binarycontrol import GaiaBinaryControl
```
mwargers: You need to keep the 2 newlines here, right?

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
```diff
@@ -79,6 +79,10 @@ class Settings(Base):
 
     _main_title_locator = (By.ID, 'main-list-title')
 
+    def __init__(self, marionette):
+        Base.__init__(self, marionette)
+        self._airplane_checkbox = HtmlBinaryControl(self.marionette, self._airplane_checkbox_locator)
+
```
mwargers: The __init__ things you added are causing all the failures on Jenkins, it seems.
The failures happen because when a Settings instance is called, for instance, it doesn't have to be launched yet.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=455fa9aa04ddcdcd4a6455ad6b0045e195a085b8

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5a4a6a7208f0c7bab7f34a77333b444d0841d0e4

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c2c7283256649612b53c515882fc51237d1aacb6

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1f783216dabf5e79c993a8ca9ece5408364485a1

tests/python/gaia-ui-tests/gaiatest/apps/contacts/regions/settings_form.py
```diff
@@ -35,18 +36,13 @@ def __init__(self, marionette):
...
+        self._order_by_last_name_switch = GaiaBinaryControl(self.marionette, self._order_by_last_name_switch_locator)
 
-    def tap_order_by_last_name(self):
-        last_name = Wait(self.marionette).until(
-            expected.element_present(*self._order_by_last_name_switch_locator))
-        Wait(self.marionette).until(expected.element_displayed(last_name))
-        initial_state = self.is_custom_element_checked(last_name)
-        last_name.tap()
-        self.wait_for_custom_element_checked_state(last_name, checked=not(initial_state))
+    def enable_order_by_last_name(self):
```
mwargers: Perhaps we could just do remove this method and just use contacts_settings.order_by_last_name_switch.enable() directly in the test?
This just feels like an unnecessary inbetween step to me.

tests/python/gaia-ui-tests/gaiatest/apps/contacts/regions/settings_form.py
```diff
@@ -35,18 +36,13 @@ def __init__(self, marionette):
...
+        self._order_by_last_name_switch.enable()
 
-    @property
-    def order_by_last_name(self):
-        return self.marionette.find_element(*self._order_by_last_name_switch_locator).is_selected()
+    def disable_order_by_last_name(self):
```
mwargers: Same here.

tests/python/gaia-ui-tests/gaiatest/apps/cost_control/app.py
```diff
@@ -29,13 +30,25 @@ class CostControl(Base):
...
+    def _mobile_tracking_switch(self):
+        return GaiaBinaryControl(self.marionette, self._mobile_data_tracking_locator)
 
     @property
     def is_wifi_data_tracking_on(self):
-        wifiswitch = self.marionette.find_element(*self._wifi_data_tracking_locator)
-        return wifiswitch.is_selected()
+        return self._wifi_tracking_switch.is_checked
+
+    def enable_wifi_data_tracking(self):
```
mwargers: Same here.

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
```diff
@@ -92,28 +92,35 @@ def switch_to_settings_app(self):
...
-        return self.marionette.find_element(*self._usb_storage_checkbox_locator).is_selected()
+        return self._usb_checkbox.is_checked
+
+    @property
+    def _usb_checkbox(self):
+        class UsbSwitch(InvisibleHtmlBinaryControl):
+            def _toggle(self):
+                # There are 2 parts on that entry. The left part allows you to choose the
+                # USB transfer protocol, and the right part is the switch.
+                right_part_of_entry = self._element_to_tap.size['width']-5
```
mwargers: Use .rect if you can, please.

tests/python/gaia-ui-tests/gaiatest/form_controls/binarycontrol.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def disable(self):
+        self._toggle_and_verify_state(final_state=False)
+
+    def _toggle_and_verify_state(self, final_state):
+        Wait(self.marionette).until(expected.element_enabled(self.root_element))
+        Wait(self.marionette).until(lambda m: self.is_checked is not final_state)
+        self._toggle()
+        Wait(self.marionette).until(lambda m: self.is_checked is final_state)
+        Wait(self.marionette).until(expected.element_enabled(self.root_element))
```
mwargers: I guess this Wait is necessary for cases like the Wifi checkbox in settings disable, right?

tests/python/gaia-ui-tests/gaiatest/form_controls/form_control.py
```diff
@@ -0,0 +1,12 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from marionette_driver import expected, Wait
+from gaiatest.apps.base import PageRegion
+
+
+class FormControl(PageRegion):
+    def __init__(self, marionette, locator):
+        element = Wait(marionette).until(expected.element_present(*locator))
+        PageRegion.__init__(self, marionette, element)
```
mwargers: hrm, that makes me wondering why we don't let PageRegion accept a locator as well, because then this wouldn't be necessary.

tests/python/gaia-ui-tests/gaiatest/apps/contacts/regions/settings_form.py
```diff
@@ -35,18 +36,13 @@ def __init__(self, marionette):
         Base.__init__(self, marionette)
         view = self.marionette.find_element(*self._settings_view_locator)
         Wait(self.marionette).until(lambda m: view.location['y'] == 0)
```
mwargers: I guess while you're here, you could change this to use view.rect as well, instead of view.location.

tests/python/gaia-ui-tests/gaiatest/apps/cost_control/regions/ftu_step3.py
```diff
@@ -27,15 +28,18 @@ def __init__(self, marionette):
         view = self.marionette.find_element(*self._view_locator)
         Wait(self.marionette).until(lambda m: view.location['x'] == 0)
```
mwargers: Here also, use .rect instead of .location if you feel like it.

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/wifi.py
```diff
@@ -4,8 +4,8 @@
 
 from marionette_driver import expected, By, Wait
 from marionette_driver.errors import StaleElementException
-
```
mwargers: We still need to keep this newline, or am I wrong?

tests/python/gaia-ui-tests/gaiatest/tests/functional/contacts/test_sort_contacts.py
```diff
@@ -24,20 +22,9 @@ def setUp(self):
...
 
-        # if "Order by last name" switch is on, turn off it
-        contacts_settings = contacts_app.tap_settings()
-        self.assertEqual(contacts_settings.order_by_last_name, order_by_last_name_default)
-        # This test starts from when order_by_last_name is false
-        if order_by_last_name_default:
-            contacts_settings.tap_order_by_last_name()
-        contacts_settings.tap_done()
-
-        # here we wait for contacts once the correct order has been set
```
mwargers: This test had the flexibility that you could just flip the order_by_last_name_default boolean to have the test pass again if order by last name would be suddenly changed.
But I don't think it's a big deal, this is some nice clean-up and when the order by last name thing suddenly changes, then it still should be quite easy to fix the test.

tests/python/gaia-ui-tests/gaiatest/form_controls/binarycontrol.py
```diff
@@ -0,0 +1,58 @@
...
+
+from abc import ABCMeta, abstractproperty
+from marionette_driver import Wait, expected
+from gaiatest.form_controls.form_control import FormControl
+
+
+class BinaryControl(FormControl):
+    # This represent elements like Switches and Checkboxes
+
+    __metaclass__ = ABCMeta
```
mwargers: What is this?

tests/python/gaia-ui-tests/gaiatest/form_controls/binarycontrol.py
```diff
@@ -0,0 +1,58 @@
...
+class HtmlBinaryControl(BinaryControl):
+    @property
+    def is_checked(self):
+        return self.root_element.is_selected()
+
+
+class InvisibleHtmlBinaryControl(HtmlBinaryControl):
+    # Sometimes the checkboxes are present in the DOM, but they are invisible.
+    # In this case, you have to tap to another element.
+
```
mwargers: This class doesn't seem to have a is_checked property.
Doesn't that give problems in the is_alarm_active case?

tests/python/gaia-ui-tests/gaiatest/form_controls/binarycontrol.py
JohanLorenzo: This is needed to use the `@abstractproperty` . This is [documented standard python](https://docs.python.org/2.7/library/abc.html?highlight=abstractmethod#abc.abstractmethod)

tests/python/gaia-ui-tests/gaiatest/form_controls/binarycontrol.py
JohanLorenzo: Right

tests/python/gaia-ui-tests/gaiatest/form_controls/binarycontrol.py
JohanLorenzo: This class inherits from `HtmlBinaryControl`. Hence, this class has an `is_checked` attribute, which is defined in `HtmlBinaryControl`

tests/python/gaia-ui-tests/gaiatest/form_controls/form_control.py
JohanLorenzo: You're right, I can move this in `PageRegion`

tests/python/gaia-ui-tests/gaiatest/tests/functional/contacts/test_sort_contacts.py
JohanLorenzo: I'll fix it

tests/python/gaia-ui-tests/gaiatest/form_controls/binarycontrol.py
mwargers: Oops! You're right.

tests/python/gaia-ui-tests/gaiatest/apps/contacts/regions/settings_form.py
JohanLorenzo: I didn't want to break the [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter). To sum up, The law of Demeter says:
```python
my_object1.object2.object3.do_something()
```
is a bad practice, because if one day you change the internal behavior between object2 and object3, this will impact more parts of the code, where as if you just do:
``` python
my_object1.do_something_with_object2()
```

To be honest, I was unsure if we should break this rule too, as most of the time, we just do
```python
def enable_stuff()
    self._stuff.enable()
```

But, sometimes, we have to check another place of the page, [like there](). So I think, for consistency (and the Law of Demeter), it's better to have `enable_stuff()`.

tests/python/gaia-ui-tests/gaiatest/apps/cost_control/regions/ftu_step3.py
JohanLorenzo: Done

tests/python/gaia-ui-tests/gaiatest/apps/contacts/regions/settings_form.py
JohanLorenzo: Done.

tests/python/gaia-ui-tests/gaiatest/form_controls/form_control.py
JohanLorenzo: After taking a closer look at it, this might be a big task to deal with. I'd rather to keep this PR not too big. I filed this [follow up bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1203527).

tests/python/gaia-ui-tests/gaiatest/tests/functional/contacts/test_sort_contacts.py
JohanLorenzo: After a closer look at the test, I think there is some value of not having the default value set at the beginning of the test. It prevents from mental mapping. That is to say:
One day, your test fails at this line:
``` python
toggle_stuff()
```

By just looking at the stacktrace, you have no idea what should be the state of `stuff`. So you go to the code, to check what the test was supposed to do, and you see:
``` python
# more line of code above
if default_stuff_state:
    toggle_stuff()
# more lines there
toggle_stuff()
```

You still don't know what's going on, so you look for that variable, and then you remember that `stuff` should be true. Then you read the test again:
``` python
default_stuff_state = True

if default_stuff_state:
    toggle_stuff()
# ...
toggle_stuff()
```
In which state should `stuff` at the last line? To answer this question, you have to think about it for (at least) 3-4 seconds, and do the conversion each time you toggle `stuff`. This is not trivial.

I agree on the fact that putting the original state is kind of factorization, which is good. But I (personally) think an end-to-end test should be close to English sentences, with no mental mapping. I think this has to be discussed in the best practice document.

PS : I didn't think my comment would be that verbose. Sorry for the ranting.

tests/python/gaia-ui-tests/gaiatest/apps/settings/app.py
JohanLorenzo: Done.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ec6eeb464c8b3e626ee91669a3f4ca213a608206

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/wifi.py
JohanLorenzo: There is no PEP8 rule that requires that. I deleted it, but we can put it back. I'm fine with both

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/wifi.py
JohanLorenzo: Done.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=69ba7022f068502f45d32544efe076b7d5b8719e

tests/python/gaia-ui-tests/gaiatest/apps/contacts/regions/settings_form.py
mwargers: So I guess what you mean is that you consider all the elements on a page to be internal objects that should not be visible to outside objects?
I guess that if they then did a redesign of the page, where they didn't use a checkbox, then we only have to change do_something_with_object2() as opposed to fix all the instances my_object1.object2.do_something() (I don't know what you mean with this object3 thing).
I'm not sure how much trouble this would be in practice. In general, at most we use these .object2 things like 5 times maybe.

PR #31774, opened 2015-09-09
Bug 1202921 - sockit-to-me does not compile on node v4.0.0

KevinGrandon:
https://bugzilla.mozilla.org/show_bug.cgi?id=1202921


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b0a1be47b8ce3363fd9670ab5608349dc32784bf

package.json
```diff
@@ -6,7 +6,7 @@
   "author": "The Gaia Team <dev-gaia@lists.mozilla.org>",
   "engineStrict": "true",
   "engines": {
-    "node": ">=0.10",
+    "node": ">=4.0",
```
eliperelman: My suggestion is to make this `^4.0.0` and change the version number of this package to `4.0.0`.

tests/jsmarionette/client/sockit-to-me/binding.gyp
```diff
@@ -1,16 +1,15 @@
 {
   'variables': {
-    'gpp': '<!(type g++ | grep "is" > /dev/null && echo "yep" || echo "no")',
-    'installed_node_version': '<!(node -v | cut -c4,5)'
+    'gpp': '<!(type g++ | grep "is" > /dev/null && echo "yep" || echo "no")'
```
eliperelman: +1

package.json
eliperelman: Oops, wrong spot for the version number change.

tests/jsmarionette/client/sockit-to-me/package.json
```diff
@@ -6,7 +6,7 @@
   "main": "build/Release/sockit",
 
   "devDependencies": {
-    "node-gyp": "~0.10.6"
+    "node-gyp": "3.0.1"
```
eliperelman: *Here* I would suggest versioning sockit-to-me to `4.0.0` and specifying the engines to `^4.0.0`.

package.json
KevinGrandon: I think this is the right spot? I agree that there should be a sockit-to-me version bump, and I'm fine tracking against node versions if we want to.

eliperelman: But to change the engines along with sockit, shouldn't Gaia also version change? Idk, just wondering. :)

KevinGrandon: There would be a conflict reported in npm commands, so yeah - we probably should bump the gaia requested version and engines as well.

nullaus: I think changing the node engine version to >=4.0 is fine. There's no reason to not support future versions if it keeps on compiling magically. When it breaks at a certain version we can cap which version we go up to.

Changing the gaia version can be done in this file as well, I would just bump it to 0.0.4 personally. I don't see any value in making it 4.0.0.

tests/jsmarionette/client/sockit-to-me/package.json
nullaus: Also don't see any value in version 4.0.0 here. engines should be 0.12.* and >=4.0 it works as advertised on 0.12.

nullaus:
The build folder should not be checked in, it's generated by node-gyp.

tests/jsmarionette/client/sockit-to-me/package.json
nullaus: I also wouldn't tie sockit-to-me to that specific version of node-gyp, instead, look at the version shipped typically with 0.12 and use that as the base version.

eliperelman: @nullaus My only contention is that since sockit-to-me really only changes whenever there is a new Node.js version, that its version stay in line with Node's. That way if someone can't upgrade to Node.js 4.0.0, they can use the 0.x branch of sockit-to-me, and if someone can't use Node.js 5.0.0, they can use the 4.x branch of sockit-to-me. Does that sound reasonable?

I don't have any opinions about engines really, more just that sockit-to-me should be more semver reliable.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4c2acba39da598d367b98be7b72062c2f3929fdf

tests/jsmarionette/client/sockit-to-me/package.json
nullaus: Sure, I think what I'm suggesting would accomplish that just fine. NPM will hand over the right version based on the engine being used. We just need to put the right engine definitions in (which are the ones I stated). If we're really paranoid about a new node version breaking sockit-to-me then I guess marking it as running with node 0.12.x and ^4.0.0 would do the trick. No need to get crazy with version tracking.

nullaus:
lgtm! ship it!

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4f804a4b1cd4192eb3cbb23a83db1590727199a0

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f2bbd833ce5774b52d1851315f38c34f7c8d76b3

PR #31780, opened 2015-09-10
Bug 1169842 - Suspend/Resume camera button listener based on screen l…

scottwu:
…ock state r=snowmantw

### DIFF
--- a/apps/system/lockscreen/js/lockscreen_state_manager.js
+++ b/apps/system/lockscreen/js/lockscreen_state_manager.js
@@ -94,7 +94,13 @@
         'secure-appopened',
         'secure-appclosing',
         'secure-appterminated',
-        'holdcamera'
+        'holdcamera',
+        // XXX: this is necessary only when LockScreen is not an iframe.
+        // Since for an iframe it will not listen to any event after unlocking,
+        // namely when it is pushed into the background.
+        // Nevertheless we need to do the same suspending/resuming
+        // thing in this way. (Bug 1169842)
+        'lockscreen-appopened'
       ],
       observers: {
         'lockscreen.passcode-lock.enabled':
@@ -105,6 +111,7 @@
           this.getSettingsObserverCallback('unlockSoundEnabled')
       }
     };
+    this.suspendedEvents = {};
 
     // Default values
     this.lockScreenDefaultStates = {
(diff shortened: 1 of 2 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=fe55d3998d44eb8f5635ba5052b14c5ac753210a

apps/system/lockscreen/js/lockscreen_state_manager.js
```diff
@@ -542,10 +564,13 @@
           this.onUnlock();
         }
         break;
+      case 'lockscreen-appopened':
       case 'lockscreen-request-lock':
+        this.resumeEvents(['holdcamera']);
```
snowmantw: You can just use `this.resumeEvents('holdcamera')` and save it in the map. To have an array is not necessary for the arguments.

apps/system/lockscreen/js/lockscreen_state_manager.js
```diff
@@ -542,10 +564,13 @@
...
         this.onLock(detail);
         break;
       case 'lockscreen-appclosed':
+        this.suspendEvents(['holdcamera']);
```
snowmantw: The same.

apps/system/lockscreen/js/lockscreen_state_manager.js
snowmantw: Since in the scenario we need to suspend multiple events, we can call it in a loop, and push into the inner array of suspending events one by one.

apps/system/lockscreen/js/lockscreen_state_manager.js
```diff
@@ -494,6 +496,26 @@
...
+      if (this.suspendedEvents[ename]) {
+        this.suspendedEvents[ename] = false;
+        window.addEventListener(ename, this);
+      }
+    });
+  };
+
+  LockScreenStateManager.prototype.suspendEvents =
+  function lssm_suspendEvents(suspendEvents) {
+    suspendEvents.forEach((ename) => {
```
snowmantw: I've commented a related issue at line 569.

apps/system/lockscreen/js/lockscreen_state_manager.js
```diff
@@ -82,6 +82,7 @@
         'screenchange',
         'lockscreen-notification-request-activate-unlock',
         'lockscreen-request-unlock',
+        'lockscreen-appopened',
```
snowmantw: It's better to add a workaround comment:

    // XXX: this is necessary only when LockScreen is not an iframe.
    // Since for an iframe it will not listen to any event after unlocking, namely when it
    // is pushed into the background. Nevertheless we need to do the same suspending/resuming
    // thing in this way.

You can find a suitable place to put the sentences like that. It elaborates the motivation why we need to do that now.

apps/system/lockscreen/js/lockscreen_state_manager.js
```diff
@@ -494,6 +496,26 @@
...
+  function lssm_resumeEvents(resumeEvents) {
+    resumeEvents.forEach((ename) => {
+      if (this.suspendedEvents[ename]) {
+        this.suspendedEvents[ename] = false;
+        window.addEventListener(ename, this);
+      }
+    });
+  };
+
+  LockScreenStateManager.prototype.suspendEvents =
```
snowmantw: You can use the method comments to describe the similar thing that I've commented at line 85. And plus why 'holdcamera' is the only event we now deal with (because hardware event will be fired even after the window get hidden).

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=fff71499dc43712dec3aac3dabbc14a6a7850d68

apps/system/lockscreen/js/lockscreen_state_manager.js
```diff
@@ -94,7 +94,13 @@
         'secure-appopened',
         'secure-appclosing',
         'secure-appterminated',
-        'holdcamera'
+        'holdcamera',
+        // XXX: this is necessary only when LockScreen is not an iframe.
+        // Since for an iframe it will not listen to any event after unlocking,
+        // namely when it is pushed into the background.
+        // Nevertheless we need to do the same suspending/resuming
+        // thing in this way. (Bug 1169842)
```
snowmantw: Well done.

PR #31797, opened 2015-09-11
Bug xxx use conversation service in window

azasypkin:


autolander:

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5ec6d7d6f1183d09acf03e05e0dcc391a8b7c1bf

apps/sms/services/js/bridge_client_mixin.js
```diff
@@ -0,0 +1,144 @@
+/*global bridge */
+/*jshint esnext: true */
+
```
azasypkin: It's veryyy wip, but the idea is to have code that deals with bridge client, upgrades, appInstanceId and can be shared across different ServiceXClients

apps/sms/services/js/bridge_client_mixin.js
```diff
@@ -0,0 +1,144 @@
...
+    initClient({ appInstanceId, endpoint, plugins }) {
+      if (!appInstanceId) {
+        throw new Error('Application instance identifier is required!');
+      }
+
+      this[priv.appInstanceId] = appInstanceId;
+      this[priv.initBridge](endpoint || exports, plugins);
+    },
+
+    clientMethod(...params) {
```
azasypkin: Those methods are just to track promises for pending operations.

@wilsonpage Is it possible (out of the box) to destroy client/service so that it automatically waits for the pending operations to complete? 

More context, in this PR we want to perform the following actions:
* Run app using service hosted in main window (to not pay SharedWorker initialization cost);
* On "visuallyLoaded" we'd like to spawn the same service in a SharedWorker;
* Starting from this moment all new method/stream calls will be forwarded to service hosted in a SharedWorker;
* Since window service stream is still active (basically until "fullyLoaded") we don't want to interrupt it (and to have less complexity), but rather wait for its completion and only then shutdown old bridge.client and bridge.service.

apps/sms/services/js/bridge_client_mixin.js
```diff
@@ -0,0 +1,144 @@
...
+
+        this[method] = clientMixin[method];
+      }, target);
+
+      target[priv.client] = null;
+      target[priv.appInstanceId] = null;
+      target[priv.pendingOperations] = null;
+      target[priv.serviceName] = serviceName;
+
+      target[priv.initBridge] = function(endpoint, plugins) {
```
azasypkin: will move this functions to somewhere else

apps/sms/services/js/conversation/conversation_client.js
```diff
@@ -32,36 +27,35 @@
...
      * Initialized conversation service client bridge.
      * @param {string} appInstanceId Unique identifier of app instance where
      * client resides in.
      */
     init(appInstanceId) {
-      if (!appInstanceId) {
-        throw new Error('AppInstanceId is required!');
-      }
+      // We use window based ConversationService initially, so let's init it.
+      exports.ConversationService.init();
```
azasypkin: That's what I'm not sure of, but one of the ideas is to keep all this switching, window-context-hack knowledge inside client only... as well path to the SharedWorker's main file.

apps/sms/services/js/conversation/conversation_service.js
```diff
@@ -125,12 +123,7 @@
         return getThreadsStream.cancel(reason);
       };
 
-      var draftsPromise = Drafts.request().then(() => {
```
azasypkin: I tend to move this part to a separate bug, it turned out to be complex to land conversation service in master with new way of draft handling

apps/sms/services/js/bridge_client_mixin.js
wilsonpage: Right now calling `.destroy()` calls `.disconnect()` which will cancel any pending calls. If you don't want this behaviour you should probably wait for the stream to close before calling `.destroy()`.

azasypkin: Okay, good to know, that's what I'm doing here..

PR #31800, opened 2015-09-11
Bug 1201016 - [Messages][NG] Migrate the current Message manager event handling to NGA, r=julienw azasypkin

steveck-chung:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=62af252a0b01be37227d942b278c07a01cd7e34d

apps/sms/views/shared/js/startup.js
```diff
@@ -89,14 +89,15 @@ var Startup = exports.Startup = {
 
       // Init UI Managers
       TimeHeaders.init();
-      ConversationView.init();
-      MessagingClient.init(App.instanceId);
       MozMobileConnectionsClient.init(App.instanceId);
-      Information.initDefaultViews();
+      MessagingClient.init(App.instanceId).then(() => {
```
steveck-chung: Here I change the view init timing because the event is registered only when service is connected with app instance ID. Not sure if there is any better why to initialize the event listener because it must have longer gap between DOM event fired from iframe and view actually register the handler(at least it's async and anything could happen during this gap).

apps/sms/services/js/messaging/messaging_service.js
```diff
@@ -65,6 +70,17 @@ var MessagingService = {
     this.initService();
   },
 
+  registerEvent(appInstanceId) {
+    var mobileMessageClient = MozMobileMessageClient.forApp(appInstanceId);
+
+    EVENTS.forEach((event) => {
+      mobileMessageClient.on(event, (...args) => {
+        this.broadcast(event, ...args);
```
azasypkin: mm, I believe we don't need ```...``` here since we can have only one argument in the event handler usually, moreover  it won't work correctly in case of multiple arguments as 3rd argument of ```service.broadcast``` should be an array of clients to broadcast to :)

Speaking about third argument, looks like we need to revise our ```forApp``` and ```appInstanceId``` logic here - currently you broadcast to _all_ clients, so if we have two clients (app and activity) and we receive message, both iframes broadcast ```message-received``` and ```MessageService``` will get two events and .... broadcast them to both ```MessagingClient``` instances :) Likely we need ```Map<appInstanceId, bridgeClient>``` logic somewhere as well.

Also I see that ```alwaysLowered``` will not help here as well since both main app and iframe will have its own hidden window (need to talk to Francisco and Etienne to understand if it's expected that hidden window is tied to originating instance):(

apps/sms/views/shared/js/startup.js
azasypkin: I'm wondering why we should wait until ```MessagingService``` confirms that event listeners are registered? Can't we just call ```MessagingService.register()``` (maybe we can choose more generic method name like 'register' instead of 'registerEvents') and don't wait for the return result?

Your approach looks good enough for me, the only alternative option (it's likely worse :), but I put it on the table anyway) I had in mind is to invert dependency here: once iframe/alwaysLowered-window/shim is created it establishes connection with MessagingService by itself and transmits events via this dedicated _pipe_ (including appInstanceId to indicate target for the event). MessagingService _optionally_ can have additional dedicated internal service to handle this _event pipe_. But as I said in the beginning it looks awkward and complex :p

steveck-chung: I think I mess it with my previous implementation that assign the client's on method to service's on method and it will need to wait for the service is actually ready. So yes you're right that client side doesn't need to wait for the service side event registration ready with current implementation.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0c8e181b7823b1964fc4c0990b2f77df15a59202

apps/sms/services/js/moz_mobile_message/moz_mobile_message_shim.js
```diff
@@ -71,6 +71,22 @@ function cloneThread(thread) {
...
+function clone(objectToBeCloned) {
+  // Basis checking.
+  if (!(objectToBeCloned instanceof Object)) {
+    return objectToBeCloned;
+  }
+
+  var objectClone = {};
+
+  // Clone each property.
+  for (var prop in objectToBeCloned) {
```
julienw: actually I'm not sure this works; I think properties for messages are not enumerable :/ (but please try it, maybe it is now)

julienw: maybe with https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyNames ? but I wonder if some properties come from the prototype ... or some properties that we don't want..

apps/sms/services/js/messaging/messaging_service.js
```diff
@@ -65,6 +70,20 @@ var MessagingService = {
     this.initService();
   },
 
+  register(appInstanceId) {
+    var mobileMessageClient = MozMobileMessageClient.forApp(appInstanceId);
+
+    EVENTS.forEach((event) => {
+      mobileMessageClient.on(event, (data) => {
+        // Workaround to avoid the duplicated events by inserting appInstanceId
+        // into event name in bridge mixin.
```
julienw: can you explain where the duplicated events come from ?

steveck-chung: From main instance and activity instance if it exists. In this case iframe in both instances will fire the event to messaging service, so both instance will receive same message event twice. In this workaround we gave the event with specific app instance id to make sure the event only triggered once to the corresponding instance.

apps/sms/services/js/moz_mobile_message/moz_mobile_message_shim.js
steveck-chung: Well, these properties are actually Inherited enumerable, so you can't list it even with `Object.getOwnPropertyNames`. The only way is using `for...in` in this case. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Enumerability_and_ownership_of_properties for more details.

julienw: OK, I'm fine if it works like this; I thought it was not enumerable at all.

apps/sms/services/js/messaging/messaging_service.js
julienw: ok, got it; this should not happen once we have one lowered window , right ?

steveck-chung: If there will be one lowered window for all the instances' iframe, then yes. But it seem that lowered window opened in different instance will still duplicated per my experiment.

julienw: yeah, but this should be fixed, but anyway this is for later.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9a5a2d103e1cd9e0f33c8f1ea935093824fcecae

apps/sms/services/js/conversation/conversation_service.js
```diff
@@ -52,6 +52,9 @@
     ):
     () => {};
 
+  const MESSAGE_EVENTS = ['message-sending', 'message-received',
+    'threads-deleted'];
```
julienw: nit: maybe put this on 5 lines

apps/sms/services/js/conversation/conversation_service.js
```diff
@@ -102,6 +106,20 @@
...
+      MESSAGE_EVENTS.forEach((event) => {
+        mobileMessageClient.on(event, (data) => {
+          // Workaround to avoid the duplicated events by inserting
+          // appInstanceId into event name in bridge mixin.
+          data.appInstanceId = appInstanceId;
+          this.broadcast(event, data);
+        });
+      });
+      return;
+    },
```
julienw: In https://github.com/mozilla-b2g/gaia/blob/master/apps/sms/services/README.md#conversationservice we did want to send events only when receiving a message, and when changing an existing message.

That said I understand why you want to change this, but then we need to change the README as well, and maybe think this through. Maybe we don't want message-sending/message-received but message-created ?

julienw: also I think we should send back the full affected message together with the event.

PR #31814, opened 2015-09-12
Bug 1191011 - Add IntlHelper

zbraniecki:
https://bugzilla.mozilla.org/show_bug.cgi?id=1191011

### DIFF
--- a/apps/sharedtest/test/unit/intl_helper_test.js
+++ b/apps/sharedtest/test/unit/intl_helper_test.js
@@ -1,0 +1,280 @@
+'use strict';
+
+/* global IntlHelper */
+
+require('/shared/js/intl_helper.js');
+
+suite('IntlHelper', function() {
+
+  beforeEach(function() {
+    IntlHelper._resetObjectCache();
+  });
+
+  suite('define', function() {
+    test('should work for a known type', function() {
+      assert.doesNotThrow(function() {
+        IntlHelper.define('shortDate', 'datetime', {
+          'month': 'numeric',
+          'year': '2-digit'
+        });
+      });
+    });
+    test('should throw for an unknown type', function() {
+      assert.throws(function() {
+        IntlHelper.define('shortDate', 'unknownType', {
+          'month': 'numeric',
+          'year': '2-digit'
+        });
+      });
+    });
+  });
+
+  suite('get', function() {
+    setup(function() {
+      IntlHelper.define('shortDate', 'datetime', {
+        'month': 'numeric',
+        'year': '2-digit'
+      });
+    });
+
+    test('should work for a defined name', function() {
+      assert.doesNotThrow(function() {
+        IntlHelper.get('shortDate');
+      });
+    });
+    test('should return a formatter with the right type', function() {
+      var formatter = IntlHelper.get('shortDate');
+      assert.isTrue(formatter instanceof Intl.DateTimeFormat);
+    });
+    test('should return a formatter with the right options', function() {
+      var formatter = IntlHelper.get('shortDate');
+      var options = formatter.resolvedOptions();
+      assert.equal(options.month, 'numeric');
+      assert.equal(options.year, '2-digit');
+      assert.strictEqual(options.hour, undefined);
+      assert.strictEqual(options.hour12, undefined);
+    });
+    test('should throw for an undefined name', function() {
+      assert.throws(function() {
+        IntlHelper.get('unknownName');
+      });
+    });
+  });
+
+  suite('observe', function() {
+    setup(function() {
+      IntlHelper.define('shortDate', 'datetime', {
+        'month': 'numeric',
+        'year': '2-digit'
+      });
+    });
+
+    test('should work for a defined name', function() {
+      assert.doesNotThrow(function() {
+        IntlHelper.observe('shortDate', () => {});
+      });
+    });
+    test('should throw for an undefined name', function() {
+      assert.throws(function() {
+        IntlHelper.observe('undefinedName', () => {});
+      });
+    });
+    test('should throw if one of the names is undefined', function() {
+      assert.throws(function() {
+        IntlHelper.observe(['shortDate', 'undefinedName'], () => {});
+      });
+    });
+    test('should fire when affected', function(done) {
+      IntlHelper.observe('shortDate', done);
+
+      IntlHelper.handleEvent({type: 'languagechange'});
+    });
+    test('should fire once even when defined twice', function() {
+      var firedCount = 0;
+      function test() {
+        firedCount++;
+      }
+      IntlHelper.observe('shortDate', test);
+      IntlHelper.observe('shortDate', test);
+
+      IntlHelper.handleEvent({type: 'languagechange'});
+
+      assert.equal(firedCount, 1);
+    });
+    test('should next event even when the the first one errors', function() {
+      var firedCount = 0;
+      function test() {
+        var x = unknownVariable + 1; // jshint ignore:line
+        firedCount++;
+      }
+      function test2() {
+        firedCount++;
+      }
+      IntlHelper.observe('shortDate', test);
+      IntlHelper.observe('shortDate', test2);
+
+      // be nice to our log, avoid spamming our error
+      sinon.stub(console, 'error');
+      IntlHelper.handleEvent({type: 'languagechange'});
+      console.error.restore();
+
+      assert.equal(firedCount, 1);
+    });
+  });
+
+  suite('unobserve', function() {
+    setup(function() {
+      IntlHelper.define('shortDate', 'datetime', {
+        'month': 'numeric',
+        'year': '2-digit'
+      });
+    });
+
+    test('should not fire after unobserved', function() {
+      var wasFired = false;
+
+      function test() {
+        wasFired = true;
+      }
+      IntlHelper.observe('shortDate', test);
+      IntlHelper.unobserve('shortDate', test);
+      IntlHelper.handleEvent({type: 'languagechange'});
+
+      assert.isFalse(wasFired);
+    });
+  });
+
+  suite('DateTime', function() {
+    setup(function() {
+      IntlHelper.define('shortTime', 'datetime', {
+        'hour': 'numeric',
+        'minute': 'numeric'
+      });
+
+      IntlHelper.define('shortDate', 'datetime', {
+        'month': 'numeric',
+        'year': '2-digit'
+      });
+    });
+
+    test('should only fire hour formatter on timeformatchange', function() {
+      var setTimeFired = false;
+      var setDateFired = false;
+      function setTime() {
+        setTimeFired = true;
+      }
+
+      function setDate() {
+        setDateFired = true;
+      }
+
+      IntlHelper.observe('shortTime', setTime);
+      IntlHelper.observe('shortDate', setDate);
+
+      IntlHelper.handleEvent({type: 'timeformatchange'});
+
+      assert.isTrue(setTimeFired);
+      assert.isFalse(setDateFired);
+    });
+    test('should fire if at least one name is affected', function() {
+      var setValuesFired = false;
+      function setValues() {
+        setValuesFired = true;
+      }
+
+      IntlHelper.observe(['shortDate', 'shortTime'], setValues);
+      IntlHelper.handleEvent({type: 'timeformatchange'});
+
+      assert.isTrue(setValuesFired);
+    });
+    test('should not fire if timeformatchange and no hour', function() {
+      var setValuesFired = false;
+      function setValues() {
+        setValuesFired = true;
+      }
+
+      IntlHelper.observe('shortDate', setValues);
+      IntlHelper.handleEvent({type: 'timeformatchange'});
+
+      assert.isFalse(setValuesFired);
+    });
+    test('should add hour12 to options', function() {
+      var expectedValue = false;
+      function setValues() {
+        var formatter = IntlHelper.get('shortTime');
+        assert.equal(formatter.resolvedOptions().hour12, expectedValue);
+      }
+
+      IntlHelper.observe('shortTime', setValues);
+      navigator.mozHour12 = false;
+      IntlHelper.handleEvent({type: 'timeformatchange'});
+
+      navigator.mozHour12 = true;
+      expectedValue = true;
+      IntlHelper.handleEvent({type: 'timeformatchange'});
+    });
+  });
+
+  suite('Number', function() {
+    setup(function() {
+      IntlHelper.define('downloadPercent', 'number', {
+        'style': 'percent',
+      });
+    });
+
+    test('should fire formatter on languagechange', function() {
+      var setValuesFired = false;
+      function setValues() {
+        setValuesFired = true;
+      }
+
+      IntlHelper.observe('downloadPercent', setValues);
+      IntlHelper.handleEvent({type: 'languagechange'});
+
+      assert.isTrue(setValuesFired);
+    });
+    test('should not fire formatter on timeformatchange', function() {
+      var setValuesFired = false;
+      function setValues() {
+        setValuesFired = true;
+      }
+
+      IntlHelper.observe('downloadPercent', setValues);
+      IntlHelper.handleEvent({type: 'timeformatchange'});
+
+      assert.isFalse(setValuesFired);
+    });
+  });
+
+  suite('Collator', function() {
+    setup(function() {
+      IntlHelper.define('timezoneSorter', 'collator', {
+        'usage': 'sort',
+        'sensitivity': 'variant',
+      });
+    });
+
+    test('should fire formatter on languagechange', function() {
+      var setValuesFired = false;
+      function setValues() {
+        setValuesFired = true;
+      }
+
+      IntlHelper.observe('timezoneSorter', setValues);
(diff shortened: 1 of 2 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=95ab940b2da6aae8923fc0f7848d0850c7aba402

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9c0f376edff1162bf5e020ad9a21f6418015e2d2

shared/js/intl_helper.js
```diff
@@ -0,0 +1,148 @@
...
+    var formatter = IntlHelper.getObject('shortDate');
+    console.log(formatter.format(new Date()));
+  }
+
+  setValue();
+
+  IntlHelper.on('shortDate', setValue);
+  
+   ****************************/
+  var objectCache;
```
zbraniecki: I switched it to 'objects' instead of 'formatters' because Collator is not a formatter, and I want to also support other intl objects in the future (DisplayNames may be one) etc.

shared/js/intl_helper.js
```diff
@@ -0,0 +1,148 @@
...
+            break;
+          default:
+            throw new Error('Unknown type: ' + obj.type);
+        }
+      }
+      return obj.object;
+    },
+    on: function(names, cb) {
+      if (typeof names === 'string') {
+        names = [names];
```
stasm: What's the use-case for the polymorphic `names` arg?

stasm: Oh, I think I see it now:  `emit` takes the `type`, but `on` takes the `name`.  So instead of:

    IntlHelper.on('datetimechange', ...);

you now do:

    IntlHelper.on(['shortDate', 'longDate'], ...);

I'm not sure I have an opinion on this, but I do have two comments:

- you should make sure all `names` have been defined, otherwise I think line 95 will break (`obj.listeners` wil be undefined),

- you might want to pass the name of the object to the callback.  Looking at the sample code in the comment at the top of the file, what happens when I listen to more than one object changes?

```javascript
IntlHelper.on(['shortDate', 'longDate'], setValue);

function setValue() {
  var formatter = IntlHelper.getObject('shortDate');      // ← shortDate is hardcoded here.
  console.log(formatter.format(new Date()));
}
```

shared/js/intl_helper.js
```diff
@@ -0,0 +1,148 @@
...
+
+      objectCache.forEach((obj, name) => {
+        if (obj.type === type) {
+          obj.listeners.forEach(cb => affectedCallbacks.add(cb));
+        }
+      });
+
+      affectedCallbacks.forEach(cb => {
+        cb(...args);
+      });
```
stasm: Nit: you don't need curly braces if there's only one expression.

    affectedCallbacks.forEach(cb => cb(...args));

However, should we somehow handle callback which throw?

julienw: +1

shared/js/intl_helper.js
```diff
@@ -0,0 +1,148 @@
...
+
+        var obj = objectCache.get(name);
+
+        obj.listeners.add(cb);
+      }
+    },
+    off: function(name, cb) {
+      var obj = objectCache.get(name);
+
+      obj.listeners.remove(cb);
```
julienw: > you should make sure all names have been defined, otherwise I think line 95 will break (obj.listeners wil be undefined),

actually, same here

shared/js/intl_helper.js
julienw: > you should make sure all names have been defined, otherwise I think line 95 will break (obj.listeners wil be undefined)

there is the "if" below (but maybe you wrote your comment before gandalf changed it :) )

shared/js/intl_helper.js
```diff
@@ -0,0 +1,148 @@
...
+            console.log('new Format');
+            console.log(options);
+            obj.object = Intl.DateTimeFormat(
+              navigator.languages,
+              options
+            );
+            break;
+          default:
+            throw new Error('Unknown type: ' + obj.type);
+        }
```
julienw: nit: maybe move this switch/case in a separate function

shared/js/intl_helper.js
stasm: Ah, no, I must have missed it!

shared/js/intl_helper.js
```diff
@@ -0,0 +1,148 @@
...
+  window.addEventListener('timeformatchange', () => {
+    resetDateTime();
+    global.IntlHelper.emit('datetime', {reason: 'timeformatchange'});
+  });
+
+  window.addEventListener('languagechange', () => {
+    resetDateTime();
+    resetNumber();
+    global.IntlHelper.emit('number', {reason: 'languagechange'});
+    global.IntlHelper.emit('datetime', {reason: 'languagechange'});
```
julienw: are we emitting datetime twice if the user changes a language that also changes the time format ? (eg: en-US -> fr-FR) ? Should we throttle ?

shared/js/intl_helper.js
```diff
@@ -0,0 +1,148 @@
...
+   ****************************/
+  var objectCache;
+
+  var knownObjects = [
+    'datetime',
+    'number',
+  ];
+
+  global.IntlHelper = {
+    defineObject: function(name, type, options) {
```
julienw: The term "Object" is really vague here; maybe it's just as good to have "define": `IntlHelper.define`.

shared/js/intl_helper.js
```diff
@@ -0,0 +1,148 @@
...
+
+      objectCache.set(name, {
+        options,
+        type,
+        object: undefined,
+        listeners: new Set()
+      });
+    },
+
+    getObject: function(name) {
```
julienw: same here, maybe "get" is good enough ?

shared/js/intl_helper.js
zbraniecki: right now yes, but soon I'll land bug 1187539 which makes us not fire timeformatchange on languagechange, so I'm counting on that.

shared/js/intl_helper.js
julienw: not sure I like that we listen for the object _names_ (as opposed to the types). It seems to me that we care about the types more, the names are only a handle.

zbraniecki: So, the reason I went for names, is that when you write a function that handles reformatting, the function covers specific set of formatters, not formatter types.

What I mean by this is that if you imagine using IntlHelper in SMS, for example, you may have a function that formats date/time in Conversation view.
That function will handle one or two DateTimeFormatters with specific names.

Now, if IntlHelper.on takes `type` then it will fire your reformatting function whenever *any* datetimeformat changes. If it takes a list of names, it will fire only when one of the affected datetimeformats change. It seems to me that desigining the API around names allows me to be smarter in the future.

One example use case I see is where there are two datetimeformatters, one with month/year/day and another with hour/minute. When timeformatchanges, I don't need to redefine the former one, so any callback that uses only the former one doesn't have to be fired at all - because month/year/day is not affected. But any callback that covers the latter should be - hour is affected.

If I follow 'types' then I don't have this kind of control. I have to fire all callbacks and trigger all reformatting every time any datetimeformat changes.

zbraniecki: @julienw - if you think it's a piece of API worth further discussion, I'll be happy to pause here and move the conversation into the bug or to dev.gaia to find the best solution.

julienw: OK, now I understand better the technical problem you want to solve.
It's difficult to put words on what I don't like; mostly that the event names are the same than the object names. Maybe the name "on" is not correct anymore and something else could work better; maybe "onChange" ?

but I think it's really bikeshedding so please go ahead with your solution.

stasm: > If I follow 'types' then I don't have this kind of control. I have to fire all callbacks and trigger all reformatting every time any datetimeformat changes.

Is this how you want the code to work right now or how you want it to work in the future?  Because looking at line 107, it seems like you're firing all callbacks of specified type anyways?

In the future, we could probably use simple heuristics like: if the `reason` is `timeformatchange` and `'hour' in obj.object.resolvedOptions()` is true, then fire the callback.  Is this what you have in mind?

stasm: If so, perhaps a better way to think about this is not that we're listening to events (`on`, `off`) but that we're observing changes (`observe`, `disconnect` like in mutation observers or something like that).

stasm: However, I'm terrible at naming, so I'll leave the exact choice to you! :)

julienw: > Is this how you want the code to work right now or how you want it to work in the future?

As I understood, this was "in the future"; so that the API does not change later :)


"observe" could work for me too.

zbraniecki: > In the future, we could probably use simple heuristics like: if the reason is timeformatchange and 'hour' in obj.object.resolvedOptions() is true, then fire the callback. Is this what you have in mind?

That's exactly a kind of scenario I wanted to address.

zbraniecki: @stasm: "you might want to pass the name of the object to the callback. Looking at the sample code in the comment at the top of the file, what happens when I listen to more than one object changes?"

Not sure if I understand correctly your question, but the way it's supposed to work is almost like a decorator. You say "when Intl object with name 'shortDate' or 'longHour' changes, fire that callback". In emitting, I also use Set which means that I will not fire the same callback multiple times even if you do sth like:

'''
IntlHelper.observe(['shortDate', 'shortHour'], setValues);
IntlHelper.observe(['longHour', 'shortDate'], setValues);
'''

It will still get fired once.

Passing names of the objects to the callback is possible but I'm not sure if I see a use case (our favorite word recently!). If you defined the observer on IntlHelper, and IntlHelper fired your function, you should reformat all datetimes/numbers/collators inside. You shouldn't ever have to do any 'if (names.has("shortDate")){...}'.
Does it make sense?

stasm: Hmm, I'm not sure if I see the value of passing multiple names to `observe`.  Afer all, inside of the `setValues` callback in your example, I still need to call `IntlHelper.getObject('shortDate');` and `IntlHelper.getObject('shortHour');` separately.  Might just as well register two observers for each name, no?

zbraniecki: my reasoning here is that you will request those objects in different lines of code, maybe in different branches of if or for loop. But if you'll have to register two/three/more observers you will do this line by line and it would be tedious.
Since there's a chance that you will want one function to observe multiple Intl Objects, I wanted to design the API around that.

Does that make sense?

stasm: OK, I see your point. I'll leave this up to you, since I don't want to block.  My suggestion would be to start with a more strict API and see how it's used and only then potentially introduce the polymorphic array attribute.  It's easy to extend this API in this direction, but not the other way around.

It's a tiny detail though.

zbraniecki: I think I'm pretty certain it should look like this, but let's revisit it when I have a patch for review. I'll take into account that it may be altered and it shouldn't affect the code too much :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f0f8a04200a273b21e0f0aa61d018be9a4f59ca1

shared/js/intl_helper.js
```diff
@@ -0,0 +1,170 @@
...
+
+  setValue();
+
+  IntlHelper.observe('shortDate', setValue);
+  
+   ****************************/
+  var objectCache;
+
+  const knownObjects = {
+    'datetime': {
```
stasm: Nit: You don't need the quotes around keys which are valid identifiers.

shared/js/intl_helper.js
```diff
@@ -0,0 +1,170 @@
...
+      set: function(options) {
+        const customOptions = Object.assign({}, options);
```
stasm: Are you sure you want to use `const` in this helper?  Will this require `;version=1.8` when linking to it?

shared/js/intl_helper.js
```diff
@@ -0,0 +1,170 @@
...
+        throw new Error('Intl Object with name "' + name + '" is not defined');
+      }
+      const obj = objectCache.get(name);
+      
+      if (!knownObjects.hasOwnProperty(obj.type)) {
+        throw new Error('Unknown type: ' + obj.type);
+      }
+
+      if (obj.object === undefined) {
+        obj.object = knownObjects[obj.type].set(obj.options);
```
stasm: I thought `knownObjects[obj.type]` was a `Set` instance because of the `.set` method. Maybe rename it to something less ambiguous (`create`?).

shared/js/intl_helper.js
```diff
@@ -0,0 +1,170 @@
...
+    var formatter = IntlHelper.get('shortDate');
+    console.log(formatter.format(new Date()));
+  }
+
+  setValue();
+
+  IntlHelper.observe('shortDate', setValue);
+  
+   ****************************/
+  var objectCache;
```
stasm: As you know, 1) I'm not a fan of calling things "objects" and 2) I'm bad at naming.  But perhaps this could called `helperCache`?  Compare:

    var obj = objectCache.get(name);
    obj.object = ...

with:

    var helper = helperCache.get(name);
    helper.intlObject = ...

shared/js/intl_helper.js
```diff
@@ -0,0 +1,170 @@
...
+
+        var obj = objectCache.get(name);
+
+        obj.listeners.add(cb);
+      }
+    },
+
+    unobserve: function(name, cb) {
+      var obj = objectCache.get(name);
+
```
stasm: Nit: needless blank line.  Also, :+1 for the name `unobserve`!

shared/js/intl_helper.js
julienw: I'm sure this is not needed for "const" :)

shared/js/intl_helper.js
```diff
@@ -0,0 +1,170 @@
...
+      isAffected: function(obj, reason) {
+        return true;
+      }
+    },
+  };
+
+  const affectedTypes = {
+    'languagechange': ['datetime', 'number', 'collator'],
+    'timeformatchange': ['datetime'],
+  };
```
stasm: Can we consider removing `affectedTypes`?  You can achieve the same thing via `isAffected` (which I really like because it encapsulates the logic inside of `knownObjects`) and the code could be simpler and easier to maintain.

shared/js/intl_helper.js
```diff
@@ -0,0 +1,170 @@
...
+      objectCache = undefined;
+    }
+  };
+
+  function resetObjects(types, reason) {
+    const affectedObjects = new Set();
+
+    objectCache.forEach((obj, name) => {
+      if (types.indexOf(obj.type) !== -1 &&
+          knownObjects[obj.type].isAffected(obj, reason)) {
```
stasm: Yeah, since you're iterating over `objectCache` anyways here, I'd just remove `affectedTypes` altogether and rely on `isAffected` to do the filtering.

shared/js/intl_helper.js
stasm: If so, that's great :) How about we make `objectCache` a const Map too, then, and use the `clear` method to reset it?

apps/sharedtest/test/unit/intl_helper_test.js
```diff
@@ -0,0 +1,264 @@
...
+    test('should throw if one of the names is undefined', function() {
+      assert.throws(function() {
+        IntlHelper.observe(['shortDate', 'undefinedName'], () => {});
+      });
+    });
+    test('should fire when affected', function(done) {
+      function test() {
+        done();
+      }
+      IntlHelper.observe('shortDate', test);
```
stasm: Put `done` directly as the callback to `observe`?

shared/js/intl_helper.js
```diff
@@ -0,0 +1,170 @@
...
+    },
+
+    unobserve: function(name, cb) {
+      var obj = objectCache.get(name);
+
+      obj.listeners.delete(cb);
+    },
+
+    _fireObservers: function(affectedObjects, ...args) {
+      var affectedCallbacks = new Set();
```
stasm: `const`?

stasm: (Also in other places.)

shared/js/intl_helper.js
```diff
@@ -0,0 +1,170 @@
...
+
+      affectedObjects.forEach((obj, name) => {
+        obj.listeners.forEach(cb => affectedCallbacks.add(cb));
+      });
+
+      affectedCallbacks.forEach(cb => cb(...args));
```
stasm: What about error handling?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=03bf7e3d95151c861b1fcb96aa8d9a28ec3d271a

shared/js/intl_helper.js
zbraniecki: I guess I'm paranoid about memory usage and wanted to keep it undefined unless someone uses it. But then, if someone is loading the helper, they should use it, right? I'll make the change.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=50fa6d1dc7532ece58ad689ccfabe10c14797b98

PR #31820, opened 2015-09-13
Bug xxx use conversation and shim services in window

azasypkin:


autolander:

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=185ce44b89cfd2fa8b6096ecc1770483436c0fb8

apps/sms/index.html
```diff
@@ -84,6 +85,21 @@
...
+    <script defer src="lib/bridge/plugins/stream/service.js"></script>
+    <script defer src="lib/bridge/plugins/finalize/client.js"></script>
+    <script defer src="services/js/bridge_service_mixin.js"></script>
+    <script defer src="services/js/moz_mobile_connections/moz_mobile_connections_shim.js"></script>
+    <script defer src="services/js/moz_mobile_message/moz_mobile_message_shim.js"></script>
+    <script defer src="services/js/moz_settings/moz_settings_shim.js"></script>
+    <script defer src="services/js/moz_mobile_message/moz_mobile_message_client.js"></script>
+    <script defer src="services/js/conversation/conversation_service.js"></script>
+    <script defer src="services/js/conversation/conversation_client.js"></script>
+    <script defer src="services/js/messaging/messaging_service.js"></script>
```
azasypkin: I've moved MessagingService to the main file list to just check how bad it's in terms of performance to have MessaginService-in-window right away. So measuring. If it's bad we can move it to lazy loading back.

steveck-chung: I don't think it will bring obvious penalty since we'll uglyfy and concatenate the scripts in production build. Just wondering it's just for or we'll need the messaging service in the main startup call path in this patch?(sorry I didn't read whole patch but I can find why we must have this in window context)

azasypkin: I'm not sure yet, likely not (only if we want to show something in Inbox if message is failed to send or something similar), I did it just to check if I can generalize ServiceManager to support as much services/clients as possible, but yeah there is no any reason to initialize it on startup.

apps/sms/services/js/service_manager.js
```diff
@@ -0,0 +1,239 @@
...
+
+  const priv = {
+    upgradeKey: Symbol('upgradeKey')
+  };
+
+  const UPGRADE_KEY = 'upgrade-key';
+
+  // Allow old clients to gracefully finish all pending operations before
+  // we destroy them.
+  var ConversationServiceInitializer = {
```
steveck-chung: It seems like Initializers have initialize/upgrade methods but doing slightly different things. I'm thinking of spiting the service/shim into different prototype or mixin, not sure if it could reduce the duplicate part.

apps/sms/services/js/threads.js
```diff
@@ -6,16 +6,24 @@
...
-    var key;
-
-    for (var i = 0; i < length; i++) {
-      key = Thread.FIELDS[i];
-      this[key] = thread[key];
-    }
+  function extractSupportedFields(
+    { id, body, timestamp, participants, lastMessageType, status, isDraft }
+  ) {
+    return {
```
steveck-chung: Seems `lastMessageSubject` is missing?

apps/sms/services/js/threads.js
```diff
@@ -6,16 +6,24 @@
...
+      id,
+      body,
+      timestamp,
+      participants,
+      lastMessageType,
+      status,
+      isDraft: !!isDraft
```
steveck-chung: BTW since you move all the status like flag into status, how about `isDraft` ?

apps/sms/views/conversation/js/startup.js
```diff
@@ -59,7 +53,14 @@
...
 
     MessageManager.getThreads({
-      each: (thread) => Threads.set(thread.id, thread),
+      each: (thread) => Threads.set(thread.id, {
+        id: thread.id,
+        participants: thread.participants,
+        body: thread.body,
+        timestamp: thread.timestamp,
+        status: { hasUnread: thread.unreadCount > 0, hasNewError: false },
+        lastMessageType: thread.lastMessageType
```
steveck-chung: 'lastMessageSubject'?

apps/sms/services/js/service_manager.js
```diff
@@ -0,0 +1,239 @@
...
+
+  const serviceInitializers = new Map([
+    ['conversation', ConversationServiceInitializer],
+    ['messaging', MessagingServiceInitializer],
+    ['moz-mobile-connections', MozMobileConnectionsServiceInitializer],
+    ['moz-settings', MozSettingsServiceInitializer]
+  ]);
+
+  function initializeShims(endpoint, upgradeKey) {
+    if (endpoint.MozMobileMessageShim) {
```
steveck-chung: Seems like we can run this three shim init within loop.

apps/sms/services/js/service_manager.js
```diff
@@ -0,0 +1,239 @@
...
+      debug('Upgrade is requested');
+
+      if (this[priv.upgradeKey]) {
+        return Promise.resolve();
+      }
+
+      var upgradeKey = App.instanceId;
+
+      return getShimEndpoint().then((endpoint) => {
+        initializeShims(endpoint, upgradeKey);
```
steveck-chung: Not sure if it's a proper timing for shim init because we run upgrade when fully loaded. Some api like settings.setReadAheadThreadRetrieval might be needed before fully loaded, but we can definitely change the timing of calling API.

steveck-chung: And, dose conversation service getAllConversations method really work for the if we init the mobile message shim here?

apps/sms/services/js/threads.js
azasypkin: I remember @julienw was against "lastMessageSubject" since we don't use it, so I removed it in this patch :) I can return it back if you guys think it's necessary.

apps/sms/services/js/threads.js
azasypkin: mmm, yeah maybe, I don't have preference. I didn't touch it in this patch because I'm not sure how we're going to handle it later - my initial idea was to replace this property with real "draft" property that will be returned from ConversationService, but it's too big change for this patch.

So just let me know if you want me to move this property to ```status``` :)

apps/sms/services/js/service_manager.js
azasypkin: Yep, why not, we can have an array with all related info (ServiceClass and related MozAPI)

apps/sms/services/js/service_manager.js
azasypkin: Can you explain a bit more what you have in mind? I just roughly combined related functionality in "initializers", but I'm sure there is much better option somewhere :)

Here "initialize()" is the method that initializes service (either via SharedWorker or directly if service is hosted in the same context) and "upgrade" is for "proper" upgrade of the window servce/client to sharedWorker/alwaysLowered-based  one.

apps/sms/services/js/service_manager.js
azasypkin: > Not sure if it's a proper timing for shim init because we run upgrade when fully loaded.

We run "upgrade" on ```visuallyLoaded```. :)

> And, dose conversation service getAllConversations method really work for the if we init the mobile message shim here?

Yep, it woks :)

> Some api like settings.setReadAheadThreadRetrieval might be needed before fully loaded, but we can definitely change the timing of calling API. 

Sorry I had to explain general idea in advance, it's not that easy to see what is going on looking at the code only :/

So the idea is the following:
* We have _all_ necessary clients/services in window on startup; In this case all shims and clients listen on ```window``` only (no ```BC``` or ```MC``` involved);
* On visually loaded we run "upgrade" process:
  * Create alwaysLowered window - in this case we'll have copies of all shims, but they listen on ```BC``` - so no conflict here with existing in-window shims, nobody uses these services yet;
  * Once we have new shims ready (inside alwaysLowered window), we start to upgrade existing window clients:
    * Destroy window clients: with the help of my brand-new finalize (not the best name though) bridge client plugin, every client has ```finalize``` method that returns promise that is resolved once all pending operations (method calls, streams) are finished and client is disconnected from the corresponding service;
    * Immediately create "upgraded" client (either ```SharedWorker``` based like ```ConversationClient``` or alwaysLowered based like ```MozMobileConnectionsClient```). So at this step, we have two clients simultaneously: old that's waiting for pending operations to complete, and the new "upgraded" one that will handle all next requests.
    * Since we can have pending operations (we run upgrade on visuallyLoaded so ```getAllConversations``` stream is still active), we wait for those operations to complete _and only then_ we destroy/disconnect window client and corresponding window based service that was backing window client.
* Now all new requests are handled by "upgraded" clients.

So all this not-that-easy machinery and its temporary and workaround-y nature made me think that having this in single file is a good idea, but I maybe completely wrong :)

apps/sms/services/js/threads.js
steveck-chung: If so, why do we need to set the lastMessageSubject while creating the thread in fromDraft & fromMessage?

azasypkin: Hmm, right! Need to remove it from everywhere then.

apps/sms/services/js/service_manager.js
steveck-chung: Thanks for the explanation! I think I missed the part that it was existed in window context at first. I have one more question about the service manager upgrade that you didn't upgrade in the conversation view. I think we still need it for conversation view?

azasypkin: >  I have one more question about the service manager upgrade that you didn't upgrade in the conversation view. I think we still need it for conversation view?

Yep! I think I just forgot to about it, I've only tested for the current setup (shared/js/startup.js) and didn't bother much about split view until I get your feedback guys :)

autolander:

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7345ad85d8dac6c2f7eef6a4e0ab56bb006efee8

apps/sms/services/js/service_manager.js
```diff
@@ -0,0 +1,239 @@
...
+  };
+
+  const serviceInitializers = new Map([
+    ['conversation', ConversationServiceInitializer],
+    ['messaging', MessagingServiceInitializer],
+    ['moz-mobile-connections', MozMobileConnectionsServiceInitializer],
+    ['moz-settings', MozSettingsServiceInitializer]
+  ]);
+
+  function initializeShims(endpoint, upgradeKey) {
```
julienw: I guess this is more a nit, but I'd move this to ServiceManager as a method

apps/sms/services/js/threads.js
julienw: I'm not so sure now TBH :)

apps/sms/services/js/threads.js
julienw: I think it's good as a "top level" property; this is really one of the main characterstics of the object

apps/sms/views/conversation/js/startup.js
```diff
@@ -59,7 +53,14 @@
     var deferred = Utils.Promise.defer();
 
     MessageManager.getThreads({
-      each: (thread) => Threads.set(thread.id, thread),
+      each: (thread) => Threads.set(thread.id, {
```
julienw: I'd move this to Thread actually; maybe Thread.fromApiThread ?

apps/sms/desktop-mock/navigator_moz_mobilemessage_shim.js
```diff
@@ -7,12 +7,12 @@
 
 *********************************************************** */
 (function(window) {
- 
-  if (!parent.navigator.mozMobileMessage) {
+
+  if (!opener || !opener.navigator.mozMobileMessage) {
```
julienw: not really sure we should use opener; sorry if I miss something but where will this script run ?

azasypkin: mm, good question, I don't really remember already :) IIRC it was intended for alwaysLowered window

julienw: ok, in that case, I think we should not use this shim but the real mock; because the opener might just eventually disappear.

but maybe we should just use the mock everywhere -- after all we're in the desktop here, it's just for mocking.

apps/sms/services/js/conversation/conversation_service.js
```diff
@@ -89,8 +88,7 @@
       body: thread.body,
       timestamp: thread.timestamp,
       status: { hasUnread: thread.unreadCount > 0, hasNewError: false },
-      lastMessageType: thread.lastMessageType,
-      draft: Drafts.byThreadId(thread.id)
+      lastMessageType: thread.lastMessageType
```
julienw: is this temporary to try stuff ?

apps/sms/services/js/moz_mobile_message/moz_mobile_message_client.js
```diff
@@ -28,71 +31,98 @@
...
+  /**
+   * Destroys all clients.
+   * @static
+   */
+  MobileMessageClient.destroy = function() {
+    var finalizers = [];
+
+    MobileMessageClient[priv.instances].forEach(
+      (instance) => finalizers.push(instance[priv.client].finalize())
+    );
```
julienw: nit: use map ?

apps/sms/services/js/service_manager.js
```diff
@@ -0,0 +1,239 @@
...
+
+  const debug = 1 ?
+    (arg1, ...args) => console.log(`[ServiceManager] ${arg1}`, ...args):
+    () => {};
+
+  const priv = {
+    upgradeKey: Symbol('upgradeKey')
+  };
+
+  const UPGRADE_KEY = 'upgrade-key';
```
julienw: I admit I don't properly understand all these "upgrade key" variables or parameters. Are they serving different purposes ?

apps/sms/services/js/threads.js
```diff
@@ -6,16 +6,24 @@
...
-    var length = Thread.FIELDS.length;
-    var key;
-
-    for (var i = 0; i < length; i++) {
-      key = Thread.FIELDS[i];
-      this[key] = thread[key];
-    }
+  function extractSupportedFields(
+    { id, body, timestamp, participants, lastMessageType, status, isDraft }
+  ) {
```
julienw: nit: i'm not sure of this, I don't find this so readable; I'd rather keep Thread.FIELDS... (but you can keep the function though)

apps/sms/views/inbox/js/inbox.js
```diff
@@ -326,13 +327,13 @@ var InboxView = {
...
         if (!thread.isDraft) {
-          var isRead = thread.unreadCount > 0;
-          var l10nKey = isRead ? 'mark-as-read' : 'mark-as-unread';
-
           params.items.push(
             {
-              l10nId: l10nKey,
-              method: this.markReadUnread.bind(this, [threadId], isRead)
+              l10nId: thread.status.hasUnread ?
+                'mark-as-read' : 'mark-as-unread',
```
julienw: you didn't like the extra var ?

apps/sms/views/inbox/js/inbox.js
```diff
@@ -374,10 +375,7 @@ var InboxView = {
       var hasUnreadselected = selected.selectedList.some((id) => {
         var thread  = Threads.get(id);
 
-        if (thread && thread.unreadCount) {
-          return thread.unreadCount > 0;
-        }
-        return false;
+        return thread ? thread.status.hasUnread : false;
```
julienw: maybe `return thread && thread.status.hasUnread` works good enough ?

apps/sms/services/js/service_manager.js
azasypkin: private method or you have the idea where we want to call this method outside of ServiceManager internals?

apps/sms/services/js/threads.js
azasypkin: Ha-ha, okay I'll return it back everywhere and then re-confirm :)

apps/sms/views/conversation/js/startup.js
azasypkin: Yep, I think Thread will be the right place for this!

apps/sms/desktop-mock/navigator_moz_mobilemessage_shim.js
azasypkin: I think with little-browser, we don't have to care about this much :) This mock will be only in top level frame, and the rest should use it instead.

apps/sms/services/js/conversation/conversation_service.js
azasypkin: Mmm, what exactly?

apps/sms/services/js/moz_mobile_message/moz_mobile_message_client.js
azasypkin: mm, nope, ```MobileMessageClient[priv.instances]``` is a ```Map``` that has only ```forEach``` :/

apps/sms/services/js/service_manager.js
azasypkin: Well, ```updagrade key``` is another name for ```app instance id``` :) All these dances around the fact that we want to have window#1->SharedWorker->window#1#shim communication channel or if say in another way: we want SharedWorker to request shim in the same window that requested SharedWorker :)

I feel like we should ask Bridge to simplify this problem for us somehow - it's becoming crazy :/ I tried once, but didn't push strong enough :)

apps/sms/services/js/threads.js
azasypkin: Sure, I don't have strong opinion on that anyway :)

apps/sms/views/inbox/js/inbox.js
azasypkin: Yeah, as you know, I don't like extra var's :p Will return it back :)

apps/sms/views/inbox/js/inbox.js
azasypkin: Mmm yep :) Not sure why I have this...

PR #31821, opened 2015-09-13
Bug 1203999 - Subscribing to settings observer on show instead of creation r=gmarty

albertopq:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2d420bff0c39f025ef49933a8fe2851c0853eb61

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d011c2c2a087aa6cce31895e94ad4b24d3e6d371

apps/system/js/app_window.js
```diff
@@ -1026,6 +1026,7 @@
       }
       // Send event instead of call crash reporter directly.
       this.publish('crashed');
+      this.hideContextMenu();
```
vingtetun: Does this code is called when the app is killed from the card view ? If no, you may want to move that to destroy() instead.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a2af80acb5ee506ffbf5601a39ad921e4bacbed6

apps/system/js/browser_context_menu.js
```diff
@@ -51,6 +50,8 @@
         return;
       }
 
+      SettingsListener.observe(PINNING_PREF, '', this._setPinning.bind(this));
+
```
vingtetun: You can probably move this call right before the comment that says: // Notify the embedder we are handling ...

The reason for it is because we may do some early returns in various cases, and in this case I'm not sure it is useful to have the observer.

vingtetun: If you follow this comment, you may also want to move the code that remove the observer after the:
     if (!this.contextMenuView.isShown()) {
        return;
      } 
code in the hide method of BrowserContextMenu.

All that said I wonder if this may not generate a race if the settings takes some time to come back from the mozSettings database. In this case you may want to keep the initial observe code in the constructor as it was before this code. And the call to this.hideContextMenu you have added in appWindow will take care of cleaning it (I think it will be better to have a more explicit destroy() method in contextmenu that calls both hide and remove the observer so you will not add/remove the observer all the time in show/hide and explicitely kill the contextmenu to let it do its own cleanup).

albertopq: The problem here will be the same than the original bug. The browser_context_menu module get's instantiated many times, even before shown, so we can have several listeners to settings running on the same preference.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=fc12133ed48371db1c3bc289c12a565d5a40ea9e

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=574b01452d12a81d279e6dc1f49cfa8c478bef25

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f1dc42f05012f7323f8bc676d486f06dbfc6ed46

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f72ffc077045d5d52e718c42344517cb6dbe9693

apps/system/test/unit/browser_context_menu_test.js
```diff
@@ -173,7 +176,9 @@ suite('system/BrowserContextMenu', function() {
 
     var stubStopPropagation =
       this.sinon.stub(fakeContextMenuEvent, 'stopPropagation');
-
```
etiennesegonzac: I think the mock for mozSettings has a "sync" mode. Can you use this instead of stubing an internal method?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=50899e61a71bc2384508147cb7fc8847de6e28c6

apps/system/test/unit/browser_context_menu_test.js
```diff
@@ -304,10 +332,17 @@ suite('system/BrowserContextMenu', function() {
 
     var md1ShowStub = this.sinon.stub(md1.contextMenuView, 'show');
     var md2ShowStub = this.sinon.stub(md2.contextMenuView, 'show');
+    this.sinon.stub(md1, '_getPinningEnabled', function(cb) {
+      cb();
+    });
+    this.sinon.stub(md2, '_getPinningEnabled', function(cb) {
+      cb();
+    });
```
etiennesegonzac: we don't need the internal stubs here, this test should look like this
```js
  test('bookmark/share buttons hidden in private browser', function(done) {
    var app1 = new AppWindow(fakePrivateConfig);
    var md1 = BaseModule.instantiate('BrowserContextMenu', app1);
    md1.start();
    var app2 = new AppWindow(fakeBrowserConfig);
    var md2 = BaseModule.instantiate('BrowserContextMenu', app2);
    md2.start();

    var md1ShowStub = this.sinon.stub(md1.contextMenuView, 'show');
    var md2ShowStub = this.sinon.stub(md2.contextMenuView, 'show');
    Promise.all([
      md1.showDefaultMenu(),
      md2.showDefaultMenu()
    ]).then(() => {
      var md1Items = md1ShowStub.getCall(0).args[0];
      var md2Items = md2ShowStub.getCall(0).args[0];
      // We should not show the bookmark or share buttons.
      assert.equal(md2Items.length - md1Items.length, 2);
      done();
    });
    MockNavigatorSettings.mReplyToRequests();
  });

```

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=94df068d43420adc4fd8d1b52d33a11d9d5764de

PR #31837, opened 2015-09-14
Bug 1159862 - Fix contact divider margin

nrbrd:
Please, take a look if this is the correct solution.

The contact image: **li aside.pack-end** had a **2rem margin-right**. So the bottom border was using this same margin and then overlapping with the scrollbar.

So I moved the margin-right from **li aside** to **li**.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=528181d6ccd1835615847e7b0a95d56d4849c5da

PR #31841, opened 2015-09-14
Bug 1203060 - Make the launch code also work for the browser app

mwargers:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f0d1f2ef50317101c6a048c35990c95e30ca2b70

tests/python/gaia-ui-tests/gaiatest/tests/functional/browser/test_browser_launch_from_background.py
```diff
@@ -0,0 +1,37 @@
...
+
+        search.launch()
+
+        browser = search.go_to_url(self.test_url)
+        browser.switch_to_content()
+        Wait(self.marionette).until(lambda m: m.title == 'Mozilla')
+
+        self.device.touch_home_button()
+
+        search.launch()
```
ShakoHo: if you add search.go_to_url after line 33, it should raise exception. Because it cannot tap the search url after the browser already surf any website. but you can patch the app.py under search folder. the code as below, then it should work..

    def go_to_url(self, url):
        # The URL bar shown is actually in the system app not in this Search app.
        # We switch back to the system app, then tap the panel, but this will only
        # work from Search app which embiggens the input bar
        if self.is_element_present(*self._url_bar_locator):
            self.marionette.find_element(*self._url_bar_locator).tap()
        else:
            self._url_bar_locator = (By.CSS_SELECTOR, '.urlbar .title')
            self._root_element = self.marionette.find_element(*self._browser_app_locator)
            self._root_element.find_element(*self._url_bar_locator).tap()
        from gaiatest.apps.homescreen.regions.search_panel import SearchPanel
        search_panel = SearchPanel(self.marionette)
        return search_panel.go_to_url(url)

tests/atoms/gaia_apps.js
```diff
@@ -280,24 +270,17 @@ var GaiaApps = {
...
-          origin: origin};
+        let result = GaiaApps.getDisplayedApp();
         marionetteScriptFinished(result);
       };
 
-      if (GaiaApps.getDisplayedApp().origin == origin) {
+      let displayedApp = GaiaApps.getDisplayedApp();
+      if (displayedApp.manifestURL == manifestURL ||
+          (!displayedApp.manifestURL && displayedApp.origin == origin)) {
         console.log('app with origin \'' + origin + '\' is already running');
```
zapion: We compare manifestURL, probably need to log more than origin since it's secondary condition.

tests/python/gaia-ui-tests/gaiatest/tests/functional/browser/test_browser_launch_from_background.py
mwargers: Ok, I will do this. Note that this will pollute browser locators to the search app.
But I think this whole distinction between Browser and Search is not really logical, they are almost the same.
I'll file a bug on removing the distinction between the two.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4f37ed10705bd3b9445cee5c3b8e14ce1ea84bbc

PR #31894, opened 2015-09-17
Bug 1204002 - View source cannot find file in calendar app r=jrburke

lambdabaa:
So the issue was that calendar was using fake urls like `/month/` and `/event/add` in concert with this page.js library. With page.js, you create a bunch of "routes" that say stuff like when you notice that the url is `/month/`, call this function. Prior to page.js version 1.3, hashes (fragment ids) weren't supported. As of 1.4, there's this neat option called `hashbang` that you can set on page.js that makes it so that
1. All anchor elements with href='/xyz' go to `/#!/xyz` instead of `/xyz`.
2. Any time you tell the page.js router to navigate to `/xyz` it will navigate to `/#!/xyz` instead.

This seems to make all of the navigation happen correctly in the url fragment so that we're still always at `index.html` and view source can map back to an actual "server route".


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8c8b3a455536cd092cc6ccb8e96ae5fb1fa7f0e4

PR #31896, opened 2015-09-17
Bug 1191773 - Implement History Adapter for FxSync feature.

weilonge:


autolander:

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e16ac72565c3fa7863ff9919d5ff04aef72b87de

autolander:

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=684b0d8e2c318c935e7be7710292f41cc71f7de9

autolander:

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3229b2e5b2681be5b645c74288846e03d3f7abc2

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9332c41e1e2490da0cd1504a9856ac21f621d289

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,98 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+/* global IAC,
```
ferjm: Where is IAC coming from? Did you forget to `git add` something maybe?

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,98 @@
...
+  SyncEngine
+*/
+
+var HistoryHelper = {
+  set syncedCollectionModifiedTime(mtime) {
```
ferjm: This is a long variable name :D

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,98 @@
...
+    window.asyncStorage.setItem('LastSyncedStatus::Collection::mtime', mtime);
```
ferjm: Use a const for the item key, please.

`const MTIME = 'LastSyncedStatus::Collection::mtime'`

Also the name is too generic. I guess we may need to store a different value for bookmarks and/or passwords, right?

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,98 @@
...
+  },
+  get syncedCollectionModifiedTime() {
+    return new Promise((resolve, reject) => {
```
ferjm: nit: reject is not used.

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,98 @@
...
+      window.asyncStorage.
+      getItem('LastSyncedStatus::Collection::mtime', (mtime) => {
```
ferjm: nit: single argument don't need parenthesis `mtime => {...}`

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,98 @@
...
+        resolve(mtime);
+      });
```
ferjm: nit: you can write this like `asyncStorage.getItem(MTIME, resolve);`

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,98 @@
...
+      args: [places]
+    });
+  },
+
+  addPlace(place) {
+    return IAC.request('sync-history', {
+      method: 'addPlace',
+      args: [place]
+    });
+  }
```
ferjm: Can we get rid of `addPlace` and always use `addPlaces`? It seems that the only difference is the number of items in the `place`/`places` arguments.

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,98 @@
...
+};
+
+SyncEngine.DataAdapterClasses.history = {
+  _fullSync(kintoCollection, lastModifiedTime) {
```
ferjm: Could you add a comment explain what this algorithm does and why is it doing it that way, please?

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,98 @@
...
+    function updateHistoryCollection(list) {
+      var historyRecords = list.data;
+      var places = [];
+      var i;
+      for (i = 0; i < historyRecords.length; i++) {
+        if (historyRecords[i].last_modified <= lastModifiedTime) {
```
ferjm: I wonder if kinto.js could handle this for us. `.list()` already take filters [1]. Could you check if it's possible to get the entries with `last_modified > lastModifiedTime` and if it isn't could you file a bug to fix that, please?

[1] http://kintojs.readthedocs.org/en/latest/api/#listing-records

ferjm: In any case, IIUC `.list()` gives us the history entries in reverse `last_modified` order. I guess you could do
     
     var i = 0;
     while ((i < historyRecords.length) && (historyRecords[i].last_modified > lastModifiedTime)) {
        historyRecords.shift();
        i++;
     }

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,98 @@
...
+          break;
+        }
+      }
+      var partialRecords = historyRecords.slice(0, i);
+      partialRecords.forEach((decryptedRecord) => {
+        var record = decryptedRecord.payload;
+        if (!record.histUri || !record.visits || !record.visits[0]) {
+          return;
```
ferjm: console.warn this case, please.

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,98 @@
...
+          visits.push(Math.floor(elem.date / 1000));
+        });
+
+        var place = {
+          url: record.histUri,
+          title: record.title,
+          visits: visits,
+          fxsyncId: record.id
+        };
+        places.push(place);
```
ferjm: nit: no need for `place` var, you can simply do `places.push({/* place object properties */});`

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,98 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+/* global IAC,
+  SyncEngine
+*/
+
+var HistoryHelper = {
```
michielbdejong: Should this helper not go inside a closure?

apps/sync/js/adapters/history.js
michielbdejong: nit: indent 4 spaces here?

apps/sync/js/adapters/history.js
michielbdejong: The `lastModifiedTime` here refers to the DataStore, not to the `kintoCollection`, right?

apps/sync/js/adapters/history.js
ferjm: Also, now all records have `last_modified` because we are taking everything from the server (readonly). But it would be good to shield it just in case we have records with a missing `last_modified` property.

From kinto.js doc: `Records with last_modified attribute were sync'ed on a server.`

michielbdejong: Why the `for` loop followed by a `slice` and a `forEach`? Maybe you can use just one `for` loop to do the work currently done by the `forEach` loop, and `break` from that when you reach a record that's old enough?

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,98 @@
...
+      partialRecords.forEach((decryptedRecord) => {
+        var record = decryptedRecord.payload;
+        if (!record.histUri || !record.visits || !record.visits[0]) {
+          return;
+        }
+
+        var visits = [];
+        record.visits.forEach((elem) => {
+          visits.push(Math.floor(elem.date / 1000));
+        });
```
michielbdejong: May be nicer to use `Array.map` instead of `Array.forEach` plus `Array.push`? Something like:
````js
var visits = record.visits.map(elem => Math.floor(elem.date / 1000));
````

File: apps/sync/js/adapters/history.js [hunk 14]
```diff
@@ -0,0 +1,98 @@
...
+      var historyRecords = list.data;
+      var places = [];
+      var i;
+      for (i = 0; i < historyRecords.length; i++) {
+        if (historyRecords[i].last_modified <= lastModifiedTime) {
+          break;
+        }
+      }
+      var partialRecords = historyRecords.slice(0, i);
+      partialRecords.forEach((decryptedRecord) => {
```
ferjm: nit: the fact that records are decrypted at this point is not very relevant.
s/decryptedRecord/record and s/record/payload please

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,98 @@
...
+        var record = decryptedRecord.payload;
+        if (!record.histUri || !record.visits || !record.visits[0]) {
```
ferjm: s/!record.visits[0]/!record.visits.length

apps/sync/js/adapters/history.js
michielbdejong: Or even directly:
````js
places.push({
  url: record.histUri,
  title: record.title,
  visits: record.visits.map(elem => Math.floor(elem.date / 1000)),
  fxsyncId: record.id
});
````

File: apps/sync/js/adapters/history.js (hunk 10 above)
Comment (ferjm) [inline apps/sync/js/adapters/history.js]: I guess this is a corrupted record, do we need to delete it?

File: apps/sync/js/adapters/history.js (hunk 7 above)
Comment (ferjm) [inline apps/sync/js/adapters/history.js]: In fact, it seems that `addPlace` is not used anywhere.

File: apps/sync/js/adapters/history.js [hunk 16]
```diff
@@ -0,0 +1,98 @@
...
+    function updateHistoryCollection(list) {
+      var historyRecords = list.data;
+      var places = [];
+      var i;
+      for (i = 0; i < historyRecords.length; i++) {
+        if (historyRecords[i].last_modified <= lastModifiedTime) {
+          break;
+        }
+      }
+      var partialRecords = historyRecords.slice(0, i);
```
ferjm: I think you can `return Promise.resolve(false)` here `if !partialRecords.length`

apps/sync/js/adapters/history.js
michielbdejong: Is it a good idea to use a setter for an asynchronous operation? Maybe just use a function

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,98 @@
...
+        };
+        places.push(place);
+      });
+
+      if (places.length === 0) {
+        return Promise.resolve(false);
+      }
+
+      return HistoryHelper.addPlaces(places).then(() => {
+        if (partialRecords.length > 0) {
```
ferjm: If we early return at line 50, you can get rid of this.

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,98 @@
...
+          HistoryHelper.syncedCollectionModifiedTime =
+            partialRecords[0].last_modified;
+        }
+        return Promise.resolve(false);
+      });
+    }
+
+    return kintoCollection.list().then(list => {
+      return updateHistoryCollection(list);
+    });
```
ferjm: `return kintoCollection.list().then(updateHistoryCollection);`

apps/sync/js/adapters/history.js
michielbdejong: With this getter, the value of `HistoryHelper.syncedCollectionModifiedTime` is a promise, that's confusing to me. I would use a normal function instead.

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,98 @@
...
+      if (places.length === 0) {
+        return Promise.resolve(false);
+      }
+
+      return HistoryHelper.addPlaces(places).then(() => {
+        if (partialRecords.length > 0) {
+          HistoryHelper.syncedCollectionModifiedTime =
+            partialRecords[0].last_modified;
+        }
+        return Promise.resolve(false);
```
michielbdejong: Notice you're resolving before the setter completed. It's unlikely to lead to race conditions in real usage, but it still feels incorrect.

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,98 @@
...
+      });
+    }
+
+    return kintoCollection.list().then(list => {
+      return updateHistoryCollection(list);
+    });
+  },
+  update(kintoCollection) {
+    return HistoryHelper.syncedCollectionModifiedTime.then((mtime) => {
```
michielbdejong: This is where the way you use the getter looks funny to me. :)

apps/sync/js/adapters/history.js
michielbdejong: Maybe you should handle `deleted: true` here? Or at least add a comment about why we're importing additions but not deletions.

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,98 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
```
michielbdejong: Maybe you can add a comment block with a description of:
* The fields of a record in the Kinto collection, and their meaning (including the `deleted: true` case).
* The fields of a Place object, and their meaning.
* The mapping we do between them.

It seems the only thing is converting visits from format:
`{ data: 1234567890123 }`
to format:
`1234567890`,
right?

apps/system/js/core.js
```diff
@@ -124,7 +124,8 @@
             'js/devtools/remote_debugger.js',
             'js/devtools/developer_hud.js',
             'js/devtools/devtools_auth.js',
-            'js/devtools/devtools_auth_dialog.js'
+            'js/devtools/devtools_auth_dialog.js',
+            'js/sync/sync_iac_api.js'
```
michielbdejong: Why add Sync IAC API in this PR? I would expect you add an IAC API just for Places?

apps/system/js/sync/sync_iac_api.js
```diff
@@ -0,0 +1,80 @@
...
+(function() {
+
+  function sendPortMessage(portName, message) {
+    var port = IACHandler.getPort(portName);
+    if (port) {
+      port.postMessage(message);
+    }
+  }
+
+  function syncHistroyRequest(id, method, args){
```
michielbdejong: Type 'Histroy'

michielbdejong: I would not add this in the System app. The only change needed in the System app afaics is that its current Places code needs to be exposed through IAC.

Actually, wait, not even that, right? Wasn't the latest discussion that Places code should be shared, and its `writeInProgress` variable should be removed, and replaced by using the DataStore correctly?

Then we don't need IAC at all in this PR.

apps/sync/test/unit/adapters/history_test.js
```diff
@@ -0,0 +1,174 @@
...
+  requireApp,
+  setup,
+  sinon,
+  suite,
+  teardown,
+  test,
+  SyncEngine
+*/
+
+var MockasyncStorage = {
```
ferjm: There are already a few asyncStorage mocks that you can reuse. There's actually one in shared that may be extended for your needs.

https://mxr.mozilla.org/gaia/find?text=&string=mock_async_storage

apps/system/manifest.webapp
```diff
@@ -147,6 +147,12 @@
       "rules": {
         "minimumAccessLevel": "certified"
       }
+    },
+    "sync-history": {
```
michielbdejong: If you want to do it with IAC, I would expose a connection called something like "places" (maybe investigate overlap with "search-results"), but preferably, I would avoid IAC and use the DataStore interface from the adapter, using shared code to make sure the data format of a Place is only defined in one place.

apps/sync/test/unit/adapters/history_test.js
```diff
@@ -0,0 +1,174 @@
...
+};
+
+var MockIAC = {
+  request(channel, data) {
+    return Promise.resolve();
+  }
+};
+
+window.SyncEngine = {
+  DataAdapterClasses:{}
```
ferjm: nit: space after `:`

apps/system/js/sync/places_adapter.js
```diff
@@ -0,0 +1,54 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/* globals appWindowManager */
+/* exported PlacesAdapter */
+
+'use strict';
+
+(function(exports) {
+  var PlacesAdapter = {
```
michielbdejong: This is the code that is shared between the System app and the History adapter, right? If you're newly adding this code in the systems app, should it not also be removed somewhere? Aren't `addPlace` and `addPlaces` already there?

apps/sync/test/unit/adapters/history_test.js
ferjm: Do this on the suite setup after saving the real SyncEngine and make sure you restore it on the suite teardown, please.

apps/sync/test/unit/adapters/history_test.js
```diff
@@ -0,0 +1,174 @@
...
+};
+
+requireApp('sync/js/adapters/history.js');
+
+suite('sync/adapters/history >', () => {
+  const SYNCED_STATUS_MTIME = 'LastSyncedStatus::Collection::mtime';
+  var realIAC, realasyncStorage, testCollectionData, iacRequestSpy;
```
ferjm: nit: s/realasyncStorage/realAsyncStorage

apps/sync/test/unit/adapters/history_test.js
```diff
@@ -0,0 +1,174 @@
...
+          visits: [{
+            date: (initDate + i * 10) * 1000, type: 3
+          }]
+        }
+      });
+    }
+    return list;
+  }
+
+  setup(() => {
```
ferjm: I think this should be `suiteSetup` if you only want it to happen once.

apps/sync/test/unit/adapters/history_test.js
```diff
@@ -0,0 +1,174 @@
...
+    realIAC = window.IAC;
+    realasyncStorage = window.asyncStorage;
+    window.IAC = MockIAC;
+    window.asyncStorage = MockasyncStorage;
+    testCollectionData = [];
+    iacRequestSpy = sinon.spy(MockIAC, 'request');
+  });
+
+  teardown(() => {
```
ferjm: Same here with `suiteTeardown`.

apps/sync/test/unit/adapters/history_test.js
```diff
@@ -0,0 +1,174 @@
...
+      list.unshift({
+        id: 'UNIQUE_ID_' + i,
+        last_modified: initDate + i * 10,
+        payload: {
+          id: 'UNIQUE_ID_' + i,
+          histUri: 'http://example' + i + '.com/',
+          title: 'Example ' + i + ' Title',
+          visits: [{
+            date: (initDate + i * 10) * 1000, type: 3
+          }]
```
ferjm: Can we test with more than one visit, please?

ferjm: Also, can we test missing fields?

apps/sync/test/unit/adapters/history_test.js
```diff
@@ -0,0 +1,174 @@
...
+    assert.equal(iacRequestSpy.args[0][0], 'sync-history');
+    assert.equal(iacRequestSpy.args[0][1].method, 'addPlaces');
+    return iacRequestSpy.args[0][1].args[0];
+  }
+
+  function verifyPlaces(collectionItem, placesItem) {
+    var payload = collectionItem.payload;
+    assert.equal(payload.histUri, placesItem.url);
+    assert.equal(payload.title, placesItem.title);
+    for(var i = 0; i < payload.visits.length; i++){
```
ferjm: nit: space after `for`

apps/system/js/sync/places_adapter.js
```diff
@@ -0,0 +1,54 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/* globals appWindowManager */
+/* exported PlacesAdapter */
```
ferjm: This code should end up being part of HistoryAdapter from the Sync App. The whole point of HistoryAdapter is to transform the data from Syncto into Places DS data and viceversa. We are spreading that logic across scripts (and even worse, across apps).

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2e0a01edcdf44df6b4ccfa72263b9267bb83deaa

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f4874442234826962f6bd6a29ee2f25681d65350

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=da34b05de5f0b9a7769fb7ead6a143c3477f6693

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d6d2615544d1c6aa42e683074c80f84d58e1537e

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a0223012280ce207afe0f58c3b62e0c702482bb3

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=13291fd5d66511b7056248077259ac96049c58d3

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=cb862299b857841d41b59d1191a181c19e91d645

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9dd0ecd0dac1ded68df197d46ab5987d38d69a48

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d00362266d7399f6d43c64d7153e2df7970617e7

apps/sync/js/adapters/history.js
weilonge: The example data records have updated [here](https://github.com/mozilla-b2g/gaia/pull/31896/files#diff-4d73ff90f2479c9b5d1f53457e85e25dR204). It includes "deleted: true" as well.

apps/sync/js/adapters/history.js
weilonge: It's updated [below](https://github.com/mozilla-b2g/gaia/pull/31896/files#diff-4d73ff90f2479c9b5d1f53457e85e25dR260).

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,309 @@
...
+  Service and merge to Places DataStore which has the same functionality, and
+  the adapter is based on SyncEngine's design to handle 'update' and
+  'handleConflict' request. History Data Adapter provides Read-Only version
+  currently, and the basic ideas are included:
+  - Traverse all records from History Collection (HC).
+  - Add a new or existed place to Places DataStore (PDS).
+    - Convert the records from HC to PDS format.
+    - Merge a new record and the existed one then write to PDS.
+  - Remove an existed place when receiving a "deleted: true" record.
+**/
```
ferjm: Thank you for adding the explanation!

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,309 @@
...
+
+'use strict';
+
+/* global
+  DataAdapters,
+  LazyLoader
+*/
+
+const HISTORY_COLLECTION_MTIME = 'LastSyncedStatus::HistoryCollection::mtime';
```
ferjm: Let's simplify the name a little bit: `collections::history::mtime`

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,309 @@
...
+  var placesStore;
+  function _ensureStore() {
+    if (placesStore) {
+      return Promise.resolve(placesStore);
+    }
+    return new Promise(resolve => {
+      navigator.getDataStores('places').then(stores => {
+        placesStore = stores[0];
+        resolve(placesStore);
+      });
```
ferjm: No need to wrap this within a new Promise. You can do:
   
    return navigator.getDataStores('places').then(stores => {
       placesStore = stores[0];
       return placesStore;
    });

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,309 @@
...
+    });
+  }
+
+  function setSyncedCollectionMtime(mtime) {
+    return new Promise(resolve => {
+      LazyLoader.load(['shared/js/async_storage.js'], () => {
+        window.asyncStorage.setItem(HISTORY_COLLECTION_MTIME, mtime, resolve);
```
ferjm: uber-nit: no need to use `window`

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,309 @@
...
+      });
+    });
+  }
+
+  function getSyncedCollectionMtime() {
+    return new Promise(resolve => {
+      LazyLoader.load(['shared/js/async_storage.js'], () => {
+        window.asyncStorage.getItem(HISTORY_COLLECTION_MTIME, resolve);
```
ferjm: ditto

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,309 @@
...
+  };
+})();
+
+DataAdapters.history = {
+/**
+    KintoCollection.list() provides a list contains all records sorted by
+  "last_modified" property which indicates the data-modified time in
+  descending order. The "last_modified" value can be recongized as which
+  records have already synchronized before. It is used to stop early and avoid
+  the redundant operations which had handled.
```
ferjm: `KintoCollection.list() provides a list containing all the remotely retrieved Firefox Sync records sorted by "last_modified" property in descending order. After each sync request we save the "last_modified" property of the last processed record so we avoid going through the same records on following operations.`

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,309 @@
...
+
+    When the new adding/updating records are coming from History Collection in
+  the format [2], it should be converted to the format [1] then stored into PDS.
+  Besides, the ID from HC records should be written into PDS records as well
+  for deleting record usage.
+
+    When the new deleting records are coming, the ID can help us to find which
+  record has to be removed.
```
ferjm: "History records are stored locally in a DataStore with format [1] while history records coming from Firefox Sync (via Kinto collection) have format [2]. We need to convert from [1] to [2] and viceversa. Also, we need to add the record ID from the Kinto collection to its corresponding match in the local DataStore, so we can remove the local record afterwards if it is remotely deleted."

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,309 @@
...
+      "visits": [
+         // NOTICE: date/time with ms
+        { "date": 1442247252490018, "type": 2 },
+        { "date": 1442247250001234, "type": 2 }
+      ]
+    },
+    "_status": "synced"
+  }
+
+  3. Delete Records from History Collection (HC): {
```
ferjm: Maybe 1.1 instead of 3?

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,309 @@
...
+          Promise.resolve(false);
+        });
+      });
+    }
+
+    return kintoCollection.list().then(updateHistoryCollection);
+  },
+  update(kintoCollection) {
+    return HistoryHelper.getSyncedCollectionMtime().then(mtime => {
+      return this._fullSync(kintoCollection, mtime);
```
ferjm: nit: s/_fullSync/_update/

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,309 @@
...
+    });
```
ferjm: You can avoid creating the `updateHistoryCollection` helper within `_fullSync` (to be renamed to `_update`)

    var mtime;
    return HistoryHelper.getSyncedCollectionMtime().then(mtime => {
      mtime = mtime;
      return collection.list();
    }).then(localRecords => {
      return _update(localRecords, mtime);
    });

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,309 @@
...
+    "sortindex": 100,
+    "last_modified": 1441985077970,
+    "payload": {
+      "id": "_Avscjx5srFy",
+      "deleted": true
+    },
+    "_status": "synced"
+  }
+**/
+  _fullSync(kintoCollection, lastModifiedTime) {
```
ferjm: s/_fullSync/_update/g

ferjm: Also, let's try to avoid specifying technology names in vars. For example, rename `kintoCollection` to `collection` or even more clear `remoteHistory`.

I think it's more clear if we use `remoteHistory` or `firefoxSyncHistory` to refer to the Kinto collection and `localHistory` or `deviceHistory` to refer to the DataStore. Same with records related var names. Instead of using `historyRecords` let's use `remoteRecords` or `localRecords`, we already know we are within the history adapter.

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,309 @@
...
+    function updateHistoryCollection(list) {
```
ferjm: You can remove this helper. Check https://github.com/mozilla-b2g/gaia/pull/31896/files#r40064079

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,309 @@
...
+        var payload = record.payload;
+        if (payload.deleted) {
+          places.push({
+            deleted: true,
+            fxsyncId: payload.id
+          });
+          return;
+        }
+        if (!payload.histUri || !payload.visits || !payload.visits.length) {
+          console.warn('Incorrect payload? ', JSON.stringify(payload));
```
ferjm: nit: I think you don't need to JSON.stringify anymore, the console already takes care of that for you.

apps/sync/js/adapters/history.js
ferjm: I still would like to see https://github.com/mozilla-b2g/gaia/pull/31896/files#r39845952 :) It saves some memory.

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,309 @@
...
+    return new Promise(resolve => {
+      places.reduce((cur, next) => {
+        return cur.then(() => {
+          if (next.deleted) {
+            return deletePlaceByFxSyncId(next.fxsyncId);
+          } else {
+            return addPlace(next);
+          }
+        });
+      }, Promise.resolve()).then(resolve);
```
ferjm: return new Promise(resolve => {
      places.reduce((current, next) => {
        ...
      }, resolve);
    });

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,309 @@
...
+    });
+  }
+
+  function updatePlaces(places) {
+    return new Promise(resolve => {
+      places.reduce((cur, next) => {
+        return cur.then(() => {
+          if (next.deleted) {
+            return deletePlaceByFxSyncId(next.fxsyncId);
+          } else {
```
ferjm: No need for `else` you are already early returning

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,309 @@
...
+    var deletingTask;
+
+    function each(task) {
+      if (task.data && task.data.fxsyncId === fxsyncId) {
+        deletingTask = task;
+      }
+      return Promise.resolve();
+    }
+
+    return traverseRecords(each).then(() => {
```
ferjm: `traverseRecords` takes two arguments. You are missing `revisionId`

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,309 @@
...
+    });
+  }
+
+  function deletePlace(id) {
+    return _ensureStore().then(placesStore => {
+      return placesStore.remove(id);
+    });
+  }
+
+  function deletePlaceByFxSyncId(fxsyncId) {
```
ferjm: Having that many helpers (with no clear order) makes the code a little bit hard to follow. Rename this to `deletePlace`, please.

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,309 @@
...
+            runNextTask(cursor);
+          });
+        }
+      });
+    });
+  }
+
+  function deletePlace(id) {
+    return _ensureStore().then(placesStore => {
+      return placesStore.remove(id);
```
ferjm: Remove this helper and move this line to where the helper is called. It's the only caller and you already ensured that the store is there.

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,309 @@
...
+    function each(task) {
+      if (task.data && task.data.fxsyncId === fxsyncId) {
+        deletingTask = task;
+      }
+      return Promise.resolve();
+    }
+
+    return traverseRecords(each).then(() => {
+      if (deletingTask) {
+        return deletePlace(deletingTask.id);
```
ferjm: `return placesStore.remove(deletingTask.id);`

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,309 @@
...
+      }
+    }).catch(e => {
+      console.error(e);
+      return;
+    });
+  }
+
+  function updatePlaces(places) {
+    return new Promise(resolve => {
+      places.reduce((cur, next) => {
```
ferjm: Rename `cur` to `reduced` and `next` to `current` please

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=43512ccd6cd4bc62f55e1a3a4f86edaf393155a9

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a53fa0b37d375923f44c2d8250819630fd0d172e

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=fdcd88975c9445f930055062872e84688ba588f3

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6fa7eb78badec0566d11078c72de8a5c8fabf56e

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d1b977bc7341416633b695469950392ee70cd506

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,283 @@
...
+
+  function getSyncedCollectionMtime() {
+    return new Promise(resolve => {
+      LazyLoader.load(['shared/js/async_storage.js'], () => {
+        asyncStorage.getItem(HISTORY_COLLECTION_MTIME, resolve);
+      });
+    });
+  }
+
+  function setSynctoIdMap(synctoId, dsId) {
```
michielbdejong: The function name suggests this sets the entire map, not one item in the map. See also https://github.com/Kinto/kinto.js/issues/173 which means we can maybe remove this whole function in a later version.

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,283 @@
...
+    return new Promise(resolve => {
+      LazyLoader.load(['shared/js/async_storage.js'], () => {
+        asyncStorage.setItem(SYNCTOID_PREFIX + synctoId, dsId, resolve);
+      });
+    });
+  }
+
+  function getSynctoIdMap(synctoId) {
```
michielbdejong: Same as a above. Maybe just call it `[g/s]etDataStoreId`?

apps/sync/js/adapters/history.js
michielbdejong: I created a follow-up bug for this: https://bugzilla.mozilla.org/show_bug.cgi?id=1207468

apps/sync/index.html
```diff
@@ -3,6 +3,10 @@
   <head>
     <title>Sync</title>
     <meta charset="utf-8">
+    <!--Shared
```
ferjm: It seems that you can remove this comment. You already have the one in line 7.

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,283 @@
...
+
+var HistoryHelper = (() => {
+  var placesStore;
+  function _ensureStore() {
+    if (placesStore) {
+      return Promise.resolve(placesStore);
+    }
+    return navigator.getDataStores('places').then(stores => {
+      placesStore = stores[0];
+      return Promise.resolve(placesStore);
```
ferjm: `return placesStore` is enough

apps/sync/js/adapters/history.js
ferjm: s/dsId/dataStoreId/

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,283 @@
...
+        asyncStorage.getItem(SYNCTOID_PREFIX + synctoId, resolve);
+      });
+    });
+  }
+
+  function mergeRecordsToDataStore(existed, newPlace) {
+    if (existed.url !== newPlace.url) {
+      // The existed record has different url(id) with the new record.
+      throw new Error('Inconsistent records');
+    }
```
ferjm: if (!existed || !newPlace ||
        existed.url !== newPlace.url ||
        existed.fxsyncId !== newPlace.fxsyncId) {
       console.error('Inconsistent records', existed, newPlace);
       throw new Error('Cannot merge records');
    }

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,283 @@
...
+    if (!existed.fxsyncId) {
+      existed.fxsyncId = newPlace.fxsyncId;
+    } else if(existed.fxsyncId !== newPlace.fxsyncId) {
```
ferjm: Check https://github.com/mozilla-b2g/gaia/pull/31896/files#r40185725

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,283 @@
...
+      // Two records have different fxsyncId but have the same url(id).
+      throw new Error('Inconsistent records');
+    }
+
+    existed.visits = existed.visits || [];
+    if (existed.visits.length === 0 && newPlace.title) {
+      existed.title = newPlace.title;
+    } else if (newPlace.visits[0] >= existed.visits[0]) {
+      existed.title = newPlace.title;
```
ferjm: You are doing the same thing...

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,283 @@
...
+
+  function getSynctoIdMap(synctoId) {
+    return new Promise(resolve => {
+      LazyLoader.load(['shared/js/async_storage.js'], () => {
+        asyncStorage.getItem(SYNCTOID_PREFIX + synctoId, resolve);
+      });
+    });
+  }
+
+  function mergeRecordsToDataStore(existed, newPlace) {
```
ferjm: Maybe rename `existed` to `localRecord` and `newPlace` to `remoteRecord`. Or at least s/existed/existing

apps/sync/js/adapters/history.js
ferjm: Can you add a comment explaining this logic, please? Actually a comment explaining the whole merging logic would be great.

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,283 @@
...
+      return b - a;
+    });
+
+    return existed;
+  }
+
+  function addPlace(place) {
+    // 1. get place by url(id of DataStore)
+    // 2.A merge the existed one and new one if it's an existed one,
+    //     and update the places.
```
ferjm: s/existed/existing/g

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,283 @@
...
+    // 2.B Add a new record with RevisionId.
```
ferjm: ... and add the DataStore record ID into the localID <-> remoteID matching table.

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,283 @@
...
+    var id = place.url;
+    var revisionId;
+    return _ensureStore().then(placesStore => {
+      revisionId = placesStore.revisionId;
+      return placesStore.get(id);
+    }).then(existedPlace => {
+      if (existedPlace) {
+        var newPlace = mergeRecordsToDataStore(existedPlace, place);
+        return placesStore.put(newPlace, id, revisionId);
+      } else {
```
ferjm: No need for `else`

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,283 @@
...
+        return placesStore.add(place, id, revisionId).then(() => {
+          return setSynctoIdMap(place.fxsyncId, id);
+        });
+      }
+    }).catch(e => {
+      console.error(e);
+      return;
```
ferjm: no need to return here

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,283 @@
...
+          if (current.deleted) {
+            return deletePlaceByFxSyncId(current.fxsyncId);
+          }
+          return addPlace(current);
+        });
+      }, Promise.resolve()).then(resolve);
+    });
+  }
+
+  function deletePlaceByFxSyncId(fxsyncId) {
```
ferjm: s/deletePlaceByFxSyncId/deletePlace/g

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,283 @@
...
+    "sortindex": 100,
+    "last_modified": 1441985077970,
+    "payload": {
+      "id": "_Avscjx5srFy",
+      "deleted": true
+    },
+    "_status": "synced"
+  }
+**/
+  _update(historyRecords, lastModifiedTime) {
```
ferjm: s/historyRecords/remoteRecords/

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,283 @@
...
+
+    if (places.length === 0) {
+      return Promise.resolve(false);
+    }
+
+    return HistoryHelper.updatePlaces(places).then(() => {
+      var latestMtime = historyRecords[0].last_modified;
+      return HistoryHelper.setSyncedCollectionMtime(latestMtime);
+    }).then(() => {
+      return Promise.resolve(false);
```
ferjm: Shouldn't this be `return true`? It's worth a comment explaining what the boolean means.

apps/sync/test/unit/adapters/history_test.js
```diff
@@ -0,0 +1,329 @@
...
+'use strict';
+
+/* global
+  assert,
+  asyncStorage,
+  DataAdapters,
+  HISTORY_COLLECTION_MTIME,
+  HistoryHelper,
+  require,
+  requireApp,
```
ferjm: alphabetically ordered, please

apps/sync/test/unit/adapters/history_test.js
```diff
@@ -0,0 +1,329 @@
...
+    window.asyncStorage.mTeardown();
+  });
+
+  test('update - empty records', done => {
+    var historyAdapter = DataAdapters.history;
+    historyAdapter.update(kintoCollection).then((result) => {
+      assert.equal(result, false);
+      assert.equal(updatePlacesSpy.callCount, 0);
+      assert.equal(asyncStorage.mItems[HISTORY_COLLECTION_MTIME], null);
+    }).then(done, done);
```
ferjm: In case of rejection, we need to fail the test.
    .then(done, reason => {
        assert.ok(false, reason);
    });

Do this here and in the rest of the tests, please.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a3aed15445add9651a9d992040b87ef167fc1199

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1b9aac49a692e99695acca802bb0ea29545014e2

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,283 @@
...
+    }
+    return navigator.getDataStores('places').then(stores => {
+      placesStore = stores[0];
+      return placesStore;
+    });
+  }
+
+  function setSyncedCollectionMtime(mtime) {
+    return new Promise(resolve => {
+      LazyLoader.load(['shared/js/async_storage.js'], () => {
```
daleharvey: Are we gaining much by lazy loading sync_storage on each usage? seems like it is used a bunch and so would be cleaner to just load inside the setup promise

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    return HistoryHelper.updatePlaces(places).then(() => {
+      var latestMtime = remoteRecords[0].last_modified;
+      return HistoryHelper.setSyncedCollectionMtime(latestMtime);
+    }).then(() => {
+      return Promise.resolve(false);
+    });
+  },
+  update(remoteHistory) {
```
daleharvey: tinynit, new lines

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=694efdbd10af651634d17468b19999d3b15e3b1e

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,285 @@
...
+    if (places.length === 0) {
+      return Promise.resolve(false);
+    }
+
+    return HistoryHelper.updatePlaces(places).then(() => {
+      var latestMtime = remoteRecords[0].last_modified;
+      return HistoryHelper.setSyncedCollectionMtime(latestMtime);
+    }).then(() => {
+      // Always return false for a read-only operation.
+      return Promise.resolve(false);
```
weilonge: @michielbdejong Could you help to explain here? In my memory, you gave me a suggestion that returning false to avoid the second round update when there is no any collection modification. Thanks.

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,285 @@
...
+    }
+    if (!localRecord.fxsyncId && typeof remoteRecord.fxsyncId === 'string') {
+      /* When a localRecord is existed without fxsyncId, assign fxsyncId to it
+         from a remoteRecord. This case always happens at first synchronization
+         or merging two records with the same URL. */
+      localRecord.fxsyncId = remoteRecord.fxsyncId;
+    } else if(localRecord.fxsyncId !== remoteRecord.fxsyncId) {
+      // Two records have different fxsyncId but have the same url(id).
+      throw new Error('Inconsistent records');
+    }
```
weilonge: From line 68 to 81, the code readability would be not good enough if we merge the three condition into one 'if' statement, and I add more comments here to describe the details.

ferjm: I disagree :)

    if (!localRecord || !remoteRecord ||
        localRecord.url !== remoteRecord.url ||
        localRecord.fxsyncId !== remoteRecord.fxsyncId) {
        throw...
     }

     if (!localRecord.fxsyncId && typeof remoteRecord.fxsyncId === 'string') {
         localRecord.fxsyncId...
     }

The first condition evaluates required preconditions, the second transforms the data in a specific case.

ferjm: But it's just a suggestion, nothing major.

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,285 @@
...
+
+    localRecord.visits = localRecord.visits || [];
+    // If a localRecord is without any visit records or with older visit
+    // than remoteRecord, its title will be replaced by remoteRecord's.
+    if (localRecord.visits.length === 0 && remoteRecord.title) {
+      localRecord.title = remoteRecord.title;
+    } else if (remoteRecord.visits[0] >= localRecord.visits[0]) {
+      localRecord.title = remoteRecord.title;
+    }
```
ferjm: if ((localRecord.visits.length === 0 && remoteRecord.title) ||
        (remoteRecord.visits[0] >= localRecord.visits[0])) {
      localRecord.title = remoteRecord.title;
    }

apps/sync/js/adapters/history.js
ferjm: Ok, got it https://bugzilla.mozilla.org/show_bug.cgi?id=1191773#c31

`return Promise.resolve(false /* no writes done into kinto */)`

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,285 @@
...
+        localRecord.url !== remoteRecord.url) {
+      // The local record has different url(id) with the new one.
+      throw new Error('Inconsistent records');
+    }
+    if (!localRecord.fxsyncId && typeof remoteRecord.fxsyncId === 'string') {
+      /* When a localRecord is existed without fxsyncId, assign fxsyncId to it
+         from a remoteRecord. This case always happens at first synchronization
+         or merging two records with the same URL. */
+      localRecord.fxsyncId = remoteRecord.fxsyncId;
+    } else if(localRecord.fxsyncId !== remoteRecord.fxsyncId) {
```
ferjm: nit: space after `if`

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,285 @@
...
+    return new Promise(resolve => {
+      asyncStorage.getItem(SYNCTOID_PREFIX + synctoId, resolve);
+    });
+  }
+
+  function mergeRecordsToDataStore(localRecord, remoteRecord) {
+    if (!localRecord || !remoteRecord ||
+        localRecord.url !== remoteRecord.url) {
+      // The local record has different url(id) with the new one.
+      throw new Error('Inconsistent records');
```
ferjm: We still need to console.error the inconsistent records

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,285 @@
...
+    }
+    if (!localRecord.fxsyncId && typeof remoteRecord.fxsyncId === 'string') {
```
ferjm: Did you find the case where remoteRecord.fxsyncId !== 'string' ?

apps/sync/js/adapters/history.js
```diff
@@ -0,0 +1,285 @@
...
+    }).then(() => {
+      // Always return false for a read-only operation.
+      return Promise.resolve(false);
+    });
+  },
+
+  update(remoteHistory) {
+    var mtime;
+    return new Promise(resolve => {
+      LazyLoader.load(['shared/js/async_storage.js'], resolve);
```
ferjm: LazyLoader already returns a Promise as well, so you can do:

     return LazyLoader.load(['shared/js/async_storage.js'])
     .then(HistoryHelper.getSyncedCollectionMtime).then(_mtime => {
     })...

apps/sync/js/adapters/history.js
weilonge: I did tried the statement you mentioned above, but unit-test is failed [here](https://github.com/mozilla-b2g/gaia/pull/31896/files#diff-be636796bffc1d301d73a8a0e16bea68R312). :P

apps/sync/js/adapters/history.js
weilonge: here is to avoid  !remoteRecord.fxsyncId case, but remoteRecord.fxsyncId === 'string' is better.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b58ebf074ca25e36eed610220b7594ec483bd4bd

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=250527747edce6941594e5ede226d31b57ac00e0

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c5ba9aeeb65969a8e3954816c50297c1d05bd960

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5d55dc10103105d3e68a37dd02a7d6812fa8f508

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ca245e65cd28b61679efc1899cc56880eca8088e

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=99f8ffc816ca38afd643f628899e87fffc90cda4

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=063537968d4a85f45a198b14cd73d8db23bc7d90

PR #31897, opened 2015-09-17
Bug 1206928 - Refactor Task Manager.

mcav:
This patch refactors Task Manager to be smaller, smoother, and more
maintainable.

A brief video of this patch is here: https://youtu.be/Fkd7EQa0BxM

As I began working on Task Manager, I noticed that we've accumulated a lot of
technical debt, which would make the planned 2.5 (and beyond) changes difficult
and error-prone. This patch removes most of that debt and should reduce the risk
of problems going forward.

I've tried to refrain from doing too much in one patch, but in the process of
cleaning up the code, a couple enhancements naturally followed the work:
- The swipe-up-to-close gesture now feels smoother and natural.
  - Rather than using a fixed swipe-up threshold, we now use velocity to decide
    whether or not the user intended to kill the app. This better reflects
    the intent of the user.
  - The logic for swiping up has been extracted into a separate class, improving
    testability. This leaves the `Card` class largely inert.
  - Cards behave with proper easing, falling naturally back into place.
  - We have axis-lock, allowing scrolling side-to-side or swipe-up, but not
    both at the same time.
  - I did not attempt to use native vertical scrolling. While native scrolling
    may provide better FPS, I think a majority of the "this feels janky" feeling
    was due to logic/easing problems rather than actual scroll FPS. In my
    testing, swiping while following the finger seems very reasonable on Aries
    and Flame, perhaps because Task Manager is fully opaque.
- CardsHelper has been renamed to TaskManagerUtils to better reflect its role;
  it now contains additional standalone functions for ease of testing.
- Our interaction with StackManager is now much more straightforward,
  reducing the need to maintain error-prone integer-based indexes.
- Filtering ("browser-only") is now only a few lines of code.

The rest of the changes are straightforward: cleaning up logic flows, reducing
complexity, and updating the code to modern standards. Similarly, the unit tests
have also been combed over to ensure they remain maintainable as well.

The diff may be hard to follow due to the nature of the refactoring; for ease of
review, I would suggest viewing `task_manager.js`, `card.js`, and
`task_manager_utils.js` independently, followed by a scan of the diff to see if
anything stands out. I've tried to structure the code in such a way that it can
be easily understood from top-to-bottom.

I have a patch in progress for adding the "new sheet" buttons, but I would like
to base that on top of this work.


autolander:

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=265411c4a07366c785648cc0790c51e19bf6b377

autolander:

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=feee16149d283596687068e0b288b44e386ded68

autolander:

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9b2d54a7a75fcf64625ae6e5d90b73843e7ff1a7

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9c0f536011a345be8f05493f714effe279b24368

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d7bded2cf6fa44d04426cb46b7f027973abbd54e

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ba9e2c4f0639ed7e71d9a1875c6a9536aab2aef3

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a2d5851ce3c8f74508099abecb164f2bf495a32f

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4e0023faa3b51202047c8a40cf9ddac718569364

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=87d45966f1cf095e4f816c62957690cd507cac2c

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b4bbcf7967392f42c2a7ea4f5753366fab7d8be7

apps/system/js/stack_manager.js
```diff
@@ -37,6 +37,9 @@
       return this._stack[this.position].getActiveWindow();
     },
     outOfStack: function sm_outOfStack() {
+      if (this.position === -1) {
+        return true;
+      }
       return (this._currentFromStack() !== this.getCurrent());
```
etiennesegonzac: nit: lets add a tiny unit test

apps/system/js/task_manager.js
```diff
@@ -1,900 +1,519 @@
...
-    var stack = this.stack;
-    stack.forEach(function(app, position) {
-      this.addCard(position, app);
-    }, this);
+  /**
+   * Update our stack and list of cards to match StackManager.
+   * Call this whenever we think the stack has changed; this function will
+   * take care of updating the layout and card positions.
+   */
+  updateStack() {
```
etiennesegonzac: :+1:

apps/system/js/task_manager_utils.js
```diff
@@ -0,0 +1,119 @@
...
+    }).then(() => {
+      element.classList.remove('pending');
+    });
+  },
+
+  /**
+   * Load a screenshot of the app into an element, only if the app is active.
+   * If it isn't, just use -moz-element.
+   */
+  loadAppScreenshot(app, element) {
```
etiennesegonzac: we need test coverage for this one (and the `isActive` special case)

etiennesegonzac: I mean unit-test coverage BTW, should be straightforward enough

apps/system/js/task_manager_utils.js
```diff
@@ -0,0 +1,119 @@
...
+
+  waitForAppToClose(app) {
+    if (!app) {
+      return Promise.resolve();
+    } else if (app.isHomescreen) {
+      return eventSafety(window, 'homescreenclosed', 400);
+    } else {
+      return eventSafety(window, 'appclosed', 400);
+    }
+  },
```
etiennesegonzac: same here, (you can stub `eventSafety`)

shared/js/event_safety.js
```diff
@@ -2,32 +2,57 @@
...
+        /*jshint validthis:true */
+        resolve(callback.apply(this, arguments));
+      } else {
+        resolve(e);
+      }
+    }
+    finishTimeout = setTimeout(done, timeout);
+    obj.addEventListener(event, done);
+  });
 }
```
etiennesegonzac: please add unit test coverage for the returned promise (the lib tests are in `apps/sharedtest/test/unit/event_safety_test.js`)

apps/system/test/unit/task_manager_test.js
```diff
@@ -1,1377 +1,570 @@
...
-                    'gets event with detail object');
-        assert.equal(event.detail.newStackPosition, 1,
-                    'event detail reflects taskManager.newStackPosition');
-        delete taskManager.newStackPosition;
-      }, failOnReject).then(done, done);
-      taskManager.hide();
-      this.sinon.clock.tick(101);
+    test('click when empty takes you home', () => {
+      sinon.spy(tm, 'hide');
+      tm.handleEvent(new CustomEvent('click'));
```
etiennesegonzac: please do a "real" `dispatchEvent` instead, covers more and no less readable :)

etiennesegonzac: I usually keep variables pointing at a few of the fake dom nodes created for the test to ease this

apps/system/test/unit/task_manager_test.js
```diff
@@ -1,1377 +1,570 @@
...
-                      '2nd card.destroy was called once');
-        destroyStub1.restore();
-        destroyStub2.restore();
-        done();
-      }, failOnReject);
-      taskManager.hide();
-      this.sinon.clock.tick(101);
+    test('takes you home after closing all cards', (done) => {
+      sinon.spy(tm, 'hide');
+      tm.handleEvent(new CustomEvent('appterminated'));
```
etiennesegonzac: same, dispatched from the app's frame, you might need to move this test to the "With Populated Task Manager" suite (which is more realistic too)

apps/system/test/unit/task_manager_test.js
```diff
@@ -1,1377 +1,570 @@
...
-        appConfigs.game
-      ].map((config) => {
-        var app = new AppWindow(config);
-        apps[app.name] = app;
-        return app;
+    test('Wheel handling for accessibility', (done) => {
+      var card = tm.currentCard;
+      var killAppStub = sinon.stub(card.app, 'kill');
+      sinon.stub(card.app, 'killable', () => true);
+      tm.handleEvent({
```
etiennesegonzac: if this event is hard to synthesized I can live with a direct `handleEvent` call here :)

apps/system/test/unit/task_manager_test.js
```diff
@@ -1,1377 +1,570 @@
...
-    });
 
-    test('lock orientation when showing', function() {
-      var orientation = MockService.mockQueryWith('defaultOrientation');
-      this.sinon.stub(screen, 'mozLockOrientation');
-      showTaskManager(this.sinon.clock);
-      assert.isTrue(screen.mozLockOrientation.calledWith(orientation));
+    test('kill cards and update', function() {
+      MockStackManager.mStack.length = 2;
+      tm.handleEvent(new CustomEvent('appterminated'));
```
etiennesegonzac: real dispatch

apps/system/test/unit/task_manager_test.js
```diff
@@ -1,1377 +1,570 @@
...
+      clock.tick(1000);
+      assert.equal(
+        document.querySelectorAll('.card').length,
+        MockStackManager.mStack.length
+      );
     });
```
etiennesegonzac: I'd like a second test here dispatching a click event on the card's close button to close instead to trigger the card removal

apps/system/test/unit/task_manager_test.js
```diff
@@ -1,1377 +1,570 @@
...
-    test('taskManager should not be active', function() {
-      taskManager.hide();
-      assert.isFalse(taskManager.isShown(), 'isShown is false');
-      assert.isFalse(cardsView.classList.contains('active'),
-                    'no .active class');
+    test('Should not respond to holdhome if already showing', () => {
+      var showSpy = sinon.spy(tm, 'show');
+      tm.respondToHierarchyEvent(new CustomEvent('holdhome'));
+      assert.isFalse(showSpy.called);
     });
```
etiennesegonzac: please add a test also covering that we _do_ respond the `holdhome` if we're not showing yet (I don't think there's one)

apps/system/test/unit/task_manager_test.js
```diff
@@ -1,1377 +1,570 @@
...
 
-    test('removes classes', function() {
-      taskManager.hide();
-      assert.isFalse(screenNode.classList.contains('cards-view'));
+    test('home takes you home', () => {
+      sinon.spy(tm, 'hide');
+      tm.respondToHierarchyEvent(new CustomEvent('home'));
+      assert.ok(tm.hide.calledOnce);
     });
```
etiennesegonzac: I think calling `tm.hide()` from the test is perfectly reasonable to easily test the promise chain, but I'd also like to see "hiding tests" triggered by
* dispatching a click event on the currently centered app
* on one of the other apps(`panToApp` code path)

apps/system/test/unit/task_manager_test.js
```diff
@@ -1,1377 +1,570 @@
...
+      assert.isTrue(tm.isShown());
+      assert.ok(document.querySelector('#cards-view.active:not(.empty)'));
 
-        waitForEvent(window, 'cardviewclosed').then(function() {
-          assert.isTrue(stub.calledOnce, 'home was open');
-        }, failOnReject)
-        .then(function() { done(); }, done);
+      assert.equal(
+        MockStackManager.getCurrent(),
+        tm.currentCard.app);
```
etiennesegonzac: can we assert on the aria attributes here too?

apps/system/test/unit/task_manager_test.js
```diff
@@ -1,1377 +1,570 @@
...
-        this.element.dispatchEvent(new CustomEvent('_opened'));
+    test('wheel left/right event', function() {
+      this.sinon.spy(tm, 'updateScrollPosition');
+      tm.handleEvent({
+        type: 'wheel',
+        deltaMode: 2,
+        DOM_DELTA_PAGE: 2,
+        deltaX: 1
       });
+      assert.ok(tm.updateScrollPosition.calledOnce);
```
etiennesegonzac: instead of simply spying I think we should inspect the aria attributes on the dom here

apps/system/test/unit/task_manager_test.js
etiennesegonzac: and for this test, instead of just spying on `show` I'd like to also assert on the active app (homescreen and not lockscreen) to make sure we're calling `close` properly and setting the special transition classes `from-home` correctly

apps/system/test/unit/task_manager_test.js
etiennesegonzac: oh and I'd like to cover the transition classes (`home-from-cardview`...) here too

apps/system/js/task_manager.js
```diff
@@ -1,900 +1,519 @@
...
-   * Default event handler
-   * @memberOf TaskManager.prototype
-   * @param  {DOMEvent} evt The event.
-   */
-  TaskManager.prototype.handleEvent = function cv_handleEvent(evt) {
-    var app;
-    switch (evt.type) {
-      case 'click':
-        this.handleTap(evt);
+      case 'card-will-drag':
```
etiennesegonzac: we need basic unit testing for the `card-will-drag` / `card-dropped` events (with real dispatch ;))

apps/system/test/unit/task_manager_test.js
etiennesegonzac: ...and make sure we dispatch the `cardviewshown` event properly, it's part of our "API" to the rest of the system we need a test

etiennesegonzac: oh we need one for `cardviewbeforeshow` too

apps/system/test/unit/card_test.js
```diff
@@ -82,368 +76,173 @@ suite('system/Card', function() {
...
-        origin: 'app://search.gaiamobile.org',
-        url: 'app://search.gaiamobile.org/newtab.html'
-      });
-      this.sinon.stub(app, 'isBrowser').returns(true);
-      var appCard = new Card({
-        app: app,
-        manager: mockManager
-      });
-      appCard.render();
+  });
```
etiennesegonzac: let's add a small test for the `role` / `aria-labelledby` attributes

apps/system/test/unit/card_test.js
```diff
@@ -82,368 +76,173 @@ suite('system/Card', function() {
...
+  }
 
-    test('cleans up references', function() {
-      this.card.destroy();
-      assert.ok(!this.card.manager, 'card.manager reference is falsey');
-      assert.ok(!this.card.app, 'card.app reference is falsey');
-      assert.ok(!this.card.element, 'card.element reference is falsey');
-    });
+  test('swipe up to kill', function() {
+    var dropStub = this.sinon.stub();
```
etiennesegonzac: if (and only if) it's as simple as a search and replace I'd like to use real dispatch events here too

etiennesegonzac: and assert that we called `stopPropagation`

apps/system/test/unit/card_test.js
etiennesegonzac: and that `loadAppIcon` is called (stub is fine)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f69ca7d5ab3d15e5735858f0a9752fdc9d83bae5

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e08fadb2e73115e1e51d685553af803fb7b1044c

etiennesegonzac:
We need to skip the transition for the first call to `translate` on a card.
Or add a parameter.

The issue is, when entering the card view, you can sometime see the card transition to their position instead of already being there.

shared/js/event_safety.js
```diff
@@ -2,32 +2,57 @@
...
 
-  obj.addEventListener(event, done);
-  finishTimeout = setTimeout(done, timeout);
+  return new Promise((resolve) => {
+    var finishTimeout;
+    function done(e) {
+      // transitionend events bubble by default, so we filter them by element.
+      if (e && e.type === 'transitionend' && e.target !== obj) {
+        return;
+      }
```
etiennesegonzac: nit maybe a follow up bug for `animationend` too (good mentored bug)

apps/system/test/unit/task_manager_test.js
```diff
@@ -1,1377 +1,783 @@
...
-      setup(function() {
-        MockStackManager.mStack = [
-          apps.sms
-        ];
-        MockStackManager.mCurrent = MockStackManager.mStack.length -1;
-        showTaskManager(this.sinon.clock);
-      });
+        var handleShown = () => {
+          window.removeEventListener('cardviewshown', handleShown);
+          try {
```
etiennesegonzac: why the `try` ?

mcav: Ah, I didn't realize that mocha was smart enough to catch the appropriate uncaught exception in a test. I guess it's not necessary after all.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=34e94144318fee3468118f3c352f215ba24e9048

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1b0de0dad126d5695fbf28d80eba97bad4320433

PR #31903, opened 2015-09-18
Bug 1205933 - Fix UTF-8 StringConversion functions

weilonge:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=70de8c9795e561ae62f36740c809efbc14b3a65c

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=afc9b6de3bda499fe32a6aad1df5bf816f796bb9

apps/sync/js/crypto/fxsyncwebcrypto.js
```diff
@@ -272,7 +271,7 @@ undle for collection ${collectionName}`);
     } catch(e) {
       return Promise.reject('Record cannot be JSON-stringified');
     }
-    const cleartext = StringConversion.rawStringToByteArray(cleartextStr);
+    const cleartext = StringConversion.utf16StringToByteArray(cleartextStr);
```
Natim: Make sure you are building an UTF-8 ByteArray rather than a UTF-16 ByteArray

apps/sync/js/crypto/stringconversion.js
```diff
@@ -23,6 +23,19 @@ const StringConversion = Object.freeze((() => {
...
 
+    utf16StringToByteArray(str) {
+      if (typeof str != 'string') {
+        throw new Error('Not a string');
+      }
+      var string = unescape(encodeURIComponent(str)),
+          charList = string.split(''),
+          uintArray = [];
+      for (var i = 0; i < charList.length; i++) {
+          uintArray.push(charList[i].charCodeAt(0));
```
Natim: http://stackoverflow.com/questions/14592364/utf-16-to-utf-8-conversion-in-javascript/14601808#14601808

apps/sync/js/crypto/fxsyncwebcrypto.js
```diff
@@ -210,8 +210,7 @@ ${collectionName}`);
                                    payload.ciphertext)
           .then(recordArrayBuffer => {
         var recordObj;
-        const recordJSON = String.fromCharCode.apply(
-            null,
+        const recordJSON = StringConversion.byteArrayToUtf16String(
```
michielbdejong: As discussed on irc, since the SyncStorage service always encodes BSOs in UTF8 (see https://docs.services.mozilla.com/storage/apis-1.5.html#basic-storage-object) we should not interpret them as UTF16.

apps/sync/js/crypto/fxsyncwebcrypto.js
michielbdejong: Same as above, since the SyncStorage service always encodes BSOs in UTF8 (see https://docs.services.mozilla.com/storage/apis-1.5.html#basic-storage-object) we should not upload UTF16.

apps/sync/js/crypto/fxsyncwebcrypto.js
michielbdejong: Or maybe we should be using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCodePoint? (just found out about that)

apps/sync/js/crypto/fxsyncwebcrypto.js
weilonge: utf16StringToByteArray does convert the string from UTF-16 to UTF-8 byte array. I will modify the function name to utf16StringToUtf8ByteArray.

apps/sync/test/unit/crypto/stringconversion_test.js
```diff
@@ -85,6 +99,22 @@ suite('utils', () => {
           throw(Error);
     });
   });
+  suite('byteArrayToUtf16String', () => {
+    test('converts an Uint8Array to a UTF-16 string', () => {
+      var array = new Uint8Array(3);
+      array[0] = 0xe2;
+      array[1] = 0x82;
+      array[2] = 0xac;
```
michielbdejong: Those are the codes for UTF-8, the codes for UTF-16 are 0x20AC, see http://www.fileformat.info/info/unicode/char/20ac/index.htm

michielbdejong: So I think that solves the mystery! :)

* You're right, the implementation of StringConversion in master is wrong and this patch is valid in what it fixes.
* But the reason it's broken, is because it's incorrectly implementing UTF-8. It's actually implementing ASCII-only! :/
* The tests for `rawStringToByteArray` are also wrong. Sorry about that, I should have spotted that [19](https://github.com/mozilla-b2g/gaia/blob/master/apps/sync/test/unit/crypto/stringconversion_test.js#L28) is not likely to be the correct char code for '✓'!).
* They encryption/decryption still pass, because I think what you implemented is are the correct conversion algorithms for UTF-8, not UTF-16. :) Switching to UTF-16 would have broken those tests on the current fixture, that's why I was confused.
* The tests for encryption/decryption pass despite StringConversion being broken, because [the fixture](https://github.com/mozilla-b2g/gaia/blob/master/apps/sync/test/unit/fixtures/fxsyncwebcrypto.js#L53-L62) only tests them on ASCII, not on UTF-8 (we should fix that test as well to be a bit more challenging...).
* As I understand it now, UTF-16 has nothing to do with any of this! :)

weilonge: That's no doubt that [0xe2, 0x82,0xac] is an byte array in UTF-8 code.
I mentioned [here](https://github.com/mozilla-b2g/gaia/pull/31903/files#r41014832) that "byteArrayToUtf16String" is converting UTF8 byte array to UTF-16. 
So '€' in UTF-16 has to be verified at [here](https://github.com/mozilla-b2g/gaia/pull/31903/files#diff-5a3d47b065aef47f4604389e7e708062R110)

I think this issue is still related to the converter of UTF-8 <-> UTF-16.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=39bd64b4411105ffe976dbbb353257b0478aa596

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=78cf91813e28bf5eb78c2cac491c6e16519009f8

apps/sync/test/unit/crypto/stringconversion_test.js
weilonge: The string "UTF-16" can be removed from the function name because the default string in JS is UTF-16.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=84b875a68f21492ac8cc2eba85536b069212ae36

apps/sync/test/unit/crypto/stringconversion_test.js
michielbdejong: Yes, that would save a lot of confusion. We don't care how String objects are represented internally, let's call all Strings just Strings, and call all ByteArrays Utf8ByteArray so that it's clearer.

apps/sync/js/crypto/fxsyncwebcrypto.js
```diff
@@ -42,7 +42,7 @@ const FxSyncWebCrypto = Object.freeze((() => {
     // The number 64 here comes from
     // (256 bits for AES + 256 bits for HMAC) / (8 bits per byte).
     return KeyDerivation.hkdf(kBByteArray,
-                              StringConversion.rawStringToByteArray(
+                              StringConversion.asciiStringToByteArray(
```
michielbdejong: Please rename `asciiStringToByteArray` to `stringToAsciiByteArray`, so its relationship with `stringToUtf8ByteArray` is clearer. Also please add a comment about why we're not using `stringToUtf8ByteArray` here.

apps/sync/test/unit/crypto/stringconversion_test.js
```diff
@@ -84,6 +97,22 @@ suite('utils', () => {
           throw(Error);
     });
   });
+  suite('utf8ByteArrayToString', () => {
+    test('converts an Uint8Array to a UTF-16 string', () => {
```
michielbdejong: "... to a [~~UTF-16~~] String"

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=93808b616b313540854ac76ff9c7013409d80c9a

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=78c4f1820ab5e1e19578f4bd0b72cf0b65afa6dd

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=197c4911ece996669378d09521e6295f9f0edff1

apps/sync/js/crypto/fxsyncwebcrypto.js
```diff
@@ -210,8 +212,8 @@ ${collectionName}`);
                                    payload.ciphertext)
           .then(recordArrayBuffer => {
         var recordObj;
-        const recordJSON = String.fromCharCode.apply(
-            null,
+        // The plaintext payload from FxSync is in UTF-8 encoding.
```
michielbdejong: Nit: I think you mean cleartext, not plaintext. Maybe a bit clearer if we just write:
// `recordArrayBuffer` contains UTF-8 JSON.

apps/sync/test/unit/crypto/stringconversion_test.js
```diff
@@ -17,76 +17,105 @@ requireApp('sync/js/crypto/main.js');
 requireApp('sync/test/unit/fixtures/fxsyncwebcrypto.js');
 
 suite('utils', () => {
-  suite('rawStringToByteArray', () => {
-    test('converts a raw string to a ByteArray', () => {
-     const ba = StringConversion.rawStringToByteArray('hi ✓');
```
michielbdejong: With what we know now, 'hi ' is not a good fixture here, please use `'\x00\xEA\x80\xFF'` instead, and then `expect(ba).to.deep.equal([0, 234, 128, 255]);` (or something similar, which contains bytes both above and below 128.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=66bfc6ff79767479dab8421bd7abd013fe5aec85

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d29bb7761abe921c8dc25c2ff847d68758b6126b

PR #31907, opened 2015-09-18
Bug 1179628 - [Messages][NG] Lay out Settings service structure, r=julienw azasypkin

steveck-chung:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=15a1be4d138dcc801c69cbf022a22b712ba2dd03

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2c486e5b2a9661d09601497bd9172a5409803e54

apps/sms/services/js/moz_settings/moz_settings_client.js
```diff
@@ -0,0 +1,98 @@
...
+    client = bridge.client({
+      service: SERVICE_NAME,
+      endpoint: new BroadcastChannel(broadcastChannelName),
+      timeout: TIMEOUT
+    });
+  },
+
+  /**
+   * Limitation of the MMS size .
+   * @returns {Promise} Promise of shim's method that returned the mms limit
```
azasypkin: __jsdoc:__ ```Promise.<Number>```?

apps/sms/services/js/moz_settings/moz_settings_client.js
```diff
@@ -0,0 +1,98 @@
...
+    return client.method('get', MMS_SIZE_LIMIT_KEY).then((result) => {
+      var size = result[MMS_SIZE_LIMIT_KEY];
+      return size && !isNaN(size) ?
+        size - MMS_SIZE_OVERHEAD : MMS_SIZE_LIMITATION;
+    });
+  },
+
+  /**
+   * Maximum concatenated number of our SMS.
+   * @returns {Promise} Promise of shim's method that return number of maximum
```
azasypkin: __jsdoc:__ ```Promise.<Number>```?

apps/sms/services/js/moz_settings/moz_settings_client.js
```diff
@@ -0,0 +1,98 @@
...
+    });
+  },
+
+  /**
+   * Check if device have mre than 1 active SIM.
+   * @returns {boolean} true if the device has more than 1 SIM port and at least
+   *  2 SIMs are inserted.
+   */
+  hasSeveralSim() {
+    return navigator.mozIccManager &&
```
azasypkin: __nit:__ can fit into one line

apps/sms/services/js/moz_settings/moz_settings_client.js
```diff
@@ -0,0 +1,98 @@
...
+      navigator.mozIccManager.iccIds.length > 1;
+  }
+};
+
+exports.MozSettingsClient = Object.freeze(
```
azasypkin: __nit:__ can fit into one line

apps/sms/services/js/moz_settings/moz_settings_client.js
```diff
@@ -0,0 +1,98 @@
...
+   */
+  maxConcatenatedMessages() {
+    return client.method('get', SMS_MAX_CONCAT_KEY).then((result) => {
+      var num = result[SMS_MAX_CONCAT_KEY];
+      return num && !isNaN(num) ? num : MAX_CONCATENATED_MESSAGES;
+    });
+  },
+
+  /**
+   * Check if device have mre than 1 active SIM.
```
azasypkin: __typo:__ ```more```

apps/sms/views/conversation/js/compose.js
```diff
@@ -314,7 +319,7 @@ var Compose = (function() {
   }
 
   var compose = {
-    init: function composeInit(formId) {
+    init(formId, mmsSizeLimitation, maxConcatenatedMessages) {
```
azasypkin: I really like that we inject such properties from outside now \0/ Hope we don't lose a lot of time on two settings calls when we init this

apps/sms/views/conversation/js/conversation.js
```diff
@@ -295,7 +265,48 @@ var ConversationView = {
...
 
-    this._isReady = true;
+    return this.initCompose().then(() => this._isReady = true);
+  },
+
+  initCompose() {
+    return Promise.all([
+      MozSettingsClient.mmsSizeLimitation(),
+      MozSettingsClient.maxConcatenatedMessages()
+    ]).then((values) => {
```
azasypkin: __nit:__ maybe ```then([mmsSizeLimitation, maxConcatenatedMessages])```

apps/sms/views/conversation/js/conversation.js
```diff
@@ -295,7 +265,48 @@ var ConversationView = {
...
+      //      message input
+      //  2. 1 or 2 characters may remain in the recipient
+      //      editable, which will be "assimilated"
+      //  3. If a user has made it past 1 & 2, any attempts to
+      //      select a contact from contact search results
+      //      will also jump focus to the message input field
+      //
+      // So we assimilate recipients if user starts to interact with Composer
+      Compose.on('interact', this.assimilateRecipients.bind(this));
+      return;
```
azasypkin: __nit:__ there is no need in ```return;``` statement I think.

apps/sms/views/shared/js/activity_handler.js
```diff
@@ -114,43 +117,51 @@ var ActivityHandler = {
...
-          Utils.alert({
-            id: 'attached-files-too-large',
-            args: {
-              n: activityData.blobs.length,
-              mmsSize: (Settings.mmsSizeLimitation / 1024).toFixed(0)
-            }
-          }).then(() => ActivityClient.postResult());
-          return;
-        }
+        dataPromise = MozSettingsClient.mmsSizeLimitation().then((limit) => {
```
azasypkin: __question:__ I'm not sure I really like this approach, it looks a bit messy. What about returning ```Promise.resolve(attachments)``` here and same for VCARD ```Promise.resolve([new Attachment(.....)])``` and in ```dataPromise.then``` callback, if we see that ```Array.isArray(data)``` then we retrieve size limitation and calculate the size, show alert and so on?

steveck-chung: I think this idea is good, and maybe we can even throw all the errors(including No data to share) and catch them all together in reject?

azasypkin: My head is emptied after just a few days I review patch, so I can miss something now, but handling all errors in one place sounds good to me if not  too difficult :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ec38480e72c0e524045eb9452ff82ccb3bca3335

apps/sms/views/shared/js/activity_handler.js
steveck-chung: I returned the postError directly in the end because I don't want to wrap the logic into another promise for the error cases, but please let me know if you think it's not appropriate.

apps/sms/services/js/moz_settings/moz_settings_client.js
```diff
@@ -0,0 +1,95 @@
...
+  },
+
+  /**
+   * Limitation of the MMS size .
+   * @returns Promise.<Number> Promise of shim's method that returned the mms
+   *  limit in Bytes. Default limitation is 295KB.
+   */
+  mmsSizeLimitation() {
+    return client.method('get', MMS_SIZE_LIMIT_KEY).then((result) => {
+      var size = result[MMS_SIZE_LIMIT_KEY];
```
julienw: wondering if we could retrieve result[key] in the service and just return the result value

steveck-chung: yes of course, I think settings shim could handle this part instead of simply returning the result.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2cdb62cd92c362216da17f13d1bbfb39717bbb8a

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=62866bab6c5e46f770640635e9d293758ab4db02

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f4a8f4e87b3192e5ce627f6d85e3b78be7fac753

apps/sms/services/js/moz_settings/moz_settings_client.js
```diff
@@ -0,0 +1,83 @@
+/* global bridge,
+          BroadcastChannel */
+
+(function(exports) {
+'use strict';
+
+/**
+ * Name of the service that is responsible for settings functionality.
+ * @type {string}
```
azasypkin: __optional nit (:bike:):__ you can also add ```@const``` directive to the jsdoc for the constants, but it's optional - this file is marked fully green by my IDE - it's a rare thing :D

apps/sms/services/js/moz_settings/moz_settings_client.js
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Reference to active bridge client instance.
+ *
+ * @type {Client}
+ */
+var client;
+
+var MozSettingsClient = {
+  /**
+   * Initializes connection to mozSettingsShim hosted in iframe.
```
azasypkin: __optional nit(:bike:):__ with this new planned changes about LittleBrowser and alike maybe we shouldn't be that specific and say ```hosted in iframe```, maybe just omit this or name ```hosted in a different compartment``` as you wish, but it's optional and doesn't make any difference anyway.

apps/sms/services/js/moz_settings/moz_settings_client.js
```diff
@@ -0,0 +1,83 @@
...
+    client = bridge.client({
+      service: SERVICE_NAME,
+      endpoint: new BroadcastChannel(broadcastChannelName),
+      timeout: TIMEOUT
+    });
+  },
+
+  /**
+   * Limitation of the MMS size .
+   * @returns Promise.<Number> Promise of shim's method that returned the mms
```
azasypkin: __nit:__ ```@returns {Promise.<Number>}```

apps/sms/services/js/moz_settings/moz_settings_client.js
```diff
@@ -0,0 +1,83 @@
...
+   */
+  mmsSizeLimitation() {
+    return client.method('get', MMS_SIZE_LIMIT_KEY).then((size) =>
+      size && !isNaN(size) ? size - MMS_SIZE_OVERHEAD : MMS_SIZE_LIMITATION
+    );
+  },
+
+  /**
+   * Maximum concatenated number of our SMS.
+   * @returns Promise.<Number> Promise of shim's method that return number of
```
azasypkin: __nit:__ ```@returns {Promise.<Number>}```

apps/sms/services/js/moz_settings/moz_settings_shim.js
```diff
@@ -35,10 +38,10 @@ var MozSettingsShim = {
   /**
    * Shim for mozSettings.createLock().get API.
    * @param {String} key Key name for Settings.
-   * @returns {Promise} Promise that return the settings value from settings DB.
+   * @returns {String} The requested value from settings DB.
```
azasypkin: __nit:__ maybe ```@returns {Promise.<String>}```?

apps/sms/services/test/unit/moz_settings/moz_settings_client_test.js
```diff
@@ -0,0 +1,130 @@
...
+
+    setup(function() {
+      MozSettingsClient.init(APP_INSTANCE_ID);
+    });
+
+    test('client method is called and returns valid number', function() {
+      size = 1024;
+
+      clientStub.method.withArgs('get', key).returns(Promise.resolve(size));
+      MozSettingsClient.mmsSizeLimitation().then((limit) => {
```
azasypkin: __issue:__ looks like ```done``` and ```.then(done, done)``` is missed here and in all tests below.

apps/sms/views/conversation/test/unit/compose_test.js
```diff
@@ -1190,6 +1187,9 @@ suite('compose_test.js', function() {
   ['image', 'video', 'audio', 'other'].forEach(function(type) {
     suite(type + ' attachment pre-send menu display', function() {
       setup(function() {
+        loadBodyHTML('/index.html');
```
azasypkin: __question:__ as far as I understand tests don't work with ```loadBodyHTML```, so just curious what is the reason?

apps/sms/views/conversation/test/unit/conversation_test.js
```diff
@@ -226,14 +232,34 @@ suite('conversation.js >', function() {
...
+    this.sinon.stub(MozSettingsClient, 'mmsSizeLimitation');
+    this.sinon.stub(MozSettingsClient, 'maxConcatenatedMessages');
     this.sinon.useFakeTimers();
 
     this.sinon.stub(ActivityClient);
 
     ConversationView.recipients = null;
     ConversationView._isReady = false;
-    ConversationView.init();
+    mmsSizeLimitation = 295 * 1024;
```
azasypkin: __optional nit:__ could you please add new line before ```mmsSizeLimitation = 295 * 1024;``` and ```ConversationView.init()``` just to make it a bit easier to read.

__question:__ Also what about making ```Conversation.init().then(done, done)``` as the last statement in ```setup``` and move asserts to dedicated test below (```test('View is ready after init()', ...)```), so no need in verifying private field ```assert.isTrue(ConversationView._isReady);``` ;)?

apps/sms/views/shared/js/activity_handler.js
```diff
@@ -27,6 +27,8 @@ const ActivityDataType = {
   VCARD: 'text/vcard'
 };
 
+const FILE_TOO_LARGE_ID = 'attached-files-too-large';
```
azasypkin: __optional nit(:bike:):__ maybe ```FILE_TOO_LARGE_L10N_ID``` to indicate that it's l10n id?

apps/sms/views/shared/js/activity_handler.js
```diff
@@ -88,51 +90,51 @@ var ActivityHandler = {
...
 
-    if (!dataToShare) {
-      ActivityClient.postError('No data to share found!');
-      return;
+    if (!data) {
+      return ActivityClient.postError('No data to share found!');
     }
 
-    return this.toView({ body: dataToShare });
+    // Treat the non array data (url, vcard) size as 0
```
azasypkin: __question:__ do we have bug to count size of VCard blob? Not sure if it's possible in practice, but seems real in theory :)

apps/sms/views/shared/js/activity_handler.js
```diff
@@ -88,51 +90,51 @@ var ActivityHandler = {
...
+
+      return size;
+    }, 0) : 0;
+
+    return MozSettingsClient.mmsSizeLimitation().then((limit) => {
+      if (size > limit) {
+        return Utils.alert({
+          id: FILE_TOO_LARGE_ID,
+          args: {
+            n: activityData.blobs.length,
```
azasypkin: __nit:__ maybe better ```data.length``` to be more consistent (we use ```data``` to count size)?

apps/sms/views/shared/js/activity_handler.js
```diff
@@ -88,51 +90,51 @@ var ActivityHandler = {
...
+    // Treat the non array data (url, vcard) size as 0
+    var size = Array.isArray(data) ? data.reduce(function(size, attachment) {
+      if (attachment.type !== 'img') {
+        size += attachment.size;
+      }
+
+      return size;
+    }, 0) : 0;
+
+    return MozSettingsClient.mmsSizeLimitation().then((limit) => {
```
azasypkin: __question/optional nit:__ should we call ```mmsSizeLimitation()``` if size is 0? Maybe something like this?
```
let sizeVerificationPromise;
if (size) {
  sizeVerificationPromise = MozSettingsClient.mmsSizeLimitation().then((limit) => {
    if (size > limit) {
      return Utils.alert({
        id: FILE_TOO_LARGE_ID,
        args: { n: data.length, mmsSize: (size / 1024).toFixed(0) }
      }).then(() => Promise.reject());
    }
  });
} else {
  sizeVerificationPromise = Promise.resolve();
}

return sizeVerificationPromise.then(
  () => this.toView({ body: data }),
  () => ActivityClient.postResult()
);
```

apps/sms/views/shared/js/startup.js
```diff
@@ -100,6 +101,7 @@ var Startup = exports.Startup = {
 
       InterInstanceEventDispatcher.connect();
 
+      MozSettingsClient.init(App.instanceId);
```
azasypkin: __nit:__ can we group initialization of all "clients" together, just for the sake of clarity?
```
MozSettingsClient.init(App.instanceId);
MessagingClient.init(App.instanceId);
MozMobileConnectionsClient.init(App.instanceId);

TimeHeaders.init();
Information.initDefaultViews();
ConversationView.init();
```

__question:__ also what do you think if we move perf marks and ```Navigation.setReady``` to result promise callback? To be clearer on the fact that those things don't depend only on ```ConversationView``` initalization.

```
return LazyLoader.load(
  [...this._lazyLoadScripts, ...this._lazyLoadStyles]
).then(() => {
  .....
  return ConversationView.init();
}).then(() => {
  Navigation.setReady();

  window.performance.mark('objectsInitEnd');
});
```

apps/sms/views/shared/test/unit/system_message_handler_test.js
```diff
@@ -6,7 +6,6 @@
          Navigation,
          NotificationHelper,
          Notify,
-         Settings,
```
azasypkin: __nit:__ could you please also remove Settings mock reference below as well?

apps/sms/views/shared/js/utils.js
```diff
@@ -963,6 +963,17 @@
       });
 
       return defer.promise;
+    },
+
+    /**
+     * Check if device have more than 1 active SIM.
```
azasypkin: __typo:__ ```if device has```

__typo:__ ```or mozIccManager is not supported.```

apps/sms/views/shared/js/utils.js
```diff
@@ -963,6 +963,17 @@
...
+     * @returns {boolean} True if the device has more than 1 SIM port and at
+     *  least 2 SIMs are inserted and false if device only has 1 SIM port or
+     *  mozIccManager not supported.
+     */
+    hasSeveralSim() {
```
azasypkin: IMO ```Utils``` is better place for this method \0/ I see you replaced ```Settings.hasSeveralSim``` in several places, but there are still two places where old ```Settings.hasSeveralSim``` is used (message manager and messaging service), do you think we can entirely migrate to ```Utils.hasSeveralSim```  and get rid of settings related method/tests in this patch or you'd rather prefer to change it in a follow up?

steveck-chung: IIRC the remaining `Settings.hasSeveralSim` should be in the services originally,  and it's not working to import `Utils` in the worker because of API. So I'm thinking of moving the `hasSeveralSim` from service to the client, and there would be no other API dependency inside the messaging service. wdyt?

apps/sms/services/test/unit/moz_settings/moz_settings_client_test.js
steveck-chung: Nice catch, thanks!

apps/sms/views/conversation/test/unit/compose_test.js
steveck-chung: The compose initialization needs lots of dom element... :/

apps/sms/views/shared/js/activity_handler.js
steveck-chung: Nope as far as I know. I'll verify this again and file a bug if vcard size is still missing.

apps/sms/views/shared/js/utils.js
azasypkin: > IIRC the remaining Settings.hasSeveralSim should be in the services originally, and it's not working to import Utils in the worker because of API.

Ah yeah, so we can replace in ```MessageManager``` for now.

> So I'm thinking of moving the hasSeveralSim from service to the client, and there would be no other API dependency inside the messaging service. wdyt?

Yeah, I think it should be fine

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=984e8cea3add46e4498180351d7f728d98045bf6

apps/sms/views/shared/js/startup.js
```diff
@@ -113,21 +119,18 @@ var Startup = exports.Startup = {
 
       // Init UI Managers
       TimeHeaders.init();
-      ConversationView.init();
-      MessagingClient.init(App.instanceId);
-      MozMobileConnectionsClient.init(App.instanceId);
       Information.initDefaultViews();
 
+      return ConversationView.init();
```
steveck-chung: I think this change make sense, but I can't figure out a better way to verify this in unit test(we can stub the Conversation since it's lazy loaded). Any good idea for this?

apps/sms/services/test/unit/moz_settings/moz_settings_client_test.js
```diff
@@ -0,0 +1,130 @@
...
+        sinon.assert.calledWithMatch(
+          clientStub.method,
+          'get',
+          key
+        );
+
+        assert.equal(limit, size - 5 * 1024);
```
azasypkin: __nit:__ could you please move ```5 * 1024``` to a constant with descriptive name so that it clear what it's? :)

apps/sms/views/shared/js/startup.js
```diff
@@ -113,21 +119,18 @@ var Startup = exports.Startup = {
...
       Navigation.setReady();
 
       window.performance.mark('contentInteractive');
-
-      // Fetch mmsSizeLimitation and max concat
-      Settings.init();
-
       window.performance.mark('objectsInitEnd');
+
+      return;
```
azasypkin: __nit:__ do we need this ```return;```?

apps/sms/services/test/unit/moz_settings/moz_settings_shim_test.js
```diff
@@ -30,7 +31,7 @@ suite('MozSettingsShim >', function() {
   });
 
   test('service is correctly initialized with settings provided', function() {
-    MozSettingsShim.init(MockNavigatorSettings);
+    MozSettingsShim.init(APP_ID, MockNavigatorSettings);
```
azasypkin: __nit:__ sorry just realized that we don't check that shim uses ```APP_ID`` correctly, could you please add one more assert here about BroadcastChannel like we have for other similar services/shims?

apps/sms/views/conversation/test/unit/compose_test.js
azasypkin: Nah, I think it's OK what you have now.

apps/sms/index.html
```diff
@@ -125,6 +125,7 @@
     <script defer src="services/js/activity/activity_client.js"></script>
     <script defer src="services/js/messaging/messaging_client.js"></script>
     <script defer src="services/js/moz_mobile_connections/moz_mobile_connections_client.js"></script>
+    <script defer src="services/js/moz_settings/moz_settings_client.js"></script>
```
azasypkin: __issue:__ Could you please also take care about split view, New Message view in split mode is broken with the PR (actually if you rebase your PR on master it should fail just revived split inbox integration test, if not then my test is bad :/, that's why I want to return all our integration tests as soon as possible :D)

apps/sms/views/shared/js/startup.js
azasypkin: > I think this change make sense, but I can't figure out a better way to verify this in unit test(we can stub the Conversation since it's lazy loaded).

Mmm, but in unit test it's not actually lazy loaded since with use Mock for lazy loader. Can't we just load ConversationView mock in unit test, stub ```init``` method with our ```new Promise((resolve) => resolveConversationInit = resolve)```, and then call ```resolveConversationInit``` in unit test and verify that ```Navigation.setReady``` is called after that?

apps/sms/views/conversation/js/compose.js
```diff
@@ -1,7 +1,7 @@
 /* -*- Mode: js; tab-width: 2; indent-tabs-mode: nil; c-basic-offset: 2 -*- /
 /* vim: set shiftwidth=2 tabstop=2 autoindent cindent expandtab: */
 
-/*global Settings, Utils, Attachment, MozActivity,
```
azasypkin: __issue:__ I see ```JavaScript Error: "TypeError: dom.message is null" {file: "app://sms.gaiamobile.org/views/conversation/js/compose.js" line: 496}``` in logcat if I share image via Messages app with your PR rebased on the latest master - could you please take a look?

apps/sms/services/js/moz_settings/moz_settings_client.js
```diff
@@ -0,0 +1,83 @@
...
+  },
+
+  /**
+   * Limitation of the MMS size .
+   * @returns {Promise.<Number>} Promise of shim's method that returned the mms
+   *  limit in Bytes. Default limitation is 295KB.
+   */
+  mmsSizeLimitation() {
+    return client.method('get', MMS_SIZE_LIMIT_KEY).then((size) =>
+      size && !isNaN(size) ? size - MMS_SIZE_OVERHEAD : MMS_SIZE_LIMITATION
```
azasypkin: Btw, it works, but size is a string :) It's the same in master, so fix is not mandatory :)

apps/sms/services/js/moz_settings/moz_settings_client.js
```diff
@@ -0,0 +1,83 @@
...
+  },
+
+  /**
+   * Maximum concatenated number of our SMS.
+   * @returns {Promise.<Number>} Promise of shim's method that return number of
+   *  maximum concatenated number of our SMS. Default maximum number is 10.
+   */
+  maxConcatenatedMessages() {
+    return client.method('get', SMS_MAX_CONCAT_KEY).then((num) =>
+      num && !isNaN(num) ? num : MAX_CONCATENATED_MESSAGES
```
azasypkin: Btw, it works, but size is a string :) It's the same in master, so fix is not mandatory :)

PR #31910, opened 2015-09-18
Bug 1196809 - Pick best available icon for pinned sites on homescreen

gmarty:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=012f27c42dcce8730df5ad9523bb10bb90ba90b3

apps/homescreen/js/app.js
```diff
@@ -321,6 +324,21 @@ const SETTINGS_VERSION = 0;
   }
 
   App.prototype = {
+    get iconSize() {
+      // If this._iconSize is undefined or 0, let's refresh the value.
+      if (!this._iconSize) {
+        for (var i = 0; i < this.icons.childNodes.length; i++) {
```
Cwiiis: This loop will be very slow, better to do:
var children = this.icons.children;
for (var i = 0, iLen = children.length; i < iLen; i++) { ... }

Because the child-list in gaia-container is virtual, it gets built on every access of it.

apps/homescreen/js/app.js
```diff
@@ -321,6 +324,21 @@ const SETTINGS_VERSION = 0;
...
+          var container = this.icons.childNodes[i].firstChild;
+          if (container.style.display !== 'none' && container.firstChild) {
```
Cwiiis: Do we ever add empty containers to the container? Just wondering why we have the container.firstChild check.

apps/homescreen/js/app.js
```diff
@@ -435,6 +453,21 @@ const SETTINGS_VERSION = 0;
         handleRoleChange(icon.app, container);
       } else {
         icon.bookmark = appOrBookmark;
+
+        IconsHelper.getIconBlob(icon.bookmark.url, this.iconSize,
```
Cwiiis: I would rather this was done in gaia-site-icon. Would you ever want to set a bookmark and not have this behaviour? Also, the icon already knows and caches its best size, which would make the above added block unnecessary. Is there a reason this isn't done there?

shared/elements/gaia-site-icon/script.js
```diff
@@ -279,7 +280,7 @@ window.GaiaAppIcon = (function(exports) {
         this._image = null;
         this._hasIcon = true;
 
-        if (!this._hasPredefinedIcon && !this._hasUserSetIcon) {
+        if (!this._hasPredefinedIcon) {
```
Cwiiis: Alberto wanted this change for system, but we need to handle it in homescreen - icon-loaded will always be fired now, even if the icon was set by the icon property. This means that the homescreen will unnecessarily re-save every icon on startup, which would be best avoided.

apps/homescreen/js/app.js
gmarty: I totally thought this was already optimised by the platform and this kind of optimisations weren't required anymore. I'll fix it.

apps/homescreen/js/app.js
gmarty: That should never happen. So I'll remove it.

apps/homescreen/js/app.js
gmarty: Previously in this patch the logic was into the component itself, but due to dependencies, I had to move it into Gaia tree. A discussion with @albertopq convinced me to do it that way instead.
At the moment the only icon image that the `gaia-site-icon` element knows about is the one in `bookmark.icon`. It is not optimised for the space available on the homescreen.

apps/homescreen/js/app.js
Cwiiis: I thought so too, but I'm told it's not a given and I've been asked to make this change before so I'm just propagating that... I'll try harder to find out for certain at some point, because I think it looks really ugly...

The main bit to change was the this.icons.children access though, that property is actually a function that builds the array, so I doubt that'd get cached.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=21cee1e9508f95e7a0eac0bd09e85149d463b9d5

apps/homescreen/js/app.js
```diff
@@ -1006,7 +1030,11 @@ const SETTINGS_VERSION = 0;
               (icon.app.manifestURL + '/' +
                (icon.entryPoint ? icon.entryPoint : '')) : icon.bookmark.id;
             if (id === this.iconsToRetry[i]) {
-              icon.refresh();
+              if (icon.id) {
```
Cwiiis: I think this check is reversed (bookmarks have ids, apps don't)

shared/js/icons_helper.js
```diff
@@ -150,6 +149,25 @@
...
+   * @param targetSize {number}
+   * @returns {Promise}
+   */
+  function setElementIcon(icon, targetSize) {
+    getIconBlob(icon.bookmark.url, targetSize, icon.bookmark, icon.bookmark)
+      .then(iconObj => {
+        if (iconObj.blob) {
+          icon.icon = iconObj.blob;
+        } else if (icon.icon) {
+          icon.icon = icon.bookmark.icon;
```
Cwiiis: Isn't bookmark.icon a string? Not sure I understand this else clause.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=20478e206901b975ad571230842e267d733925eb

apps/homescreen/js/app.js
gmarty: Fixed!

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=15a3c7f5d52a6282feab11737c96ba08f3877446

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f59bd033f2408e5e8eb57aa03bcb563314abf921

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7f9fe2ce698c817f262877cb013d43035de4ad0e

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a31d797b69b9cf02525f3a5f9f53902de1f122e3

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=646f1f43a2c4d9638349efc2650441f30dff8542

shared/js/icons_helper.js
```diff
@@ -150,6 +149,39 @@
...
+   * @param targetSize {number}
+   * @returns {Promise}
+   */
+  function setElementIcon(icon, targetSize) {
+    getIconBlob(icon.bookmark.url, targetSize, icon.bookmark, icon.bookmark)
+      .then(iconObj => {
+        if (iconObj.blob) {
+          icon.icon = iconObj.blob;
+        } else if (icon.bookmark.icon) {
+          return icon.bookmark.icon;
```
Cwiiis: Is there any point in returning here, rather than just embedding the below fetchIconBlob directly?

gmarty: Not really. I guess it's a refactoring that went wrong!

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=35d981407c80ce2a434efdcf43e532b3dabd176a

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e6e8927caca1eb9da14683203a912574b5990e7d

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=38a6e086ce7bbe2c71f9ef5c20e5604f45cee361

PR #31912, opened 2015-09-18
Bug 1204941 - Write a Gaia UI test for bug 1201739 (add-on installation)

mwargers:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=65988d09cfb682e6a6b13c355b6bd5aa7cca7ec3

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7ebfa3b1c38be76838ec056c311880b657e1dd89

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/addon_details.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def enable(self):
+        self._checkbox.enable()
+
+    def disable(self):
+        self._checkbox.disable()
+
+    @property
+    def affected_apps(self):
+        return self.root_element.find_element(*self._affected_apps_locator).text
```
npark-mozilla: Because it can have more than one affected apps, I guess this should either return a list, or change the name to indicate that it shows one of the affected apps?

tests/python/gaia-ui-tests/gaiatest/tests/functional/marketplace/test_marketplace_addon.py
```diff
@@ -0,0 +1,80 @@
...
+        settings.launch()
+        addons_page = settings.open_addons()
+
+        for x in addons_page.items:
+            if x.addon_name == self.addon_name_to_install:
+                addon_details = x.tap()
+                break
+
+        self.assertTrue(addon_details.is_enabled())
+        self.assertEquals(addon_details.affected_apps, 'System')
```
npark-mozilla: This will not work in non-English locale.

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/addon_details.py
mwargers: No, like we discussed on irc, this is just the visible text of an element, which lists the apps the addon can influence. In this case, it's only the System app.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9e439176926d4742e41018a8f66df5df713af7c5

tests/python/gaia-ui-tests/gaiatest/tests/functional/marketplace/test_marketplace_addon.py
```diff
@@ -0,0 +1,82 @@
...
+        element = elements[0]
+        self.assertNotEquals(element.rect['x'], 0)
+        self.assertNotEquals(element.rect['y'], 0)
+        self.assertNotEquals(element.rect['width'], 0)
+        self.assertNotEquals(element.rect['height'], 0)
+        # element.text doesn't work, because Marionette thinks the element is not displayed for some reason
+        self.assertEquals(self.marionette.execute_script('return arguments[0].innerHTML', [element]), 'M')
+
+        self.apps.switch_to_displayed_app()
+        addon_details.refresh_root_element()
```
JohanLorenzo: What's the value of refreshing the element in the test? In my opinion, this should be an implementation details, that should belong in the Page Class.
What would it take to put that call in `disable()`?

tests/python/gaia-ui-tests/gaiatest/apps/marketplace/app.py
```diff
@@ -70,3 +72,70 @@ def tap_install_button(self):
...
     def get_app_name(self):
         return self.root_element.find_element(*self._name_locator).text
+
+class MarketplaceDev(Marketplace):
+
+    name = 'Dev'
+    manifest_url = 'https://marketplace.firefox.com/app/bee10fbb-7829-4d99-8a9f-a6d0a905138e/manifest.webapp'
+
+    _enable_addons_locator = (By.ID, 'enable-addons')
+    _m_in_statusbar_locator = (By.CSS_SELECTOR, 'a[href="/addon/m-in-status-bar"]')
```
JohanLorenzo: What does this represent?

tests/python/gaia-ui-tests/gaiatest/apps/marketplace/app.py
```diff
@@ -70,3 +72,70 @@ def tap_install_button(self):
...
+    _addon_install_locator = (By.CSS_SELECTOR, '.addon-install-btn')
+
+    def install_addon(self, addon_name):
```
JohanLorenzo: This method is about 60 lines long. At a first glance, it's hard to know what are the steps in order to install an addon. How about
``` python
def install_addon(self, addon_name):
    self._open_searchbox()
    self._search(addon_name)
    self._enable_addons()
    addon_page = self._open_addon()
    addon_page.install
```
?

JohanLorenzo: BTW, is it possible to use `search()` coming from the parent class?

tests/python/gaia-ui-tests/gaiatest/tests/functional/marketplace/test_marketplace_addon.py
```diff
@@ -0,0 +1,82 @@
...
+
+        addon_details.enable()
+
+        import time
+        time.sleep(5)
+
+        self.marionette.switch_to_frame()
+        elements = status_bar.root_element.find_elements(*self._status_bar_m_locator)
+        # this is to test that re-enabling the addon doesn't add another element, which is currently failing
+        print(len(elements))
```
JohanLorenzo: What should be the expected behavior? Can we put an insert instead?

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/addons.py
```diff
@@ -44,3 +54,16 @@ def tap_first_item(self):
...
     def toggle_addon_status(self):
         self.marionette.find_element(*self._state_toggle_locator).tap()
+
+
+class Addon(PageRegion):
+
+    _name_locator = (By.CSS_SELECTOR, 'span')
+
+    @property
+    def addon_name(self):
```
JohanLorenzo: Nit: `name` is enough, the class is already called `Addon`, then the common usage would be `addon.name`, instead of `addon.addon_name`

tests/python/gaia-ui-tests/gaiatest/tests/functional/marketplace/test_marketplace_addon.py
```diff
@@ -0,0 +1,82 @@
...
+        results = marketplace.install_addon(self.addon_name_to_install)
+
+        settings = Settings(self.marionette)
+        settings.launch()
+        addons_page = settings.open_addons()
+
+        for x in addons_page.items:
+            if x.addon_name == self.addon_name_to_install:
+                addon_details = x.tap()
+                break
```
JohanLorenzo: If the addon is missing, the loop will end and then we will have an unresolved reference for `addon_details`. How about 
``` python
for addon in addons_page.items:
    if addon.name == self.addon_name_to_install:
        addon_details = addon.tap()
        break
else:
    self.fail('{} was not found in {}'.format(self.addon_name_to_install, addons_page.items))
```

tests/python/gaia-ui-tests/gaiatest/tests/functional/marketplace/test_marketplace_addon.py
```diff
@@ -0,0 +1,82 @@
...
+        marketplace = MarketplaceDev(self.marionette)
+        marketplace.launch()
+
+        results = marketplace.install_addon(self.addon_name_to_install)
+
+        settings = Settings(self.marionette)
+        settings.launch()
+        addons_page = settings.open_addons()
+
+        for x in addons_page.items:
```
JohanLorenzo: Nit: `x` is not a great variable name. For example: Does it refer to a coordinate? Let's renaming to something closer to `addon`

tests/python/gaia-ui-tests/gaiatest/tests/functional/marketplace/test_marketplace_addon.py
```diff
@@ -0,0 +1,82 @@
...
+        self.apps.switch_to_displayed_app()
+        addon_details.refresh_root_element()
+        addon_details.disable()
+
+        addon_details.wait_for_toaster_displayed()
+
+        addon_details.enable()
+
+        import time
+        time.sleep(5)
```
JohanLorenzo: why is this sleep necessary? Is it to wait for the addon to appear?

tests/python/gaia-ui-tests/gaiatest/tests/functional/marketplace/test_marketplace_addon.py
```diff
@@ -0,0 +1,82 @@
...
+
+        self.assertTrue(addon_details.is_enabled())
+
+        self.assertEquals(addon_details.description, "Adds a 'M' to the status bar")
+        self.assertEquals(self.apps.is_app_installed(addon_details.affected_apps)['name'], 'System')
+        self.apps.switch_to_displayed_app()
+
+        system = System(self.marionette)
+        status_bar = system.status_bar
+        elements = status_bar.root_element.find_elements(*self._status_bar_m_locator)
```
JohanLorenzo: Finding elements in the test is not great. I agree, creating a page class along with the others seems overkill, bug maybe it's a good thing to create the page class just above the test (in the same file). This way, the test can be ligther:
``` python
status_bar_addon = StatusBarAddon(self.marionette)
self.assertTrue(status_bar_addon.is_displayed)
self.assertEqual(status_bar_addon.text, 'M')
```

``` python
class StatusBarAddon(PageRegion):
   _root_locator = # ...
   @property
    def is_displayed(self):
        # Workaround because is_displayed() doesn't work
        return self.root_element.rect['x'] != 0 and
                   self.root_element.rect['y'] != 0 and
                   self.root_element.rect['width'] != 0 and
                   self.root_element.rect['height'] != 0
    @property
    def text(self):
        # workaround
        return self.marionette.execute_script('return arguments[0].innerHTML', [self.root_element]), 'M'
```

tests/python/gaia-ui-tests/gaiatest/tests/functional/marketplace/test_marketplace_addon.py
mwargers: I need to do this because of https://bugzilla.mozilla.org/show_bug.cgi?id=1202246 .
Ideally, this should not be necessary at all.
This is tied to the switching to the system app, then switching back to the displayed app.
I would rather keep it like this, if you don't mind.
I think any way of putting this call in disable () would make the code even more awkward.

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/addons.py
```diff
@@ -32,11 +34,19 @@ def details_screen_element(self):
         return self.marionette.find_element(*self._details_page_locator)
 
     @property
+    def items(self):
+        elements = self.marionette.find_elements(*self._items_locator)
```
JohanLorenzo: We can shorten this function into something smaller:
```python
return [Addon(self.marionette, element) for element in self.marionette.find_elements(*self._items_locator)]
```

This is called a [list comprehension](https://docs.python.org/2.7/tutorial/datastructures.html#list-comprehensions) and we use it at some other places in the page classes

tests/python/gaia-ui-tests/gaiatest/tests/functional/marketplace/test_marketplace_addon.py
mwargers: Yeah, sorry, this last part (including the time.sleep(5) is basically about testing whether disabling, then re-enabling the addon doesn't add a second 'M' in the statusbar.
Currently, this is broken, that's why print(len(elements)) returns 2, where it should return only 1.
Now that you mention it, disabling the addon doesn't work either, currently.

I guess it might be better to assert on current broken behavior and when it would be fixed, we could just update the test, when the test would suddenly fail, right?

tests/python/gaia-ui-tests/gaiatest/apps/marketplace/app.py
mwargers: Yes, I'll do that.

tests/python/gaia-ui-tests/gaiatest/tests/functional/marketplace/test_marketplace_addon.py
mwargers: Ok, I can remove the print(). But it means that this doesn't automatically fails when this suddenly gets fixed. We would then notice it directly and flip the check here.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2a828197f5ffb915f9d35190012af6f2c10a1370

tests/python/gaia-ui-tests/gaiatest/apps/marketplace/app.py
```diff
@@ -70,3 +72,57 @@ def tap_install_button(self):
...
+        # Addons page gets loaded, tap on the addon link to open that page
+        addon_page = self._open_addon(addon_name)
+        addon_page.install()
+
+class AddonPage(Base):
+
+    _addon_install_locator = (By.CSS_SELECTOR, '.addon-install-btn')
+
+    def install(self):
+        element = Wait(self.marionette, timeout=20, interval=1).until(
```
JohanLorenzo: Nit: Let's use a better name like `install_link` or `install_button`

tests/python/gaia-ui-tests/gaiatest/apps/marketplace/app.py
```diff
@@ -70,3 +72,57 @@ def tap_install_button(self):
...
+            expected.element_present(*self._addon_install_locator))
+        Wait(self.marionette).until(expected.element_displayed(element))
+        element.tap()
+
+        # Confirm the installation and wait for the app icon to be present
```
JohanLorenzo: "wait for the app icon to be present" => I don't see it in the code. Should we remove this comment?

tests/python/gaia-ui-tests/gaiatest/apps/marketplace/app.py
```diff
@@ -70,3 +72,57 @@ def tap_install_button(self):
...
+        confirm_install = ConfirmInstall(self.marionette)
+        confirm_install.tap_confirm()
+
+        # Wait for the notification to disappear
```
JohanLorenzo: This comment is not exactly true. We also wait for the notification to appear.

I would remove it, we already have this information 2 lines below.

tests/python/gaia-ui-tests/gaiatest/apps/marketplace/app.py
```diff
@@ -70,3 +72,57 @@ def tap_install_button(self):
...
+        element.tap()
+        return AddonPage(self.marionette)
+
+    def install_addon(self, addon_name):
+        self.search(':debug')
+
+        # Debug page is opened, tap on the "Enable Firefox OS Add-ons" button
+        self._enable_addons()
+
+        # Addons page gets loaded, tap on the addon link to open that page
```
JohanLorenzo: Duplication of the next 2 lines

tests/python/gaia-ui-tests/gaiatest/apps/marketplace/app.py
```diff
@@ -70,3 +72,57 @@ def tap_install_button(self):
...
+        element = Wait(self.marionette, timeout=20, interval=1).until(
+            expected.element_present(*addon_locator))
+        Wait(self.marionette).until(expected.element_displayed(element))
+        element.tap()
+        return AddonPage(self.marionette)
+
+    def install_addon(self, addon_name):
+        self.search(':debug')
+
+        # Debug page is opened, tap on the "Enable Firefox OS Add-ons" button
```
JohanLorenzo: "tap on the "Enable Firefox OS Add-ons" button" is not necessary. I would move "Debug page is opened" before, so we have something like:
```python
# Open debug page
self.search(':debug')
```

tests/python/gaia-ui-tests/gaiatest/apps/marketplace/app.py
```diff
@@ -70,3 +72,57 @@ def tap_install_button(self):
...
+
+class MarketplaceDev(Marketplace):
+
+    name = 'Dev'
+    manifest_url = 'https://marketplace.firefox.com/app/bee10fbb-7829-4d99-8a9f-a6d0a905138e/manifest.webapp'
+
+    _enable_addons_locator = (By.ID, 'enable-addons')
+
+    def _enable_addons(self):
+        element = Wait(self.marionette).until(
```
JohanLorenzo: `element` is a bit too generic. `enable_addon_button` sounds better

tests/python/gaia-ui-tests/gaiatest/apps/marketplace/app.py
```diff
@@ -70,3 +72,57 @@ def tap_install_button(self):
...
+            expected.element_present(*self._enable_addons_locator))
+        Wait(self.marionette).until(expected.element_displayed(element))
+        element.tap()
+
+    def _open_addon(self, addon_name):
+        selector = '//a[text()="%s"]' % addon_name
+        addon_locator = (By.XPATH, selector)
+
+        element = Wait(self.marionette, timeout=20, interval=1).until(
```
JohanLorenzo: `element` is a bit too generic

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/addon_details.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @property
+    def description(self):
+        return self.root_element.find_element(*self._description_locator).text
+
+    def wait_for_toaster_displayed(self):
```
JohanLorenzo: Unused function

tests/python/gaia-ui-tests/gaiatest/tests/functional/marketplace/test_marketplace_addon.py
```diff
@@ -0,0 +1,87 @@
...
+        # Workaround because is_displayed() doesn't work here
+        return (self.root_element.rect['x'] != 0 and
+                self.root_element.rect['y'] != 0 and
+                self.root_element.rect['width'] != 0 and
+                self.root_element.rect['height'] != 0)
+
+    @property
+    def text(self):
+        self.marionette.switch_to_frame()
+        # workaround because is_displayed doesn't work here
```
JohanLorenzo: Nit: text() doesn't work here

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/addon_details.py
```diff
@@ -0,0 +1,58 @@
...
+
+    #  workaround for bug 1202246.  Need to call this method after frame switching
+    def refresh_root_element(self):
+        self.root_element = self.marionette.find_element(*self._details_page_locator)
+
+    @property
+    def _checkbox(self):
+        return GaiaBinaryControl(self.marionette, self._state_toggle_locator)
+
+    def is_enabled(self):
```
JohanLorenzo: Nit: `@property`

tests/python/gaia-ui-tests/gaiatest/apps/marketplace/app.py
mwargers: I'm only using `element` locally. I find it generally easier to work then with a generic name than to use special names for variables which I only use locally and don't reuse at all.

tests/python/gaia-ui-tests/gaiatest/tests/functional/marketplace/test_marketplace_addon.py
mwargers: Well, it's because is_displayed doesn't work, because .text is internally checking if the element is displayed.

tests/python/gaia-ui-tests/gaiatest/apps/marketplace/app.py
mwargers: Yes, I copied this code from somewhere, probably from one of the homescreen tests.

tests/python/gaia-ui-tests/gaiatest/tests/functional/marketplace/test_marketplace_addon.py
JohanLorenzo: Oh, I wasn't aware of that! That's worth to be mentioned in the code as a comment, I think

tests/python/gaia-ui-tests/gaiatest/apps/marketplace/app.py
mwargers: Not sure I like it myself, but I'll change it.

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/addon_details.py
mwargers: Thanks, I'll remove it.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e723b03281d56adcc57d11c37846669af7cba0c2

PR #31919, opened 2015-09-18
Bug 1204687 - Fix bluetooth open received file race condition between…

aosmond:
… device storage and lazy loader.

### DIFF
--- a/apps/system/js/bluetooth_transfer.js
+++ b/apps/system/js/bluetooth_transfer.js
@@ -643,59 +643,58 @@ var BluetoothTransfer = {
     var storageType = 'sdcard';
     var storage = navigator.getDeviceStorage(storageType);
     var getreq = storage.get(filePath);
+    var loader = LazyLoader.load(['shared/js/mime_mapper.js']);
+    Promise.all([getreq, loader]).then(() => {
+      var file = getreq.result;
+      // When we got the file by storage type of "sdcard"
+      // use the file.type to replace the empty fileType which is given by API
+      var fileName = file.name;
+      var extension = fileName.split('.').pop();
+      var originalType = file.type || contentType;
+      var mappedType = (MimeMapper.isSupportedType(originalType)) ?
+        originalType : MimeMapper.guessTypeFromExtension(extension);
 
-    LazyLoader.load(['shared/js/mime_mapper.js']).then(() => {
-      getreq.onerror = () => {
-        var msg = 'failed to get file:' +
-                  filePath + getreq.error.name +
-                  getreq.error.name;
+      var a = new MozActivity({
+        name: mappedType == 'text/vcard' ? 'import' : 'open',
+        data: {
+          type: mappedType,
+          blob: file,
+          // XXX: https://bugzilla.mozilla.org/show_bug.cgi?id=812098
+          // Pass the file name for Music APP since it can not open blob
+          filename: fileName
+        }
+      });
+
+      a.onerror = (e) => {
+        var msg = 'open activity error:' + a.error.name;
+        this.debug(msg);
+        switch (a.error.name) {
+        case 'NO_PROVIDER':
+          UtilityTray.hide();
+          // Cannot identify MIMETYPE
+          // So, show cannot open file dialog with unknow media type
+          this.showUnknownMediaPrompt(fileName);
+          return;
+        case 'ActivityCanceled':
+          return;
+        case 'USER_ABORT':
+          return;
+        default:
+          return;
+        }
+      };
+      a.onsuccess = (e) => {
+        var msg = 'open activity onsuccess';
         this.debug(msg);
       };
-
-      getreq.onsuccess = () => {
-        var file = getreq.result;
-        // When we got the file by storage type of "sdcard"
-        // use the file.type to replace the empty fileType which is given by API
-        var fileName = file.name;
-        var extension = fileName.split('.').pop();
-        var originalType = file.type || contentType;
-        var mappedType = (MimeMapper.isSupportedType(originalType)) ?
-          originalType : MimeMapper.guessTypeFromExtension(extension);
-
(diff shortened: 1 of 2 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e2c6f5d29b40083cf65bbd77ee7dcde7fcc7ad95

apps/system/js/bluetooth_transfer.js
```diff
@@ -644,15 +644,17 @@ var BluetoothTransfer = {
     var storage = navigator.getDeviceStorage(storageType);
     var getreq = storage.get(filePath);
```
timdream: I assume this is a DOMRequest. If so it's a thenable. So better way to do this would be:

```javascript
var getreq = storage.get(filePath);
var loader = LazyLoader.load(['shared/js/mime_mapper.js']);
Promise.all([getreq, loader])
  .then(<success>, <error>)
  .catch(...);
```

The `then()` block would ensure both thenables is resolved and the `.catch()` would catch throw from both functions.

aosmond: A DOMRequest isn't a promise, so Promise.all will simply create a resolved promise which yields getreq as a parameter. It will work if I use "getreq.then()" to get an actual promise representing the transaction properly however.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=bf545f5149089653402999bd892d34348037f30c

apps/system/js/bluetooth_transfer.js
timdream: No. According to [1], elements passes into Promise.all([...]) are automatically wrapped with `Promise.resolve()` if it is not a promise.

According to [2], promise returned by `Promise.resolve()` only resolves when the `then()` of a "thenable" is called. So it's safe to pass the `getreq` into it directly and safely.

[1] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all
[2] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve

apps/system/js/bluetooth_transfer.js
```diff
@@ -643,60 +643,54 @@ var BluetoothTransfer = {
     var storageType = 'sdcard';
     var storage = navigator.getDeviceStorage(storageType);
     var getreq = storage.get(filePath);
+    var loader = LazyLoader.load(['shared/js/mime_mapper.js']);
+    Promise.all([getreq.then(), loader]).then(() => {
```
timdream: Therefore, this line should be 

`Promise.all([getreq, loader]).then(() => {`

apps/system/js/bluetooth_transfer.js
```diff
@@ -643,60 +643,54 @@ var BluetoothTransfer = {
...
+      var file = getreq.result;
+      // When we got the file by storage type of "sdcard"
+      // use the file.type to replace the empty fileType which is given by API
+      var fileName = file.name;
+      var extension = fileName.split('.').pop();
+      var originalType = file.type || contentType;
+      var mappedType = (MimeMapper.isSupportedType(originalType)) ?
+        originalType : MimeMapper.guessTypeFromExtension(extension);
+
+      var a = new MozActivity({
```
timdream: This also returns a DOMRequest :) you could chain this into the main chain if you want to.

apps/system/js/bluetooth_transfer.js
```diff
@@ -643,60 +643,54 @@ var BluetoothTransfer = {
...
-          this.debug(msg);
-        };
+      a.onsuccess = (e) => {
+        var msg = 'open activity onsuccess';
+        this.debug(msg);
       };
     }).catch((err) => {
+      var msg = 'failed to get file:' +
+                filePath + ' ' + err.name;
+      this.debug(msg);
```
timdream: It is better to do the following than put these lines in the final `catch()` callback:

```javascript
Promise.all([getreq, loader])
  .then(function() {
     .... // Handling the request
  }, function(errors) {
    if (errors[0]) { // DeviceStorage Error
      var msg = 'failed to get file:' + filePath + ' ' + errors[0].name;
      this.debug(msg);
    }
    if (errors[1]) { // LazyLoader Error
      throw errors[1];
    }
  })
  .catch(function(e) {
    console.error(e);
  });
```

By using the second `onReject` callback of `then()` method, you would retain the opportunity to handle the deviceStorage error separately.

The final `catch()` will also help us print out TypeError if there is any typo in any (rejected or resolved) functions in the chain.

It is also important to recognize a promise chain is also an error handling chain. You should always try to handle the error in your own step in the chain and throw the error to the next `catch` callback if you couldn't handle them -- or, handle the error but still throw to abort all the steps follows.

apps/system/js/bluetooth_transfer.js
aosmond: And according to the implementation in Gecko:

Promise::All implementation goes through each value in the array and calls ::Resolve to extract a promise object from it:
https://dxr.mozilla.org/mozilla-central/source/dom/promise/Promise.cpp#975

If the value *is* a Promise, it returns the same object:
https://dxr.mozilla.org/mozilla-central/source/dom/promise/Promise.cpp#729

If the value *isn't* a Promise, it returns a new promise resolved to that value:
https://dxr.mozilla.org/mozilla-central/source/dom/promise/Promise.cpp#760

It doesn't matter if the object has a .then method.

aosmond: I retract previous statement, you are correct. I totally missed the check for the .then method elsewhere in the file.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=23c821a2872b9336479f413c280ca16b30e815d9

PR #31920, opened 2015-09-18
Bug 1037620 - Late Arrival Notice in Messages App r=julien

sroucheray:
In SMS app :
Add a humanized lateness (diff between sent and received time) notice in the bubbles of conversation view
Add a humanized lateness message in report view


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f617806e8a2f074a1423bf83ae397b91c8ec7ef6

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=204bf6284bf5bae6b6131e66c44a4e6cf70e91fc

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2bdc2dbe9a495632ef77f1756a4a5fb449a14a8d

build/csslint/xfail.list
```diff
@@ -55,6 +55,7 @@ apps/settings/style/settings.css 0 9
 apps/settings/style/settings_large.css 1 1
 apps/settings/style/screen_lock.css 0 2
 apps/settings/style/downloads.css 0 6
+apps/sms/views/conversation/style/message.css 2 0
```
julienw: maybe you can try to contribute to csslint to make the tool know about margin-inline-start and margin-inline-end (and their friends) ?

julienw: at least file a bug there :)

apps/sms/locales/sms.en-US.properties
```diff
@@ -316,6 +316,39 @@ message-delivery-status-delivered.ariaLabel          = Delivered
...
+message-lateness-hours[other]   = {{ hours }}h
+
+message-lateness-minutes        = {[ plural(minutes) ]}
+message-lateness-minutes[zero]  =
+message-lateness-minutes[other] = {{ minutes }}m
+
+message-lateness-seconds        = {[ plural(seconds) ]}
+message-lateness-seconds[zero]  =
+message-lateness-seconds[other] = {{ seconds }}s
+
```
julienw: hey @zbraniecki could the Intl API help with the duration translations here ?

julienw: stéphane, maybe Zibi will contradict me here, but you need to put all "flavors" of plurals here, even if in english we don't use all of them; here is an example:
```
selected-threads         = {[ plural(n) ]}
selected-threads[zero]   = none selected
selected-threads[one]    = {{n}} selected
selected-threads[two]    = {{n}} selected
selected-threads[few]    = {{n}} selected
selected-threads[many]   = {{n}} selected
selected-threads[other]  = {{n}} selected
```

I think this is a limitation of our translation tools.

Also we need to provide a translation for "zero"; if we want nothing to be displayed, then we need to do it programmatically (either hide in CSS or not add it in JS).

zbraniecki: Funny you're asking. We just got a first draft of the spec: https://github.com/tc39/ecma402/issues/35 :)

I will be working on shimming it into `moz_intl.js` soon. For our uses I will want to offer async version (using `formatValue`) and DOM version where you pass en element and we assign `data-l10n-id`.

If you want to move this part to moz_intl.js already, feel free! I'll be happy to collaborate on moz_intl.js :)

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1840,6 +1860,46 @@ var ConversationView = {
...
+    var noticeNode = this.tmpl.lateArrivalNotice.prepare({
+        noticeL10nId: noticeL10nIds[period]
+    }).toDocumentFragment();
+
+    navigator.mozL10n.setAttributes(
+      noticeNode.querySelector('.late-arrival-notice'), 
+      noticeL10nIds[period], 
+      {
+        'lateness': (lateness - lateness % period) / period
+      }
```
julienw: proposal: you can calculate the lateness to be displayed before, and then have a simpler call to setAttributes, like this:
```javascript
lateness = Math.floor(lateness / period);
navigator.mozL10n.setAttributes(
   noticeNode.querySelector('.late-arrival-notice'), 
   noticeL10nIds[period], 
   { lateness }
);
```

alternatively:
```javascript
navigator.mozL10n.setAttributes(
   noticeNode.querySelector('.late-arrival-notice'), 
   noticeL10nIds[period], 
   { lateness: Math.floor(lateness / period) }
);
```

apps/sms/locales/sms.en-US.properties
zbraniecki: Yeah, for now we need all variants :( We plan to remove this limitation post 2.5.

apps/sms/views/conversation/js/conversation.js
```diff
@@ -79,6 +84,13 @@ var ConversationView = {
     preventLastCall: false
   },
 
+  LATE_NOTICE_OPTS: {
```
julienw: nit: instead of OPTS you can use LIMITS; and then you can remove _LIMIT from all properties :)

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1840,6 +1860,46 @@ var ConversationView = {
...
+    if (!sent || !received || lateness < opts.OFF_NOTICE_LOWER_LIMIT) {
+      return '';
+    }
+
+    if (lateness < opts.IN_HOURS_LIMIT)       { period = MINUTE; } 
+    else if (lateness < opts.IN_DAYS_LIMIT)   { period = HOUR; } 
+    else if (lateness < opts.IN_MONTHS_LIMIT) { period = DAY; }
+    
+    var noticeNode = this.tmpl.lateArrivalNotice.prepare({
+        noticeL10nId: noticeL10nIds[period]
```
julienw: nit: we increment by 2 spaces :)

apps/sms/views/conversation/js/information.js
```diff
@@ -339,6 +346,42 @@ var VIEWS = {
...
+            seconds: SECOND
+          };
+          var lateness = message.timestamp - message.sentTimestamp;
+
+          this.latenessBlock.classList.remove('hide');
+          if (
+            !message.timestamp || 
+            !message.sentTimestamp || 
+            lateness < LATENESS_DISPLAY_LOWER_LIMIT
+          ) {
```
julienw: nit: this is not very readable; maybe extract the condition to a carefully named boolean variable ? `shouldDisplayLatenessIndicator` for example :)

apps/sms/views/conversation/js/information.js
```diff
@@ -339,6 +346,42 @@ var VIEWS = {
...
+            this.latenessBlock.classList.add('hide');
+          } else {
+            var rest = 0;
+            ['months', 'days', 'hours', 'minutes', 'seconds'].forEach((val) => {
+              var output = 0;
+              var camelCaseId = Utils.camelCase('lateness-' + val);
+              if(lateness >= units[val]) {
+                rest = lateness % units[val];
+                output = (lateness - rest) / units[val];
```
julienw: you can use Math.floor here as well.

apps/sms/views/conversation/js/information.js
```diff
@@ -339,6 +346,42 @@ var VIEWS = {
...
+            !message.sentTimestamp || 
+            lateness < LATENESS_DISPLAY_LOWER_LIMIT
+          ) {
+            this.latenessBlock.classList.add('hide');
+          } else {
+            var rest = 0;
+            ['months', 'days', 'hours', 'minutes', 'seconds'].forEach((val) => {
+              var output = 0;
+              var camelCaseId = Utils.camelCase('lateness-' + val);
+              if(lateness >= units[val]) {
```
julienw: nit: a space is missing after "if"

apps/sms/views/conversation/js/information.js
```diff
@@ -339,6 +346,42 @@ var VIEWS = {
...
+          this.latenessBlock.classList.remove('hide');
+          if (
+            !message.timestamp || 
+            !message.sentTimestamp || 
+            lateness < LATENESS_DISPLAY_LOWER_LIMIT
+          ) {
+            this.latenessBlock.classList.add('hide');
+          } else {
+            var rest = 0;
+            ['months', 'days', 'hours', 'minutes', 'seconds'].forEach((val) => {
```
julienw: maybe "unit" instead of "val" ?

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1840,6 +1860,46 @@ var ConversationView = {
...
+
+    navigator.mozL10n.setAttributes(
+      noticeNode.querySelector('.late-arrival-notice'), 
+      noticeL10nIds[period], 
+      {
+        'lateness': (lateness - lateness % period) / period
+      }
+    );
+
+    return noticeNode.firstChild.outerHTML;
```
julienw: Mmm I don't really like this, we serialize to HTML so that we can inject it later on; maybe it's better to just insert the fragment at the right place ?

apps/sms/views/conversation/js/information.js
```diff
@@ -339,6 +346,42 @@ var VIEWS = {
...
+              var output = 0;
+              var camelCaseId = Utils.camelCase('lateness-' + val);
+              if(lateness >= units[val]) {
+                rest = lateness % units[val];
+                output = (lateness - rest) / units[val];
+                lateness = rest;
+              }
+
+              setL10nAttributes(
+                this[camelCaseId],
```
julienw: I don't see where this comes from, but maybe I miss something ?

julienw: ah sorry, I just saw it below

julienw: in that case I think it's better if we just query it here, because it's used only here. What do you think ?

apps/sms/views/conversation/js/information.js
```diff
@@ -400,7 +445,6 @@ var Information = function(type) {
   }
 
   this.panel = document.querySelector(this.selector);
-
```
julienw: nit: please add this empty line back :)

apps/sms/views/conversation/style/message.css
```diff
@@ -142,6 +142,20 @@
   color: #d1eeee;
 }
 
+.message.incoming .message-details .late-arrival-notice {
```
julienw: nit, I believe we only need `.late-arrival-notice` as selector ?

apps/sms/views/conversation/style/message.css
```diff
@@ -142,6 +142,20 @@
...
+  color: #026a69;
+  font-size: 1.3rem;
+  display: flex;
+  align-items: center;
+}
+
+.message.incoming .message-details .late-arrival-notice:before {
```
julienw: same here

apps/sms/views/conversation/style/message.css
```diff
@@ -142,6 +142,20 @@
...
+  content: url('images/icons/late_clock.png');
```
julienw: I think we'll want to use a background image with a background-size instead, because I don't think this works fine with high-dpi devices. I haven't tried yet though.

apps/sms/views/conversation/test/unit/conversation_test.js
```diff
@@ -2364,6 +2364,61 @@ suite('conversation.js >', function() {
...
+            var noticeNode = node.querySelector('.late-arrival-notice');
+            if(index < 2){
+              assert.isNull(noticeNode);
+            } else {
+              assert.isNotNull(noticeNode);
+              assert.equal(noticeNode.dataset.l10nId, dataL10nIds[index]);
+            }
+          });
+        })
+      ).then(() => done(), () => done());
```
julienw: this won't work: you're gobbling all errors with this then :)

i'd propose to generate 1 test for each delay, instead. See https://github.com/mozilla-b2g/gaia/blob/e21a2e5c1fb7bf329553a2ae98a8b9e6bef0028b/apps/sms/views/shared/test/unit/startup_test.js#L120 as an example

apps/sms/views/conversation/test/unit/conversation_test.js
```diff
@@ -2364,6 +2364,61 @@ suite('conversation.js >', function() {
...
+        'late-arrival-notice-in-hours',
+        'late-arrival-notice-in-days',
+        'late-arrival-notice-in-days',
+        'late-arrival-notice-in-months',
+        'late-arrival-notice-in-months'
+      ];
+
+      Promise.all(
+        delays.map((delay, index) => {
+          var message =MockMessages.sms({
```
julienw: nit: `=` should be surrounded by 1 space

apps/sms/locales/sms.en-US.properties
julienw: > If you want to move this part to moz_intl.js already, feel free! I'll be happy to collaborate on moz_intl.js :)

Mmm I don't see where moz_intl.js is. So I'd say we can work without it for now and convert later ?

build/csslint/xfail.list
sroucheray: @julienw Yes it resides in the parser-lib https://github.com/CSSLint/parser-lib/blob/master/src/css/Properties.js

apps/sms/views/conversation/js/conversation.js
sroucheray: @julienw I don't like that either :) I was wondering if the mozL10n had an API that given an ID and some data would translate and return a simple string. These string could be inserted later in the template mechanism. Maybe it is [`L10n.get`](https://developer.mozilla.org/en-US/docs/Web/API/L10n) but there is no documentation for the method signature. Or, as you said, I can return the fragment and insert it manually.
What do you think ?

apps/sms/views/conversation/style/message.css
julienw: yes I confirm this; this is not high-dpi friendly.

to test this easily on Firefox, you can rename one of the high-dpi images (for example the @2x image) to the not-suffixed name.

I don't think we can make it work with a pseudo element, so better use the usual way of using background-image and background-size (or the background shortcut of course)

apps/sms/views/conversation/js/conversation.js
sroucheray: OK, I have given a closer look at the L10n API and I will make you a Promise based proposal in the next commit.

zbraniecki: `mozL10n.get` is deprecated, so if you need this, use async `mozL10n.formatValue` please.

But I'm wondering if you could just assign this id/args to an element via `mozL10n.setAttributes` instead.

sroucheray: @zbraniecki I saw in the l10n.js that the `mozL10n.get` method was using a legacy function so I switched to using `mozL10n.formatValue` which, as a bonus, gives a much cleaner code IMO.

apps/sms/views/conversation/style/message.css
sroucheray: @julienw I would let `.message.incoming .late-arrival-notice` just for the sake of semantic because the late arrival notice only applies to incoming messages.

apps/sms/views/conversation/style/message.css
sroucheray: @julienw I have managed to kept the pseudo element and use background-size at the same time. I tested it in mulet and it seems to work but you may double check this :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=82c8e19d71515432204386d78e267a60d0da5b4c

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=76283c492b8ccdd9d30a012a05994f977c334189

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=acb49e79004ed6a48113c0376665911adce8392c

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=74e3054c1053dbe6d1e1e397b9ba0191f787436a

apps/sms/views/conversation/js/information.js
```diff
@@ -339,6 +346,51 @@ var VIEWS = {
...
+              setL10nAttributes(
+                unitElement,
+                'message-lateness-' + unit, 
+                { [unit]: output }
+              );
+            });
+            this.latenessBlock.classList.remove('hide');
+          } else {
+            this.latenessBlock.classList.add('hide');
+          }
```
julienw: I'd move all this code to an external function, what do you think ?

apps/sms/views/conversation/js/information.js
```diff
@@ -339,6 +346,51 @@ var VIEWS = {
...
+          if (showLatenessNotice) {
+            [
+              'months',
+              'days',
+              'hours',
+              'minutes',
+              'seconds'
+            ].forEach((unit) => {
+              var output = 0;
+              var unitElement = this[Utils.camelCase('lateness-' + unit)];
```
julienw: I wasn't very clear in my previous comment; what do you think of doing the querySelector here (because it's used only here) instead of binding the element at init time ?

apps/sms/views/conversation/js/conversation.js
```diff
@@ -79,6 +84,13 @@ var ConversationView = {
     preventLastCall: false
   },
 
+  LATE_NOTICE_LOWER_LIMITS: {
+    NO_NOTICE: 1 * MINUTE,
```
julienw: I feel we should not display it here for less than 5 minutes. Because it's quite current to have desynchronized clocks (for example, my phone has a 2-minute difference from my girlfriend's phone !)

apps/sms/views/conversation/js/information.js
```diff
@@ -15,6 +15,13 @@
 (function(exports) {
 'use strict';
 
+const SECOND = 1000;
+const MINUTE = 60 * SECOND;
+const HOUR = 60 * MINUTE;
+const DAY = 24 * HOUR;
+const MONTH = 30 * DAY;
+const LATENESS_DISPLAY_LOWER_LIMIT = 1 * MINUTE;
```
julienw: same comment that we should not display if it's below 5 minutes

sroucheray: @julienw OK

apps/sms/views/conversation/js/conversation.js
sroucheray: @julienw OK

apps/sms/views/conversation/js/information.js
sroucheray: OK I understand now :) I'll do the change.

BTW Wouldn't it make sense to lazy query select, replacing this :

```javascript 
this.elements.forEach(function(name) {
  this[Utils.camelCase(name)] = this.panel.querySelector('.' + name);
}, this);

//this.contactList.setAttributes...
```
...by this :
```javascript 
this.elements.forEach(function(name) {
  var element;

  this[Utils.camelCase(name)] = () => {
    if (!element) {
      element = this.panel.querySelector('.' + name);
    }
    
    return element;
  };
}, this);

//this.contactList().setAttributes..
```

julienw: we could even use defineProperty with a getter ;) but I don't think it's worth it now, unless you profile and tell me this is costly ;)

sroucheray: I forgot we can use accessors ! That would even don't change other pieces of code who require queried elements. I don't think it's worth it in the information panel but can be a good move elsewhere if it is a pattern. It would definitely require profiling testing.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=07cedd6b4982c4d40d884f19ddfab05c7371e456

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6d9dad89cca409be346dd7fd4c07b26e3e651b1a

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=301e2feea2a084f7e50e34b0dd23e19410b692bf

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=37ad63f1883cb8245d974bc87295efc7b2c1f392

julienw:
You need to rebase to latest master but it seems easy (I think only xfail.list needs a manual conflict resolution)

apps/sms/locales/sms.en-US.properties
```diff
@@ -316,6 +316,62 @@ message-delivery-status-delivered.ariaLabel          = Delivered
...
+message-lateness-minutes[many]  = {{ minutes }}m
+message-lateness-minutes[other] = {{ minutes }}m
+
+message-lateness-seconds        = {[ plural(seconds) ]}
+message-lateness-seconds[zero]  = {{ seconds }}s
+message-lateness-seconds[one]   = {{ seconds }}s
+message-lateness-seconds[two]   = {{ seconds }}s
+message-lateness-seconds[few]   = {{ seconds }}s
+message-lateness-seconds[many]  = {{ seconds }}s
+message-lateness-seconds[other] = {{ seconds }}s
```
julienw: Maybe move the strings used in Message Report dialog to the "Message report" part of the file ? so that it's easier for localizers.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=13b2de9dea45b314316a494d9487b88f62878fa5

apps/sms/locales/sms.en-US.properties
zbraniecki: Actually, could you put those strings in `./shared/locales/intl/unitformat.en-US.properties` ? We are going to work on `Intl.UnitFormat` soon and we could then use those strings and switch your use case, first to `mozIntl.UnitFormat` and then to full API once we get this in the platform (and then we can remove those strings).

The name of strings should be 'duration-hour-narrow`, `duration-minute-narrow` etc.

That would make those strings reusable for other apps and we wouldn't need to move that later for localizers.

If you'd prefer not to play with this in this patch, I'll be happy to move them in the follow up.

zbraniecki: Corresponding data for french that we're going to use once we get this in the platform is here: http://www.unicode.org/cldr/charts/28/summary/fr.html#5834

We will have also short/long, but what you have here is just `narrow` variant. one/other is plural form.

apps/sms/locales/sms.en-US.properties
```diff
@@ -308,6 +349,19 @@ message-delivery-status-delivered.ariaLabel          = Delivered
...
+late-arrival-notice-in-months         = {[ plural(lateness) ]}
+late-arrival-notice-in-months[zero]   = {{ lateness }}mth late
+late-arrival-notice-in-months[one]    = {{ lateness }}mth late
+late-arrival-notice-in-months[two]    = {{ lateness }}mths late
+late-arrival-notice-in-months[few]    = {{ lateness }}mths late
+late-arrival-notice-in-months[many]   = {{ lateness }}mths late
+late-arrival-notice-in-months[other]  = {{ lateness }}mths late
+late-arrival-notice-in-days           = {{ lateness }}d late
+late-arrival-notice-in-hours          = {{ lateness }}h late
+late-arrival-notice-in-minutes        = {{ lateness }}m late
```
zbraniecki: I hope you have appropriate space for that, since this string will probably take much more space in many languages. @flodolo ?

Is there any chance to use `Late arrival: 5 months` instead of `5mths late`? That would also allow you to jump on the `Intl.UnitFormat` for `duration-month-long` instead of custom strings.

flodolo: I think "mth" is quite confusing, several locales won't have such an abbreviation and will end up using the full "month" word, taking up a lot of space. I agree with gandalf, something like "Late arrival: XXX" should be easier to manage on the localization side.

julienw: I don't think we have the room to add "Late arrival" here. Why isn't "5 months late" possible ? Maybe we can just skip having an abbreviation for months...

flodolo: It will still get long ([example](https://transvision.mozfr.org/string/?entity=apps/calendar/calendar.properties:months-after[two]&repo=gaia)).

julienw: in our case I think it's ok, especially because we'll likely never get months for this value. But having "Late arrival" all the time is too long. This string ends up at the bottom of a message in the conversation.

So Zibi, flod, do you prefer to have 'months' without an abbreviation here? Does it make more sense? Or should we have an abbreviation for the locales that can have one (we can add a localization note to state this of course) ?

zbraniecki: abbreviation is ok, the problem is with "{{time}} late". But for now, keep it, just make sure you have enough space for much longer version :)

flodolo: +1 on what gandalf said. Localizers will still be able to use the full word is the abbreviation is awkward.

julienw: yeah I think we have enough room, but please ping if anything is wrong in some language.

apps/sms/locales/sms.en-US.properties
julienw: @sroucheray do you think you can do that ? I'm fine if you want to do this in a follow-up bug so that we can finish this first.

julienw: ok, you'll hate me ;) I'm sorry I see this only now. I'd simply use "value" everywhere as the variable, to make things simpler, both in code, and I guess it's also more future-proof with the Intl changes.

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1703,6 +1717,22 @@ var ConversationView = {
       pElement.setAttribute('data-l10n-id', 'no-attachment-text');
     }
 
+    var lateArrivalInfos = this.computeLateArrivalInfos(message, messageStatus);
+    if (lateArrivalInfos) {
+      var lateArrivalNoticeDOM = this.tmpl.lateArrivalNotice.prepare({})
```
julienw: so, I checked the existing lib, and I think it actually already supports having no options. See https://github.com/mozilla-b2g/gaia/blob/master/shared/js/template.js#L122

What was the issue you saw when you tried to remove the argument ?

apps/sms/views/conversation/js/information.js
```diff
@@ -357,6 +366,48 @@ var VIEWS = {
...
+          } else {
+            unitElement.classList.add('hide');
+          }
+        });
+        latenessBlock.classList.remove('hide');
+      } else {
+        latenessBlock.classList.add('hide');
+      }
+    },
+
```
julienw: sorry that I see this this late as well... but I see no unit test for the changes in this file ? Can you write some ?

julienw:
looks like xfail.list changed again, you'll need to rebase once again :(

apps/sms/views/conversation/js/conversation.js
sroucheray: Actually it supports having no `options` but it does not support undefined `data` argument.

apps/sms/locales/sms.en-US.properties
sroucheray: @julienw Ok no problem I can do that here :)
@zbraniecki Just a few questions :
 - Do you want me to move `late-arrival-notice-in-*` as well ? And, if yes, what would be there new names ?
 - I guess I must not move `late-notice-label.ariaLabel` and `message-lateness` but tell me if I am wrong.
 - As the `./shared/locales/intl/unitformat.en-US.properties` files does not exist yet I must create it. @zbraniecki are you OK with that ? and as a consequence, I must add this new file to `index.html`, @julienw are you OK too ?

julienw: thinking about it more, I'd really prefer to do this in a separate bug, because this one is fairly ready now.

The idea is only to move the "duration narrow" strings, because they are reusable.

In this patch, I think we should just change the "seconds", "months", etc to "value" and then we're good.

apps/sms/views/conversation/js/conversation.js
julienw: ah yeah, right ! how could I misread this...

OK, because I want this to land better sooner than later, let's keep it like this and file a separate bug to change the template lib.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5f2eb5ce44335e60c82202fe8042cd41b255464e

apps/sms/views/conversation/js/conversation.js
```diff
@@ -2009,7 +2077,8 @@ var ConversationView = {
     var bubble;
 
     do {
-      if (node.classList.contains('bubble')) {
+      // A Text node does not have a classList property
+      if (node.classList && node.classList.contains('bubble')) {
```
julienw: not a big deal, I just see this change now; did you have any issue about it ?  I think "getMessageBubble" is only called in one place, and with evt.target as parameter I think it's always an element ?

sroucheray: Yes, I saw an error being thrown when long touching the late notice inside the bubble and I reproduced as well on the text message. It those cases the evt.target for some reason was the TextElement itself, so without classList prop.

julienw: I really don't reproduce :/ I tried both on device and in Firefox... Can you check once again ?
Maybe you had the issue on unit tests or integration tests ?

And if you still have the issue, maybe do an early continue like `if (!node.classList) { continue; }` at the start of the loop (with appropriate indentation) ? What do you think ?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c799f8bcb511ed270b2a447e81d54a9e08c4becb

apps/sms/views/conversation/js/conversation.js
```diff
@@ -2009,6 +2077,9 @@ var ConversationView = {
     var bubble;
 
     do {
+      // A Text node does not have a classList property
```
julienw: please add "See Bug 1214238"

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=518a78b59e8f1a4daac18605cdc93c275ed9c075

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4261ee0cd65f2037575a8c7f09b95f728e2950d3

PR #31948, opened 2015-09-21
Bug 1171685 - Create hash of the deviceID. r=marshall.

tamarahills:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8de6ab78a567a0122c851ebef68971ae88fb53e6

apps/system/js/ftu_ping.js
```diff
@@ -358,9 +339,13 @@
           this.debug('No OS information, holding off');
           return false;
         }
-
-        this.ping();
-        return true;
```
marshall: are any tests failing because this `return true` was removed?

shared/js/telemetry.js
```diff
@@ -135,13 +134,45 @@
...
+        crypto.subtle.importKey('raw', bytes, {name:'PBKDF2'},
+                                false, ['deriveBits']).then(function (key) {
+          return crypto.subtle.deriveBits({
+            name: 'PBKDF2',
+            iterations: 1000,
+            salt: salt,
+            hash: 'SHA-1'
+          }, key, 160);
+        }).then(function (digest) {
+          var digestArray = new Uint8Array(digest).slice(0, 4);
```
marshall: Unsure if it matters, but this is only ~4.2 billion unique IDs. maybe we should be more ambitious :)

tamarahills: Hi Marshall,

This was my original thought too... There was a rationale behind this in comment 12 on the bug that it makes reversing harder:  https://bugzilla.mozilla.org/show_bug.cgi?id=1171685#c12

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5be7a038c79b279e269513a679b72c27a36cd70e

apps/system/js/ftu_ping.js
tamarahills: Hi Marshall,
So this boolean return value is just for unit testing and has no bearing on the functionality.  What finally gets clearInterval called is when ping() is called.  I had to remove it because it only returns true when it gets to the end of the function.  I could not return the true after the promise was resolved and ping was called as it would have chained me into the next clause in the promise.

marshall: k, that sounds good as long as the tests are passing

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6839ddb0a33909ca7edc66f844b9ba354a28d866

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=207c191a40a0def515fcf10f1e7c37b5602c31db

PR #31951, opened 2015-09-21
Bug 1187911 - Multiple notifications per calendar alarm r=jrburke

lambdabaa:
The strategy is simply to add a flag to each alarm that tells the calendar app that we've already issued a notification for it once we send the first notification. Then we check that flag before sending notifications to prevent the case where we issue multiple notifications for a single alarm.

### DIFF
--- a/apps/calendar/js/controllers/notifications.js
+++ b/apps/calendar/js/controllers/notifications.js
@@ -2,6 +2,7 @@ define(function(require, exports) {
 'use strict';
 
 var calc = require('common/calc');
+var co = require('ext/co');
 var core = require('core');
 var dateFormat = require('date_format');
 var debug = require('common/debug')('controllers/notifications');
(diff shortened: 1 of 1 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=59e5c335bd219f191e7a0f420c1ad531b9ac2a50

apps/calendar/js/controllers/notifications.js
```diff
@@ -17,20 +18,33 @@ exports.unobserve = function() {
...
-};
+  }
+
+  var storeFactory = core.storeFactory;
+  var alarmStore = storeFactory.get('Alarm');
+  alarm.fired = true;
+  // Save that we've fired this alarm.
+  yield alarmStore.persist(alarm);
+});
+
```
jrburke: This seems to be the core of the change, the rest is using co and the in the issueNotification, to skip it alarm.fired is set.

I am unfamiliar with the storage model in calendar, but given that, the concern I have is how alarmStore is cleaned up with the alarm stuff. My first impulse would be store if a notification was acted upon by placing that data on the calendar event itself, so that if it is ever purged or deleted, the cleanup happens naturally from there.

gaye: Good thought @jrburke. So the data model for calendar is that we have events which can have one or more occurrences (we call them busytimes). A user can set as many reminders for an occurrence as they want. If you set 2 reminders (say one at 5 minutes and one an hour before) for an occurrence, then we wouldn't want to prevent the 5 minute reminder from firing given that the 1 hour reminder had fired.

As far as cleanup goes, we have this "dependent stores" concept in our calendar data handling. Deleting a calendar event cleanly deletes all of the expanded event occurrences. Deleting an event occurrence, similarly, cleanly also deletes all of the occurrence's alarms.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e9b9c31da489ffb361a2690c0fc9f6bc923eec6c

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=301a8a1c126d0731ee7fb8f116544995ab5427f4

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=353905b1ed8fc573fc6827d2a3bf5020454d79a6

PR #31978, opened 2015-09-22
Bug 1207093 - [Messages][Dialer] Implement low storage condition popup

julienw:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=99562e080449e2f75b142c3a18be7d795ad2766c

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9af6857f5f511e46a5bc1a23c3677b1c47989933

shared/elements/gaia_lowstorage_dialog/script.js
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * The gaia-confirm component displays a dialog in which the user has a
+ * choice to confirm or cancel the action. It may be displayed along with a
+ * title, description, and image. Buttons may also be configured.
+ * @requires GaiaButtons
+ */
+window.GaiaLowstorageDialog = (function(win) {
+  // Extend from the HTMLElement prototype
+  var proto = Object.create(HTMLElement.prototype);
```
azasypkin: I'm new to WebComponents, but can't we use ```gaiaComponent.register``` here, like other web components do to hide some of the complexity [1]?

[1] https://github.com/gaia-components/gaia-checkbox/blob/cf04bb03f1b8390a6d973bdc2e37edcf246a5d61/gaia-checkbox.js#L15

shared/elements/gaia_lowstorage_dialog/examples/index.html
```diff
@@ -0,0 +1,41 @@
+<!DOCTYPE html>
```
azasypkin: __question:__ do we have convention when we use underscore and dashes in the component name, I see we have both in shared folder? I remember Wilson told me that underscore is used for old/kind-of-obsoleted components.

shared/elements/gaia_lowstorage_dialog/script.js
```diff
@@ -0,0 +1,78 @@
...
+   */
+  proto._localizeShadowDom = function() {
+    navigator.mozL10n.translateFragment(this.shadowRoot);
+  };
+
+  proto.createdCallback = function() {
+    var shadow = this.createShadowRoot();
+    this._template = template.content.cloneNode(true);
+
+    shadow.appendChild(this._template);
```
azasypkin: I see that if we insert template with other web components inside here ```createdCallback``` is not called for them (at least in Firefox Desktop) that leads to a broken component, trying to figure out why right now

azasypkin: And ```createdCallback``` seems to be called if injected via Light DOM

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c9c037c9b908fd2918f542e6f9dc8493320ec72a

shared/elements/gaia_lowstorage_dialog/examples/index.html
julienw: yep; I think dashed-components are for the new ones that come from external repositories, underscore_components are for the ones that are in gaia/shared only, and should be replaced in the future by external ones... So I choose underscore here because it uses an underscore component to keep the old style.

shared/elements/gaia_lowstorage_dialog/script.js
julienw: I didn't want to use the new style too much, I'm afraid of bad interactions between using the old gaia_confirm and new mechanisms.

shared/elements/gaia_lowstorage_dialog/script.js
julienw: yeah I don't know yet either ;) 

The new-style components are doing: https://github.com/gaia-components/gaia-component/blob/master/gaia-component.js#L131
I'll try this instead

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b07e9e5eb467a81d96ac996aa95e3819dd57ef26

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=87613399ba47ff571b5bbe42c34cefccc057ecf3

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b289150811744c2bbbb03432814786deb936ff10

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e98fae580dbf92fd61a2770d8af47541b41f8c52

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3eef2d984a46b4d18c40fb0230c46ccdc015103c

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a54dd2238f7619689e38421ccb9c226509454e1d

shared/elements/gaia_confirm/style.css
```diff
@@ -72,10 +72,13 @@ p {
   word-wrap: break-word;
   margin: 1rem 0 0;
   padding: 1rem 1.5rem;
-  border-top: 0.1rem solid #686868;
   line-height: 3rem;
 }
 
+p:first-of-type {
+  border-top: 0.1rem solid #686868;
```
julienw: I found only 1 gaia-confirm with several `<p>` (apps/keyboard/settings.html), so maybe I'm regressing it here. I suppose I fix it instead but maybe I should check with them ?

Some others used "<br>" (ex: apps/settings/elements/full_developer_mode.html, other in settings too) because they needed the same as me: having the line separator only once, but this is IMO bad.

shared/elements/gaia_confirm/style.css
```diff
@@ -85,6 +88,10 @@ p strong {
   font-weight: 300;
 }
 
+.stronger {
+  font-weight: 500;
+}
```
julienw: the font-weight "300" above is IMO wrong but I don't want to break some uses. That said, I didn't find any place where it's used so maybe I could just change the value above.

shared/elements/gaia_lowstorage_dialog/examples/locales/example.en-US.properties
```diff
@@ -0,0 +1,7 @@
+lowstorage-dialog-title=Low Device Storage
+lowstorage-dialog-generic-text=At least 10 MB of space is required. Delete apps and media files, then restart your device.
+lowstorage-dialog-learnmore=Learn More
+lowstorage-dialog-ok=OK
+
+lowstorage-explanation=Less than 5MB remain.
+lowstorage-explanation-details=Messages can not be sent, saved or received, not can Contacts be added at this time.
```
julienw: this will need to be copied in the apps that use the component; maybe I should put this in a README.

shared/elements/gaia_lowstorage_dialog/style.css
```diff
@@ -0,0 +1,33 @@
+gaia-lowstorage-dialog[hidden] {
+  /* The underlying dialog is controlled by this attribute, but we don't want to
+   * hide the outer element so that we see the transition */
+  display: initial;
+}
+
+.button-link-like {
```
julienw: wondering if I should put this elsewhere.

shared/elements/gaia_lowstorage_dialog/examples/locales/example.en-US.properties
julienw: BTW, the text is likely not correct; for example "less than 5MB remain" is not necessarily true, see the generic text above. I'll ask for a better text.

shared/elements/gaia_lowstorage_dialog/examples/index.html
KevinGrandon: Yeah, this is correct. We probably want to create a folder to split the implementations. (I still think they should all just live inside gaia and following our conventions of underscores, but I seem to be losing that battle!)

shared/elements/gaia_lowstorage_dialog/style.css
KevinGrandon: It seems fine to me here.

shared/elements/gaia_confirm/style.css
KevinGrandon: My preference would be to just change the above code. I don't think it should break anything. I'm also ok if you want to play it safe.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e74086354f7a072ad1664f0ea476837dfdda4c30

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e1e30f0a308c69b192c44cc97da4e3d8f5c48704

shared/elements/gaia_lowstorage_dialog/script.js
```diff
@@ -0,0 +1,98 @@
+'use strict';
+/* global ComponentUtils, MozActivity */
+
+/**
+ * The gaia-confirm component displays a dialog in which the user has a
```
azasypkin: __nit:__ wrong description

shared/elements/gaia_lowstorage_dialog/script.js
```diff
@@ -0,0 +1,98 @@
...
+
+  function learnmore() {
+    var activity = new MozActivity({
+      name: 'configure',
+      data: {
+        section: 'applicationStorage'
+      }
+    });
+
+    return Promise.resolve(activity);
```
azasypkin: __question:__ is it just to have ```.catch``` on return result, do you use result of this function anywhere?

shared/elements/gaia_lowstorage_dialog/script.js
```diff
@@ -0,0 +1,98 @@
...
+  };
+
+  proto.createdCallback = function() {
+    var shadow = this.createShadowRoot();
+
+    var node = document.importNode(template.content, true);
+    shadow.appendChild(node);
+    ComponentUtils.style.call(this, baseurl);
+
+    navigator.mozL10n.ready(this._localizeShadowDom.bind(this));
```
azasypkin: __optional nit:__ I have a feeling that there are 3 different callback approaches are used here, but essentially they're the same:
```
// Bind to function defined on prototype.
navigator.mozL10n.ready(this._localizeShadowDom.bind(this));

// Use closure to reference function from outer scope
learnmoreLink.addEventListener('click', learnmore);

// Use arrow function
confirm.addEventListener('confirm', () => {
  this.dispatchEvent(new CustomEvent('confirm'));
});
```

Is there any reason why you can't use only one approach?

shared/elements/gaia_lowstorage_dialog/script.js
```diff
@@ -0,0 +1,98 @@
...
+   * within the shadow dom. See also: bug 1026236.
+   */
+  proto._localizeShadowDom = function() {
+    navigator.mozL10n.translateFragment(this.shadowRoot);
+  };
+
+  proto.createdCallback = function() {
+    var shadow = this.createShadowRoot();
+
+    var node = document.importNode(template.content, true);
```
azasypkin: __nit:__ It would be great to have comment here, to understand what is going on in this block :)

shared/elements/gaia_lowstorage_dialog/script.js
```diff
@@ -0,0 +1,98 @@
...
+
+  proto._mirrorHiddenAttribute = function() {
+    if (this.hasAttribute('hidden')) {
+      this.confirm.setAttribute('hidden', '');
+    } else {
+      this.confirm.removeAttribute('hidden');
+    }
+  };
+
+  proto.attributeChangedCallback = function(name, _, newVal) {
```
azasypkin: __optional nit:__ no need to define unused parameters ```_, newVal```. Although it's arguable, so up to you :)

shared/elements/gaia_lowstorage_dialog/script.js
```diff
@@ -0,0 +1,98 @@
...
+    this._mirrorHiddenAttribute();
+  };
+
+  var template = document.createElement('template');
+  template.innerHTML = `<gaia-confirm>
+    <h1 data-l10n-id="lowstorage-dialog-title"></h1>
+    <p>
+      <strong data-l10n-id='lowstorage-explanation'></strong>
+      <br/>
+      <strong><content/></strong>
```
azasypkin: __question:__ just curious, we want to support _custom_ content via ```<content></content>``` instead of say ```custom-text-l10n-id``` attribute on web component host element just to be more extensible in the future?

shared/elements/gaia_lowstorage_dialog/script.js
```diff
@@ -0,0 +1,98 @@
...
+
+  proto.attributeChangedCallback = function(name, _, newVal) {
+    if (name !== 'hidden') {
+      return;
+    }
+
+    this._mirrorHiddenAttribute();
+  };
+
+  var template = document.createElement('template');
```
azasypkin: __question:__ what is your plan regarding lazy loading this component?

Should it be lazy loaded only in low-storage condition? Or lazy loaded in any case, but later, once app is fully loaded (in this case I'm wondering if we should make ```template``` lazily initialized as well to not occupy the memory, maybe not a big deal, but still).

shared/elements/gaia_lowstorage_dialog/script.js
gabrielesvelto: I'm kind of wary of using the ``<node>.bind(this) `` style because it inherently leaks the event listener. Generally it's not a problem as it will go away with the node but I've encountered multiple cases where we were binding more than one listener because of it.

shared/elements/gaia_lowstorage_dialog/script.js
julienw: ah yeah, I was using newVal in a previous version of the code.

shared/elements/gaia_lowstorage_dialog/script.js
julienw: In our case, I would LazyLoad the JS file only when needed, then insert the element.

shared/elements/gaia_lowstorage_dialog/script.js
julienw: yeah, I also asked myself this question, and thought it was more flexible that way without much more complexity. In some apps we may want to have different texts for different cases too.

shared/elements/gaia_lowstorage_dialog/script.js
julienw: oops

shared/elements/gaia_lowstorage_dialog/script.js
julienw: good question, I don't really know why I did this. Maybe we'd like that the caller knows when the activity is over, so that for example he could close the dialog ? (the spec does not tell what we should do here BTW, I'll ask a clarification)

shared/elements/gaia_lowstorage_dialog/script.js
julienw: good suggestion!

shared/elements/gaia_lowstorage_dialog/script.js
julienw: Answering to @azasypkin:

In my mind, they have quite different uses:
* `learnmore` is independant, it could as well be in another file, we don't really care.
* `localizeShadowDom` depends on the actual element being localized, so we need the "this". (but see below)
* for `confirm`, because the handler is so small, I think it's more readable to have the handler inline.

Answering to @gabrielesvelto:

createdCallback is always called only once but now I realize from your comment that we don't free it when the element is removed. So I think I should attach on "attachedCallback" and detach on detachedCallback (we don't know when the object is garbage collected). But... we can't detach l10n's ready event. So I think I'll either have to use a global event handler and do my own attach/detach mechanism, or implement this in l10n.js.

azasypkin: > for confirm, because the handler is so small, I think it's more readable to have the handler inline.

Yeah, but ```_localizeShadowDom``` is one-liner as well, why not to use arrow there?

julienw: Because I copied it from https://github.com/mozilla-b2g/gaia/blob/master/shared/elements/gaia_sim_picker/script.js#L60 where it's used in several places ;)

I can change this.

azasypkin: Ah, okay :) it's __bikeshedding__ aka __optional nit__ anyway, so up to you :)

shared/elements/gaia_lowstorage_dialog/examples/locales/example.en-US.properties
```diff
@@ -0,0 +1,7 @@
+lowstorage-dialog-title=Low Device Storage
```
azasypkin: __question:__ should not we have ```shared/elements/gaia_lowstorage_dialog/locales/gaia_lowstorage_dialog.en-US.properties``` for the common strings (title, learnmore, ok, generic-test)?

shared/elements/gaia_lowstorage_dialog/style.css
azasypkin: __issue:__ ":active" state looks wrong for this link-like button (tap and hold a bit)

shared/elements/gaia_lowstorage_dialog/script.js
```diff
@@ -0,0 +1,98 @@
...
+  template.innerHTML = `<gaia-confirm>
+    <h1 data-l10n-id="lowstorage-dialog-title"></h1>
+    <p>
+      <strong data-l10n-id='lowstorage-explanation'></strong>
+      <br/>
+      <strong><content/></strong>
+    </p>
+    <p data-l10n-id="lowstorage-dialog-generic-text"></p>
+    <p>
+      <button type="button"
```
azasypkin: __question:__ wondering if we should add ```role="link"``` for that button, what do you think?

azasypkin: Hmm, seems I'm stuck in Settings when I navigate there using Screen Reader (and can neither select anything to read there nor navigate back). Do you know by any chance if it's filed already? (definitely not related to this PR)

shared/elements/gaia_confirm/style.css
azasypkin: > I found only 1 gaia-confirm with several <p> (apps/keyboard/settings.html), so maybe I'm regressing it here. I suppose I fix it instead but maybe I should check with them ?

Not a big deal, but I'd rather make them aware of this change, just in case :)

shared/elements/gaia_lowstorage_dialog/script.js
azasypkin: Okay, works for me

shared/elements/gaia_lowstorage_dialog/script.js
azasypkin: Okays, sounds good

shared/elements/gaia_lowstorage_dialog/style.css
julienw: grrrr at bug 1208113 
thanks

julienw: ah and the RTL work also regressed my override of the margin...

shared/elements/gaia_lowstorage_dialog/script.js
stasm: > createdCallback is always called only once but now I realize from your comment that we don't free it when the element is removed. So I think I should attach on "attachedCallback" and detach on detachedCallback (we don't know when the object is garbage collected). But... we can't detach l10n's ready event. So I think I'll either have to use a global event handler and do my own attach/detach mechanism, or implement this in l10n.js.

As discussed with @julienw on IRC, perhaps a safer solution memory-wise would be to use `mozL10n.once()` and `window.addEventListener('localized', …)` inside of it.

Or, we could introduce the l20n-style proxy in l10n.js for `shared/` code, as suggested by @julienw in https://bugzilla.mozilla.org/show_bug.cgi?id=1203108#c7.  Both @zbraniecki  and I like this approach;  we might  be able to land it soon.

shared/elements/gaia_lowstorage_dialog/examples/locales/example.en-US.properties
julienw: I asked myself this too; maybe I should split it already and let the app developers decide.

shared/elements/gaia_lowstorage_dialog/script.js
julienw: actually I really used the "button" element because it triggers an action so I definitely didn't want a link semantic. But now that I think this further, we're actually navigating, so maybe a real link is better ?

azasypkin: Well, I have a mixed feeling as well, in this case we have kind of link "Learn More" that directly leads to another page with "back" button. E.g. in  our "Add attachment" case, it's really a button that triggers menu with the links that lead to a new page (though I'm not sure if links are used there).

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ea39177554452e5d041699678a07d70b41c964a9

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1066ab34cdab12b591df4d444906c3e2dea92346

PR #31988, opened 2015-09-22
Bug 1193576 - Add a setting to browse privately by default

jedireza:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=cc6e1b1685fb70ef28eced3fc0b03d46b0c36d2e

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4f7e3595a7ab047134ec8000f88ab0dc3a81616b

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a981fa798f3986fbd10b1cdf8ca277d4bf90bc38

apps/system/js/browser.js
```diff
@@ -1,9 +1,17 @@
...
 
 (function(exports) {
 
   'use strict';
 
+  function handleSettingChange (value) {
+    Browser.privateByDefault = value;
+  }
+
+  SettingsListener.observe('browser.private.default', false,
```
KevinGrandon: The listeners in this file are currently in the start method of the Browser "class". To keep things in line with the rest of the system app, I might recommend moving this listener to into the start() method as well - it should make it easier to write a unit test against this if we wanted to in the future. (If this goes into the start method, then privateByDefault would probably go onto Browser.prototype).

apps/system/js/app_chrome.js
```diff
@@ -749,9 +749,10 @@
     };
 
   AppChrome.prototype.setThemeColor = function ac_setThemColor(color) {
-    // Do not set theme color for private windows
+    // Overwrite color for private windows
     if (this.app.isPrivateBrowser()) {
-      return;
+      color = '#392E54';
```
KevinGrandon: Is this to prevent the theme-color from taking effect? I'm ok if we want to do that, please update the comment to say so if that's the reason.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ea5e82f61239fcec55af8410ca9d0f532ebecf8c

apps/system/js/app_chrome.js
jedireza: Updated comment

apps/system/js/browser.js
jedireza: I started doing it this way, and realized that I didn't have an instance of a `Browser` in all cases. Specifically in `apps/system/js/wrapper_factory.js`, accessing the static property of the `Browser` class works well.

As it is, `privateByDefault` is a static property and doesn't change from one instance to another.

apps/search/js/newtab.js
```diff
@@ -1,8 +1,17 @@
...
 
+  SettingsListener.observe('browser.private.default', false, function (value) {
+    if (value) {
+      exports.newtab.pretendToBePrivate();
+    }
+    else {
+      exports.newtab.stopPretendingToBePrivate();
+    }
+  });
+
```
jedireza: I know you mentioned not using an observer here, but this actually works well. If I change the setting while the app is open it will toggle the expected functionality.

apps/search/js/newtab.js
```diff
@@ -1,8 +1,17 @@
 'use strict';
-/* global asyncStorage */
+/* global asyncStorage, SettingsListener */
 
 (function(exports) {
 
+  SettingsListener.observe('browser.private.default', false, function (value) {
+    if (value) {
+      exports.newtab.pretendToBePrivate();
+    }
```
KevinGrandon: nit: prefer the `} else {` style. It's not enforced (but should be), and is that way in most gaia apps.

apps/search/js/newtab.js
KevinGrandon: Sure, if you've added a stopPretendingToBePrivate method that seems fine.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f63800c4fc38acee76d85162f843fe5cec8f6e44

apps/search/js/newtab.js
jedireza: Ah yes. Old habbit ;) Fixed.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1c6b466cc88aab35618976ef79315918a5f9adb6

PR #31993, opened 2015-09-23
Bug 1180470 - [Messages] Propose a setting to disable sending the read report

kumarrishav:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0dc2653c5bb901261bf044341c98366e4e8a5071

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=91f7e334b2aeab29c7ff611f04cfa1b0b79a4b29

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9adf4b06660de3554b870c7f18c269360c66b236

apps/sms/views/shared/js/settings.js
```diff
@@ -135,6 +135,26 @@ var Settings = {
...
   },
 
+  // return true if send-read-report switch is ON else false
+  toSendReadReport: function toSendReadReport() {
+    if (!navigator.mozSettings) {
+      return Promise.reject(new Error('The mozSettings API is not available.'));
+    }
+
+    var defer = Utils.Promise.defer();
+    var key = 'message.mms.sendReadReport.enabled';
```
steveck-chung: You can move this key to const

apps/sms/views/shared/js/settings.js
```diff
@@ -135,6 +135,26 @@ var Settings = {
...
+    var request = navigator.mozSettings.createLock().get(key);
```
steveck-chung: I think you can return the `navigator.mozSettings.createLock().get(key)` directly, but we'll refactor this into settings service anyway...

apps/sms/services/js/message_manager.js
```diff
@@ -421,33 +421,41 @@ var MessageManager = {
...
+        if (!list.length || !isRead) {
+          return deferred.resolve();
+        } else {
+          markMessage(sendReadReport);
+        }
+      };
+
+      req.onerror = function onerror() {
+        console.error('Error while marking message %d as read: %s', id, this.error.name);
+      };
```
steveck-chung: You will need to return the `deferred.promise`

apps/sms/services/js/message_manager.js
```diff
@@ -421,33 +421,41 @@ var MessageManager = {
...
+        // if isRead == false then mark only one message in each thread
+        if (!list.length || !isRead) {
+          return deferred.resolve();
+        } else {
+          markMessage(sendReadReport);
+        }
+      };
+
+      req.onerror = function onerror() {
+        console.error('Error while marking message %d as read: %s', id, this.error.name);
```
steveck-chung: For sync the behavior, you can also reject the error here.

gasolin:
remember to add the default value `"message.mms.sendReadReport.enabled": false` in build/config/comm-settings.json

apps/settings/elements/messaging.html
```diff
@@ -23,8 +23,14 @@ <h2 data-l10n-id="message-confirmation"></h2>
...
         <li id="menuItem-readReport" aria-disabled="true">
           <gaia-switch name="ril.mms.requestReadReport.enabled"/>
-            <label data-l10n-id="message-read-reports"></label>
-            <details data-l10n-id="message-read-reports-details"></details>
+            <label data-l10n-id="message-request-read-reports"></label>
+            <details data-l10n-id="message-request-read-reports-details"></details>
+          </gaia-switch>
+        </li>
+        <li id="menuItem-sendReadReport" aria-disabled="true">
+          <gaia-switch name="message.mms.sendReadReport.enabled"/>
```
julienw: nit: I'd use "messages" instead of "message" as prefix; "messages" is the official name for our application

apps/settings/elements/messaging.html
```diff
@@ -23,8 +23,14 @@ <h2 data-l10n-id="message-confirmation"></h2>
       <ul class="mmsSettings-list">
         <li id="menuItem-readReport" aria-disabled="true">
```
julienw: nit: maybe change the id to include "request" too ?

apps/sms/services/js/message_manager.js
```diff
@@ -421,33 +421,41 @@ var MessageManager = {
 
   markMessagesRead: function mm_markMessagesRead(list, isRead = true) {
     if (!this._mozMobileMessage || !list.length) {
-      return;
+      return Promise.reject(new Error('The mozMobileMessage API is not available or list is empty.'));
```
julienw: I think we should return a resolved promise:
* either the list is empty, and it's OK
* or we don't have mozMobileMessage, and it usually means we're in Firefox. So we need to do as if everything works.

apps/sms/views/shared/js/settings.js
```diff
@@ -135,6 +135,26 @@ var Settings = {
...
+    request.onsuccess = function onsuccess(e) {
+      defer.resolve(e.target.result);
+    };
+
+    request.onerror = function onerror(e) {
+      defer.reject(e.target.error);
+    };
+
+    return defer.promise;
+  },
```
julienw: in https://bugzilla.mozilla.org/show_bug.cgi?id=1180470#c19 I proposed to synchronize the setting value using similar methods that we have already (for example look `initSmsMaxConcatenatedMsg` and `keyHandlerSet` in `init`).

Then I thought that we can have a property "Settings.sendReadReport" that is a promise that resolves to the value. This way it's not resolved at startup yet, but the code in MessageManager can still request the promise and use the value when it's resolved.

So in MessageManager you can have:
`Settings.sendReadReport.then((value) => ...)`

apps/sms/views/conversation/js/conversation.js
```diff
@@ -927,13 +927,13 @@ var ConversationView = {
...
     }
 
-    MessageManager.markMessagesRead([message.id]);
-
-    this.onMessage(message);
-    this.scrollViewToBottom();
-    if (this.isScrolledManually) {
-      this.showNewMessageNotice(message);
-    }
+    MessageManager.markMessagesRead([message.id]).then(() => {
```
julienw: I don't think we want to do this. We don't need to wait that the message is mark as read before doing the rest of the work.

apps/sms/services/js/message_manager.js
```diff
@@ -421,33 +421,41 @@ var MessageManager = {
...
-        'Error while marking message %d as read: %s', id, this.error.name
-      );
+    Settings.toSendReadReport().then((result) => {
+      var sendReadReport = result['message.mms.sendReadReport.enabled'];
+      markMessage(sendReadReport);
+    });
+
+    function markMessage(sendReadReport) {
+      var id = list.pop();
+      var req = self._mozMobileMessage.markMessageRead(id, isRead, sendReadReport);
```
julienw: remember the request has a "then" method now, you can use it directly.

apps/sms/services/js/message_manager.js
kumarrishav: i think there won't be any case where list is empty and markMessageRead is called.

kumarrishav: Sorry i was wrong. ignore above case :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=863dc4ec5107ca06867ec6e104ab96c440633b05

PR #32007, opened 2015-09-23
Bug 1203993 - Add a performance marker for app launching. r=eli

Cwiiis:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=041fccecc4a35c7066dfb9787ea54a34c802a575

apps/homescreen/js/app.js
```diff
@@ -687,6 +687,11 @@ const SETTINGS_VERSION = 0;
 
           default:
             icon.launch();
+
+            // Launching an app
+            if (icon.app) {
```
eliperelman: I notice here also that the performance marker is generated after the icon launch occurs, which is different than the way the current homescreen works. I think in this instance we want the marker before launch so the numbers are consistent and have a common relationship.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2c53ed3b796ce299dc89363d9f03c34dd9fbd36e

apps/homescreen/js/app.js
Cwiiis: Done :)

PR #32022, opened 2015-09-24
Bug 1203886 - The back button in sms conversation view does not always work, r=julienw

steveck-chung:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1a6625dd19e56d009da5f26f5eef5e9be637886e

apps/sms/views/shared/js/navigation.js
```diff
@@ -751,6 +751,13 @@ var Navigation = {
       return Promise.resolve();
     }
 
+    if (navigationTransition) {
+      debug('startNavigation: previous navigation exists, deferring.');
```
steveck-chung: The reason why I did transition checking here before startNavogation is because startNavogation didn't check the currentPanel, and even we could check it inside startNavogation, it's harder to exist the navigation promise chain directly if we're in startNavogation already. So I add another checking in the toPanel.

BTW I forgot to rename the debug log... Will fix that later.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=33818ef40c431f07da32b968aa7788ef5340180e

apps/sms/views/shared/js/navigation.js
```diff
@@ -751,6 +751,13 @@ var Navigation = {
       return Promise.resolve();
     }
 
+    if (navigationTransition) {
```
steveck-chung: The reason why I did transition checking here before startNavogation is because startNavogation didn't check the currentPanel, and even we could check it inside startNavogation, it's harder to exist the navigation promise chain directly if we're in startNavogation already. So I add another checking in the toPanel.

julienw: I agree it's more difficult to move the isCurrentPanel to startNavigation, so why not.

Also we already have https://github.com/steveck-chung/gaia/blob/33818ef40c431f07da32b968aa7788ef5340180e/apps/sms/views/shared/js/navigation.js#L322-L327 but I think we need to keep it too, because we can access startNavigation from other places (like "back" or "hashchange" event).

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5234d3ded76a3f5e334da097e777cc43759c8567

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=994a741928229d1989179434512c53a254890855

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a604c5d55075c0c4ad23e36320ec24c6cca8df58

PR #32034, opened 2015-09-24
Bug 1179586 - Intermittent apps/sms/test/marionette/conversation_testjs | Conversation Panel Tests General use cases Long SMS thread User can see all messages when scrolls up. r=julien

azasypkin:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ee85f6b0356730727279e43593a886c5a2fffeb8

apps/sms/test/marionette/conversation_test.js
```diff
@@ -53,28 +53,22 @@ marionette('Conversation Panel Tests', function() {
...
-          } else {
-            // Otherwise, node may not be visible and we should scroll up to
-            // see it
-            messagesApp.Conversation.scrollUp();
-            loadedPage++;
-
-            client.helper.waitForElement(messageNode);
+          // If node is not visible, let's just scroll up and wait until it
+          // becomes visible.
+          if (!messageNode.displayed()) {
```
azasypkin: Now we don't have fragile logic that were checking page-by-page loading, but it's not what we wanted to test anyway - what we initially wanted to test is that user can scroll and see all messages.

It will fail by timeout if scroll doesn't help, works for both my fast laptop and slow vargant VM.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=cadde8ff3916fcfcc4f845f8b2506a614ee4aac2

apps/sms/test/marionette/conversation_test.js
julienw: I think it's better and I'm not afraid of the timeout because we display each message once at a time, so we need only to go back so far. I assume the messages array is ordered in the order most recent -> oldest.

PR #32035, opened 2015-09-24
Bug 1208094 - Fix failure in test_keyboard.py

mwargers:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a1e5a404f962f128c3d258c78a5b89725452f2a8

tests/python/gaia-ui-tests/gaiatest/apps/contacts/regions/contact_form.py
```diff
@@ -105,18 +105,12 @@ def comment(self):
...
-        element = self.marionette.find_element(*self._add_new_note_locator)
-        element.tap()
-        self.marionette.execute_script(
-            'arguments[0].scrollIntoView(true);', [element])
-        comment = self.marionette.find_element(*self._comment_locator)
-        Wait(self.marionette, timeout=120).until(expected.element_displayed(comment))
-        self.marionette.execute_script(
-            'arguments[0].scrollIntoView(true);', [element])
-        element.tap()
+        self.marionette.find_element(*self._comment_locator).tap()
```
ShakoHo: Suggest we could use "Wait(self.marionette).until(expected.element_present(*self._comment_locator)).tap()"
instead of just find the element and tap ...that will increase the stability of running test case on mtbf.

mwargers: That should not be necessary, because beforehand, tap_new_contact() is called:
http://mxr.mozilla.org/gaia/source/tests/python/gaia-ui-tests/gaiatest/apps/contacts/app.py#64
..which calls wait_for_new_contact_form_to_load():
http://mxr.mozilla.org/gaia/source/tests/python/gaia-ui-tests/gaiatest/apps/contacts/regions/contact_form.py#200
This should take care that the page has been loaded completely and is ready to use. The check in there doesn't look too great, I have to say, but it's probably enough.

Currently, there is actually no difference between using self.marionette.find_element and Wait(self.marionette).until(expected.element_present, only conceptually.
You should use find_element, when you are sure the element should be there (which it should in this case) and Wait(self.marionette).until(expected.element_present when you don't expect the element to be there, immediately.
But currently, find_element and Wait(self.marionette).until(expected.element_present both wait 10s before it times out and reports the element is not there.

See also discussion I had with Dave Hunt about this: https://mail.mozilla.org/private/gaia-ui-automation/2015-September/001857.html
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1201870

PR #32044, opened 2015-09-25
Bug 1180470 - [Messages] Propose a setting to disable sending the read report

kumarrishav:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0a4f63bc90c494c10468895d209eddaa1da6d2c7

apps/settings/locales/settings.en-US.properties
```diff
@@ -297,8 +297,10 @@ messagingSettings-header={{messagingSettings}}
 message-confirmation = Message Confirmation
 message-delivery-reports = Delivery Reports
 message-delivery-reports-details = Request notification of delivery for each message sent
-message-read-reports = Read Reports
-message-read-reports-details = Request read report for each message sent
+message-request-read-reports = Request Read Reports
+message-request-read-reports-details = Request read report for each message sent
+message-send-read-reports = Send Read Reports
+message-send-read-reports-details = Send read report when others request
```
julienw: note that I asked for a better text on the bug

apps/sms/services/js/message_manager.js
```diff
@@ -420,8 +420,12 @@ var MessageManager = {
   },
 
   markMessagesRead: function mm_markMessagesRead(list, isRead = true) {
+    var self = this;
+    var deferred = Utils.Promise.defer();
+
     if (!this._mozMobileMessage || !list.length) {
-      return;
+      deferred.resolve();
+      return deferred.promise;
```
julienw: you can just return `Promise.resolve()` here

apps/sms/services/js/message_manager.js
```diff
@@ -429,26 +433,32 @@ var MessageManager = {
...
+        if (!list.length || !isRead) {
+          deferred.resolve();
+          return deferred.promise;
+        } else {
+          markMessage(sendReadReport);
+        }
+      }, () => {
+        console.log('ERROR in mozMobileMessage.markMessageRead method');
+      });
+    }
```
julienw: I think you can do simpler, without using a deferred. Deferred are useful when we don't have promises in the APIs we use, but here we have promises (or then-able which is essentially the same).

How about something like this:
```javascript
if (!this._mozMobileMessage) {
  return Promise.resolve();
}

var markMessage = (shouldSendReadReport) => {
  if (!list.length) {
    return Promise.resolve();
  }

  var id = list.pop();
  var req = self._mozMobileMessage.markMessageRead(id, isRead, shouldSendReadReport);
  return req.then(
    () => isRead ? markMessage(shouldSendReadReport) : Promise.resolve()
  );
};

return Settings.sendReadReport.then(markMessage);
```

If you don't understand porperly, remember the promises are "transitive" (not sure it's the right word, hope you'll understand it).

apps/sms/views/shared/js/settings.js
```diff
@@ -67,9 +70,37 @@ var Settings = {
       }
     }
 
+    //create lock for SEND_READ_REPORT_KEY
+    keyHandlerSet[this.SEND_READ_REPORT_KEY] = this.toSendReadReport.bind(this);
+
```
julienw: nit: add it directly in keyHandlerSet above, and call the  function "initSendReadReport" like the other ones

apps/sms/views/shared/js/settings.js
```diff
@@ -67,9 +70,37 @@ var Settings = {
...
     for (var key in keyHandlerSet) {
       setHandlerMap(key);
     }
+
+    //add Observer for SEND_READ_REPORT_KEY
+    settings.addObserver(this.SEND_READ_REPORT_KEY, function (event) {
+      Settings.sendReadReportStatus = event.settingValue;
+      console.log(Settings.sendReadReportStatus);
+    });
```
julienw: you can add hte observer in "initSendReadReport" like the other "init*" functions

apps/sms/views/shared/js/settings.js
```diff
@@ -67,9 +70,37 @@ var Settings = {
...
+    return deferred.promise;
+  },
+
+  //Set sendReadReportStatus for SEND_READ_REPORT_KEY
+  toSendReadReport: function toSendReadReport(status) {
+    if (!isNaN(status)) {
+      Settings.sendReadReportStatus = status;
+      console.log(Settings.sendReadReportStatus);
+    }
   },
```
julienw: so it's not exactly what I had in mind, but it's much closer.

What I propose is that `sendReadReportStatus` is already a Promise (and we can call it simply `sendReadReport`).

So in init:
```javascript
... keyHandlerSet + init things

this._sendReadReportDefer = Utils.Promise.defer();
this.sendReadReport = this._sendReadReportDefer.promise;
```

in initSendReadReport:
```javascript
function initSendReadReport(value) {
  this._sendReadReportDefer.resolve(value);
  ...addObserver(settingName, (e) => this.sendReadReport = Promise.resolve(e.settingValue));
}
```

Tell me if this makes sense ?

kumarrishav: why we are deleting this._sendReadReportDefer,  here in initSendReadReport. Shouldn't it be deleted in init, once initSendReadReport is done and then after assigning this.sendReadReport = this._sendReadReportDefer.promise; then delete it.  
So, it should be like this: 
1.in init-> this._sendReadReportDefer = Utils.Promise.defer(); then,
2. function initSendReadReport is called [setHandlerMap(key) function] -> resolve promise here.
3. in init -> assigning this.sendReadReport = this._sendReadReportDefer.promise;
4 in init -> then delete, delete this._sendReadReportDefer;
Just going through execution sequence.
@julienw  please correct me if i am wrong here.

kumarrishav: @julienw i am just curious to know why we have  this._sendReadReportDefer = Utils.Promise.defer(); in init and resolving it in initSendReadReport.
Instead we can have this._sendReadReportDefer=Promise.resolve(value) directly in initSendReadReport.
I guess reason is , setHandlerMap(key) function in init has aysnc api (DOMrequest).
please correct it if i am wrong.

Also, few days back , steve warned me that, we shouldn't go for sync looping like foreach,while for looping having async method inside. That's why we go for chaining of the api call.
Similarly  here too, 
  <pre>
  for (var key in keyHandlerSet) {
      setHandlerMap(key); //this method inisde it has async DOM request.
    }
</pre>

kumarrishav: Regarding the chaining of api calls, in which situation, when we should go for it? 
When we are accessing IndexDB or some complex situation , we should go for chaining and at other place it's fine without it?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6ae748da71d262ecfbf6f52e3fd1e26bec5e295a

julienw:
Hey ! I'll try to answer your questions here !

> @julienw i am just curious to know why we have this._sendReadReportDefer = Utils.Promise.defer(); in init and resolving it in initSendReadReport.
Instead we can have this._sendReadReportDefer=Promise.resolve(value) directly in initSendReadReport.
I guess reason is , setHandlerMap(key) function in init has aysnc api (DOMrequest).
please correct it if i am wrong.

The reason is that I'd like to have `Settings.sendReportReport` holding a Promise at startup, before we get the value, because we might need to use it in markMessagesRead at startup (for example when we want to send a message to a contact that already has a conversation, from the Contacts app).

That's where Promises are useful: as the name says it, it's a "promise" that we'll have a value later, and we can wait for this value. We don't need promises if we have the value already !


> Also, few days back , steve warned me that, we shouldn't go for sync looping like foreach,while for looping having async method inside. That's why we go for chaining of the api call.
Similarly here too, 

> Regarding the chaining of api calls, in which situation, when we should go for it?
When we are accessing IndexDB or some complex situation , we should go for chaining and at other place it's fine without it?

IMO it depends on when you do the operation, how many items you have in your loop and whether it's read operations or write operations.

In this case, we have few items in this loop (4 or 5 if I remember correctly), it's only read operations (so it's less costly and they can usually be done in parallel in DB), and it's done in startup (we initialize a lot of stuff -- although here it's arguable :) ).

If you compare with the markMessagesRead call, we can have hundreds (even thousands !) calls, and they're write operations. Write operations are usually (maybe always) run in sequences anyway.
In markMessagesRead we could likely run 2 or 3 operations in parallel, but it's more work and not really worth it in this case IMO.


Hope I'm clear :)

kumarrishav:
Thanks @julienw you explained very clearly :) thnx for that.
But one question, [Hope you got me]

Settings.sendReportReport will hold a Promise only when settings.js (more specific) script, run before (in start up) markMessageRead is called. Lets assume (though it won't happen, as init runs very first compare to any other method), Settings.js (init) didn't run yet at startup, then it will through error in markMessageRead.  right? may be it sound hypothetical.
Also, init and initSendReadReport. runs together only (as initSendReadReport. is called inside init). You mean to say that, there might be race condition for Setting.sendReadReport, that's why we are having promise in init itself.  But, don't you think, execution of init and initSendReadReport. is exactly together/same. or you are assuming there may be delay, so lets have promise for Settings.sendReadReport so that it can wait for the value. 

BTW: You are too good and awesome on explaining stuff :+1:

kumarrishav:
Also regarding this comment:
The reason is that I'd like to have Settings.sendReportReport holding a Promise at startup, before we get the value, because we might need to use it in markMessagesRead at startup (for example when we want to send a message to a contact that already has a conversation, from the Contacts app).

You are holding a Promise at startup, what defines it: i mean, where it holding..it's in init() and initSendReadReport is also called in init() only, that don't you think, we will get the value at startup only, when you are holding the promise.?
@julienw

kumarrishav:
i just checked, value is getting set at start up time only.
So, having promise, you mean, there is possibility that it (markMessageRead) will/can call for that value, even before the init() script ran. so, if we will have promise, then it can wait for init to run and get value set.
please correct me if i am wrong :)

julienw:
It's the role of startup.js to take care that inits are run before we navigate to the first panel. Where you're right here is that we call Navigation.setReady() (this is where we load the panel if it's not the default panel) before Settings.init. Fortunately I think we're safe because we use promises everywhere in Navigation.

About init/initSendReadReport; you see that initSendReadReport is called only when we get the mozSettings.get "success" event. So it's not happening together.

kumarrishav:
"About init/initSendReadReport; you see that initSendReadReport is called only when we get the mozSettings.get "success" event. So it's not happening together."
yeah , you are right. that domrequest made difference, though it's not effecting here as both are in init and startup.js runs init() before navigating.

T

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6ae748da71d262ecfbf6f52e3fd1e26bec5e295a

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ffae34842cd7da5fa4890be3b7766d22c4289391

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3cd245c7832e5a524011c18c5549be03a056413e

apps/sms/views/shared/js/settings.js
```diff
@@ -72,6 +79,15 @@ var Settings = {
     }
   },
 
+  // return resolve promise with sendReadReport status
+  initSendReadReport: function initSendReadReport(value) {
+    this._sendReadReportDefer.resolve(value);
+    delete this._sendReadReportDefer;
+
+    navigator.mozSettings.addObserver('messages.mms.sendReadReport.enabled',
+      (e) => this.sendReadReport = Promise.resolve(e.settingValue));
```
steveck-chung: If we apply settings addObserver init pattern like MaxConcatenatedMsg and others below, why do we need to return as promise here instead of return the value directly?

apps/sms/services/js/message_manager.js
```diff
@@ -429,26 +429,23 @@ var MessageManager = {
...
-        'Error while marking message %d as read: %s', id, this.error.name
+      var id = list.pop();
+      var req = this._mozMobileMessage.markMessageRead(id, isRead, shouldSendReadReport);
+
+      return req.then(
+        () => isRead ? markMessage(shouldSendReadReport) : Promise.resolve()
       );
     };
+
+    return Settings.sendReadReport.then(markMessage);
```
steveck-chung: If we treat the sendReadReport as a simple boolean, we can just return markMessage() and no need to provide any parameter for markMessage. You can return
```
return this._mozMobileMessage.markMessageRead(
  id,
  isRead,
  Settings.shouldSendReadReport
).then(
  ...
```

apps/sms/views/shared/js/settings.js
kumarrishav: @steveck-chung  I asked same from julienw, Actually, it possible that on startup the value is not set. But that's also true that, init() runs first before any method/navigation do. (and all are navigation are promise based)
Or i guess @julienw  can explain better.

steveck-chung: Ok I see it's to prevent the request of sendReadReport start earlier than settings init ready.

steveck-chung: BTW I just wondering why don't we simply call the Settings get API directly but  use init here. I know it would be replaced once we introduce the settings client, but shouldn't it be easier to call the settings get API in sendReadReport?

kumarrishav: I guess we have to make a call between calling setting get Api or return promise from sendReadReport. [Performance wise hopefully] .

apps/sms/views/shared/js/settings.js
```diff
@@ -35,14 +36,16 @@ var Settings = {
...
+      'messages.mms.sendReadReport.enabled':
+        this.initSendReadReport.bind(this)
     };
     var settings = navigator.mozSettings;
     var conns = navigator.mozMobileConnections;
 
     function setHandlerMap(key) {
       var req = settings.createLock().get(key);
-      req.onsuccess = function settings_getSizeSuccess() {
+      req.onsuccess = function settings_createLockSuccess() {
```
julienw: if you change this, you can just as well use .then() instead : `settings.createLock().get(key).then((result) => ...)`

apps/sms/services/test/unit/message_manager_test.js
```diff
@@ -443,23 +446,29 @@ suite('message_manager.js >', function() {
...
+          );
+        });
+
+        sinon.assert.callCount(MockNavigatormozMobileMessage.markMessageRead, 4);
+      }).then(done,done);
+
     });
   });
-
+/*
```
julienw: mmm why do you comment this ?

apps/sms/services/test/unit/message_manager_test.js
```diff
@@ -443,23 +446,29 @@ suite('message_manager.js >', function() {
...
-            message.id, true
-          )
-        );
-      });
+      Settings.sendReadReport.then(function(result) {
+        messagesList.forEach(function(message) {
+          assert.ok(
+            MockNavigatormozMobileMessage.markMessageRead.calledWith(
+              message.id, true, result
+            )
```
julienw: nit: you can use `sinon.assert.calledWith(spy, ...)` instead of `assert.ok(spy.calledWith(...))`. It's easier to debug when there is a problem because the error is a lot more verbose.

apps/sms/views/shared/js/settings.js
```diff
@@ -72,6 +79,15 @@ var Settings = {
     }
   },
 
+  // return resolve promise with sendReadReport status
+  initSendReadReport: function initSendReadReport(value) {
+    this._sendReadReportDefer.resolve(value);
+    delete this._sendReadReportDefer;
+
+    navigator.mozSettings.addObserver('messages.mms.sendReadReport.enabled',
```
julienw: note: if matej likes my proposal, we'll need to reverse the meaning here; so replace "enabled" by "disabled" and resolve the promise with "!value".

Also, nit: please put the name in a constant, like `READ_AHEAD_THREADS_KEY`

apps/sms/services/test/unit/message_manager_test.js
kumarrishav: @julienw  still working on unit test. i commented  those

apps/sms/views/shared/js/settings.js
kumarrishav: @julienw  i have something better ;) i guess . will post on bug

apps/sms/views/shared/js/settings.js
julienw: I think it's a good thing to keep everything related to settings in one place.

So that means we could still have a function in Settings that requests mozSettings and returns a promise.

But I also think it's good to cache the value because currently we may call the function "markMessagesRead" several time (if we call markThreadRead for several threads).
Maybe a fix for this could be to make markThreadRead accept a list of threads so that markMessagesRead is called only once ?

In the future we'll use a function that returns a promise instead of a property that holds a promise anyway so we'll have to do this work... But I feel it's outside the scope of this bug.

From my point of view, the code is nearly ready here so I'd prefer to not change much more, except for the text change and maybe reverting the meaning (but reverting it in Settings only).

apps/sms/views/shared/js/settings.js
```diff
@@ -72,6 +79,15 @@ var Settings = {
     }
   },
 
+  // return resolve promise with sendReadReport status
+  initSendReadReport: function initSendReadReport(value) {
```
julienw: I wonder what the value is if the setting does not exist yet. Is it undefined ? Should we set a default value if it's undefined ? I think that if we reverse the meaning (disabled instead of enabled), we may want to disable by default and so set the setting to "true".

apps/sms/services/test/unit/message_manager_test.js
julienw: oki

apps/sms/services/js/message_manager.js
```diff
@@ -429,26 +429,23 @@ var MessageManager = {
...
-    var req = this._mozMobileMessage.markMessageRead(id, isRead, true);
-    // isRead == false i.e mark as unread case, marking only one message
-    // as unread is sufficient.
-    req.onsuccess = (function onsuccess() {
-      if (!list.length || !isRead) {
-        return;
-      } else if (isRead) {
-        this.markMessagesRead(list, isRead);
+    // shouldSendReadReport == true if Send-Read-Reports switch in
+    //  Messaging Settings is enabled, else false
```
julienw: nit: you have 2 spaces at the start of the line

apps/sms/services/js/message_manager.js
```diff
@@ -429,26 +429,23 @@ var MessageManager = {
...
-    }).bind(this);
 
-    req.onerror = function onerror() {
-      console.error(
-        'Error while marking message %d as read: %s', id, this.error.name
+      var id = list.pop();
+      var req = this._mozMobileMessage.markMessageRead(id, isRead, shouldSendReadReport);
+
+      return req.then(
+        () => isRead ? markMessage(shouldSendReadReport) : Promise.resolve()
```
julienw: can you please add back the comment `// if isRead == false i.e mark as unread case, marking only one message as unread is sufficient.` ?

apps/sms/views/shared/js/settings.js
kumarrishav: gasolin asked to set default value in some config file of setting.
Also, if setting doesn't exist yet. It's false. It's in init()  of settings.js

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=eadcbf71764ffb3f403e3e5c16c7d5941e32a7f4

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=cc5f06973e478bd4eee5f7956ccc319d2b4bd116

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0a1033d6d56d3bdd70cbe65ccbfcc182f395a5fa

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=09f8ead67d6ba07724490c8857399a5f7b6e45df

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=54490c95a6a69e151beea39dd9914afec1795bad

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4559ce38d1464ab04318facc4d5b800b36485764

apps/settings/elements/messaging.html
```diff
@@ -23,8 +23,14 @@ <h2 data-l10n-id="message-confirmation"></h2>
       <ul class="mmsSettings-list">
         <li id="menuItem-readReport" aria-disabled="true">
```
kumarrishav: @julienw  will change the id at last, this will led to change at other place too. SO, will do at last.

menuItem-readReport

apps/sms/services/js/message_manager.js
```diff
@@ -426,26 +426,26 @@ var MessageManager = {
...
-      console.error(
-        'Error while marking message %d as read: %s', id, this.error.name
-      );
+      var id = list.pop();
+      var req = this._mozMobileMessage.markMessageRead(
+        id, isRead, shouldSendReadReport);
+
+      // if isRead == false i.e mark as unread case,
+      // marking only one message as unread is sufficient
+      return req.then(() => {
```
kumarrishav: @julienw  it is throwing Type error. Don't know why.  It was fine with onsuccess before.
 TypeError: req.then is not a function

apps/sms/services/test/unit/message_manager_test.js
```diff
@@ -6,7 +6,6 @@
         SMIL,
         Threads
 */
-
```
kumarrishav: nits: 
will add space back again.

apps/sms/views/shared/js/settings.js
```diff
@@ -35,14 +36,16 @@ var Settings = {
     var keyHandlerSet = {
       'dom.mms.operatorSizeLimitation': this.initMmsSizeLimitation.bind(this),
       'operatorResource.sms.maxConcat':
-        this.initSmsMaxConcatenatedMsg.bind(this)
+        this.initSmsMaxConcatenatedMsg.bind(this),
+      'messages.mms.sendReadReport.enabled':
```
kumarrishav: here didn't use constant, because of two reason. (i) make consistency and look better like others two. (ii) it will passed as key i..e Settings[...] in a suite function , so if i use this.SEND_READ_REPORT_KEY. Throwing error in test.

apps/sms/views/shared/js/settings.js
kumarrishav: @julienw  didn't change to .then(), as it is effecting the createLock.get(key) mock. So, didn't get better option to resolve this. If you can help me here.?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=798e6b396a39c73e416c6d42b742660e306ce0ca

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7dfb04106973de53d64ba351788702012c9f2d5e

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8d9413408c834669c5cd3e3b890897e5342b8d45

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d38e9a6be10d80448edcbe576c2376b1a31cd136

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=45417a45cf1b8d51347c5b4dc3d9177e01883529

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=fe4966b193fdd1b63c3c3fd689766f328caa437d

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4a5d7ee7ac828237a92d860ae704003bf91945ac

apps/sms/services/js/message_manager.js
julienw: when/where is it throwing a type error ? It works fine for me on the device ?

apps/sms/views/shared/js/settings.js
julienw: change the mock ? ;)

apps/sms/views/shared/js/settings.js
```diff
@@ -72,6 +81,15 @@ var Settings = {
     }
   },
 
+  // return resolve promise with sendReadReport status
+  initSendReadReport: function initSendReadReport(value) {
+    this._sendReadReportDefer.resolve(value);
+    delete this._sendReadReportDefer;
+
+    navigator.mozSettings.addObserver(this.SEND_READ_REPORT_KEY,
+      (e) => this.sendReadReport = Promise.resolve(e.settingValue));
```
julienw: nit: please indent like we usually do in the SMS app :)
```javascript
navigator.mozSettings.addObserver(
  this.SEND_READ_REPORT_KEY,
  (e) => ...
);
```

apps/sms/views/shared/js/settings.js
```diff
@@ -51,9 +56,13 @@ var Settings = {
     this._serviceIds = [];
 
     if (!settings) {
+      this.sendReadReport = Promise.resolve(false);
```
julienw: nit: I think you can put this directly on the object, after `smsServiceId` in line 35, and then remove this line.

apps/sms/views/shared/js/settings.js
julienw: sorry, I don't follow the problem with the test... I'd very much prefer to use the constant to avoid repeat :)

apps/sms/views/shared/js/settings.js
julienw: :+1:

apps/sms/services/test/unit/message_manager_test.js
```diff
@@ -412,91 +412,126 @@ suite('message_manager.js >', function() {
...
-      });
+    test('call mark read on the correct messages', function() {
+      MessageManager.markThreadRead(threadId);
+      cursor.continue();
+    
+      sinon.assert.calledWith(
+        MockNavigatormozMobileMessage.getMessages, {
+          threadId: 1,
+          read: false
+        }, false);
```
julienw: nit: indentation in SMS is something like:
```javascript
sinon.assert.calledWith(
  MockNavigatormozMobileMessage.getMessages,
  {
    threadId: 1,
    read: false
  },
  false
);
```
(possibly you can put the object on one line)

apps/sms/views/shared/test/unit/settings_test.js
```diff
@@ -159,7 +167,12 @@ suite('Settings >', function() {
         'operatorResource.sms.maxConcat',
         10
       );
+      triggerSettingsReqSuccess('messages.mms.sendReadReport.enabled', true);
+
+      Settings.sendReadReport.then((result) => assert.equal(result, true)
+        ).then(done, done);
```
julienw: nit: indentation (and please use isTrue)

apps/sms/views/shared/test/unit/settings_test.js
```diff
@@ -159,7 +167,12 @@ suite('Settings >', function() {
...
 
+      sinon.assert.calledWith(Settings.initSendReadReport, true);
```
julienw: not sure this is useful :) the previous assertion tests the same thing, and in a better way

apps/sms/views/shared/test/unit/settings_test.js
```diff
@@ -222,9 +235,9 @@ suite('Settings >', function() {
       navigator.mozMobileConnections = [{ iccId: 'SIM 1' }, { iccId: null }];
       Settings.init();
 
-      // Four calls for
-      // maxConcatenatedMessages/mmsSizeLimitation/mmsServiceId/smsServiceId
-      sinon.assert.callCount(navigator.mozSettings.createLock, 4);
+      // Five calls for maxConcatenatedMessages/mmsSizeLimitation
+      // sendReadReport/mmsServiceId/smsServiceId/
+      sinon.assert.callCount(navigator.mozSettings.createLock, 5);
```
julienw: not really sure all these checks with createLock are useful; aren't they implied by the other checks ?

apps/sms/services/test/unit/message_manager_test.js
```diff
@@ -412,91 +412,126 @@ suite('message_manager.js >', function() {
...
+          }
+          cursor.onsuccess();
+        }
+      };
 
-      // doing one more to see if we're not called too many times
-      MockNavigatormozMobileMessage.mTriggerMarkReadSuccess();
-      assert.equal(MockNavigatormozMobileMessage.markMessageRead.callCount, 4);
+    this.sinon.stub(MockNavigatormozMobileMessage, 'getMessages').withArgs(
+      filter, false).returns(cursor);
```
julienw: nit: indentation

apps/sms/services/test/unit/message_manager_test.js
```diff
@@ -412,91 +412,126 @@ suite('message_manager.js >', function() {
...
+      var testMessages = [
         { id: 1 },
         { id: 2 },
         { id: 3 },
         { id: 4 }
       ];
 
-      MockNavigatormozMobileMessage.mTriggerMessagesRequest(messagesList);
+      var filter = {
+        threadId: threadId,
```
julienw: nit: you can just use the shorter `threadId`

apps/sms/services/test/unit/message_manager_test.js
```diff
@@ -412,91 +412,126 @@ suite('message_manager.js >', function() {
...
+      copyMsgIds = messageIds.slice();
     });
 
-    test('properly mark all ids as read', function() {
-      MessageManager.markMessagesRead(messageIds);
-      while (MockNavigatormozMobileMessage.mTriggerMarkReadSuccess()) {
-      }
+    test('mark thread read i.e mark all message ids as read', function(done) {
+      this.sinon.stub(MockNavigatormozMobileMessage, 'markMessageRead').
+        returns(Promise.resolve());
```
julienw: nit: I'd put the `.` on the next line.

apps/sms/services/test/unit/message_manager_test.js
```diff
@@ -412,91 +412,126 @@ suite('message_manager.js >', function() {
...
+    test('output an error if there is an error', function(done) {
       this.sinon.stub(console, 'error');
-      MockNavigatormozMobileMessage.mTriggerMarkReadError('UnknownError');
-      assert.isTrue(
-        console.error.firstCall.args.some(
-          (arg) => typeof arg === 'string' && arg.includes('UnknownError')
-        )
-      );
+      this.sinon.stub(MockNavigatormozMobileMessage, 'markMessageRead').
+        returns(Promise.reject('UnknownError'));
```
julienw: same here

apps/sms/services/test/unit/message_manager_test.js
```diff
@@ -412,91 +412,126 @@ suite('message_manager.js >', function() {
...
+        returns(Promise.resolve());
+
+      MessageManager.markMessagesRead(messageIds, false).then(() => {
+        sinon.assert.callCount(
+          MockNavigatormozMobileMessage.markMessageRead, 1
+        );
+
+        sinon.assert.calledWith(
+          MockNavigatormozMobileMessage.markMessageRead,
+            copyMsgIds.pop(), false, true
```
julienw: nit: indentation

apps/sms/services/js/message_manager.js
kumarrishav: umm, Sorry i didn't check on device.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8e527db386eecdf63f80914b41101e6a3f1db0d3

apps/sms/views/shared/test/unit/settings_test.js
kumarrishav: yeah, But i think it's good to keep it as previous assertion check the promise return value and this one check that function is called with expected argument. Though it means the same.
Still, if you say, it will remove it then .  @julienw

apps/sms/views/shared/js/settings.js
kumarrishav: @julienw  Why i am getting this error: Reference error : Settings is not defined
When i replaced the string with this.SEND_READ_REPORT_KEY

apps/sms/views/shared/test/unit/settings_test.js
kumarrishav: Yeah, They are implied by other checks. In my POV, it's better to have it, as it give clear understanding of test that ,what n all and where these locks created.

apps/sms/services/test/unit/message_manager_test.js
kumarrishav: Didn't get you here :)

apps/sms/views/shared/js/settings.js
kumarrishav: umm,  that's what i was trying to skip ;) Do you want me to change the mock ;) ?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=77ad52fcf153f3bd4a28c176a7dfa73786c823b0

apps/sms/views/shared/js/settings.js
julienw: you need to use `[this.SEND_READ_REPORT_KEY]`, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#Computed_property_names

apps/sms/views/shared/js/settings.js
julienw: yes please ;)

apps/sms/views/shared/test/unit/settings_test.js
julienw: I think it's testing things that are too internal; I don't think we test the external behavior with these checks. Remember the goal of tests: test the code from the outside. Then if the internal code can change but not the external behavior/interface, the test should not change.

In this case I think it's messy to test the count of createLock..

apps/sms/services/test/unit/message_manager_test.js
julienw: see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#Property_definitions

Basically:
```javascript
var o = { a: a, b: b };
```
is equivalent to
```javascript
var o = { a, b };
```

(given the variables `a` and `b` exist of course).

So you can simply use `threadId` instead of `threadId: threadId`; it's shorter.  But this is really a nit !

apps/sms/views/shared/test/unit/settings_test.js
```diff
@@ -159,7 +168,13 @@ suite('Settings >', function() {
         'operatorResource.sms.maxConcat',
         10
       );
+      triggerSettingsReqSuccess('messages.mms.sendReadReport.enabled', true);
+
+      Settings.sendReadReport.then(
+        (result) => assert.isTrue(result)
+      ).then(done, done);
 
+      sinon.assert.calledWith(Settings.initSendReadReport, true);
```
julienw: answering to:
> yeah, But i think it's good to keep it as previous assertion check the promise return value and this one check that function is called with expected argument. Though it means the same.
Still, if you say, it will remove it then . @julienw

I think this assertion is testing something internal. In the end, what we want to know is whether sendReadReport returns the right value. We don't care how this is internally implemented.

Please tell me if you need clearer explanations, maybe I'm not clear :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3dedf2c0ba6e4c8ec0c86134205ce6533de992a6

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5bf65c0da0c8cecf8d7b9a297ffbea4cca37b092

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=98c2751b2e7a3f1b7ffe70d1387ba0d1f065112b

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=59b6af99ab9c9266d9b40b2a671b1bc21099fbbf

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=245827a2e285c3592ebac85945f147cf3276ccd1

PR #32063, opened 2015-09-25
Bug 1205874 - [Messages] After starting a new conversation from previously saved draft, that conversation will be counted twice when using Select All in Inbox. Remove draft from Threads. r=schung

azasypkin:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=fa2076de3f7ce4d81b997b95a813af6b84fbacbb

apps/sms/views/inbox/js/inbox.js
```diff
@@ -974,7 +974,7 @@ var InboxView = {
     }
 
     if (thread.isDraft) {
-      this.removeThread(thread.id);
+      this.deleteThread(thread.id);
```
azasypkin: It will make redundant call to the ```Notification.get```, but we do the same for the case when we just delete draft in selection mode. Maybe we can just file bug to improve this if we really want to.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8a1bb70a0cf03d221cc0653ffeaca4ed38a4d8ea

apps/sms/views/inbox/js/inbox.js
steveck-chung: Yeah I think we can fire another bug to improve the thread deletion. BTW maybe we can rename the removeThread as well because it's misleading, removeThreadDom or something could be more clear.

azasypkin: > BTW maybe we can rename the removeThread as well because it's misleading, removeThreadDom or something could be more clear.

Yep, I thought about the same! Will rename

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7398c7a5897d5581ce3a258d5ec76c7555cc8386

apps/sms/views/inbox/js/inbox.js
azasypkin: > Yeah I think we can fire another bug to improve the thread deletion. 

Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1209963

PR #32066, opened 2015-09-25
Bug 1208108 - [Music][NGA] Clean-up before switching out Music app

justindarc:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1bdfc560b95018c7f88a3f95b5e96602288d9ed8

dev_apps/music-nga/js/endpoint.js
```diff
@@ -276,7 +276,14 @@ function getAlbum(filePath) {
     var album = song.metadata.album;
 
     return new Promise((resolve) => {
-      Database.enumerateAll('metadata.album', album, 'next', songs => resolve(songs));
+      Database.enumerateAll('metadata.album', album, 'next', (songs) => {
+        songs.sort((a, b) => {
+          return (a.metadata.discnum - b.metadata.discnum) ||
+                 (a.metadata.tracknum - b.metadata.tracknum);
+        });
```
jimporter: This doesn't belong in the front-end; ordering guarantees should probably be a part of the database backend. I think it'd make more sense in `js/db.js`.

dev_apps/music-nga/js/endpoint.js
```diff
@@ -292,7 +299,17 @@ function getArtist(filePath) {
...
     return new Promise((resolve) => {
-      Database.enumerateAll('metadata.artist', artist, 'next', songs => resolve(songs));
+      Database.enumerateAll('metadata.artist', artist, 'next', (songs) => {
+        songs.sort((a, b) => {
+          var albumSort = a.metadata.album < b.metadata.album ?
+            -1 : (a.metadata.album > b.metadata.album ?  1 : 0);
+
+          return albumSort || (a.metadata.discnum - b.metadata.discnum) ||
+                              (a.metadata.tracknum - b.metadata.tracknum);
+        });
```
jimporter: Ditto here. This should go in `js/db.js`. Also, you're reusing the album sorting code from above. We should just make them separate functions that get composed as necessary.

dev_apps/music-nga/js/endpoint.js
justindarc: This technically is the back-end though. This is not the front-end.

dev_apps/music-nga/js/endpoint.js
jimporter: We'd also probably want to sort the list of artists, but we could wait until later for that I suppose. I'm also not totally sure if sorting by album name is strictly necessary at the moment; if it is, we should probably be using `localeCompare`, since that should produce more useful results.

PR #32127, opened 2015-09-30
Bug 1206678 - Outgoing MMS contains subject that was typed by user, but then hidden, r=azasypkin

steveck-chung:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=21bb68ca546a29a67a4704a83dc36830d9bff3e1

apps/sms/views/conversation/js/conversation.js
```diff
@@ -2172,7 +2172,7 @@ var ConversationView = {
 
   sendMessage: function conv_sendMessage(opts) {
     var content = Compose.getContent(),
-        subject = Compose.getSubject(),
+        subject = Compose.isSubjectVisible ? Compose.getSubject() : null,
```
julienw: I really would prefer that Compose.js has this in Compose.getSubject() (it already has such a check in Compose.hasSubject() btw...)

steveck-chung: Oh, it was my first implementation before the draft part conflict actually ;) @azasypkin  How do you  fell about getSubject checks the visibility inside compose?

azasypkin: Yeah, I think it's better, so we can remove my temporal workaround in https://github.com/mozilla-b2g/gaia/blob/2433ba5a10ede2f3f221376a4ed7fc3c33ad266d/apps/sms/views/conversation/js/conversation.js#L2914 :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=770e525a46d6c1ba4d37c89949484562af43c592

PR #32130, opened 2015-09-30
Bug 1176297 - Hangup calls when pressing the power button. r=timdream

gabrielesvelto:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=27970db2d6fca89d90a9ec9725783e901a96e67e

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1c210c932b6855d1b733db7faa3c36e391abaffe

apps/system/js/screen_manager.js
```diff
@@ -3,6 +3,14 @@
 /* globals SettingsListener, Bluetooth, StatusBar, Service,
            ScreenBrightnessTransition, ScreenWakeLockManager */
 
+/**
+ * Counts the number of currently active calls.
+ */
+function numOpenLines() {
```
timdream: Move this into `ScreenManager`, e.g. `ScreenManager._getNumOfCalls`.

apps/system/js/screen_manager.js
```diff
@@ -292,11 +319,15 @@ var ScreenManager = {
...
       case 'callschanged':
-        if (!telephony.calls.length &&
-            !(telephony.conferenceGroup &&
-              telephony.conferenceGroup.calls.length)) {
-
-          this.turnScreenOn();
+        if (numOpenLines() === 0) {
+          if (this._turnOffWhenCallsEnd) {
+            this._turnOffWhenCallsEnd = false;
+            setTimeout(function() {
```
timdream: Why do we need this set timeout?

apps/system/js/screen_manager.js
```diff
@@ -292,11 +319,15 @@ var ScreenManager = {
...
+              this.turnScreenOff(/* instant */ true, 'powerkey');
+            }.bind(this), 2000);
```
timdream: `2000`?

apps/system/js/screen_manager.js
gabrielesvelto: That's an UX problem: when we hangup the last call we usually display the last call on the callscreen with the duration of the call, etc... for a couple of seconds, then we hide the callscreen. If we turn off the screen immediately this cannot be displayed, the user will not see the calls being terminated. Since I didn't have an UX for this bug I went for waiting 2 seconds before turning off the screen as that's the same behavior I see on iOS (haven't tried on Android though).

apps/system/js/screen_manager.js
gabrielesvelto: See above, this is because the callscreen is hidden after 2s when hanging up the last call.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=509d1f6308bb71b8298ccbfcfdbe456e59ffb61a

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=32b2fd04903d0f4337cae36a72cab76d659ca1ff

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9074c76017f3043da6305d8085289813547e2b36

PR #32131, opened 2015-09-30
 Bug 1202382 - Implement read-only Bookmarks Adapter for FxSync feature.

weilonge:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b468efa620b424ce20f2706fe39cdf15d048f882

apps/sync/js/adapters/bookmarks.js
```diff
@@ -0,0 +1,325 @@
...
+  for FxSync only which has the same functionality, and the adapter is based on
+  SyncEngine's design to handle 'update' and 'handleConflict' request. Bookmarks
+  Data Adapter provides Read-Only version currently, and the basic ideas are
+  included:
+  - Traverse all records from Bookmarks Collection (BC).
+  - Add a new or existed bookmark to Bookmarks DataStore (BDS).
+    - Convert the records from BC to BDS format.
+    - Merge a new record and the existed one then write to BDS.
+  - Remove an existed bookmark when receiving a "deleted: true" record.
+**/
```
ferjm: "The purpose of the Bookmarks adapter is to retrieve bookmark records from Firefox Sync and merge them in a local bookmarks DataStore ('firefox-sync-bookmarks') which will be hosting Firefox Sync records only. This adapter implements the API defined by SyncEngine. For now the implementation allows only to retrieve bookmarks from Sync but it is not ready for pushing any changes. The high level algorithm implemented here is the following:
* Traverse all the records coming from Firefox Sync and stored locally in Kinto.
* Convert any new bookmark from the Firefox Sync format to the 'firefox-sync-bookmarks' DataStore one.
* For existing bookmarks merge the remote with the local bookmark after converting them to the proper 'firefox-sync-bookmarks' DS format.
* Remove any existing bookmark that has been remotely deleted (these records come with a 'deleted: true' property)."

apps/sync/js/adapters/bookmarks.js
```diff
@@ -0,0 +1,325 @@
...
+'use strict';
+
+/* global
+  asyncStorage,
+  DataAdapters,
+  LazyLoader
+*/
+
+const BOOKMARKS_COLLECTION_MTIME = 'collections::bookmarks::mtime';
+const BOOKMARKS_SYNCTOID_PREFIX = 'SynctoId::bookmarks::';
```
ferjm: Let's keep everything in lower case. `synctoid::bookmarks::`

apps/sync/js/adapters/bookmarks.js
```diff
@@ -0,0 +1,325 @@
...
+
+var BookmarksHelper = (() => {
+  var bookmarksStore;
+  function _ensureStore() {
+    if (bookmarksStore) {
+      return Promise.resolve(bookmarksStore);
+    }
+    return navigator.getDataStores('firefox-sync-bookmarks').then(stores => {
+      bookmarksStore = stores[0];
```
ferjm: Could you check that we are the owners of this DataStore, please?

This is a privacy issue that is quite unlikely to happen, but just in case DataStore is ever opened to non certified apps we should be ready for it. If another app creates a DataStore with the same name, we might end up writing all the user's bookmarks in that DataStore.

ferjm: Also, I think we need to do the same for the history adapter.

apps/sync/js/adapters/bookmarks.js
```diff
@@ -0,0 +1,325 @@
...
+    });
+  }
+
+  function getSyncedCollectionMtime() {
+    return new Promise(resolve => {
+      asyncStorage.getItem(BOOKMARKS_COLLECTION_MTIME, resolve);
+    });
+  }
+
+  function setDataStoreId(synctoId, dataStoreId) {
```
ferjm: Add a comment explaining why we need this matching between IDs

apps/sync/js/adapters/bookmarks.js
```diff
@@ -0,0 +1,325 @@
...
+    return new Promise(resolve => {
+      asyncStorage.setItem(
+        BOOKMARKS_SYNCTOID_PREFIX + synctoId, dataStoreId, resolve);
+    });
+  }
+
+  function getDataStoreId(synctoId) {
+    return new Promise(resolve => {
+      asyncStorage.getItem(BOOKMARKS_SYNCTOID_PREFIX + synctoId, resolve);
```
ferjm: console.error in case `.getItem` fails, please.

apps/sync/js/adapters/bookmarks.js
```diff
@@ -0,0 +1,325 @@
...
+    });
+  }
+
+  function mergeRecordsToDataStore(localRecord, remoteRecord) {
+    if (!localRecord || !remoteRecord ||
+        localRecord.id !== remoteRecord.id ||
+        (remoteRecord.type === 'url' && localRecord.url !== remoteRecord.url)) {
+      // The local record has different url(id) with the new one.
```
ferjm: "// Local and remote records have different url(id).

apps/sync/js/adapters/bookmarks.js
```diff
@@ -0,0 +1,325 @@
...
+      console.error('Inconsistent records on url', localRecord, remoteRecord);
+      throw new Error('Inconsistent records on url');
```
ferjm: 'Inconsistent records' is enough

apps/sync/js/adapters/bookmarks.js
```diff
@@ -0,0 +1,325 @@
...
+    if (!localRecord.fxsyncId && typeof remoteRecord.fxsyncId === 'string') {
+      /* When a localRecord is existed without fxsyncId, assign fxsyncId to it
+         from a remoteRecord. This case always happens at first synchronization
+         or merging two records with the same URL. */
+      localRecord.fxsyncId = remoteRecord.fxsyncId;
+    } else if (localRecord.fxsyncId !== remoteRecord.fxsyncId) {
+      // Two records have different fxsyncId but have the same url(id).
+      console.log('Inconsistent records on FxSync ID',
+        localRecord, remoteRecord);
+      throw new Error('Inconsistent records on FxSync ID',
```
ferjm: 'Records should have the same Firefox Sync ID'

apps/sync/js/adapters/bookmarks.js
```diff
@@ -0,0 +1,325 @@
...
+
+    localRecord.name = remoteRecord.name;
+    localRecord.fxsyncPayload = remoteRecord.fxsyncPayload;
+
+    return localRecord;
+  }
+
+  function addBookmark(remoteRecord) {
+    // 1. Get bookmark by url(id of DataStore)
+    // 2.A Merge the existing one and new one if it's an existing one,
```
ferjm: 2.A. If the bookmark already exists locally, we merge it with the remote one.

apps/sync/js/adapters/bookmarks.js
```diff
@@ -0,0 +1,325 @@
...
+    //     and update the bookmark.
+    // 2.B Add a new record with RevisionId.
+    // 3. Add the DataStore record ID into LocalID <-> RemoteID matching table.
+
+    var id = remoteRecord.id;
+    var revisionId;
+    return _ensureStore().then(bookmarksStore => {
```
ferjm: s/bookmarksStore/store/g 

We already know we are managing bookmarks ;)

apps/sync/js/adapters/bookmarks.js
```diff
@@ -0,0 +1,325 @@
...
+      }, Promise.resolve()).then(resolve);
+    });
+  }
+
+  function deleteBookmark(fxsyncId) {
+    var url;
+    return getDataStoreId(fxsyncId).then(id => {
+      url = id;
+      return _ensureStore();
+    }).then(bookmarksStore => {
```
ferjm: ditto

apps/sync/js/adapters/bookmarks.js
```diff
@@ -0,0 +1,325 @@
...
+        continue;
+      }
+      var typeWithUri = ['query', 'bookmark']
+          .some(value => value === payload.type);
+      if (typeWithUri && !payload.bmkUri) {
+        console.warn('Incorrect payload? ', payload);
+        continue;
+      }
+      bookmarks.push({
+        id: payload.type === 'bookmark' ? payload.bmkUri :
```
ferjm: It may be worth documenting the format of the id somewhere

apps/sync/js/adapters/bookmarks.js
```diff
@@ -0,0 +1,325 @@
...
+          (payload.type + '|' + payload.id),
+        url: payload.bmkUri,
+        name: payload.title,
+        type: payload.type === 'bookmark' ? 'url' : 'others',
+        iconable: false,
+        icon: '',
```
ferjm: :( Should we at least try to get the favicon?

apps/sync/js/adapters/bookmarks.js
```diff
@@ -0,0 +1,325 @@
...
+        fxsyncPayload: payload,
+        fxsyncId: payload.id
+      });
+    }
+
+    if (bookmarks.length === 0) {
+      return Promise.resolve(false);
```
ferjm: `return Promise.resolve(false /* no writes done into kinto */);`

apps/sync/js/adapters/bookmarks.js
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    return BookmarksHelper.updateBookmarks(bookmarks).then(() => {
+      var latestMtime = remoteRecords[0].last_modified;
+      return BookmarksHelper.setSyncedCollectionMtime(latestMtime);
+    }).then(() => {
+      // Always return false for a read-only operation.
+      return Promise.resolve(false);
```
ferjm: `return Promise.resolve(false /* no writes done into kinto */);`

apps/sync/js/adapters/bookmarks.js
```diff
@@ -0,0 +1,325 @@
...
+    });
+  },
+
+  update(remoteBookmarks) {
+    var mtime;
+    return LazyLoader.load(['shared/js/async_storage.js'])
```
ferjm: Instead of lazy loading `async_storage.js` before the function call, lazy load it within the function implementation. Also for the setter https://github.com/mozilla-b2g/gaia/pull/32131/files#diff-6dac2f6d58d68fbc11a10804f9946d4fR45

apps/sync/test/unit/adapters/bookmarks_test.js
```diff
@@ -0,0 +1,369 @@
...
+    });
+  }
+
+  function verifyBookmarks(collectionItem, bookmarksItem) {
+    var payload = collectionItem.payload;
+    if (payload.type === 'bookmark' || payload.type === 'query') {
+      assert.equal(payload.bmkUri, bookmarksItem.url);
+      assert.equal(payload.title, bookmarksItem.name);
+    } else {
+      assert.equal(payload.type + ':' + payload.id, bookmarksItem.id);
```
ferjm: I might be missing something... but you used `|` as the separator at https://github.com/mozilla-b2g/gaia/pull/32131/files#diff-6dac2f6d58d68fbc11a10804f9946d4fR277

ferjm: Also `query` type is supposed to have a `payload.type|payload.id`, right?

apps/sync/js/adapters/bookmarks.js
```diff
@@ -0,0 +1,325 @@
...
+      if (typeWithUri && !payload.bmkUri) {
+        console.warn('Incorrect payload? ', payload);
+        continue;
+      }
+      bookmarks.push({
+        id: payload.type === 'bookmark' ? payload.bmkUri :
+          (payload.type + '|' + payload.id),
+        url: payload.bmkUri,
+        name: payload.title,
+        type: payload.type === 'bookmark' ? 'url' : 'others',
```
ferjm: Is `others` something already considered in the 'bookmarks_store' DS? If so, could you document all the types at line https://github.com/mozilla-b2g/gaia/pull/32131/files#diff-6dac2f6d58d68fbc11a10804f9946d4fR173 please?

apps/sync/test/unit/adapters/bookmarks_test.js
```diff
@@ -0,0 +1,369 @@
...
+    }
+    assert.equal(payload.id, bookmarksItem.fxsyncId);
```
ferjm: I am missing the check for some record properties (i.e. type, iconable, icon)

apps/sync/test/unit/adapters/bookmarks_test.js
```diff
@@ -0,0 +1,369 @@
...
+    return navigator.getDataStores().then(stores => {
+      return Promise.resolve(stores[0]);
+    });
+  }
+
+  function verifyBookmarks(collectionItem, bookmarksItem) {
+    var payload = collectionItem.payload;
+    if (payload.type === 'bookmark' || payload.type === 'query') {
+      assert.equal(payload.bmkUri, bookmarksItem.url);
+      assert.equal(payload.title, bookmarksItem.name);
```
ferjm: We should also check bookmarksItem.id in this case.

apps/sync/test/unit/adapters/bookmarks_test.js
ferjm: I guess this is not failing cause you are only testing `bookmark` type. Add tests for all the possible types, please.

apps/sync/test/unit/adapters/bookmarks_test.js
```diff
@@ -0,0 +1,369 @@
...
+  teardown(() => {
+    updateBookmarksSpy.restore();
+    MockDatastore._inError = false;
+    MockDatastore._records = Object.create(null);
+    window.asyncStorage.mTeardown();
+  });
+
+  test('update - empty records', done => {
+    var bookmarksAdapter = DataAdapters.bookmarks;
+    bookmarksAdapter.update(kintoCollection).then((result) => {
```
ferjm: nit: single arguments don't need parenthesis `result => {}`

apps/sync/test/unit/adapters/bookmarks_test.js
```diff
@@ -0,0 +1,369 @@
...
+      }
+
+      list.unshift({
+        id: 'UNIQUE_ID_' + i,
+        last_modified: initDate + i * 10,
+        payload: {
+          id: 'UNIQUE_ID_' + i,
+          bmkUri: 'http://example' + i + '.com/',
+          title: 'Example ' + i + ' Title',
+          type: 'bookmark'
```
ferjm: We need to test all possible types. Also, missing and malformed fields.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=16ae0d9943084a99893138d571b7f20e4c7ccfc4

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f158ba807ab3d88aeadc7d3ab1233a69f7137208

apps/sync/js/adapters/bookmarks.js
weilonge: Dale gave a suggestion here:
https://github.com/mozilla-b2g/gaia/pull/31896#discussion_r40199678

I think 'update' is the entry to ensure AsyncStorage has to be ready to use, and LazyLoader is at here.

apps/sync/js/adapters/bookmarks.js
weilonge: asyncStorage.getItem has no error case. shall we handle the case as well?
https://github.com/mozilla-b2g/gaia/blob/master/shared/js/async_storage.js#L93

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=723c018588925e7d166ed642135623d7e78996c9

apps/sync/js/adapters/bookmarks.js
weilonge: I trace the code to use 'url' type at here:
https://github.com/mozilla-b2g/gaia/blob/master/apps/system/js/app_chrome.js#L1043
I did not see any other types existing, and 'others' is for other types without a valid URL.

apps/sync/test/unit/adapters/bookmarks_test.js
weilonge: Got it! It should be added. :)

apps/sync/js/adapters/bookmarks.js
weilonge: ni Vivien to see if there is any suggestion. :)

apps/sync/js/adapters/bookmarks.js
```diff
@@ -0,0 +1,334 @@
...
+const BOOKMARKS_COLLECTION_MTIME = 'collections::bookmarks::mtime';
+const BOOKMARKS_SYNCTOID_PREFIX = 'synctoid::bookmarks::';
+
+var BookmarksHelper = (() => {
+  var store;
+  function _ensureStore() {
+    if (store) {
+      return Promise.resolve(store);
+    }
+    return navigator.getDataStores('firefox-sync-bookmarks').then(stores => {
```
michielbdejong: Can we not use the existing `bookmarks_store`? https://github.com/mozilla-b2g/gaia/blob/master/apps/system/manifest.webapp#L179-L182 If we diverge now, it might be harder to make them come back together further down the road.

apps/sync/js/adapters/bookmarks.js
```diff
@@ -0,0 +1,334 @@
...
+      store = stores[0];
+      return store;
+    });
+  }
+
+  function setSyncedCollectionMtime(mtime) {
+    return new Promise(resolve => {
+      asyncStorage.setItem(BOOKMARKS_COLLECTION_MTIME, mtime, resolve);
```
michielbdejong: Can't we rely on the DataStore's built-in changelog mechanism for this?

apps/sync/js/adapters/bookmarks.js
michielbdejong: You can mention that this is a temporary work-around until we implement the solution from bug 1207468.

apps/sync/js/adapters/bookmarks.js
```diff
@@ -0,0 +1,334 @@
...
+      });
+    }).catch(e => {
+      console.error(e);
+    });
+  }
+
+  function updateBookmarks(records) {
+    return new Promise(resolve => {
+      records.reduce((reduced, current) => {
+        return reduced.then(() => {
```
michielbdejong: I think you should set initialValue to `Promise.resolve()` here? Otherwise I think it will fail of `records.length <= 1`? See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce#Description

apps/sync/js/adapters/bookmarks.js
```diff
@@ -0,0 +1,334 @@
...
+        return setDataStoreId(remoteRecord.fxsyncId, id);
+      });
+    }).catch(e => {
+      console.error(e);
+    });
+  }
+
+  function updateBookmarks(records) {
+    return new Promise(resolve => {
+      records.reduce((reduced, current) => {
```
michielbdejong: I think the use of `Array.reduce` is confusing here, since you're not accumulating anything other than `promise.then().then().then()....`. Maybe something more like:
````js
return Promise.all(records.map(record => {
  if(record.deleted) {
    return deleteBookmark(current.fxsyncId);
  }
  return addBookmark(current);
});
````

File: apps/sync/js/adapters/bookmarks.js [hunk 27]
```diff
@@ -0,0 +1,334 @@
...
+  when any deleting record requests with the format [3] are coming from FxSync.
+
+  [1] Records stored in Bookmarks DataStore (BDS): {
+    "id": "http://mozilla.org/", // KEY in PDS
+    "url": "http://mozilla.org/",
+    "name": "Mozilla",
+    "type": "url",
+    "iconable": false,
+    "icon": "http://www.lego.com/favicon.ico",
+    "fxsyncPayload": Object, // payload from BC
```
michielbdejong: Why store this?

apps/sync/js/adapters/bookmarks.js
```diff
@@ -0,0 +1,334 @@
...
+    "fxsyncId": "REMOTE_ID" // [1.1]
```
michielbdejong: Maybe we can add `fxsyncLastModified` here and get rid of `BOOKMARKS_COLLECTION_MTIME`. Not sure which is better, though.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5bcf69b1c03f96bf5b7a409682def75d19859caf

apps/sync/test/unit/adapters/bookmarks_test.js
weilonge: Added.
https://github.com/mozilla-b2g/gaia/pull/32131/files#diff-92d20696e312b942fc6c1050c7fc6ce4R274

apps/sync/js/adapters/bookmarks.js
weilonge: Please see the [comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1207725#c1) to discuss creating a new data store for FxSync Bookmarks.

apps/sync/js/adapters/bookmarks.js
weilonge: Promise.all won't guarantee the execution after the previous one is resolved.
Please see the [example](http://jsbin.com/rucupanilu/edit?js,console), and case3 is what I expected.

apps/sync/js/adapters/bookmarks.js
weilonge: We store this to keep the data is compatible to FxSync Bookmarks even it's stored in DataStore.
From the adapter's perspective, it is just stored for the consumer to handle it.

apps/sync/js/adapters/bookmarks.js
weilonge: This timestamp is for Bookmarks Collection rather than DataStore, and it's necessary to be recorded what records had been synchronized.

apps/sync/js/adapters/bookmarks.js
weilonge: Based on my [comment](https://github.com/mozilla-b2g/gaia/pull/32131/files#r40929357), I would like to use BOOKMARKS_COLLECTION_MTIME. fxsyncLastModified does not have to be stored for each record, and the latest modified time is sufficient for recognizing the records which are not synchronized yet.

apps/sync/js/adapters/bookmarks.js
weilonge: Do you agree that the consumer of bookmark datastore to get the favicon instead of the adapter to do this?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c9bac78cf90b255f5a50b26a76d5869e002aff4f

apps/sync/js/adapters/bookmarks.js
```diff
@@ -0,0 +1,347 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+  Bookmarks Data Adapter
+
+    The purpose of the Bookmarks adapter is to retrieve bookmark records from
+  Firefox Sync and merge them in a local bookmarks DataStore
+  ('firefox-sync-bookmarks') which will be hosting Firefox Sync records only.
```
ferjm: I know I proposed this naming, but thinking about twice it's probably better to keep the same naming as 'bookmarks_store', so this could be 'sync_bookmarks_store'.

apps/sync/js/adapters/bookmarks.js
```diff
@@ -0,0 +1,347 @@
...
+  function mergeRecordsToDataStore(localRecord, remoteRecord) {
+    if (!localRecord || !remoteRecord ||
+        localRecord.id !== remoteRecord.id ||
+        (remoteRecord.type === 'url' && localRecord.url !== remoteRecord.url)) {
+      // Local and remote records have different url(id).
+      console.error('Inconsistent records', localRecord, remoteRecord);
+      throw new Error('Inconsistent records');
+    }
+    if (!localRecord.fxsyncId && typeof remoteRecord.fxsyncId === 'string') {
+      /* When a localRecord is existed without fxsyncId, assign fxsyncId to it
```
ferjm: When localRecord has no fxsyncId...

apps/sync/js/adapters/bookmarks.js
ferjm: Let's leave it with others. We can take the type from fxsyncPayload if needed.

apps/sync/js/adapters/bookmarks.js
ferjm: Ok

apps/sync/js/adapters/bookmarks.js
```diff
@@ -0,0 +1,347 @@
...
+    return Promise.resolve(conflict.remote);
+  },
+
+  specialRecords: [
+    'menu',
+    'mobile',
+    'toolbar',
+    'places',
+    'unfiled'
+  ]
```
ferjm: This is not needed anymore

apps/sync/js/adapters/bookmarks.js
ferjm: Fair enough. Still a bit unsafe though...

apps/sync/test/unit/adapters/bookmarks_test.js
```diff
@@ -0,0 +1,505 @@
...
+      assert.equal(payload.title, bookmarksItem.name);
+      assert.equal('others', bookmarksItem.type);
+    } else if (payload.type === 'separator') {
+      assert.equal(payload.type + '|' + payload.id, bookmarksItem.id);
+      assert.equal('others', bookmarksItem.type);
+    } else if (payload.type === 'microsummary') {
+      console.warn('microsummary is OBSOLETED ', payload);
+    } else {
+      assert.ok(false, 'Unknown type');
+    }
```
ferjm: I think this would be more clear with a switch

apps/sync/js/adapters/bookmarks.js
ferjm: You are right. Yes, I think it's worth to console.warn the null case on async_storage.js.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b7cbb6b202eaac5875c70dc4f682bc7884553c9b

PR #32136, opened 2015-09-30
Bug 1209562 - test_fmradio_frequency_dialer.py : UnicodeEncodeError: …

JohanLorenzo:
…'decimal' codec can't encode character u'\u2068' in position 0: invalid decimal Unicode string

### DIFF
--- a/tests/python/gaia-ui-tests/gaiatest/apps/fmradio/app.py
+++ b/tests/python/gaia-ui-tests/gaiatest/apps/fmradio/app.py
@@ -2,6 +2,8 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import re
+
 from marionette_driver import expected, By, Wait
 from marionette_driver.marionette import Actions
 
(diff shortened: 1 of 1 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d0d83367d1142daeb7543a7765e40f5eb114fffd

tests/python/gaia-ui-tests/gaiatest/apps/fmradio/app.py
```diff
@@ -81,7 +83,14 @@ def is_power_button_on(self):
 
     @property
     def frequency(self):
-        return float(self.marionette.find_element(*self._frequency_display_locator).text)
+        raw_frequency = self.marionette.find_element(*self._frequency_display_locator).text
+        return float(self._crop_trailing_mhz_and_invisible_characters(raw_frequency))
+
+    @staticmethod
+    def _crop_trailing_mhz_and_invisible_characters(raw_frequency):
+        match = re.search(r'\d{2}\.\d', raw_frequency)
```
npark-mozilla: Hmm, in this case, we always assumes that the FM frequency is in following format: XX.XMhz.  But since FM frequency can go above 3 digits after the decimal point, (eg. "107.5MHz") it'll not work properly in such case.  If we're only going to check the 87.5Mhz and have no need for general method, perhaps we should make it clear in the method name.  Could we remove anything that is non-digit and decimal point instead?  like, this would work:
match = re.search(r'\d+\.\d',raw_frequency)

JohanLorenzo: oops, that's right! I'll update it

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0777d5c5bfa4687cb5f8033afd4757a1baa34760

PR #32138, opened 2015-09-30
Bug 1208532 - [Messages][Refactoring] Extract SystemMessageHandler from ActivityHandler. r=schung

azasypkin:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b4d653f86afc9e7792c9bd9285948245e2e544ea

apps/sms/views/shared/js/startup.js
```diff
@@ -17,6 +17,7 @@
          Navigation,
```
azasypkin: It's just WIP, so I only cared about non-split version to get early feedback :)

apps/sms/views/shared/js/system_message_handler.js
```diff
@@ -0,0 +1,341 @@
...
+        if (Settings.hasSeveralSim() && message.iccId) {
+          return Utils.getSimNameByIccId(message.iccId).then(
+            (simName) => ({
+              id: 'dsds-notification-title-with-sim',
+              args: { sim: simName, sender: sender }
+            })
+          );
+        }
+
+        return { raw: sender };
```
azasypkin: This thing fixes the "funny" issue I mentioned in https://bugzilla.mozilla.org/show_bug.cgi?id=1208532#c2

apps/sms/views/shared/js/system_message_handler.js
```diff
@@ -0,0 +1,341 @@
...
+    onSmsReceivedSystemMessage(message) {
+      // Acquire the cpu wake lock when we receive an SMS.  This raises the
+      // priority of this process above vanilla background apps, making it less
+      // likely to be killed on OOM.  It also prevents the device from going to
+      // sleep before the user is notified of the new message.
+      //
+      // We'll release it once we display a notification to the user.  We also
+      // release the lock after 30s, in case we never run the notification code
+      // for some reason.
+      this.requestWakeLock();
```
azasypkin: I've extracted this logic into separate methods ```requestWakeLock``` and ```releaseWakeLock``` initially because I was calling ```releaseWakeLock``` in several different places like currently in master, but I then I decided to always release wake lock here (there were several cases in old activity_handler when we didn't call release explicitly and were relying on timeout, now we rely on timeout only if something unexpected happened and ```handleReceivedMessagePromise``` is rejected).

So in the end I think it's good enough to have separate methods instead of inline logic, but tell me if you think it's overkill and I move it back inline here :)

apps/sms/views/shared/js/system_message_handler.js
```diff
@@ -0,0 +1,341 @@
...
+
+      var handleReceivedMessagePromise;
+      if (message.messageClass === 'class-0') {
+        handleReceivedMessagePromise = this.onMessageZeroClassReceived(message);
+      } else {
+        handleReceivedMessagePromise = SilentSms.checkSilentModeFor(
```
azasypkin: I keep ```SilentSms.checkSilentModeFor``` in this method instead of ```onMessageNonZeroClassReceived``` because I feel it's a separate thing, but tell me what you think!

apps/sms/views/shared/js/utils.js
```diff
@@ -942,6 +942,21 @@
       return promise.then(
         () => shimHostIframe.contentWindow.bootstrap(appInstanceId)
       );
+    },
+
+    onceDocumentIsVisible() {
```
azasypkin: Maybe the name is not the best, but I feel like we need something like this for a long time already :) What do you think?

apps/sms/views/shared/js/system_message_handler.js
steveck-chung: It's an edge case but still a good catch :)

apps/sms/views/shared/js/utils.js
steveck-chung: Yeah I think it's good timing to make this visibility promise into an independent util function.

apps/sms/views/shared/js/startup.js
steveck-chung: ok

azasypkin: Hmm, I see we don't initialize ```ActivityHandler``` neither in split Inbox, nor in split Conversation - likely we'll have separate bug to enable ```ActivityHandler``` and ```SystemMessageHandler``` in those views once we fix https://bugzilla.mozilla.org/show_bug.cgi?id=818000

apps/sms/views/shared/js/system_message_handler.js
steveck-chung: I agree that it's 2 different things , but I think it's not possible to handle silent sms, so maybe we can check the silent message inside the onMessageNonZeroClassReceived safely. Anyway both works for me since `onSmsReceivedSystemMessage` is not complicate.

apps/sms/views/shared/js/system_message_handler.js
steveck-chung: No, I don't think it's overkill and it looks clean in `onSmsReceivedSystemMessage`

apps/sms/views/shared/js/system_message_handler.js
```diff
@@ -0,0 +1,341 @@
...
+    },
+
+    /**
+     * Processes received class-0 message.
+     * @param {MozMmsMessage|MozSmsMessage} message Received class-0 message.
+     * @returns {Promise.<void>} Promise that is resolved once message is fully
+     * handled.
+     * @private
+     */
+    onMessageZeroClassReceived(message) {
```
steveck-chung: nit: Since we also call class 0 message as flash message, maybe we can name it like `onFlashMessageReceived`?

azasypkin: Oh "flash", sounds better, maybe then we can have ```onFlashMessageReceived``` and just ```onMessageReceived``` then?

apps/sms/views/shared/js/system_message_handler.js
azasypkin: Since both work for you, I'll try to keep it like it's currently (to keep ```onMessageReceived``` simpler for now) and will see how it goes :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e88113f26430831d7b29a7ea8122b183f7b913d0

apps/sms/test/marionette/message_received_test.js
```diff
@@ -0,0 +1,145 @@
+/* global require, marionette, setup, suite, test, __dirname */
```
azasypkin: Added just few basic tests here :)

apps/sms/views/shared/js/system_message_handler.js
```diff
@@ -0,0 +1,341 @@
...
+    onNotificationSystemMessage(notification) {
+      // When notification is removed from notification tray, notification
+      // system message will still be fired, but "clicked" property will be
+      // equal to false. This should change once bug 1139363 is landed. When
+      // user clicks on notification we'll get two system messages, first to
+      // notify app that notification is clicked and then, once we show
+      // Conversation view to the user, we remove that notification from the
+      // tray that causes the second system message with "clicked" set to false.
+      if (!notification.clicked || !notification.data) {
+        return Promise.resolve();
```
azasypkin: I return ```Promise``` here just for the return result consistency and to make tests easier, but not sure if you like it :)

apps/sms/views/shared/test/unit/activity_handler_test.js
```diff
@@ -45,71 +26,41 @@ require('/views/shared/test/unit/mock_utils.js');
...
 suite('ActivityHandler', function() {
   mocksHelperForActivityHandler.attachTestHelpers();
 
-  var realSetMessageHandler;
-  var realWakeLock;
-  var realMozApps;
-  var realMozL10n;
-
-  var isDocumentHidden;
+  var navigatorMocks = new Map([
```
azasypkin: It's my new experiment with navigator mocks, I always didn't like old approach, so trying to make it easier (it's better visible in system_message_handler_tests.js - we have several navigator mocks there, here I use Map for one mock only, just to be consistent with system_message_handler_tests). It was actually quite cumbersome to make old mocks work with Promises (e.g. mozApps mock), that is one of the reasons I've tried to use new approach :)

Would be happy to hear your thoughts or other ideas!

apps/sms/views/shared/test/unit/mock_activity_handler.js
```diff
@@ -2,11 +2,5 @@
 'use strict';
 
 var MockActivityHandler = {
-  init: function() {},
```
azasypkin: All other methods are private :)

apps/sms/views/shared/test/unit/system_message_handler_test.js
```diff
@@ -0,0 +1,778 @@
...
+  'SMIL',
+  'Utils'
+]).init();
+
+suite('SystemMessageHandler', function() {
+  mocksHelperForActivityHandler.attachTestHelpers();
+
+  var isDocumentHidden;
+
+  var navigatorMocks = new Map([
```
azasypkin: Here is a better example :)

apps/sms/views/shared/test/unit/system_message_handler_test.js
```diff
@@ -0,0 +1,778 @@
...
+      setup(function() {
+        SilentSms.checkSilentModeFor.returns(Promise.resolve(true));
+      });
+
+      test('does not notify user in any way', function(done) {
+        var handlingPromise = SystemMessageHandler.onSmsReceivedSystemMessage(
+          MockMessages.sms()
+        );
+
+        assertWakeLockIsRequested();
```
azasypkin: I test for the correct wake lock request/release only in a subset of tests where it makes sense and tried to keep other tests simpler without this check :)

apps/sms/views/shared/test/unit/system_message_handler_test.js
```diff
@@ -0,0 +1,778 @@
...
+        setup(function(done) {
+          this.sinon.stub(MessageManager, 'getMessage').withArgs(
+            smsMessage.id
+          ).returns(smsMessage);
+
+          SystemMessageHandler.onSmsReceivedSystemMessage(smsMessage).then(
+            () => {
+              // We don't want to use "yield" here because we'd like to have
+              // promise-based result of yielded function.
+              notificationClickHandler = notificationStub.addEventListener.
```
azasypkin: Didn't find better approach :/ It would be great that ```yield``` could return result of the yielded function. @julienw do you know if it's possible to get result of yielded function somehow?

julienw: I don't know, but you could maybe spy the function and get its result ?

The idea is that normally when you "yield" you act on this stub, but getting the result of the yielded function is something internal that you should not care about.
But maybe in this case where we really need it, then you're interested in the function itself, and you should spy it.

Does it make sense ?

azasypkin: > I don't know, but you could maybe spy the function and get its result ?

Well, it's a tiny anonymous event handler that returns Promise-based result :) Okay it's not really an ordinary use case. But agree with the idea in general, thanks!

julienw: I don't have a better idea ;)

apps/sms/views/shared/js/system_message_handler.js
steveck-chung: Looks good!

apps/sms/views/shared/test/unit/utils_test.js
```diff
@@ -1155,27 +1155,27 @@ suite('Utils', function() {
 
   suite('Utils.throttle', function() {
     setup(function() {
-      // Starting fake timers clock at Unix epoch (timestamp of 0) would
```
steveck-chung: Sorry that I didn't check the tailing whitespace carefully... :/

azasypkin: Ah, np! It's hard, WebStorm does it for me automagically :p

apps/sms/views/shared/test/unit/activity_handler_test.js
```diff
@@ -45,71 +26,41 @@ require('/views/shared/test/unit/mock_utils.js');
...
-    navigator.mozApps = MockNavigatormozApps;
-
-    realMozL10n = navigator.mozL10n;
-    navigator.mozL10n = MockL10n;
-
-    Object.defineProperty(document, 'hidden', {
-      configurable: true,
-      get: function() {
-        return isDocumentHidden;
+    var mockedNavigatorKeys = navigator[Symbol.for('mocks')] = [];
```
steveck-chung: Why do we need `mockedNavigatorKeys`?  Seems like you only need either `navigator[Symbol.for('mocks')]` or `mockedNavigatorKeys`(in global)

apps/sms/views/shared/test/unit/activity_handler_test.js
steveck-chung: I'm fine with the changes since the original navigator mocks did a lot of similar action repeatedly and we can rewire it in simpler way apparently. Just a question that you didn't preserve the original navigator API(and no way to resume obviously). I think for the most of the case it should work since almost all the navigatior API will not be accessible in the test, and it seems meaningless to resume while suiteTeardown. So just want to confirm that you did this for specific reason.

apps/sms/views/shared/test/unit/system_message_handler_test.js
```diff
@@ -0,0 +1,778 @@
...
+    });
+
+    test('nothing is done when notification is removed', function(done) {
+      SystemMessageHandler.onNotificationSystemMessage({
+        // When notification is removed "clicked" property is "false".
+        clicked: false,
+        data: { id: smsMessage.id, threadId: smsMessage.threadId }
+      }).then(() => {
+        sinon.assert.notCalled(appStub.launch);
+        sinon.assert.notCalled(Navigation.toPanel);
```
steveck-chung: very, very optional nit: Since you would like to insert a newline in front of the `Navigation.toPanel` assertion, maybe you would like to sync the convention here?

apps/sms/views/shared/test/unit/activity_handler_test.js
azasypkin: Heh, I just declared it for the usage inside loop below :) But agree, I can use ```navigator[Symbol.for('mocks')].push(key)``` directly, since it should not affect anything - it's just test code :)

apps/sms/views/shared/test/unit/activity_handler_test.js
azasypkin: You mean we don't need to recover original navigator based API so we don't need ```navigator[Symbol.for('mocks')].forEach((key) => delete navigator[key]);``` in ```suiteTeardown``` at all (and then ```Symbol.for('mocks')``` magic as well :))?

If so, then I don't have any particular reason to do so, I just had a mixed feelings about it and if you think that we don't need to recover Navigator APIs in ```suiteTeardown``` I tend to agree and happily will simplify this piece of code :)

apps/sms/views/shared/test/unit/system_message_handler_test.js
azasypkin: Totally! Will fix it :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=bc7f24aeb016097b86e32daf6f44fb9922e096df

PR #32141, opened 2015-09-30
Bug 1210151 - [email] Prevent pasting of rich text in contenteditable fields

jrburke:
editor_mixins is used both by compose and settings_signature, the two contenteditables in email, so this catches both cases: they both call _bindEditor to set up the editor functionality.

Tested with picture, rich text, and plain text and they work as expected.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=204a4b0c55eafbb20dfaa233fbbf2579a8f81915

PR #32152, opened 2015-10-01
Bug 1205325 - fix JS error from |getScreenshot| and use card type 'Application' for bookmark. r=rexboy,yifan.

schien:
This patch contains following fix:
1. can add card for new web page on home screen
2. can open page in new tab while browser app is launched already.
3. can filter by "website" category on home screen
4. can open page in new tab while browser app is not launched in the first place.

### DIFF
--- a/tv_apps/browser/js/awesomescreen.js
+++ b/tv_apps/browser/js/awesomescreen.js
@@ -1792,13 +1792,13 @@ var Awesomescreen = {
         }
         if(this.checkBox.classList.contains('true')){
           title = this.inputArea.value;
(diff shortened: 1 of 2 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=34b2bcf8a6228562334f2178cd9fd4ba8e7e8e97

rexboy7:
Looks good to me but I found the screenshot size made is not big enough. Although the size of screenshot is set to 658 x 336px but the screenshot we got actually is 411 x 210px. I tried to do a little trace but didn't find the reason. We may need to keep track of this issue.

PR #32166, opened 2015-10-01
Bug 1209860 - Clear the gecko metrics data after merging the data to …

tamarahills:
…avoid duplication.

### DIFF
--- a/apps/system/js/advanced_telemetry.js
+++ b/apps/system/js/advanced_telemetry.js
@@ -267,6 +267,7 @@
       // If it's a merge request mark it as completed.
       if (self.merge) {
         self.merge = false;
+        self.clearPayload(false);
       }
       // if it's a request to transmit, transmit the merged metrics.
       else if (self.collecting && navigator.onLine) {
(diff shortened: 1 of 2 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=809a51d26c1d4b92aa5f4400d8b8a6c63262ce68

apps/system/js/advanced_telemetry.js
```diff
@@ -267,6 +267,7 @@
       // If it's a merge request mark it as completed.
       if (self.merge) {
         self.merge = false;
```
russnicoletti: From my reading of the code, `this.merge` is not needed. `handleGeckoMessage` always merges ("We always need to merge the Metrics when we get this." I don't anywhere where `this.merge` is used to make a decision, other than to set it to false if it's true.

tamarahills: So, a different action is taken based on whether it was a merge or a transmit timeout.  If it was a merge timeout, we just need to merge/save to asynch/clear gecko metrics.  If it was a transmit timeout, we need to merge/transmit/delete asynch/clear gecko metrics.

The mergeTimeStart gets set when we ask gecko for the payload.  Are you suggesting to move that to after the mergeMetrics completes?

PR #32179, opened 2015-10-01
Bug 1207792 - make video and poster positioning size-independent

davidflanagan:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b1a3f7257336f99f877d045e283034fc86e649d4

shared/js/media/video_player.js
```diff
@@ -396,111 +399,83 @@ function VideoPlayer(container) {
...
+    // The video size doesn't matter, so let's pick an arbitrary width
+    // to make these computations easier.
+    var dummyVideoWidth = 1080;
+    var dummyVideoHeight = dummyVideoWidth / aspectRatio;
+    var scale = Math.min(containerWidth / dummyVideoWidth,
+                         containerHeight / dummyVideoHeight);
+
+    // These are the maximum playback dimensions
+    playbackWidth = Math.round(dummyVideoWidth * scale);
+    playbackHeight = Math.round(dummyVideoHeight * scale);
```
aosmond: Love the size independent scaling of the poster/video elements; never struck my mind. I think the logic above can be simplified to avoid division and reduce the number (and at least to be me, easier to understand what's going on):

    // If the container aspect ratio is larger than the poster/video
    // aspect ratio, then the playback size is constrained by the height
    // in order to display the entire content without clipping. If it is
    // smaller, then it is constrained by the width.
    if (containerWidth > aspectRatio * containerHeight) {
      playbackWidth = Math.round(containerHeight * aspectRatio);
      playbackHeight = containerHeight;
    } else {
      playbackWidth = containerWidth;
      playbackHeight = Math.round(containerWidth / aspectRatio);
    }

davidflanagan: Thanks for this. I started trying to do something like what you've got here, then got myself confused and wasn't positive it work, so I went with the scale code that I already new and trusted. But I agree that this is better, and I've convinced myself that it works, so I'll make this change.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=770c47a287d5211f0c3ba034b47916defe741d41

PR #32196, opened 2015-10-02
Bug 1207094 - [Messages] Disable appropriate controls when in low storage condition, r=julienw azasypkin

steveck-chung:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c58f68491b5831ef7ca34c1ebcf04e09ea8ca88d

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7f61876e0e2062c2998301ea990445db067c5801

apps/sms/services/js/bridge_service_mixin.js
```diff
@@ -72,6 +72,11 @@
         throw new Error('Event "' + eventName + '" is not allowed!');
       }
 
+      if (data.appInstanceId) {
```
steveck-chung: Move the broadcast de-dup workaround to this patch

apps/sms/services/js/device_storage/device_storage_client.js
```diff
@@ -0,0 +1,71 @@
...
+     * @param {string} iccId Id that could map to the target SIM.
+     * @returns {Promise} Promise of shim's method that returned when sim
+     *  switching is ready or failed.
+     */
+    on(event, func) {
+      LOW_STORAGE_ENABLE &&
+        this[priv.client].on(`${event}-${this[priv.appInstanceId]}`, func);
+    },
+
+    lowDiskSpace: LOW_STORAGE_ENABLE && navigator.getDeviceStorage('apps') &&
```
steveck-chung: It was defined as promise that calling the API in shim originally, but I realized that lowDiskSpace checking could happen earlier than service init ready. Even we could still do with promise return and pending the click action in inbox view, it will also make the checking in compose view much more complex. So I made an exception that called the API directly in client(I know it looks bad)... But any good suggestion is welcome here.

apps/sms/views/conversation/js/conversation.js
```diff
@@ -596,7 +597,11 @@ var ConversationView = {
       var emailThread = Settings.supportEmailRecipient &&
         this.activeThread.participants.some(Utils.isEmailAddress);
 
-      Compose.setupLock({ forceType: () => emailThread ? 'mms' : null });
+      Compose.setupLock({
```
steveck-chung: I simply reuse the setupLock to make the changes as less as possible, but it might be difficult if we lowDiskSpace is returned as async.

apps/sms/views/conversation/js/conversation.js
```diff
@@ -3015,6 +3022,37 @@ var ConversationView = {
     threadMessagesClass.remove('editable-select-mode');
     node.focus();
     window.getSelection().selectAllChildren(node);
+  },
+
+  initLowStorageHandling: function() {
```
steveck-chung: There functions are reused between 2 views, maybe I  should do a device storage mixin for view(or even replace the device storage client/service)?

apps/sms/views/shared/style/sms.css
```diff
@@ -261,3 +261,8 @@ html[dir="rtl"] .contact-header aside.pack-end {
   overflow: hidden;
   text-overflow: ellipsis;
 }
+
+/* Only change the opacity low storage disabled element */
+.low-storage-disabled {
```
steveck-chung: I tried to add disable attribute to the button when low storage, but we might still need the button click-able for dialog and overriding the pointer-events css will need !important because the element might be a gaia-component . So I simply apply an reduced opacity here, but the style will be slightly different then real disabled.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4fb25ae44bbd1385b171f99350d0c5d90532c28f

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4f02735963c0742aca7f4247ac1fd44888e61658

apps/sms/services/js/device_storage/device_storage_shim.js
```diff
@@ -0,0 +1,59 @@
...
+      return;
+    }
+
+    deviceStorage = realDeviceStorage;
+
+    EVENTS.forEach((event) => {
+      deviceStorage.addEventListener(event, (evt) => {
+        // Workaround to avoid the duplicated events by inserting
+        // appInstanceId into event name in bridge mixin.
+        evt.appInstanceId = appInstanceId;
```
azasypkin: Sorry, I'm not sure I understand why we need this, could you please clarify it a bit? I thought that using app-instance-bound-broadcast-channel is enough....

apps/sms/services/js/device_storage/device_storage_shim.js
```diff
@@ -0,0 +1,59 @@
...
+    return deviceStorage.lowDiskSpace;
+  }
+};
+
+exports.DeviceStorageShim = Object.seal(
+  BridgeServiceMixin.mixin(
+    DeviceStorageShim,
+    SERVICE_NAME,
+    { events: EVENTS,
+      methods: METHODS }
```
azasypkin: __nit:__ a bit unusual braces alignment :) maybe ```{ events: EVENTS, methods: METHODS }``` or
```
{
  events: EVENTS,
  methods: METHODS
}
```

apps/sms/services/js/device_storage/device_storage_client.js
```diff
@@ -0,0 +1,70 @@
...
+  'use strict';
+
+  /**
+   * Name of the service that is responsible for device storage functionality.
+   * @type {string}
+   */
+  const SERVICE_NAME = 'device-storage-shim';
+
+  // TODO: Remove the flag once the low storage feature is completed.
+  const LOW_STORAGE_ENABLE = false;
```
azasypkin: __nit:__ agree with Julien that it would be better to keep this in Settings (if possible).

apps/sms/services/js/device_storage/device_storage_client.js
```diff
@@ -0,0 +1,70 @@
...
+      });
+    },
+
+    /**
+     * Bypass the event handler function to service with specific appInstanceId.
+     * @param {string} event Event name.
+     * @param {Function} func Handler function corresponding to event.
+     */
+    on(event, func) {
+      LOW_STORAGE_ENABLE &&
```
azasypkin: __question:__ same question here, isn't app-instance-bc enough?

apps/sms/services/js/device_storage/device_storage_client.js
```diff
@@ -0,0 +1,70 @@
...
+        endpoint: new BroadcastChannel(broadcastChannelName)
+      });
+    },
+
+    /**
+     * Bypass the event handler function to service with specific appInstanceId.
+     * @param {string} event Event name.
+     * @param {Function} func Handler function corresponding to event.
+     */
+    on(event, func) {
```
azasypkin: __optional nit:__ honestly I'd rather use ```EventDispatcher``` here to have all these on/once/off/allowedEvents stuf instead of directly delegating this to bridge client (like in ActivityClient), but it's arguable of course.

apps/sms/views/shared/style/sms.css
azasypkin: __nit:__ I think if we finally decide to do some changes via CSS in low storage condition maybe we can have smth like single ```low-storage``` css class applied to the body instead and use it to style particular elements?

__question:__ wondering if ```NewMessageView.beforeEnter``` can check low storage condition instead, show dialog and cancel navigation/disable controls (I assume we can do that if we return rejected promise in before enter)?

Or it doesn't work for all NewMessageView-related cases?

I also haven't look into the spec yet, but is it UX requirement to kind of disable Composer button instead of disabling controls in New Message View?

apps/sms/services/js/device_storage/device_storage_client.js
azasypkin: Even if we access MozAPI here directly I'd still have _promise based method_ to set interface in stone. At least in Conversation View it seems OK as both ```beforeEnterComposer``` and ```beforeEnterThread``` are promise-based. Also if we decide to go with service-in-window on startup - it will be easier to migrate. But it's IMO.

apps/sms/views/conversation/js/conversation.js
azasypkin: __thinking aloud:__ I don't know why exactly yet :), but I feel that ```App.js``` can work with ```DeviceStorageClient``` instead (or even replace it, not sure on this one), as it seems app specific thing. Then app.js can append specific CSS class to the body as it does with ```.js-app-ready``` (in this case we can remove similar code from Inbox entirely) and broadcast event if anyone else needs more work (like ConversationView here).

Just thought.

apps/sms/services/js/device_storage/device_storage_shim.js
steveck-chung: I think you are right that we don't need the workaround here. I was confused by the message event broadcasting case, but it's because the messaging client and service was sharing with same channel. For the storage client - shim case, it already had unique channel and we don't need to worry about receiving the event from other instance.

apps/sms/services/js/device_storage/device_storage_client.js
steveck-chung: I think it's good timing to think about this case. I used this in the message event because we don't need to do any further logic for handling the returns from service. But we can discuss about this for sure.

PR #32207, opened 2015-10-02
Bug 1205042 - Tracking Protection FTE Popup r=mhenretty

albertopq:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e414cb9a6abe5fe062c94c6f4e4dd4b4704f5fc1

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=307f8e6a2bbfd58f28e4d9f57fae3f59ad2ab113

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d478cc2d0538ddc5e50192d5d8af53ed0dcfea09

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b38b1a3cff47ebba7e63b9e6de34b104ab132ca2

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=24aa72190f63b49176f64735e52217989de51418

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0455d14a93765a7b737b4690b7b59c55a23138e5

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d407322493c4ae2ca83db9516a42739fb3a08cfe

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=70d343c16c9f395234f468c7544014b1f6a51474

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2209c3347882542d07416a600c43d91902a79157

apps/bookmark/test/marionette/client_options.js
```diff
@@ -2,6 +2,7 @@
 module.exports = {
   settings: {
     'devtools.qps.enabled': false,
-    'language.current': 'en-US'
+    'language.current': 'en-US',
+    'privacy.trackingprotection.shown': true
```
mikehenrty: There's probably a way to make `privacy.trackingprotection.shown' default to true for integration tests. I think we have to figure out the right way to modify [1]. This way, we don't have to modify so many files, and future system app browser tests don't need to worry about this pop-up. Let's figure out how to do this.

1.) https://github.com/mozilla-b2g/gaia/blob/master/tests/jsmarionette/runner/mozilla-profile-builder/test/fixtures/b2g-profile/settings.json

apps/system/index.html
```diff
@@ -344,6 +346,8 @@
       <!-- XXX: Before we can make LockScreen as a iframe or app, we need this to split LockScreen and System part in this way (Bug 1057198) -->
       <div id="lockscreen-frame-placeholder"></div>
 
+      <div id="fullscreen-overlay" data-z-index-level="fullscreen-overlay"></div>
+
```
mikehenrty: I don't like the idea that we are adding something to DOM that won't be used for most of the life of the phone. Let's move this overlay and it's associated styles completely into the Notice module.

apps/system/js/tracking_notice.js
```diff
@@ -0,0 +1,152 @@
...
+      self.confirmNotice();
+    });
+  };
+
+  TrackingNotice.prototype._fetchElements = function spl_initElements() {
+    this.element = document.querySelector('#' + this.instanceID);
+    this.confirm = this.element.querySelector('#tracking-notice-confirm');
+    this.setting = this.element.querySelector('gaia-switch');
+    this.description = this.element.querySelector('#tracking-notice-copy');
+    this.description.addEventListener('click', function() {
```
mikehenrty: It's pretty surprising to click anywhere on the description and have it navigate the underlying browser window. The biggest change I want here is to make only clicking on the link perform the action. Secondly, I would prefer that it would open a new window rather than navigate my already open one. But that second one I won't block on.

apps/system/js/tracking_notice.js
```diff
@@ -0,0 +1,152 @@
...
+   * @param {options} object for attributes `onShow`, `onHide` callback.
+   * @extends SystemDialog
+   */
+  var TrackingNotice = function(controller) {
+    this.instanceID = 'tracking-notice';
+    this.controller = controller || {};
+    this.options = {};
+    this._settings = navigator.mozSettings;
+    var req = navigator.mozSettings.createLock().get(TRACKING_NOTICE_KEY);
+    req.onsuccess = (function() {
```
mikehenrty: nit: use fat arrow

apps/system/js/tracking_notice.js
```diff
@@ -0,0 +1,152 @@
...
+      var alreadyShown = req.result[TRACKING_NOTICE_KEY];
+      this._alreadyShown = alreadyShown;
+      if (!alreadyShown) {
+        this.renderDialog();
+      }
+    }.bind(this));
+
+    req.onerror = (function() {
```
mikehenrty: nit: fat arrow

apps/system/js/tracking_notice.js
```diff
@@ -0,0 +1,152 @@
...
+
+  TrackingNotice.prototype.DEBUG = false;
+
+  TrackingNotice.prototype.name = 'TrackingNotice';
+
+  TrackingNotice.prototype.containerElement =
+    document.getElementById('fullscreen-overlay');
+
+  TrackingNotice.prototype.renderDialog = function() {
+    LazyLoader.load('shared/elements/gaia_switch/script.js', function() {
```
mikehenrty: nit: fat arrow

apps/system/style/tracking_notice/tracking_notice.css
```diff
@@ -0,0 +1,182 @@
...
+  background-image: url("images/tp_screen.png");
+  background-repeat: no-repeat;
+}
+
+#tracking-notice .setting {
+  position: relative;
+  top: .6rem;
+}
+
+#tracking-notice .setting > * {
```
mikehenrty: extremely slow selector. using * as the key selector causes the css engine to grab a list of all elements on the page, iterate over them and check if their parent structure matches the parent selectors. can we be more explicit about what elements we are applying position: absolute to?

apps/system/style/tracking_notice/tracking_notice.css
```diff
@@ -0,0 +1,182 @@
...
+  pointer-events: all;
+  visibility: hidden;
+  opacity: 0;
+  transition: opacity 300ms, visibility 0s 300ms
+}
+
+#tracking-notice:not(.hidden) {
+  animation-name: gaia-dialog-entrance;
+  animation-duration: 300ms;
+  animation-timing-function: cubic-bezier(0.175, 0.885, 0.320, 1.275);
```
mikehenrty: fancy!

apps/system/locales/system.en-US.properties
```diff
@@ -609,6 +609,24 @@ copy-link=Copy Link
...
 
+#Tracking Protection Notice
+
+# LOCALIZATION NOTE (tracking-notice-header): This string is used as
+# title of the dialog that explains this new feature to new users.
+tracking-notice-header=New! Tracking Protection
+# LOCALIZATION NOTE (tracking-notice-description): This string is used as
+# introduction to the new feature. Link to the feature description in mdn.
+tracking-notice-description=This new feature protects your browsing activity against potential trackers. <a class="learnMore"> Learn More </a>
+tracking-notice-learn-more-url=https://support.mozilla.org/en-US/products/firefox-os/privacy-and-security
```
mikehenrty: Let's not put the url in the localization file. If we just use the `https://support.mozilla.org/products/firefox-os/privacy-and-security`, mozilla's servers we redirect to the right place. Let's put the href directly in the markup.

apps/system/js/app_core.js
```diff
@@ -59,7 +59,8 @@
           'ActivityWindowManager',
           'PermissionManager',
           'Rocketbar',
-          'ActivityHandler'
+          'ActivityHandler',
+          'TrackingNotice'
```
mikehenrty: Did you try having TrackingNotice lazy loaded by the browser? It makes me sad we are putting TrackingNotice on the same level as thinks like Rocketbar and Activity handler. Better would be to have Browser check the `privacy.trackingprotection.shown` setting, and only lazyload the tracking notice module if necessary. But I realize that is asking a lot for such a short deadline, so I won't block the review on that.

apps/system/js/tracking_notice.js
```diff
@@ -0,0 +1,152 @@
...
+      this.render();
+      this.boundShowIfNeeded = this.showIfNeeded.bind(this);
```
mikehenrty: It seems weird to me that we create this bound function after we call render, yet use this bound function inside of `_fetchElements` (albeit asynchronously). I think instead of using a "saved" bound function like this, we should just do `window.addEventListener('appopened', this);` and use `handleEvent`. That way you can just `removeEventListener(..., this)` from anywhere and don't have to create a handle for a bound function.

apps/system/js/tracking_notice.js
```diff
@@ -0,0 +1,152 @@
...
+      navigator.mozL10n.formatValue(TRACKING_NOTICE_LEARN_MORE_URL)
+      .then((url) => {
+        var activeApp = Service.query('AppWindowManager.getActiveWindow');
+        this.hide();
+        activeApp.navigate(url);
+        var element = activeApp.element;
+        // When the user clicks 'Learn More', go to that URL and display
+        // back the dialog when navigates to any other page.
+        element.addEventListener('_locationchange', function onLoadLearnMore() {
+          element.removeEventListener('_locationchange', onLoadLearnMore);
```
mikehenrty: This `removeEventListener` isn't working since you are listening with a bound function. You will either need to name the bound function with a `var onLoadLearnMore = function9) {..}.bind(this);` or save off `var self = this`.

apps/system/js/tracking_notice.js
```diff
@@ -0,0 +1,152 @@
...
+    });
+  };
+
+  TrackingNotice.prototype._fetchElements = function spl_initElements() {
+    this.element = document.querySelector('#' + this.instanceID);
+    this.confirm = this.element.querySelector('#tracking-notice-confirm');
+    this.setting = this.element.querySelector('gaia-switch');
+    this.description = this.element.querySelector('#tracking-notice-copy');
+    this.description.addEventListener('click', function() {
+      navigator.mozL10n.formatValue(TRACKING_NOTICE_LEARN_MORE_URL)
```
mikehenrty: Again, let's use Mozilla's servers to do this. Let's just put an 'a' tag in the markup that points to `https://support.mozilla.org/products/firefox-os/privacy-and-security`

apps/system/js/tracking_notice.js
```diff
@@ -0,0 +1,152 @@
...
+  };
+
+  TrackingNotice.prototype.confirmNotice = function() {
+    this._settings.createLock().set({
+      'privacy.trackingprotection.enabled': this.setting.checked,
+      'privacy.trackingprotection.shown': true
+    });
+    window.removeEventListener('appopened', this.boundShowIfNeeded);
+    window.removeEventListener('_locationchange', this.boundShowIfNeeded);
+    this.hide();
```
mikehenrty: maybe we should `destroy` here?

apps/system/js/tracking_notice.js
mikehenrty: also nit: fat arrow

apps/system/style/tracking_notice/tracking_notice.css
mikehenrty: just spoke with :bz about this in #content. so, in practice this might not be too slow because we use ancestor filters to optimize it. but wildcards as the right most selector will cause all things under both .setting, and #tracking-notice to be restyled on every match, and so if either of these change a lot, we are adding a lot to the style matching work. again, this might not be noticeable to the user, but if we can avoid it we should.

apps/bookmark/test/marionette/client_options.js
albertopq: That makes sense! Thanks!

apps/system/index.html
albertopq: The thing is that this dialog is a little bit different. It needs to be in top of the statusbar, but not covering the SHB. Not other overlay has this properties. I tried to put it in several places, but didn't find a way to accommodate the needs of this dialog. So this div would serve as a container to any other dialog with this properties (in the same way the dialog-overlay works)

apps/system/js/tracking_notice.js
albertopq: Sorry, I bound the function afterwards, and forgot about it. Good catch!

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=49f75a83529cd692dc25e78691219efc3c335a9d

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d283811ced1bbf7b33ed9d0ee865cdacb05a7154

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1c309d0477a5f7909aa4f97bb768bd61a9d96a28

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=610753317e7f672c6f45b7c4e5ed043dfb1e2c7a

apps/system/js/system_dialog_manager.js
```diff
@@ -318,6 +320,26 @@
...
+      }
+    };
+
+    req.onerror = () => {
+      this._includeTrackingNotice();
+    };
+  };
+
+  SystemDialogManager.prototype._includeTrackingNotice = function() {
+    LazyLoader.load('/js/tracking_notice.js').then(()  => {
```
mikehenrty: :+1:

apps/system/js/system_dialog_manager.js
```diff
@@ -318,6 +320,26 @@
...
+      this.trackingNotice = new TrackingNotice();
```
mikehenrty: nit: we should pass `SystemDialogManager` into `TrackingNotice`
```
this.trackingNotice = new TrackingNotice(SystemDialogManager);
```

apps/system/js/tracking_notice.js
```diff
@@ -0,0 +1,156 @@
...
+   * @class TrackingNotice
+   * @param {options} object for attributes `onShow`, `onHide` callback.
+   * @extends SystemDialog
+   */
+  var TrackingNotice = function(controller) {
+    this.instanceID = 'tracking-notice';
+    this.controller = controller || {};
+    this.options = {};
+    this._settings = navigator.mozSettings;
+    var req = this._settings.createLock().get(TRACKING_NOTICE_KEY);
```
mikehenrty: Do we need to read the setting again here? It feels redundant to me. If we are instantiating the `TrackingNotice`, we know that it hasn't been shown yet. What's to prevent us from calling `TrackingNotice.show()` directly from `SystemDialogManager`?

apps/system/js/tracking_notice.js
```diff
@@ -0,0 +1,156 @@
...
+    });
+    window.addEventListener('appopened', this.boundShowIfNeeded);
+  };
+
+  TrackingNotice.prototype._fetchElements = function spl_initElements() {
+    this.element = document.querySelector('#' + this.instanceID);
+    this.confirm = this.element.querySelector('#tracking-notice-confirm');
+    this.setting = this.element.querySelector('gaia-switch');
+    this.learnMore = this.element.querySelector('#tracking-notice-learn-more');
+    var boundShowIfNeeded = this.showIfNeeded.bind(this);
```
mikehenrty: I *really* don't like having this bound function when the convention is just to use 
```
 element.addEventListener('_locationchange', this);
```
and `handleEvent`. I won't block the review on it, but I suggest this is a cleaner way to go.

albertopq: Sorry! I forgot this chgange!

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6103d7668cb6804ee716eec5080181c188242be7

PR #32223, opened 2015-10-04
Bug 1204894 - Check for entry_point as well as manifestURL in wait_to_be_displayed/wait_to_not_be_displayed

mermi:
I am not sure but what I did; 
I tried something simple and basic as a first test but I think this is not the optimal solution


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5ca7e4d3c006113191f6fb572fb7740336e5c078

PR #32238, opened 2015-10-05
Bug 1210737 - Return metadata progressively. r=gmarty

Cwiiis:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1252491abdb8a0bfd532715a8ae0330fc7a7c649

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=62871c43ce3272238aa485b4036edee02eb04ecd

apps/homescreen/js/app.js
```diff
@@ -164,141 +164,44 @@ const SETTINGS_VERSION = 0;
...
-      (e) => {
-        console.error('Failed to initialise metadata db', e);
-        resolve();
-      });
-    }).then(this.firstRun ?
+          console.error('Failed to initialise metadata db', e);
+          return Promise.resolve();
+        }).then(this.firstRun ?
         LazyLoader.load(['js/firstrun.js'],
           () => {
```
gmarty: What about the following code?
```js
}).then(!this.firstRun ? Promise.resolve() :
LazyLoader.load(['js/firstrun.js'],
```
It seems more readable to me to get the short alternative first (`Promise.resolve()`), then the longer code, like in the Promise above.
Also maybe add a comment to draw the attention to the `!this.firstRun` (different to the `this.firstRun` above).

apps/homescreen/js/app.js
```diff
@@ -307,17 +210,134 @@ const SETTINGS_VERSION = 0;
...
+        });
+      })
+    ]).then(() => {
+      // Add any applications that were missed and are in the startup metadata
+      var id;
+      var pendingIcons = this.pendingIcons;
+      this.pendingIcons = {};
+      for (id in pendingIcons) {
+        this.addAppIcon.apply(this, pendingIcons[id]);
+      }
```
gmarty: Is the code above covered by unit tests? I know the code that removes unknown entries from the DB is, so this one should be unit tested too, if that's not already the case.

apps/homescreen/js/app.js
```diff
@@ -395,24 +427,33 @@ const SETTINGS_VERSION = 0;
...
+    addAppIcon: function(appOrBookmark, entryPoint) {
+      var id = this.getIconId(appOrBookmark, entryPoint);
+      var entry = -1;
+
+      if (this.startupMetadata !== null) {
+        entry = this.startupMetadata.findIndex((element) => {
+          return element.id === id;
+        });
+        if (entry === -1) {
+          this.pendingIcons[id] = Array.slice(arguments);
```
gmarty: Was about to suggest spread operator here (i.e. `this.pendingIcons[id] = [...arguments];`), but Firefox is not up to spec on arguments iterator :-(

apps/homescreen/js/app.js
```diff
@@ -540,6 +577,12 @@ const SETTINGS_VERSION = 0;
 
         this.pageHeight = pageHeight;
         this.pendingGridHeight = gridHeight;
+
+        if (!this.visualLoadComplete &&
+            Math.floor(visibleChildren / (this.small ? 4 : 3)) * iconHeight >=
+              scrollHeight) {
+          this.onVisualLoad();
+        }
```
gmarty: I don't entirely agree here. Only the icons on the first visible page are populated, but the icons on the first row of the next page are usually visible (on a Z3C screen aspect ratio at least). We should really call `onVisualLoad()` when all the icons above the fold are visible.
Have you tried `Math.floor(visibleChildren / (this.small ? 5 : 4))`? If so does it result in significantly worse result?

apps/homescreen/js/app.js
Cwiiis: I agree, but I think it's a trade-off between having the negation and the ordering - I'd rather keep it consistent and have this slightly odd layout than have the negation and the inconsistent comparison.

apps/homescreen/js/app.js
Cwiiis: Good call, thanks for keeping on top of this :) This may be a bit difficult to test, but I'll figure it out.

apps/homescreen/js/app.js
Cwiiis: That's why I use Math.floor and scrollHeight here - this should mean that you get an extra row if it's at all visible.

gmarty: That's weird because in 4 columns configuration, the icons on the first page + the ones on the top row of next page appear at the same time.
But in 3 columns mode, I can clearly see the 3 icons of next page loading one after another after the first page icons.
Do you see that too?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=28f52f74a312845626d678f221fe68cd0600a76c

apps/homescreen/js/app.js
gmarty: Why do we need 2 separate Promises? I really find that bit difficult to read.

apps/homescreen/js/app.js
Cwiiis: That is weird... I don't see that, but I'm also not using the software home button. I say let's fix this in a follow-up, I'm pretty sure the maths checks out so this could be a slightly different problem...

apps/homescreen/js/app.js
Cwiiis: Actually, the first-run stuff could go further up instead of the Promise.resolve(), which would cut down on redundant Promise.resolves and this.firstRun comparisons... I'll see about doing that.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=bffad461249388ac72587553fdc6e858b314331b

apps/homescreen/js/app.js
Cwiiis: Ok, now I do see it and it's quite obvious - I'm going to take a look and see what it is.

Cwiiis: Weird, so onVisualLoad is getting called at the right time, but for some reason, you still see the icons load in - It could be platform not loading the image in time... There's code in gaia-site-icon that's meant to make the image fade in on load, but it seems that the image isn't actually ready to display when receiving the onload signal a lot of the time...

Cwiiis: Indeed, this is what's happening - I think we're best fixing gaia-site-icon so that the fade animation it's meant to run actually works (somehow). I'll see if there's anyone on platform that knows the heuristics of image loading/painting and signals, but let's address this in follow-up. (note, this affects homescreen even without this patch, just this patch makes it appear *almost* correct, coincidentally :))

PR #32247, opened 2015-10-05
Bug 1204894 - Check for entry_point as well as manifestURL in wait_to_be_displayed/wait_to_not_be_displayed

mwargers:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=76f1eb97800b9afa865504761fd05e428553e0dd

tests/atoms/gaia_apps.js
```diff
@@ -296,8 +297,11 @@ var GaiaApps = {
             function() {
               // wait for the displayed app to have expected manifestURL/origin
               let displayedApp = GaiaApps.getDisplayedApp();
+              console.log('*****n****'+displayedApp.entryPoint);
+              console.log('*******'+displayedApp.manifestURL + '::::' + manifestURL);
+              console.log('*******'+displayedApp.entryPoint + '::::' + entryPoint);
               if (displayedApp.manifestURL) {
```
JohanLorenzo: I'd add another check here:
```js
if (displayedApp.manifestURL && displayedApp.manifestURL) {
    return displayedApp.manifestURL == manifestURL && displayedApp.entryPoint == entryPoint;
} else if (displayedApp.manifestURL) {
    return displayedApp.manifestURL == manifestURL
} else {
```
because some apps don't have entry points.

tests/atoms/gaia_apps.js
```diff
@@ -362,13 +366,28 @@ var GaiaApps = {
...
+
     console.log('app with manifestURL \'' + app.manifestURL + '\' displayed');
     let result = {
       frame: (app.browser) ? app.browser.element : app.frame.firstChild,
       src: (app.browser) ? app.browser.element.src : app.iframe.src,
       name: app.name,
       origin: app.origin,
-      manifestURL: app.manifestURL
+      manifestURL: app.manifestURL,
+      entryPoint: currentEntryPoint
```
JohanLorenzo: Do we want entryPoint to be null when an app has no entry point at all?

tests/atoms/gaia_apps.js
mwargers: I don't think any apps  have entryPoints except contacts and dialer. In all cases displayedApp.entryPoint should be null otherwise, except contacts and dialer. And that would then be equal to entryPoint, which is also most of the times null.

tests/atoms/gaia_apps.js
mwargers: Yes, entryPoint is often null. This is also the case in the launch code.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4c41cfda79ab1f7adfd08c5be0b69d2a78271a4c

tests/atoms/gaia_apps.js
```diff
@@ -279,10 +279,12 @@ var GaiaApps = {
       };
 
       let displayedApp = GaiaApps.getDisplayedApp();
-      if (displayedApp.manifestURL == manifestURL ||
+      if ((displayedApp.manifestURL == manifestURL &&
+          displayedApp.entryPoint == entryPoint) ||
           (!displayedApp.manifestURL && displayedApp.origin == origin)) {
         console.log('app with origin \'${displayedApp.origin}\' and ' +
-          'manifestURL \'${displayedApp.manifestURL}\' is already running');
+          'manifestURL \'${displayedApp.manifestURL}\' and ' +
```
JohanLorenzo: For this line, `adb logcat | grep 'Content JS LOG'` returns 
`I/GeckoConsole( 9414): Content JS LOG: app with origin '${displayedApp.origin}' and manifestURL '${displayedApp.manifestURL}' has launched`

This line was buggy, let's change it to:
```js 
console.log("app with origin '" + displayedApp.origin"' and " +
"manifestURL '" + displayedApp.manifestURL "' and " +
"entryPoint '" + displayedApp.entryPoint+ "' is already running');
```

tests/atoms/gaia_apps.js
```diff
@@ -290,14 +292,15 @@ var GaiaApps = {
           waitFor(
             function() {
               console.log('app with origin \'${displayedApp.origin}\' and ' +
-                'manifestURL \'${displayedApp.manifestURL}\' has launched');
+                'manifestURL \'${displayedApp.manifestURL}\' and ' +
```
JohanLorenzo: Same comment here about the buggy output

tests/atoms/gaia_apps.js
mwargers: Thanks, I should have checked this code. Also need to check this with jslint, because that is being checked by treeherder as well.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2f73768dc36cf5d2c19982bd1d00cf9ac9660886

PR #32251, opened 2015-10-05
Bug 1211621 - Fix failures in tests in wait_for_cards_view

mwargers:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d124e935626d2361b2e7cbefd86f927dd9fe7124

tests/python/gaia-ui-tests/gaiatest/apps/system/regions/cards_view.py
```diff
@@ -14,7 +14,7 @@
 class CardsView(Base):
 
     # Home/Cards view locators
-    _cards_view_locator = (By.ID, 'cards-view')
+    _task_manager_locator = (By.ID, 'task-manager')
```
JohanLorenzo: The cards view is a PageRegion of the system app. I think we can rename as the root_locator. What do you think?

tests/python/gaia-ui-tests/gaiatest/apps/system/regions/cards_view.py
```diff
@@ -34,12 +34,12 @@ def cards(self):
 
     @property
     def is_cards_view_displayed(self):
```
JohanLorenzo: This can be renamed as `is_displayed`. So we can call this function this way: `cards_view.is_displayed`

tests/python/gaia-ui-tests/gaiatest/apps/system/regions/cards_view.py
mwargers: Yes, I agree. I noticed lots of these regions do need to be converted to use PageRegion like that, right?

PR #32256, opened 2015-10-05
Bug 910576 - Use TextDecoder in BlobView

jimporter:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e5207466ecc4d4f5772bb7e3e604dced78107ef9

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a449e11cb75953125ea305a0dd637618ee06e385

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5ba42d9feb2691e021c933e7a0f930de65e76066

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f32f737f245d4f4b643338ad74d9047b18e547cd

shared/js/blobview.js
```diff
@@ -260,114 +260,26 @@ var BlobView = (function() {
 
     getASCIIText: function(offset, len) {
       var bytes = new Uint8Array(this.buffer, offset + this.viewOffset, len);
-      return String.fromCharCode.apply(String, bytes);
+      var decoder = new TextDecoder('ascii');
+      return decoder.decode(bytes);
```
davidflanagan: As you found in image_size.js, this new implementation is not equivalent to the old. Arguably, the original method should have been getBinaryString().  Feel free to rename it if you want to. But I'm reluctant to just change this method. Is there any reason that TextDecoder('ascii') is actually needed here? I'd guess that String.fromCharCode() is actually the faster option here, and would recommend that you just don't change this at all.

shared/js/blobview.js
```diff
@@ -260,114 +260,26 @@ var BlobView = (function() {
...
     },
 
     readUTF8Text: function(len) {
-      try {
-        return this.getUTF8Text(this.index, len);
-      }
-      finally {
-        this.index += len;
-      }
+      var s = this.getUTF8Text(this.index, len);
```
davidflanagan: Are you sure that getUTF8Text() will never throw an exception now? If so it is fine. If not I'd recommend that you keep the try/finally structure to ensure that the index gets incremented even on failure. That try/finally looks like something that got added to fix a bug...

shared/js/blobview.js
jimporter: Yes, there is. The ID3 (v1 and v2) parsers expect Latin-1 strings, which is one of the more-common places this functin is used. For any non-ASCII Latin-1 character, we'd want to convert it to the correct Unicode code point so we don't get mojibake.

`TextDecoder` is also about 20% faster (at least on Desktop).

shared/js/blobview.js
jimporter: It might still throw, but that behavior was wrong.

I'd expect this function to follow the *strong exception safety* guarantee; namely that, if an exception is thrown, the function has no side effects. I'll double check to make sure that this wasn't added to fix a bug somewhere else, and if it was, I'll move the fix to wherever the problem really was.

jimporter: Upon further examination, that code has been there since BlobView was created, so it probably wasn't in response to a bug someone found (unless you found it in your earliest tests of the music app's metadata parser).

shared/js/blobview.js
```diff
@@ -376,30 +288,31 @@ var BlobView = (function() {
...
+        bytes = bytes.subarray(2);
+
         if (BOM === 0xFEFF) {
           le = false;
-        }
-        else {
+        } else if (BOM === 0xFFFE) {
           le = true;
+        } else {
+          fail('expected BOM');
```
davidflanagan: You've added a new failure case, so this isn't the same as the old implementation. Could be a source of bugs.  Probably better would be to use le=true in this case, but not do the subarray, so the first two bytes don't get truncated.

What would be best is is TextDecoder supports "utf-16" and can handle the BOM automatically. Then all you have to do is pick the encoding based on the le argument.

Or, another option is to just not bother with TextDecoder here, since this is infrequently used and probably pretty fast, and it can't really fail the way it is written.

shared/js/blobview.js
```diff
@@ -420,33 +333,36 @@ var BlobView = (function() {
...
-      var s = '';
-      for (var i = 0; i < size; i++) {
-        var charcode = this.view.getUint8(this.index + i);
-        if (charcode === 0) {
-          i++;
-          break;
-        }
-        s += String.fromCharCode(charcode);
+    // more than size bytes.  And return as an ASCII string
+    readNullTerminatedASCIIText: function(size) {
```
davidflanagan: I can't remember if there was any reason for ASCII vs Latin1. Looking at it now, it seems to me that "BinaryString" would be more accurate in both cases.  To avoid possible changes in semantics I'd recommend sticking with the String.fromCharCode() implementation. But if you want to rename the method to use BinaryString that's okay with me.

shared/js/blobview.js
```diff
@@ -420,33 +333,36 @@ var BlobView = (function() {
...
+
+      var nil = bytes.indexOf(0);
+      if (nil !== -1) {
+        bytes = bytes.subarray(0, nil);
       }
-      this.index += i;
+
+      var decoder = new TextDecoder('ascii');
+      var s = decoder.decode(bytes);
+      this.index += (nil !== -1) ? nil + 1 : size;
```
davidflanagan: It would be easier to just use `this.index += bytes.length;` here

shared/js/blobview.js
```diff
@@ -420,33 +333,36 @@ var BlobView = (function() {
...
-        var charcode = this.view.getUint8(this.index + i);
-        if (charcode === 0) {
-          i++;
-          break;
-        }
-        s += String.fromCharCode(charcode);
+    // more than size bytes.  And return as an ASCII string
+    readNullTerminatedASCIIText: function(size) {
+      var bytes = new Uint8Array(this.buffer, this.viewOffset + this.index,
+                                 size);
```
davidflanagan: What happens if size extends beyond the end of the buffer? Do both implementations fail in that case?

shared/js/blobview.js
```diff
@@ -420,33 +333,36 @@ var BlobView = (function() {
...
-      for (var len = 0; len < size; len++) {
-        if (this.view.getUint8(this.index + len) === 0) {
-          break;
-        }
-      }
-      var s = this.readUTF8Text(len);
-      if (len < size) {    // skip the null terminator if we found one
-        this.advance(1);
+      var bytes = new Uint8Array(this.buffer, this.viewOffset + this.index,
+                                 size);
```
davidflanagan: Same question about failure modes: do both implementations fail if size extends beyond the end of the buffer (and fi there is no null before then?)

shared/js/blobview.js
davidflanagan: Nevermind. Your code is correctly. bytes.length doesn't include the nul-terminator. Sorry.

shared/js/blobview.js
jimporter: Yes. `Uint8Array` will throw if `size` is too big.

shared/js/blobview.js
```diff
@@ -457,26 +373,15 @@ var BlobView = (function() {
...
+      for (var len = 0; len < size; len += 2) {
+        if (this.getUint16(this.index + len, le) === 0) {
           break;
         }
-        s += String.fromCharCode(charcode);
       }
-      this.index += i;
+
+      var s = this.getUTF16Text(this.index, len, le);
+      this.index += len + (len < size) ? 2 : 0;
```
davidflanagan: the RHS is a little hard to read. How about `(len < size) ? len+2 : len;`

shared/js/blobview.js
```diff
@@ -457,26 +373,15 @@ var BlobView = (function() {
...
-          i += 2;
+      for (var len = 0; len < size; len += 2) {
+        if (this.getUint16(this.index + len, le) === 0) {
           break;
         }
-        s += String.fromCharCode(charcode);
       }
-      this.index += i;
+
+      var s = this.getUTF16Text(this.index, len, le);
```
davidflanagan: The code is simpler but we now have to loop through the bytes twice instead of once as before. I'm guessing that the performance is still fine, but wanted to make sure that you noticed you'd doubled the number of array reads here.

shared/js/media/image_size.js
```diff
@@ -29,7 +29,9 @@ function getImageSize(blob, callback, error) {
       error('corrupt image file');
       return;
     }
-    var magic = data.getASCIIText(0, 8);
+
+    var bytes = data.getUnsignedByteArray(0, 8);
+    var magic = String.fromCharCode.apply(String, bytes);
```
davidflanagan: I'd prefer it if you didn't have to make the change.  I'm okay with renaming the method, but I'd rather not change the semantics of the method.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1703c57d190bb294c724d2bc072e1fd5665f0b42

apps/music/js/metadata/vorbis_picture.js
```diff
@@ -15,29 +15,27 @@ var VorbisPictureComment = (function() {
...
     }
 
-    // We ignore these next two fields
-    len = view.readUnsignedInt();
-    view.readNullTerminatedUTF8Text(len); // desc
-    // skip w, h, bpp and indexed colours.
-    view.advance(16);
+    // Skip desc, w, h, bpp and indexed colours.
+    var desc_len = view.readUnsignedInt();
+    view.advance(desc_len + 16);
```
davidflanagan: This is a potential change in behavior. Is the string actually null-terminated? If so, the old code does not necessarily advance by desc_len bytes.  I don't know what the spec says, but it looks to me like there might have been a bug before, and it is gone now.  But you might double-check this.

shared/js/blobview.js
```diff
@@ -420,63 +350,53 @@ var BlobView = (function() {
...
-    readNullTerminatedLatin1Text: function(size) {
-      var s = '';
-      for (var i = 0; i < size; i++) {
-        var charcode = this.view.getUint8(this.index + i);
-        if (charcode === 0) {
-          i++;
-          break;
-        }
-        s += String.fromCharCode(charcode);
+    // more than size bytes.  And return as an ASCII string
```
davidflanagan: Should be Latin1 in the comment, right?

shared/js/blobview.js
```diff
@@ -420,63 +350,53 @@ var BlobView = (function() {
...
+    readNullTerminatedLatin1Text: function(size, advance_by_size = false) {
```
davidflanagan: Please add a comment documenting this new optional argument you've added.

shared/js/blobview.js
```diff
@@ -420,63 +350,53 @@ var BlobView = (function() {
...
+                                 size);
+
+      var nil = bytes.indexOf(0);
+      if (nil !== -1) {
+        bytes = bytes.subarray(0, nil);
       }
-      this.index += i;
+
+      var s = getDecoder('latin1').decode(bytes);
+      this.index += (nil !== -1 && !advance_by_size) ? nil + 1 : size;
```
davidflanagan: I'd prefer just a plain old if/else here. This is pretty hard to read.

shared/js/blobview.js
```diff
@@ -420,63 +350,53 @@ var BlobView = (function() {
...
+      var bytes = new Uint8Array(this.buffer, this.viewOffset + this.index,
+                                 size);
+
+      var nil = bytes.indexOf(0);
+      if (nil !== -1) {
+        bytes = bytes.subarray(0, nil);
       }
+
+      var s = getDecoder('utf-8').decode(bytes);
+      this.index += (nil !== -1 && !advance_by_size) ? nil + 1 : size;
```
davidflanagan: Same comments here about if/else and documenting the new argument.

davidflanagan: You could factor out some shared code from this and the NullTerminatedLatin1 function into a getNullTerminatedByteArray() utility function, if you want to.

shared/js/blobview.js
```diff
@@ -6,6 +6,15 @@ var BlobView = (function() {
...
   }
 
+  var decoderCache = {};
+  function getDecoder(encoding) {
+    if (encoding in decoderCache) {
+      return decoderCache[encoding];
+    }
+    var decoder = decoderCache[encoding] = new TextDecoder(encoding);
+    return decoder;
+  }
```
davidflanagan: There might be a tradeoff between performance and memory usage here. You're going for higher peformance with potentially higher memory usage.  I don't know anything about the TextDecoder implementation, though, so maybe there isn't really any real memory downside to keeping decoder instances around.  On the other hand, I don't know if they are at all expensive to construct new instances. It might be fine to just call new TextDecoder() each time.

I'm fine with doing it this way, but would be finer with it if you had some data (or knowledge of the impl) to justify using this cache.

apps/music/js/metadata/vorbis_picture.js
jimporter: Yes, the old code was wrong, and I've fixed it.

shared/js/blobview.js
jimporter: There's definitely some performance overhead to creating lots of `TextDecoder`s. I tried without the cache, and scanning my music collection went from 35 seconds to 45 seconds.

shared/js/blobview.js
jimporter: Fixed.

shared/js/blobview.js
jimporter: Done.

shared/js/blobview.js
jimporter: Fixed.

shared/js/blobview.js
jimporter: I suppose I could, but I wanted to avoid encouraging anyone to use a function like `getNullTerminatedByteArray`.

shared/js/blobview.js
jimporter: Yeah, the others are like that too (they have an `indexOf` call), but I think this is still fast in practice. It's also only called for older versions of ID3 tags.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=28b1f8dfb647d32292372724597ab0af41612ddc

PR #32258, opened 2015-10-06
Bug 1211668 - Migrate Cost Control from l10n_date to Intl API

zbraniecki:
https://bugzilla.mozilla.org/show_bug.cgi?id=1211668


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=aad101cf793b16feb238145fd210a896cfa8ebe8

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e183e7cdb8761aaff1e5fc283ecce1392ac963ee

apps/costcontrol/js/views/telephony.js
```diff
@@ -112,8 +112,10 @@ var TelephonyTab = (function() {
       billingCycle.hidden = true;
     } else {
       billingCycle.hidden = false;
-      var content = Formatting.getFormattedDate(settings.nextReset,
-        _('short-date-format'));
+      var content = Formatting.getFormattedDate(settings.nextReset, {
```
lodr: What about exposing the object replacing `_('short-date-format')` as a getter in the `Formatting` object? For instance:

```js
Formatting.getFormattedDate(settings.nextReset, Formatting.formats.shortDate)
```

apps/costcontrol/test/unit/formatting_test.js
```diff
@@ -142,36 +142,35 @@ suite('Formatting Test Suite >', function() {
...
-      var dateSpy = this.sinon.spy(MockMozL10n.DateTimeFormat.prototype,
-                        'localeFormat');
 
       assert.equal(f.formatTime(), 'never');
-      f.formatTime(oneTuesday, '%a');
-      assert.ok(dateSpy.calledWith(oneTuesday, '%a'));
+      f.formatTime(oneTuesday, {
+        weekday: 'short'
+      });
+      //assert.ok(dateSpy.calledWith(oneTuesday, '%a'));
```
lodr: If this is not needed, remove the comment.

apps/costcontrol/test/unit/formatting_test.js
```diff
@@ -200,26 +199,24 @@ suite('Formatting Test Suite >', function() {
...
-      assert.ok(dateSpy.calledWith(oneTuesday, '%a'));
-      assert.ok(dateSpy.calledWith(oneTuesday,  'shortTimeFormat12'));
 
-      dateSpy.reset();
-      window.navigator.mozHour12 = false;
+      var calledTimes = 0;
+
+      var realDateTimeFormat = Intl.DateTimeFormat;
+
+      Intl.DateTimeFormat = function() {
```
lodr: Use a sinon spy instead, please:

```js
sinon.spy(Intl.DateTimeFormat.prototype, 'format');
```

apps/costcontrol/test/unit/formatting_test.js
```diff
@@ -200,26 +199,24 @@ suite('Formatting Test Suite >', function() {
...
+            calledTimes++;
+          }
+        };
+      };
+      window.navigator.mozHour12 = true;
       f.formatTimeSinceNow(oneTuesday);
-      assert.ok(dateSpy.calledTwice);
-      assert.ok(dateSpy.calledWith(oneTuesday, '%a'));
-      assert.ok(dateSpy.calledWith(oneTuesday,  'shortTimeFormat24'));
+      assert.equal(calledTimes, 2);
```
lodr: Then assert, using sinon API.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9ae90ac91b84518ced6d1e3705e66eadd0a40040

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=103d5b00910934f360317a4119a64039c535e5f3

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4f850afb6bf120c52ce9966c8775471b4eaade68

apps/costcontrol/js/views/telephony.js
zbraniecki: Ok, I cached the formatters and added code to reset them on language/timeformat changes.

apps/costcontrol/test/unit/formatting_test.js
zbraniecki: I would love to, but it doesn't work this day. `DateTimeFormat` doesn't work this way. `format` is not a method on a class, it's dynamically generated function and there's no way to spy on it :(
I tried with the new approach to set a spy on `Formatting.formats.shortTime.format` but that doesn't work either.
It seems that it's a getter, because `console.log(Formatting.formats.shortTime.format);` returns null, but executing this function works.

It's a more common problem and in many places in our code we have to deal with it.

Eventually, I realized that there's no need to test that. We are testing if Formatting methods return the right values, and it shouldn't be tested *how* they do this. The fact that Intl API is called twice inside it shouldn't matter if I'm correct as long as the returned value is correct.

So because with this change I am able to test the returned value against real data (instead of stubs of `mozL10n.DateTimeFormat`) I just removed that code.

Hope you'll like it:)

apps/costcontrol/js/utils/formatting.js
```diff
@@ -208,6 +207,34 @@ var Formatting = (function() {
     formatTimeHTML: formatTimeHTML,
 
     // Given the API information compute the human friendly minutes
-    computeTelephonyMinutes: computeTelephonyMinutes
+    computeTelephonyMinutes: computeTelephonyMinutes,
+
+    formats: {}
```
lodr: Oh, I see you cached the formatters, not the _formats_. What I was asking was to cache the formats. Caching formatters mean you need to listen for those [events in lines 239 and 240](#diff-f80fe1f3ae9362ef69f590f6898c42d2R239) although it mean to save some overhead when displaying a date. I'm ok with caching formatters but then change the map name from `formats` to `formatters`.

apps/costcontrol/test/unit/formatting_test.js
```diff
@@ -134,44 +134,17 @@ suite('Formatting Test Suite >', function() {
...
-
-      dateSpy.reset();
-      window.navigator.mozHour12 = false;
-      f.formatTime(oneTuesday);
-      assert.ok(dateSpy.calledTwice);
-      assert.ok(dateSpy.calledWith(oneTuesday, '%a'));
-      assert.ok(dateSpy.calledWith(oneTuesday,  'shortTimeFormat24'));
-
-      dateSpy.restore();
-      window.navigator.mozHour12 = originalMozHour12;
```
lodr: I'm assuming you're removing all of this because it's not actually testing a public API. That's great!

apps/costcontrol/js/utils/formatting.js
zbraniecki: deal! :) Yeah, I misunderstood you, but I think the outcome is good. We are caching what should be cached :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0c91f6e410250c361964a0f0d4e8b1104a8821de

PR #32275, opened 2015-10-06
Bug 1212007 - making sure that pin dialog is trully hidden from a11y …

yzen:
…when it has hidden style.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=20689fcf2346752db1944407f4bf73673683131f

PR #32289, opened 2015-10-07
Bug 1211258 - PopupWindow#request{Open|Close} should set visibility of it's rearWindow

timdream:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ee61af7d8b7f6a86279e327f3a703a3a1732d727

apps/system/js/child_window_factory.js
```diff
@@ -143,7 +143,7 @@
     };
     var childWindow = new PopupWindow(configObject);
     childWindow.element.addEventListener('_closing', this);
```
etiennesegonzac: can't we just add a
```js
childWindow.element.addEventListener('_opened', this);
```
listener here and benefit from the mechanism we already have in place for activities?
(see `_handle_child__opened` / `_handle_child__closing`)

timdream: I see. That make sense.

PR #32304, opened 2015-10-07
Bug 1188286 - [Aries][Camera][Media Frame] Use small EXIF preview to fix flash seen while viewing images in camera preview

punamdahiya:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ffa093ba63ac25e11a539b64e68303bd02496fc5

apps/camera/js/views/preview-gallery.js
```diff
@@ -27,7 +27,7 @@ require('gaia-header');
 var SWIPE_DISTANCE_THRESHOLD = window.innerWidth / 3; // pixels
 var SWIPE_VELOCITY_THRESHOLD = 1.0;                   // pixels/ms
 
-var SWIPE_DURATION = 250;   // How long to animate the swipe
+var SWIPE_DURATION = 150;   // How long to animate the swipe
```
punamdahiya: IMO 250ms swipe duration is long and reason grey screen is seen between images while swiping.

shared/js/media/media_frame.js
```diff
@@ -47,6 +47,8 @@ function MediaFrame(container, includeVideo, maxImageSize) {
   this.image = new Image();
   this.image.className = 'image-view';
   this.image.style.opacity = 0;                        // Start off hidden.
+  this.image.style.transitionProperty = 'opacity';
+  this.image.style.transitionDuration = '100ms';
   this.image.onload = function () {                    // When image loads...
```
punamdahiya: For smooth transition effect when opacity changes

shared/js/media/media_frame.js
```diff
@@ -47,6 +47,8 @@ function MediaFrame(container, includeVideo, maxImageSize) {
   this.image = new Image();
   this.image.className = 'image-view';
   this.image.style.opacity = 0;                        // Start off hidden.
+  this.image.style.transitionProperty = 'opacity';
+  this.image.style.transitionDuration = '100ms';
```
punamdahiya: Add transition property and duration for smooth transition effect when opacity changes

davidflanagan: Put the styles in a stylesheet please. See apps/camera/style.preview-gallery.css:43.
If you think this change is good for gallery too, feel free to change gallery.css as well.

shared/js/media/media_frame.js
```diff
@@ -245,7 +247,20 @@ MediaFrame.prototype.displayImage = function displayImage(blob,
     }
   }
   else {
-    noPreview();
+    // If we have valid small EXIF preview use it for css background image
+    // This will help to load large downsampled preview images without flash
+    // See bug 1188286
+    // ( TBD: preview & preview.start might not be enough.
+    // Need more robust check for small valid EXIF preview )
+    if (preview && preview.start) {
```
davidflanagan: I think maybe I'd combine this if with the else above to have an else if and I'd do
else if (preview && preview.width && preview.height && preview.start) {
Note that if you don't check the aspect ratio then it might be possible to get a flash of a distorted preview before the full image loads, and that would probably be bad.  I actually have some test images like this in apps/gallery/test/images/: images where the EXIF preview has the wrong aspect ratio and so is not used. This patch would make them display that bad preview briefly.

So maybe refactor usePreview() into two methods. The first would just be isValidPreview() that would do all the checks  (preview exists, size is non-zero, aspect ratio is correct) except for the size is big enough check.  The second would be isBigEnoughPreview() which would do the size check. Then the code here could be refactored to look like this:
```
var isValid = isValidPreview(preview);
var isBigEnough = isBigEnoughPreview(preview);
if (isValid && isBigEnough) {
  // this is the existing usePreview() code
}
else if (isValid) {
  // this is your new case where you're setting the background image.
}
else {
  // this is the existing nopreview case
}
```
if (isValidPreview(preview) && isBigEnough(preview))

shared/js/media/media_frame.js
```diff
@@ -245,7 +247,20 @@ MediaFrame.prototype.displayImage = function displayImage(blob,
...
+      // Create a blob URL for the bg preview
+      self.bgPreviewURL = URL.createObjectURL(
```
davidflanagan: Please initialize this.bgPreviewURL to null up at the start of displayImage() where we initialize other things that later get cleared. Maybe after this.fullImageURL. Then add a comment there explaining what the field is for.

shared/js/media/media_frame.js
davidflanagan: Actually, I think the new case would have to be:
```
else if (isValid && preview.start)
```

shared/js/media/media_frame.js
davidflanagan: thinking about writing a comment to explain this field, it occurs to me that it might be easier to explain what this is and how it differs from this.fullImageURL and this.previewurl if you called it this.tinyPreviewURL. The size is more relevant than the fact that it is being displayed using background-image, I think.

shared/js/media/media_frame.js
davidflanagan: Actually, making this animated might mean that calling clear() (which sets opacity back to 0) will trigger an animation. If it works okay without this, it would be safer to just leave it out.  But if you need it, please just test carefully

apps/camera/js/views/preview-gallery.js
davidflanagan: This was purposely set long enough to hide some kind of image decoding ugliness. But if a smaller value works well, please do change. it.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d1b4757a15fcd59a6714354bcb8a59df9def8913

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=bd325e3da17ce9d56d41c08366987ba8a4e65a37

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=bc7345af4bdc76e3bcf8c128e7abc6719e5a2d4e

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f118bb57f156b696ae6118d0a45fe78af8432aa6

shared/js/media/media_frame.js
```diff
@@ -244,6 +257,17 @@ MediaFrame.prototype.displayImage = function displayImage(blob,
...
+  else if (isValid) {
+    // If we have valid small EXIF preview use it for css background image
+    // This will help to load large downsampled preview images without flash
+    // See bug 1188286
+
+    // Create a blob URL for the tiny preview
+    self.tinyPreviewURL = URL.createObjectURL(
+      blob.slice(preview.start, preview.end, 'image/jpeg'));
+
+    noPreview(self.tinyPreviewURL);
```
davidflanagan: I think you can use this here and on line 266 instead of self, can't you?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=409c80a342dd612abfafa809cf3a99ce174d7471

shared/js/media/media_frame.js
punamdahiya: good catch, will update.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=34b253120dd27af22e572af196cc534a999866ad

PR #32306, opened 2015-10-07
Bug 1208553 - Convert gaia ui tests for music-nga app 

nojunpark:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d0ad95f72c07e871649af99d4b6ab5f2793b5df3

tests/python/gaia-ui-tests/gaiatest/apps/music/app.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def wait_for_music_tiles_displayed(self):
+        self.marionette.switch_to_frame(self.marionette.find_element(*self._active_view_locator))
+        Wait(self.marionette).until(expected.element_displayed(*self._music_tiles_locator))
+        self.apps.switch_to_displayed_app()
+
+    def wait_for_empty_message_to_load(self):
+        empty_overlay = self.marionette.find_element(*self._empty_overlay_shadow_DOM_locator)
+        self.marionette.switch_to_shadow_root(empty_overlay)
+        Wait(self.marionette).until(lambda m: self.empty_music_title == 'Add songs to get started')
```
mwargers: Fwiw, this will not work in other locales.

npark-mozilla: True, that was what the previous script was doing, but I should think of doing it differently.

tests/python/gaia-ui-tests/gaiatest/apps/music/app.py
```diff
@@ -0,0 +1,98 @@
...
+    _artists_tab_locator = (By.CSS_SELECTOR, '[data-l10n-id="artists-tab"]')
+    _title_locator = (By.ID, 'title-text')
+
+    def launch(self):
+        Base.launch(self)
+        Wait(self.marionette).until(expected.element_not_displayed(
+            *self._loading_spinner_locator))
+
+    def wait_for_music_tiles_displayed(self):
+        self.marionette.switch_to_frame(self.marionette.find_element(*self._active_view_locator))
```
mwargers: What is this? This is an iframe withing the Music app? 
You use that quite a bit. Perhaps you could make this a method of the Music app, e.g. self.switch_to_active_view() or something like that?

npark-mozilla: yes, in music app, each view is enclosed within an iframe, so in order to search anything inside the view I have to switch into the iframe.  (Except in case of tabs or audio object).  I was thinking about creating a separate method, but then again it would just be a one-line method.  It would perhaps make the code easier to read, but only to those who doesn't know what this iframe is for.  i can change it though.
There are a number of different iframes within the music app as well.

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/list_view.py
```diff
@@ -0,0 +1,65 @@
...
+from gaiatest.apps.music.regions.player_view import PlayerView
+
+
+class ListView(Base):
+
+    _albums_view_locator = (By.CSS_SELECTOR, 'iframe.active[src*="/views/albums/index.html"]')
+    _artists_view_locator = (By.CSS_SELECTOR, 'iframe.active[src*="/views/artists/index.html"]')
+    _songs_view_locator = (By.CSS_SELECTOR, 'iframe.active[src*="/views/songs/index.html"]')
+    _playlists_view_locator = (By.CSS_SELECTOR, 'iframe.active[src*="/views/playlists/index.html"]')
+    _view_locator = (By.ID, 'list')
```
JohanLorenzo: Unused locator

tests/python/gaia-ui-tests/gaiatest/apps/music/app.py
```diff
@@ -0,0 +1,98 @@
...
+        self.apps.switch_to_displayed_app()
+
+    def wait_for_empty_message_to_load(self):
+        empty_overlay = self.marionette.find_element(*self._empty_overlay_shadow_DOM_locator)
+        self.marionette.switch_to_shadow_root(empty_overlay)
+        Wait(self.marionette).until(lambda m: self.empty_music_title == 'Add songs to get started')
+        Wait(self.marionette).until(lambda m: self.empty_music_text == 'Load songs on to the memory card.')
+        self.apps.switch_to_displayed_app()
+
+    def wait_for_view_displayed(self, view_name):
```
mwargers: This method is not used.

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/list_view.py
```diff
@@ -0,0 +1,65 @@
...
+    _list_item_locator = (By.CLASS_NAME, 'gfl-item')
+
+    def __init__(self, marionette, view_type):
+        Base.__init__(self, marionette)
+        if view_type is 'albums':
```
JohanLorenzo: How about breaking this List views into smaller classes? Something like:
``` python
class ListView(Base):
    def __init__(self, marionette):
         self._wait_for_main_view_to_be displayed()

    def media():
        ....

    def get_view_by_type_locator(self, type)
        return By.CSS_SELECTOR, 'iframe.active[src*="/views/{}/index.html"]'.format(type)

class ArtistView(ListView):
    def _wait_for_main_view_to_be displayed(self):
       Wait(self.marionette).until(expected.element_displayed(*self.get_view_by_type_locator('artists')))

class AlbumView(ListView):
    ...
```

What do you think?

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/player_view.py
```diff
@@ -0,0 +1,102 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import time
+
+from marionette_driver import expected, By, Wait
+from gaiatest.apps.base import Base
+
+
+class PlayerView(Base):
+    _audio_locator = (By.CSS_SELECTOR, '#audio')
```
JohanLorenzo: Nit: By.ID

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/player_view.py
```diff
@@ -0,0 +1,102 @@
...
+    _active_view_locator = (By.CSS_SELECTOR, 'iframe.active[src*="/views/player/index.html"]')
+    _player_seek_elapsed_locator = (By.ID, 'elapsed-time')
+    _player_controls_shadow_DOM_locator = (By.ID, 'controls')
+    _player_controls_play_locator = (By.ID, 'toggle')
+    _cover_image_shadow_DOM_locator = (By.ID, 'artwork')
+    _seek_bar_shadow_DOM_locator = (By.ID, 'seek-bar')
+    _cover_share_locator = (By.CSS_SELECTOR, '[data-l10n-id="share-song"]')
+    _rating_view_locator = (By.CSS_SELECTOR, '#rating')
```
JohanLorenzo: Nit: By.ID

npark-mozilla: actually, strangely enough, when I used 'By.ID', it didn't find the element.  I think it's either some kind of marionette bug or shadow dom issue.

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/player_view.py
npark-mozilla: same here, when i used By.ID, it couldn't find the element. (I double checked)

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/list_view.py
npark-mozilla: Hmm, makes sense, I was trying to preserve the existing format so we won't need to modify the script itself, but maybe this is a better idea.

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/player_view.py
```diff
@@ -0,0 +1,102 @@
...
+    def player_elapsed_time(self):
+        self.marionette.switch_to_frame(self.marionette.find_element(*self._active_view_locator))
+        self.marionette.switch_to_shadow_root(self.marionette.find_element(*self._seek_bar_shadow_DOM_locator))
+        Wait(self.marionette).until(expected.element_displayed(
+            Wait(self.marionette).until(expected.element_present(*self._player_seek_elapsed_locator))))
+        elapsed_time = time.strptime(self.marionette.find_element(*self._player_seek_elapsed_locator).text, '%M:%S')
+        self.apps.switch_to_displayed_app()
+        return elapsed_time
+
+    def is_player_playing(self):
```
JohanLorenzo: Oh I remember this function having intermittency. Could we change it to something like:
``` py
def wait_until_time_stamp_has_changed(self):
    base_timestamp = self.player_current_timestamp
    Wait(self.marionette).until(lambda m: self.player_current_timestamp > base_timestamp)
```
Hence we can solve [bug 1201880](https://bugzilla.mozilla.org/show_bug.cgi?id=1201880), directly?

npark-mozilla: I'm fine with that, Martijn, does it sound good for you as well?

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/sublist_view.py
```diff
@@ -0,0 +1,39 @@
...
+class SublistView(Base):
+
+    _album_sublist_view_locator = (By.CSS_SELECTOR, 'iframe.active[src*="/views/album-detail/index.html"]')
+    _artist_sublist_view_locator = (By.CSS_SELECTOR, 'iframe.active[src*="/views/artist-detail/index.html"]')
+    _album_list_locator = (By.ID, 'list')
+    _first_song_locator = (By.CLASS_NAME, 'gfl-item first')
+
+    def __init__(self, marionette, sublist_type):
+        Base.__init__(self, marionette)
+        if sublist_type == 'album':
```
JohanLorenzo: I wonder if we can also break down this class into smaller ones.

npark-mozilla: for consistency, if we break down list view, we should break down sublist view as well (the src are distinct too)

tests/python/gaia-ui-tests/gaiatest/tests/functional/music/test_music_share_ringtone.py
```diff
@@ -37,6 +37,8 @@ def test_music_share_ringtone(self):
         # wait until the player view is shown, then tap the share button
         play_time = time.strptime('00:01', '%M:%S')
         Wait(self.marionette).until(lambda m: player_view.player_elapsed_time >= play_time)
+        #player_view.tap_play()
```
JohanLorenzo: Nit: Commented code.

tests/python/gaia-ui-tests/gaiatest/tests/functional/music_oga/manifest.ini
```diff
@@ -0,0 +1,26 @@
+[parent:../manifest.ini]
```
JohanLorenzo: I wonder if we really need to keep the tests of Music OGA. If we had to roll back to the OLD OGA, A backout of this commit would suffice. What do you think?

npark-mozilla: I'm sure we probably would deprecate it eventually, but I thought perhaps we should maintain it at least until 2.5 is released, since if we just backout this commit, any fixes we (might) do on oga test would be lost.

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/player_view.py
```diff
@@ -0,0 +1,102 @@
...
+    _player_controls_play_locator = (By.ID, 'toggle')
+    _cover_image_shadow_DOM_locator = (By.ID, 'artwork')
+    _seek_bar_shadow_DOM_locator = (By.ID, 'seek-bar')
+    _cover_share_locator = (By.CSS_SELECTOR, '[data-l10n-id="share-song"]')
+    _rating_view_locator = (By.CSS_SELECTOR, '#rating')
+    _stars_on_locator = (By.CSS_SELECTOR, '.rating-star.star-on')
+
+    def __init__(self, marionette):
+        Base.__init__(self, marionette)
+        self.marionette.switch_to_frame(self.marionette.find_element(*self._active_view_locator))
```
JohanLorenzo: Why do we need to switch to this frame, each time we want to dive in the shadow dom?

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/player_view.py
```diff
@@ -0,0 +1,102 @@
...
+
+from marionette_driver import expected, By, Wait
+from gaiatest.apps.base import Base
+
+
+class PlayerView(Base):
+    _audio_locator = (By.CSS_SELECTOR, '#audio')
+    _active_view_locator = (By.CSS_SELECTOR, 'iframe.active[src*="/views/player/index.html"]')
+    _player_seek_elapsed_locator = (By.ID, 'elapsed-time')
+    _player_controls_shadow_DOM_locator = (By.ID, 'controls')
```
JohanLorenzo: Nit: shadow_dom (all lowecase)

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/player_view.py
npark-mozilla: because the shadow dom is inside this iframe.  and in order to check elements within an iframe, you have to switch frames or otherwise it won't be visible at displayed_app context.

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/player_view.py
npark-mozilla: true, will change that

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/player_view.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def tap_play(self):
+        self.marionette.switch_to_frame(self.marionette.find_element(*self._active_view_locator))
+        self.marionette.switch_to_shadow_root(self.marionette.find_element(*self._player_controls_shadow_DOM_locator))
+        self.marionette.find_element(*self._player_controls_play_locator).tap()
+        self.apps.switch_to_displayed_app()
+
+    def tap_cover_in_player_view(self):
+        self.marionette.switch_to_frame(self.marionette.find_element(*self._active_view_locator))
+        self.marionette.switch_to_shadow_root(self.marionette.find_element(*self._cover_image_shadow_DOM_locator))
```
JohanLorenzo: I think the `cover_image` is worth to be extracted in something like a `wigdet`. We switch into it many times in this class. Having it extracted would lighten the code in PlayerView

It might be something to do in a follow up bug, though.

tests/python/gaia-ui-tests/gaiatest/apps/music/app.py
```diff
@@ -0,0 +1,98 @@
...
+    @property
+    def views(self):
+        return self.marionette.find_element(*self._views_locator)
+
+    @property
+    def tabs(self):
+        return self.marionette.find_element(*self._tabs_locator)
+
+    def tap_albums_tab(self):
+        self.apps.switch_to_displayed_app()
```
JohanLorenzo: I saw the DOM of Music NGS. I understand the technical reason why we have to switch to the Music app frame each time we want to tap a tab. 

I think we can reduce the duplicate by creating a class called `TabElement`, which will be in charge of making sure we're in the right frame. Something like:
```py
class TabElement(PageRegion):
    def tap(self):
        self.apps.switch_to_displayed_app()
        Wait(self.marionette).until(expected.element_displayed(self.root_element))
        self.root_element.tap()
        # I'm still unsure on where should the return be. We can use the `_get_class_by_name()` function we had, but it sounds overkill for 5 functions. 
```

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/player_view.py
npark-mozilla: Happy to do it in a follow up bug. could you also point me to a sample of widget case?

tests/python/gaia-ui-tests/gaiatest/apps/music/app.py
npark-mozilla: So having this element would mean that at the end of other methods, I shouldn't put self.apps.switch_to_displayed_app()?  I think we still should, because there are some element that are at the app level, like _audio_locator element in player_view.py.  
Also, if we use this method, wouldn't we we add extra work to the test script writers by making aware of such class?  personally i think test script should be as simple as it can be.

tests/python/gaia-ui-tests/gaiatest/apps/music/app.py
JohanLorenzo: A separate method in `Base` sounds fine. Buddy up is also engineered this way, we'd need it there too.

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/player_view.py
JohanLorenzo: I think it's worth to file a bug against Marionette and link it here. This looks related to shadow dom.

tests/python/gaia-ui-tests/gaiatest/apps/music/app.py
JohanLorenzo: Yeah, that might be overkill.

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/list_view.py
```diff
@@ -0,0 +1,65 @@
...
+            self._view = self._albums_view_locator
```
mwargers: Perhaps you can you do self._view = self.marionette.find_element(*self._albums_view_locator) or will that cause problems at some point due to marionette frame switching and https://bugzilla.mozilla.org/show_bug.cgi?id=1202246 ?

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/player_view.py
```diff
@@ -0,0 +1,102 @@
...
+    _seek_bar_shadow_DOM_locator = (By.ID, 'seek-bar')
+    _cover_share_locator = (By.CSS_SELECTOR, '[data-l10n-id="share-song"]')
+    _rating_view_locator = (By.CSS_SELECTOR, '#rating')
+    _stars_on_locator = (By.CSS_SELECTOR, '.rating-star.star-on')
+
+    def __init__(self, marionette):
+        Base.__init__(self, marionette)
+        self.marionette.switch_to_frame(self.marionette.find_element(*self._active_view_locator))
+        element = self.marionette.find_element(*self._cover_image_shadow_DOM_locator)
+        Wait(self.marionette).until(lambda m: element.location['x'] == 0)
```
mwargers: Use .rect please, also could you add a element.is_displayed() here? Because invisible elements also have .rect['x'] == 0.

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/list_view.py
npark-mozilla: yeah... that's what I initially tried, but after switching frames, it became stale, so I decided to set the locator only.

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/sublist_view.py
```diff
@@ -0,0 +1,39 @@
...
+            self._view = self._album_sublist_view_locator
+        elif sublist_type == 'artist':
+            self._view = self._artist_sublist_view_locator
+
+        element = self.marionette.find_element(*self._view)
+        Wait(self.marionette).until(expected.element_displayed(element))
+        self.marionette.switch_to_frame(element)
+
+        element = self.marionette.find_element(*self._album_list_locator)
+        Wait(self.marionette).until(lambda m: element.location['x'] == 0)
```
mwargers: Use .rect please, also could you add a element.is_displayed() here? Because invisible elements also have .rect['x'] == 0.

tests/python/gaia-ui-tests/gaiatest/tests/functional/music/test_music_share_ringtone.py
```diff
@@ -37,6 +37,8 @@ def test_music_share_ringtone(self):
...
+        #self.assertFalse(player_view.is_player_playing(), 'The player did not stop playing')
```
mwargers: This is commented out, either you need to remove it or uncomment it.

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/list_view.py
mwargers: And perhaps self._active_view is a better name here?

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/player_view.py
mwargers: The body innerHTML of the payer is this:
  <music-artwork src="/img/artwork/artwork-8.png" album="Unknown album" artist="Unknown artist" shuffle="off" repeat="off" id="artwork"></music-artwork>
  <music-seek-bar id="seek-bar"></music-seek-bar>
  <music-controls id="controls"></music-controls>

music-rating id="rating" is shadow dom part of music-artwork.

Supposedly you should be able to use By.ID.
This test should cover that it works:
https://dxr.mozilla.org/mozilla-central/source/testing/marionette/client/marionette/tests/unit/test_shadow_dom.py#54
https://dxr.mozilla.org/mozilla-central/source/testing/marionette/client/marionette/www/test_shadow_dom.html

Anyway, this is something to look into in a follow-up bug, indeed.

tests/python/gaia-ui-tests/gaiatest/apps/music/app.py
npark-mozilla: oh, this would be used for imagecompare tests, will update this test on a separate PR where I migrate the imagecompare tests that access music apps.

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/list_view.py
npark-mozilla: removed

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/list_view.py
npark-mozilla: I'll subclass this view as Johan suggested, and rename the locators

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/player_view.py
npark-mozilla: fixed

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/player_view.py
npark-mozilla: ahh, actually this won't work, because this method (is_player_playing) also should work for when the player is NOT playing, and if we use the Wait() call, then it'll return error instead.

npark-mozilla: what we could do is to introduce another method is_player_stopped() and selectively use the methods, but i think that's better to address it in bug 1201880.

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/sublist_view.py
npark-mozilla: fixed

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/sublist_view.py
npark-mozilla: created subclasses for sublistview

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/player_view.py
npark-mozilla: This works now, changed.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=75bda47da9b119fe9018b7ab1ab2f4e1768a4af3

tests/python/gaia-ui-tests/gaiatest/apps/music/regions/player_view.py
```diff
@@ -0,0 +1,104 @@
...
+    _active_view_locator = (By.CSS_SELECTOR, 'iframe.active[src*="/views/player/index.html"]')
+    _player_seek_elapsed_locator = (By.ID, 'elapsed-time')
+    _player_controls_shadow_dom_locator = (By.ID, 'controls')
+    _player_controls_play_locator = (By.ID, 'toggle')
+    _cover_image_shadow_dom_locator = (By.ID, 'artwork')
+    _seek_bar_shadow_dom_locator = (By.ID, 'seek-bar')
+    _cover_share_locator = (By.CSS_SELECTOR, '[data-l10n-id="share-song"]')
+    _rating_view_locator = (By.ID, 'rating')
+
+    #_rating_view_locator = (By.CSS_SELECTOR, '#rating')
```
mwargers: This one is commented out now? So I guess you could use By.ID after all?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2824d7de70008f2c37b140358a90f39d76f5a845

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9f382ae5edffcaebe99af402454daad56f1bfea5

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=bc2fc269166297b871df278e8d9f5fa5b32bf5cc

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a7f499c6aa1b72d4d81d86a046d455c764a08bcf

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=84bfdcde0e40dbe1eec29350ac1885b58e53a324

PR #32319, opened 2015-10-08
Bug 1192263 - [Messages] We load Inbox before going to the notification conversation when app is run via notification click, r=julienw

steveck-chung:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b4026ec03c9196104ff2c3e0aa93b130415fa2ee

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=91c3acf56e3ea471aa9d068c1dbd3ed7e58dd46a

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=28fd00ba3325a0e838b1d4094147c2e8ee2a02a4

apps/sms/views/shared/js/system_message_handler.js
```diff
@@ -329,6 +329,11 @@
       // Conversation view to the user, we remove that notification from the
       // tray that causes the second system message with "clicked" set to false.
       if (!notification.clicked || !notification.data) {
+        // Force close the app for notification removal case when app is hidden
+        // and navigation state is invalid(early reject while initializing).
+        if (document.hidden && !Navigation.isValidCurrentState()) {
+          return Promise.reject(window.close());
```
julienw: not sure why you reject with the result of window.close ? I think we should always try to reject with an error.

apps/sms/views/shared/js/system_message_handler.js
```diff
@@ -329,6 +329,11 @@
       // Conversation view to the user, we remove that notification from the
       // tray that causes the second system message with "clicked" set to false.
       if (!notification.clicked || !notification.data) {
+        // Force close the app for notification removal case when app is hidden
+        // and navigation state is invalid(early reject while initializing).
+        if (document.hidden && !Navigation.isValidCurrentState()) {
```
julienw: my problem is that isValidCurrentState will only be true at the first toPanel... this happens after a mozApps.getSelf(), wait that the document is visible (`onceDocumentIsVisible`) (so this should always be true if document.hidden is false), and that we retrieve the message information with `MessageManager.get(id)`.

Too many possible races...

IMO the only way if you don't want to use the task runner is to add a state like a boolean "this.handlingNotification". This would be good for 2.5 but I'd like that we convert using the task runner post-2.5.

apps/sms/views/shared/js/system_message_handler.js
steveck-chung: If we don't close the app, the other alternative solution is finish the navigation init again or toPanel(inboxView).

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=77901fc97686ae13f57426d47ffd8d27f896d7e1

steveck-chung:
@julienw  I changed the checking with additional `hasPendingInit` that could let us know if we finish the init step or not. The main reason I didn't apply task runner because I thought they are 2 different stories(even we introduce task runner we still need to check the pending notification message to avoid this issue). So I also create another patch for the task runner version(but not working) in https://github.com/mozilla-b2g/gaia/pull/32588.

apps/sms/views/shared/js/system_message_handler.js
```diff
@@ -329,6 +329,12 @@
       // Conversation view to the user, we remove that notification from the
       // tray that causes the second system message with "clicked" set to false.
       if (!notification.clicked || !notification.data) {
+        // Initialize the navigation again for notification removal case
+        // because navigation didn't init correctly if pending notification
+        // message existed.
+        if (Navigation.hasPendingInit()) {
+          return Navigation.init();
```
julienw: so I guess here it's either this or `window.close()` ?
What do you prefer ?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=043ab64fc02536517c5dc0d3ae83ce3cc635a88a

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=48a93863e4f6e2d90ec0cd3de357a40118e65e66

apps/sms/views/shared/js/system_message_handler.js
```diff
@@ -329,6 +332,12 @@
       // Conversation view to the user, we remove that notification from the
       // tray that causes the second system message with "clicked" set to false.
       if (!notification.clicked || !notification.data) {
+        // Force closing the window for notification removal case
+        // because navigation didn't init correctly if pending notification
+        // message existed.
+        if (Navigation.hasPendingInit()) {
+          return Promise.reject(window.close());
```
steveck-chung: @julienw I would prefer and close the app more if we can make sure it's safe to do so. For this change we will need to change the hasPendingInit flag in toPanel as well.

apps/sms/views/shared/js/system_message_handler.js
```diff
@@ -270,11 +270,14 @@
...
         //Validate if message still exists before opening message thread
         //See issue https://bugzilla.mozilla.org/show_bug.cgi?id=837029
-        return Utils.onceDocumentIsVisible().then(
-          () => MessageManager.getMessage(id)
-        ).then(
+        return MessageManager.getMessage(id).then(
           (message) => Navigation.toPanel('thread', { id: message.threadId }),
-          () => Utils.alert('deleted-sms')
+          () => {
+            if (Navigation.hasPendingInit()) {
```
steveck-chung: I don't think the thread removed case is common, I thought user rarely have the chance to see the notification still exist after thread removed. But I still init the navigation here if it didn't finish the initialization with pending notification message case.

apps/sms/views/shared/js/system_message_handler.js
julienw: yep

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=da63be3e8b5c855b1f1b3314746780d01052982a

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4c477b2a175b7d504c4e22cf36ced66506d1b32e

apps/sms/test/marionette/lib/messages.js
```diff
@@ -112,6 +112,11 @@ var ParticipantsAccessor = require('./views/participants/accessors');
           newMessageView.addNewRecipient(recipient);
         },
 
+        getActivePanelName: function() {
+          var cls = client.findElement('.panel-active').getAttribute('class');
+          return cls.match(/(panel-[A-Z][a-zA-Z]+)/g);
```
julienw: This is working by chance: `match` returns an array, but the conversation array -> string returns the correct value in the assertion.

I think we should return something more expected here.

You can do this:
```javascript
var matchResult = /\bpanel-\w+\b/g.exec(cls);
return matchResult && matchResult[0];
```

or even:
```javascript
var matchResult = /\bpanel-(\w+)\b/g.exec(cls);
return matchResult && matchResult[1];
```
=> and you get `ConversationView` instead of `panel-ConversationView`. What do you think ?

apps/sms/views/shared/test/unit/navigation_test.js
```diff
@@ -166,6 +169,34 @@ suite('navigation >', function() {
...
+      });
+
+      test('Clear pending when launched by notification and enter other panel',
+        function(done) {
+
+        navigator.mozHasPendingMessage.withArgs('notification').returns(true);
+
+        Navigation.init().catch(() => {}).then(() => {
+          assert.isTrue(Navigation.hasPendingInit());
+        }).then(() => Navigation.toPanel('composer')).then(() => {
```
julienw: you can simply add `return Navigation.toPanel('composer')` in the previous then handler

julienw: also you can maybe merge this test and the previous one ?

apps/sms/views/shared/test/unit/navigation_test.js
```diff
@@ -711,6 +743,34 @@ suite('navigation >', function() {
...
+          assert.isFalse(Navigation.hasPendingInit());
+        }).then(done, done);
```
julienw: same comments here

apps/sms/views/shared/test/unit/system_message_handler_test.js
```diff
@@ -546,7 +546,7 @@ suite('SystemMessageHandler', function() {
         setup(function(done) {
           this.sinon.stub(MessageManager, 'getMessage').withArgs(
             smsMessage.id
-          ).returns(smsMessage);
+          ).returns(Promise.resolve(smsMessage));
```
julienw: indeed :)

apps/sms/views/shared/test/unit/system_message_handler_test.js
```diff
@@ -728,6 +730,24 @@ suite('SystemMessageHandler', function() {
...
+      Navigation.hasPendingInit.returns(true);
+      this.sinon.stub(window, 'close');
+
+      SystemMessageHandler.onNotificationSystemMessage({
+        // When notification is removed "clicked" property is "false".
+        clicked: false,
+        data: { id: smsMessage.id, threadId: smsMessage.threadId }
+      }).then(
+        () => new Error('Should not enter resolved case'),
+        () => {
```
julienw: can we assert that Navigation.toPanel was not called ?

apps/sms/views/shared/js/system_message_handler.js
julienw: yeah, I think this happens if the user deletes the received SMS from an activity or something like this.

apps/sms/views/shared/js/system_message_handler.js
```diff
@@ -270,11 +270,14 @@
 
         //Validate if message still exists before opening message thread
         //See issue https://bugzilla.mozilla.org/show_bug.cgi?id=837029
-        return Utils.onceDocumentIsVisible().then(
-          () => MessageManager.getMessage(id)
-        ).then(
+        return MessageManager.getMessage(id).then(
           (message) => Navigation.toPanel('thread', { id: message.threadId }),
```
julienw: tell me what you think, I think we should even do this _before_ calling app.launch(). `toPanel().then(() => app.launch())`.

julienw: but maybe this takes too long, we can keep this for a separate bug post-2.5

apps/sms/views/shared/js/system_message_handler.js
julienw: like before, I don't get whi you reject with the result of window.close ?

why not doing this:
```javascript
window.close();
return Promise.reject(new Error('The notification has been dismissed.'));
```

apps/sms/test/marionette/lib/messages.js
steveck-chung: Thanks for the advice, I'm really bad at Regex :p

apps/sms/views/shared/test/unit/system_message_handler_test.js
steveck-chung: Not sure if it's meaningful to do so, because previously we didn't call toPanel while removing the notification either.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9eaeebc20e71b916843e93306c813f2668b0bef5

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5bb574f23aeb3c963a07f2ae345e5e58d210f98d

apps/sms/views/shared/js/system_message_handler.js
```diff
@@ -329,6 +338,23 @@
       // Conversation view to the user, we remove that notification from the
       // tray that causes the second system message with "clicked" set to false.
       if (!notification.clicked || !notification.data) {
+        // Force closing the window for notification removal case
+        // because navigation didn't init correctly if pending notification
+        // message existed.
+        if (Navigation.hasPendingInit()) {
+          return Notification.get().then((notifications) => {
```
steveck-chung: So here's the new approach: Check if we have any other notification. It seems like the notifications we can get would be restricted to the message notification, so I only need to check the number and decide the next step. If we don't have other notification, close app directly, otherwise we check the isNavigating flag and resolve if next toPanel already ongoing/start navigation init if we launch the app directly. I add the onceDocumentIsVisible to make sure it does enter the navigation.toPanel, but it seems not necessary because it would be set even without waiting for visibility. BTW we can also replace the Navigation.init with window.close.

apps/sms/views/shared/js/system_message_handler.js
steveck-chung: Maybe we can do some experiment later

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c3912863f07dd9bb0c3a8da21ae58257d39e7911

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c3912863f07dd9bb0c3a8da21ae58257d39e7911

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=21197923a161305d1016ace941ac41dd98c9fd70

PR #32334, opened 2015-10-08
Bug 1212726 - SpatialNavigator should check if an element is navigable before focusing it

luke-chang:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d9f65dba529d35e364b0bd536f9020a92802cd27

PR #32344, opened 2015-10-08
Bug 1212949 - [Music][NGA] Hammering next/previous buttons leads to u…

justindarc:
…nresponsiveness

### DIFF
--- a/apps/music/elements/music-controls.js
+++ b/apps/music/elements/music-controls.js
@@ -113,8 +113,8 @@ proto.createdCallback = function() {
     }
   );
 
-  this.els.container.addEventListener('click', (evt) => {
-    var button = evt.target.closest('button');
+  this.els.container.addEventListener('click', throttle((evt) => {
+    var button = evt.originalTarget.closest('button');
     switch (button.id) {
       case 'previous':
       case 'next':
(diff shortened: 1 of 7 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f1005901c04a79a8848d0cb55cabc68ecd9a0640

apps/music/js/endpoint.js
```diff
@@ -497,4 +497,24 @@ function open(blob) {
...
+        fn.apply(this, args);
+      }, ms);
+    }
+
+    else {
+      last = now;
+      fn.apply(this, args);
+    }
+  };
+}
```
jimporter: I see you duplicated this function above. Would it make sense to add a utils.js file or something?

justindarc: Yes. I was thinking this when I posted this patch. There are more things than just this that we can probably consolidate though, so I'm gonna hold off and just file another "cleanup" bug to DRY things up a bit :-)

apps/music/elements/music-controls.js
```diff
@@ -113,8 +113,8 @@ proto.createdCallback = function() {
     }
   );
 
-  this.els.container.addEventListener('click', (evt) => {
-    var button = evt.target.closest('button');
+  this.els.container.addEventListener('click', throttle((evt) => {
```
jimporter: I've been thinking about this; do we actually *need* to throttle the next/previous buttons to fix this bug? I can see a user wanting to tap "next" twice to quickly skip two songs ahead.

PR #32364, opened 2015-10-09
Bug 1213381 - [Music][NGA] Remove A-Z picker until we have a UX spec …

justindarc:
…for additional locales

### DIFF
--- a/apps/music/views/albums/view.js
+++ b/apps/music/views/albums/view.js
@@ -30,10 +30,6 @@ var AlbumsView = View.extend(function AlbumsView() {
   this.list.minScrollHeight = `calc(100% - ${searchHeight}px)`;
 
   this.list.configure({
-    getSectionName: (item) => {
-      return item.sectionName;
-    },
-
     getItemImageSrc: (item) => {
       return this.getThumbnail(item.name);
     }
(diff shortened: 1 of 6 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=cf4fb8770dc0dc9c5ea3e54ec1927aed679fbb0a

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4484132c7606f42c4450fe609e6f1414c68ef746

apps/music/views/albums/view.js
```diff
@@ -64,31 +60,19 @@ AlbumsView.prototype.render = function() {
 };
 
 AlbumsView.prototype.getAlbums = function() {
-  return document.l10n.formatValues('unknownAlbum', 'unknown')
-    .then(([unknownAlbum, unknown]) => {
+  return document.l10n.formatValues('unknownAlbum', 'unknownArtist')
```
jimporter: I'm not sure I understand why unknownArtist was added in this patch. Is that just another bug that you're fixing at the same time here?

apps/music/views/songs/view.js
```diff
@@ -71,24 +67,18 @@ SongsView.prototype.render = function() {
 
 SongsView.prototype.getSongs = function() {
   console.time('getSongs');
-  return document.l10n.formatValue('unknown')
-    .then((unknown) => {
+  return document.l10n.formatValues('unknownTitle', 'unknownArtist')
```
jimporter: Ditto here about the unknownArtist string.

apps/music/views/albums/view.js
justindarc: see below on line 73.. we show the `artist` as the sub-text in the list items and we weren't falling back to the 'unknownArtist' string if that field was blank

apps/music/views/songs/view.js
justindarc: Same reasons as above.

PR #32367, opened 2015-10-09
Bug 1213355 - GaiaHeader.go_back shows ElementNotVisibleException

mwargers:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=00c28dfb86819630f64244030ab535d389550098

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=40e25a4d1d61ac7178ce3adb218e0fad4d0c7fea

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/display.py
```diff
@@ -9,7 +9,7 @@
 
 class Display(Base):
 
-    _page_locator = (By.ID, 'sound')
```
JohanLorenzo: That's weird that we lived with this error for nearly 2 years. This line was added on november 2013 and never modified since.

In this class, we only use it in `screen_element`, which is used in many image_compare test. Was is it working by coincidence? Like the sound panel was always open right before the display panel so we return a coherent value? pinging @npark-mozilla who has a better view on the image compare tests.

tests/python/gaia-ui-tests/gaiatest/form_controls/form_control.py
```diff
@@ -8,5 +8,8 @@
 
 class Widget(PageRegion):
     def __init__(self, marionette, locator):
-        element = Wait(marionette).until(expected.element_present(*locator))
+        if type(locator) is tuple:
```
JohanLorenzo: I agree, that would be the way we do in JavaScript. But in python, we can avoid to force a parameter in lieu of another:
```python
def __init__(self, marionette, locator=None, element=None):
   if locator is None and element is None:
        raise AttributeError('Either the locator or the element should be provided')
   if element is None:
        element =  Wait(marionette).until(expected.element_present(*locator))
   PageRegion.__init__(self, marionette, element)
```

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/privacy_controls.py
```diff
@@ -57,12 +57,13 @@ def tap_trans_control(self):
...
-        if html_header_locator:
-            # Used by test_settings_PS_RTL.py
-            HTMLHeader(self.marionette, html_header_locator).go_back()
-        else:
-            GaiaHeader(self.marionette, self._header_locator).go_back()
+    def return_to_prev_menu(self, parent_view, exit_view, html_header=False):
+        from gaiatest.apps.settings.app import Settings
+        Settings(self.marionette).return_to_prev_menu(parent_view, exit_view, html_header=html_header)
+
+    def return_to_prev_menu2(self, parent_view, exit_view):
```
JohanLorenzo: How about `return_to_prev_menu_legacy()`?

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/display.py
npark-mozilla: Wait, I don't think this line was there  for 2 years- I think I added it during my RTL work recently (https://github.com/mozilla-b2g/gaia/commit/7a0ebe3d2ffac3162417b92c57c272286b2b5673#diff-d6a18b8bf566e3f2ed25c705ae628f7f), I found this at the same time martijn did, and he corrected it while testing it on the entire RTL test

JohanLorenzo: Oops, Pycharm tricked me then. It showed me that it was added in bug 909718, when I loaded the git log for that single line. Sorry about that

npark-mozilla: pycharm can show github info? I should look for that option

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=52fe68191402b4c8c958fbcbd45c785361a06b2a

npark-mozilla:
Martijn, I pushed my chages, but since my master was more recent than yours, we have those commits in between.  Could you squash/rebase those commits?  (I'm afraid that if I do it, i might mess things up) I checked that everything is passing now.

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/display.py
JohanLorenzo: It can show git info. I don't remember if I installed a plugin or not for that. The steps I did were:
1. highlight the line I'm interested in
2. right click -> git history -> show history for selection

PR #32379, opened 2015-10-10
Bug 942837 - [User Story] Show Alarm Time on Lock Screen R=prateekjadhwani

prateekjadhwani:
User Story:

As a user I want to view the time of my next alarm while my phone is locked - this enables me to verify that it is set without needing to unlock my phone and navigate to the clock app.

Acceptance Criteria:
1. When my phone is locked (but not in sleep mode) I can see that my alarm is set, and the time and date of my next alarm.
2. Only my next alarm is shown, for example if I have an alarm set for 8am every day I can only see tomorrows alarm.
3. If my alarm is removed, It is no longer shown on the lock screen.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=25e504c388eb756eeec47c3484fbe11d0daf87a9

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=35b8b50f4d2ad1779eecb8e8600f61e5fa852be1

apps/clock/js/alarm.js
```diff
@@ -175,6 +250,56 @@ define(function(require, exports, module) {
...
       });
+    },
+
+    /** Sort all the alarms and return the next upcoming
+     *  alarm from the list
+     */
+    sortAlarms: function(storeAlarms) {
+      storeAlarms.sort(function(a,b) {
+        return new Date('1970/01/01 '+a.hour+':'+a.minute)
+          - new Date('1970/01/01 ' + b.hour+':'+b.minute);
```
mcav: Is there a reason not to just compare `hour` and `minute` directly? Constructing a Date in this way seems unnecessary.

apps/clock/js/alarm.js
```diff
@@ -175,6 +250,56 @@ define(function(require, exports, module) {
           this._notifyChanged(/* removed = */ true);
         });
       });
+    },
+
+    /** Sort all the alarms and return the next upcoming
+     *  alarm from the list
+     */
+    sortAlarms: function(storeAlarms) {
```
mcav: This function should be named something a little more descriptive -- e.g. `getNextAlarm`

apps/clock/js/alarm.js
```diff
@@ -156,6 +156,81 @@ define(function(require, exports, module) {
...
+            if(a.registeredAlarms.normal) {
+              //Check for Repeated Alarms
+              if(Object.keys(a.repeat).length != 0) {
+                var hasToBePushed = false;
+                for (var n in a.repeat) {
+                  var day = n;
+                  if(now.getDay() == 6 && n < 6) {
+                    day = 7 + parseInt(n);
+                  }
+                  if(day == now.getDay() || day == now.getDay() + 1) {
```
mcav: You can get rid of most of this logic, and maybe some from sortAlarms, by using `getNextAlarmFireTime()`.

prateekjadhwani: I don't think `getNextAlarmFireTime()` does what it is supposed to do. If I have 4 alarms set in the clock app, say 1st one for 2 days from now, and others for 10 min from now, it always returns the first alarm, even though it is supposed to set off 2 days from today.
Or may be am I using it wrong?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=23bf94e20637e22f9974036beb6c4c03433562fb

apps/system/lockscreen/js/widgets/clock/lockscreen_clock_widget.js
```diff
@@ -50,6 +52,44 @@
     this.logger.debug('Clock updated', now);
   };
 
+	LockScreenClockWidget.prototype.updateAlarm =
+	function() {
+    var self = this;
+		// returns alarm info from DataStore
+		navigator.getDataStores('alarms')
+			.then( function(stores){
+        stores[0].getLength().then(function(len){
```
snowmantw: Could you just clear up the nested structure with some named functions registered in the widget? Also, for the multiple levels of `if...else`, I feel each branch deserves one method.

It's also necessary to split the data fetching functions out of rendering logic. A fetching method returns Promise should already provide enough interface to chaining the following rendering actions after data fetching in an elegant way, rather than just put a large callback in a `Promise.then` like the old style. For example:

    LockScreenClockWidget.prototype.fetchAlarmData = function() {
        .... return the promise of fetching data, although it's only a one liner.
    }

    LockScreenClockWidget.prototype.renderAlarm = function(fetchingResult) {
      if (fetchingResult.data /*...... to put the conditions here*/) {
        renderNoAlarm();   // there is no alarm according to your code
      } else {
        renderAlarmDateTime();
      }
    }

    // in the using section
   widget.fetchAlarmData()
   .then(widget.renderAlarm)

apps/system/lockscreen/js/widgets/clock/lockscreen_clock_widget.js
```diff
@@ -50,6 +52,44 @@
...
+          if(len > 0) {
+            stores[0].get(1)
+              .then( function(data) {
+                if(data.data) {
+                  if(data.data.hour && data.data.minute) {
```
snowmantw: If the `if(data.data)` only has one child branch, and it's `else` just do nothing, you can write it as

    if (!data.data) {
        return;
    }
    if (data.data.hour && data.data.minute) {
        // ....
    }

In this way we reduce the unnecessary sections just because you want to do nothing if there is no data after all.

snowmantw: And also it's very confused to see code like

    if (data.data.....)

Could you please change it as

    if (fetchingResult.data)

or something like that?

apps/system/lockscreen/js/widgets/clock/lockscreen_clock_widget.js
```diff
@@ -50,6 +52,44 @@
     this.logger.debug('Clock updated', now);
   };
 
+	LockScreenClockWidget.prototype.updateAlarm =
+	function() {
+    var self = this;
```
snowmantw: Please stop using the old `self` trick while you can use `bind` or arrow function `() => {}`.

apps/system/lockscreen/js/widgets/clock/lockscreen_clock_widget.js
```diff
@@ -50,6 +52,44 @@
...
+		navigator.getDataStores('alarms')
+			.then( function(stores){
+        stores[0].getLength().then(function(len){
+          if(len > 0) {
+            stores[0].get(1)
+              .then( function(data) {
+                if(data.data) {
+                  if(data.data.hour && data.data.minute) {
+                    var type = 'AM',
+                        hour = parseInt(data.data.hour);
```
snowmantw: It's a good convention to give the radix of `parseInt`.

apps/system/lockscreen/js/widgets/clock/lockscreen_clock_widget.js
```diff
@@ -50,6 +52,44 @@
...
+                      type = 'PM';
+                      hour = hour - 12;
+                    }
+
+                    var fullAlarmTime = hour + ':' + data.data.minute + type;
+                    self.resources.elements.alarmtime.textContent =
+                      fullAlarmTime;
+                    self.resources.elements.alarm.classList.remove('no-alarms');
+                    self.logger.debug('Alarm updated to :', fullAlarmTime);
+                    return data;
```
snowmantw: If the main purpose of this function is to render the thing, you don't need to return the data.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=bf331cfe0b7dc9afda251376b31d2c936313d4d9

apps/clock/js/alarm.js
```diff
@@ -156,6 +156,48 @@ define(function(require, exports, module) {
       // Only update the application if this alarm was actually saved
       // (i.e. it has an ID).
       if (this.id) {
+        var self = this,
```
snowmantw: Please, stop using this...you can check what I've said at

https://github.com/mozilla-b2g/gaia/blob/master/apps/system/lockscreen/js/lockscreen_inputpad.js#L63-L66

It's an example to use bind instead of the old trick of refering `this`.

Or, if you're willing to learn more, and which I feel it's more proper, you can take a look at ArrowFunction:

https://github.com/mozilla-b2g/gaia/blob/master/apps/system/lockscreen/js/lockscreen_inputpad.js#L77-L81

The MDN links are:

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_objects/Function/bind
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions

snowmantw: And although it's in Clock app, the reason of avoiding refering `self` regards JavaScript practice, is that `self` is in fact a occupied keyword which refer to `window` in main thread and `global` in a worker. So it's not really good to shadow it in the code. Some people choose to use `_this` or `THIS` or `that` or some other tricks. They're still unnecessary when you can use `bind` or  Arrow Functions, but better to redefine the exiting `self`.

apps/system/lockscreen/js/widgets/clock/lockscreen_clock_widget.js
```diff
@@ -50,6 +52,52 @@
...
+      if(hour > 12) {
+        type = 'PM';
+        hour = hour - 12;
+      }
+
+      var fullAlarmTime = hour + ':' + alarmData.minute + type;
+      this.resources.elements.alarmtime.textContent = fullAlarmTime;
+      this.resources.elements.alarm.classList.remove('no-alarms');
+      this.logger.debug('Alarm updated to :', fullAlarmTime);
+    }).catch(() => {
```
snowmantw: Please don't use `then` and `catch` as control flow...

apps/system/lockscreen/js/widgets/clock/lockscreen_clock_widget.js
```diff
@@ -50,6 +52,52 @@
...
+  /*
+   * Gets the Alarm Data from the DataStore.
+   */
+  LockScreenClockWidget.prototype.fetchAlarmData = function() {
+    return new Promise((resolve, reject) => {
+      navigator.getDataStores('alarms')
+      .then((stores) => {
+        stores[0].getLength()
+        .then((len) => {
+          if(len > 0) {
```
snowmantw: Please invert this as I mentioned:

    if (len === 0) {
      return;
    }

Since it's an only one-branch conditional statement.

apps/system/lockscreen/js/widgets/clock/lockscreen_clock_widget.js
```diff
@@ -50,6 +52,52 @@
...
+    }).catch(() => {
+      this.resources.elements.alarm.classList.add('no-alarms');
+    });
+	};
+
+  /*
+   * Gets the Alarm Data from the DataStore.
+   */
+  LockScreenClockWidget.prototype.fetchAlarmData = function() {
+    return new Promise((resolve, reject) => {
```
snowmantw: You really don't need to wrap another Promise when the inner methods are Promises. And it's not good to use `reject` as control flow.

apps/system/lockscreen/js/widgets/clock/lockscreen_clock_widget.js
```diff
@@ -50,6 +52,52 @@
...
+      navigator.getDataStores('alarms')
+      .then((stores) => {
+        stores[0].getLength()
```
snowmantw: Rather to create nested structure like this, it should be:

    navigator.getDataStores('alarms')
    .then((stores) => {
      return stores[0].getLength();
    }).then((len) => {
        if (0 === len) { return; }
        // others
        return stores[0].get(1);
    }).then((fetchedData) => {
        // do data handling
    })

In a `then` you can definitely return another Promise, and the step will be blocked until the inner Promise returns. So we can flatten the whole structure.

apps/system/lockscreen/js/widgets/clock/lockscreen_clock_widget.js
snowmantw: You can design a empty object or something represents `null` data or no data, but an Error should be catch should really be an Error, not acceptable values.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8e7847c12eef3ca9fbf2f00bec6f1ab3d4f9154e

snowmantw:
Then you may use this tactic: 

    navigator.getDataStores('alarms')
    .then((stores) => {
      return [stores[0].getLength(),
                 stores[0]];
    }).then(([len, store]) => {
        if (0 === len) { return; }
        // others
        return store.get(1);
    }).then((fetchedData) => {
        // do data handling
    })

Since it's legit to destruct array in the parameter list. Namely, you can do this:

   var f = ([a, b]) => a + b
   f([1,2])    // 3

Although I think the real issue is we need to do two asynchronous queries before we perform the data handling, but since it's somewhat API issue so let's give a nice and quick fix. If you have other concerns please comment it here, thanks.

prateekjadhwani:
Creating a fresh PR in https://github.com/mozilla-b2g/gaia/pull/32955

PR #32397, opened 2015-10-12
Bug 1213946 - Use |now.toString()| instead of |now| in month_day_agen…

jandem:
…da_test.js

In bug 1187233 we're changing |new Date(now)| to create a copy (which will preserve milliseconds). To keep the current behavior, the test should use |new Date(now.toString())| instead.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b94995626714cfa4e6317aa807ecc8dab42034be

jandem:
Closing as I opened a new one.

PR #32410, opened 2015-10-13
Bug 1212960 - Remove position fixed breaking style from fullscreen an…

mikehenrty:
…cestors

### DIFF
--- a/apps/system/style/window.css
+++ b/apps/system/style/window.css
@@ -591,6 +591,18 @@ html[dir="rtl"] .appWindow.invoked {
   100% { transform: translateY(calc(-100%)); }
 }
 
+/*
+  When a mozbrowser iframe requests fullscreen with SHB enabled,
+  Gecko will make that element fixed and take up the entire viewport.
+  However, due to bug 879443 the transform property will make fixed
+  relative to this ancestor rather than the viewport, so we unset
+  the transform for this case.
+  See: https://bugzilla.mozilla.org/show_bug.cgi?id=1212960
+*/
+#screen:-moz-full-screen-ancestor > #windows > .appWindow:not(.homescreen):not(.lockScreenWindow):not(.attentionWindow):not(.lockScreenInputWindow).active {
+  transform: unset;
+}
+
 /* Edges */
 #windows > .appWindow:not(.homescreen):not(.lockScreenWindow):not(.attentionWindow):not(.lockScreenInputWindow) {
   /* not in the viewport */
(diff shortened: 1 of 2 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3446a4336c2ff6b0317272506c65491251c6f1ff

apps/system/js/layout_manager.js
```diff
@@ -194,6 +203,16 @@
             this.publish('system-resize', systemResizeEventDetail);
           }
           break;
+
+        case 'mozfullscreenchange':
+          this.screen.classList.toggle('fullscreen-ancestor',
+            document.mozFullScreen);
+          if (evt.type === 'keyboardhide') {
```
etiennesegonzac: not sure I understand this part, we're `switch`ing on `evt.type` already

apps/system/style/window.css
```diff
@@ -591,6 +591,18 @@ html[dir="rtl"] .appWindow.invoked {
...
+  Gecko will make that element fixed and take up the entire viewport.
+  However, due to bug 879443 the transform property will make fixed
+  relative to this ancestor rather than the viewport, so we unset
+  the transform for this case.
+  See: https://bugzilla.mozilla.org/show_bug.cgi?id=1212960
+*/
+#screen.fullscreen-ancestor > #windows > .appWindow:not(.homescreen):not(.lockScreenWindow):not(.attentionWindow):not(.lockScreenInputWindow).active {
+  transform: unset;
+}
+
```
etiennesegonzac: It's going to be pretty confusing to have `#screen:-moz-full-screen-ancestor` rules and `#screen.fullscreen-ancestor` rules :)

What's the difference? Maybe we can find a better name.

apps/system/js/layout_manager.js
mikehenrty: Whoops this is a mistake. I was just copying what happens in the `default` case to maintain current behavior for `'mozfullscreenchange':` (and I don't like not have a `break` and having case statements fall through). The publish for system-resize should be enough.

apps/system/style/window.css
mikehenrty: In effect there is no difference, and that is what we are going for. According to [1], the -moz-full-screen-ancestor is going away, so I was proposing an alternative to that. I'm happy to use -moz-full-screen-ancestor since it's more consistent and we can then change everything over at once later.

1.) https://bugzilla.mozilla.org/show_bug.cgi?id=1212960#c11

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9ae132e5a0df05b91f79233f47c0d76b6d59b7c8

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0bbdfe474e8eb5d52820b171bbce6a047b54bd59

PR #32413, opened 2015-10-13
Bug 1210731 - [SMS] CSS Refactoring for RTL

autra:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=bb22df736c11e49087b6d167eeeac1b5c6ce9af3

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1e7b55606664f568d4a2732e7961f5073c2dface

apps/sms/views/conversation/style/attachment.css
```diff
@@ -127,15 +125,23 @@
   overflow: hidden;
   text-overflow: ellipsis;
 }
+/* .file-name has a dir="auto". Moreover, sms.css in not included in attachment
```
autra: For the `text-align` changes that follows, it is not exactly equivalent to before, but I think it is more consistent now.

**Before patch**
conversation_attachment_01

**After**
conversation_attachment_01

(Please don't pay attention to the phone icon, it will be fixed in the bug for gaia-icons)

apps/sms/views/conversation/style/conversation.css
```diff
@@ -157,7 +157,3 @@
 #messages-header-text .ellipsis-dir-fix {
   width: auto;
 }
-
-.conversation-view .panel {
```
autra: I'm pretty sure this is dead code.

apps/sms/views/shared/style/composer.css
```diff
@@ -186,16 +186,21 @@ form[role="search"] button[type="submit"]:after {
...
+html[dir="ltr"] #messages-send-button {
+  background-image: url(images/icons/actionicon_sms_send_30x30.png);
   /* When we put the icon at the center, an optical illusion makes it look to be
    * slightly to the left, so we offset the button by 2px to the right to
    * counteract this illusion */
   transform: translate(.2rem);
 }
+html[dir="rtl"] #messages-send-button {
+  background-image: url(images/icons/actionicon_sms_send_30x30_rtl.png);
+  transform: translate(-.2rem);
```
autra: This rule did not exist before. I feel that *if* we reckon it appears slightly to the left in LTR, then it *also* appears slightly to the right in RTL.

TBH, I'd prefer 0.1 rem or nothing :-) It is quite subjective imo (and quite subtle also, so that's ok I guess).

build/csslint/xfail.list
```diff
@@ -64,9 +64,16 @@ apps/settings/style/lists.css 2 1
 apps/settings/style/screen_lock.css 0 2
 apps/settings/style/settings.css 0 9
 apps/settings/style/settings_large.css 1 1
-apps/sms/views/inbox/style/inbox.css 0 2
+apps/sms/views/conversation/style/conversation.css 4 0
```
autra: I wish b2g_sdk would update... :-/

apps/sms/views/conversation/style/attachment.css
```diff
@@ -153,8 +159,11 @@
   width: 100%;
   height: 100%;
 
-  background: url('images/attachments.png') no-repeat left top / 10rem;
+  background: url('images/attachments.png') no-repeat;
+  background-size: 10rem;
 }
+html[dir="ltr"] .thumbnail-placeholder { background-position: left top; }
+html[dir="rtl"] .thumbnail-placeholder { background-position: right top; }
```
julienw: are you sure of this change ? isn't it a css sprite as the next ones ?

julienw: left and right are the same here actually :)

apps/sms/views/conversation/style/attachment.css
```diff
@@ -171,23 +180,33 @@
...
-  background: url('images/attachment-vcard-incoming.png') no-repeat left top / 4rem;
+  background: url('images/attachment-vcard-incoming.png') no-repeat;
+  background-size: 4rem;
+}
+html[dir="ltr"] .incoming.received .vcard-placeholder {
+  background-position: left top;
+}
+html[dir="rtl"] .incoming.received .vcard-placeholder {
+  background-position: right top;
 }
```
julienw: can't we just use `.incoming.received .vcard-placeholder { background-url: url(right url); }` and get rid of the other properties ?

apps/sms/views/conversation/style/conversation.css
julienw: not sure, I think it's used in views/conversation/index.html :) `git grep conversation-view` is your friend

apps/sms/views/conversation/style/message.css
```diff
@@ -35,11 +40,15 @@
...
+  background: #39c3c2;
+}
+html[dir="ltr"] .message.incoming .bubble {
   float: left;
-
   border-radius: 0 1.5rem 1.5rem 1.5rem;
-
-  background: #39c3c2;
+}
+html[dir="rtl"] .message.incoming .bubble {
```
julienw: just wondering why we don't use `:-moz-dir(rtl)` ?

autra: Here in particular, or in general?

In general, that's a convention we took with the rest of the team. Basically, that's the best css equivalent of "document.dir", aka the direction of the language. The semantic of :-moz-dir is a bit different, and I'm pretty sure it is also useful is some case (but here, I don't see the difference).

I do like it because it protects your layout to be messed up if a `dir` is added in a parent element, including dir="auto". As I've seen some `dir = ltr` used in place of `text-align` or `float` or... quite a few times even in current master, I feel it is more robust this way. 

And it might be better performance-wise, because `:-moz-dir(rtl)` might need to walk the parents of current element until it found one with a 'dir' (I'm really not sure about that though. Maybe this info is available on the current node anyway).

Did I understand your question well?

apps/sms/views/conversation/style/attachment.css
autra: Yep. I'll revert this and put a comment instead.

apps/sms/views/conversation/style/message.css
julienw: I don't know how `:-moz-dir(rtl)` works but I know for sure that adding another component to a selector has a performance impact.

I was just curious, I'm not saying one is better than the other. I guess that in _some_ cases it will be obvious that one is better than the other, and in most cases we don't really care.

apps/sms/views/conversation/style/conversation.css
autra: I meant that you have a 
```css
.panel {
  /* ...*/
  left: 0;
  /* ... */ 
}
 ```
In `./apps/sms/views/shared/style/panel.css:1`. Dead code in the sense of a liveliness analysis, if this makes sense :-) I might have missed something though.

**Comment by autolander** [comment]
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=eb89f0101fde2a802d0e13695344391e821a83f7

File: apps/sms/views/conversation/style/conversation.css (hunk 2 above)
Comment (julienw) [inline apps/sms/views/conversation/style/conversation.css]: mmm got you, you're likely right. The rule you're talking about was likely added in the recent bug 1162030. I'll look closer later :)

**Comment by autolander** [comment]
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=01337717eb6425e055df94a50c35ec0f3c255346

**Comment by autolander** [comment]
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2509365bdfa39bf0e9d0536d016b6fff7597d318

File: build/csslint/xfail.list (hunk 4 above)
Comment (julienw) [inline build/csslint/xfail.list]: it's moving forward :)

File: apps/sms/views/conversation/style/conversation.css (hunk 2 above)
Comment (julienw) [inline apps/sms/views/conversation/style/conversation.css]: can you remove the class "conversation-view" in views/conversation/index.html then ?

File: apps/sms/views/inbox/style/edit-mode.css [hunk 8]
```diff
@@ -17,24 +17,20 @@
...
+  transform: translateX(-5rem);
+}
 
 .edit #threads-container aside.pack-end {
-  -moz-margin-start: 4rem;
+  margin-inline-start: 4rem;
 }
 
-.edit #threads-container[data-type="list"] aside.pack-end span {
+html[dir="ltr"] .edit #threads-container[data-type="list"] aside.pack-end span {
```
julienw: what an enormous sselector :(

julienw: but nothing you can really do

apps/sms/views/inbox/style/inbox.css
```diff
@@ -106,9 +106,13 @@
 }
 
 [data-type="list"] .unread aside.icon-unread {
-  background: url("images/unread.png") no-repeat left center / 1rem;
+  background-image: url("images/unread.png");
+  background-repeat: no-repeat;
```
julienw: I think you can keep the background property for the image and repeat values.
rationale: it sets all other properties to their initial values (so we don't risk being bitten by some inherited stuff here)

apps/sms/views/inbox/style/inbox.css
```diff
@@ -122,6 +126,12 @@
   background: url("images/draft.png") no-repeat left center / 1.2rem;
```
julienw: don't you want to change this as well ?

apps/sms/views/inbox/style/inbox.css
```diff
@@ -122,6 +126,12 @@
...
   visibility: visible;
 }
+html[dir="ltr"] [data-type="list"] .draft .icon-draft {
+  background-position: left center;
+}
+html[dir="rtl"] [data-type="list"] .draft .icon-draft {
+  background-position: right center;
+}
```
julienw: wondering it we can have only one set of rules for both .icon-draft and .icon-unread by using simply .icon.

julienw:
When in RTL, the text is aligned left in the composer, but right in the subject. Is it expected ?

apps/sms/views/new_message/style/recipients.css
```diff
@@ -143,56 +143,25 @@ Do not remove.
...
-   */
-  margin-left: -0.5rem;
-  margin-right: 0;
+html[dir="rtl"] .recipient[contenteditable=true]:focus,
+html[dir="rtl"] .recipient[contenteditable=true]:hover,
+html[dir="rtl"] .recipient[contenteditable=true]:empty:focus {
+  margin-left: 0.5rem;
+  padding-left: 0.3rem;
 }
+
```
julienw: you did a really good job here, this is a very tricky part of the SMS app and I can't find any issue !

Still i'd like that you try to keep as many comments as you can, because they give useful explanations about some choices we made here.

apps/sms/views/shared/style/composer.css
julienw: yes I think it's better like this after looking closely once more :)

apps/sms/views/shared/style/composer.css
```diff
@@ -186,16 +186,21 @@ form[role="search"] button[type="submit"]:after {
...
+}
+html[dir="ltr"] #messages-send-button {
+  background-image: url(images/icons/actionicon_sms_send_30x30.png);
   /* When we put the icon at the center, an optical illusion makes it look to be
    * slightly to the left, so we offset the button by 2px to the right to
    * counteract this illusion */
   transform: translate(.2rem);
 }
+html[dir="rtl"] #messages-send-button {
+  background-image: url(images/icons/actionicon_sms_send_30x30_rtl.png);
```
julienw: so you prefer having 2 icons instead of 1 that is mirrored ? I can understand, it makes things quite messy with the sim indicator ;)

apps/sms/views/inbox/style/edit-mode.css
autra: Oh can't I just throw `.edit` out? As #threads-container is an id, it should be unnecessary right?

apps/sms/views/shared/style/panel.css
```diff
@@ -5,34 +5,51 @@
...
 
   animation: .2s both linear;
 }
+html[dir="ltr"] .panel { transform: translate(100%); }
+html[dir="rtl"] .panel { transform: translate(-100%); }
 
-.panel-active {
+/**
+ * adding html[dir] to make the specificity of this rule high enough to
+ * overwrite the default translation for element with class panel, without
```
julienw: maybe you can add "just above" so that it's easier to understand you're talking about the rules that are just above.

apps/sms/views/shared/style/panel.css
```diff
@@ -5,34 +5,51 @@
...
+  animation-name: old-slide-right
+}
 
-.panel-will-deactivate.panel-animation-reverse {
+html[dir="ltr"] .panel-will-deactivate.panel-animation-reverse {
   animation-name: old-slide-right;
 }
+html[dir="rtl"] .panel-will-deactivate.panel-animation-reverse {
+  animation-name: old-slide-left;
+}
```
julienw: Mmm I don't think these classes are used at all, actually. I think this comes from one of the earlier versions of bug 1162030... I changed using `style.animationName` because it was faster in one of the latest iterations of the patch, and that's why these css rules are still there.

Maybe you could simply remove all this ?

apps/sms/views/inbox/style/edit-mode.css
julienw: no no, `.edit` is like a boolean here (like dir="ltr"). Sometimes it's here, sometimes it's not.

apps/sms/views/conversation/style/message.css
autra: Yeah, probably. Actually I spent some time reading about selector performances yesterday, and one thing that seems to stand out every time is that we shouldn't try to optimize CSS selectors until it is obvious we need to :-)

julienw: I think we're at a point where we need to optimize from different directions -- and fixing some of our awful selectors is one of the directions.

autra:
> When in RTL, the text is aligned left in the composer, but right in the subject. Is it expected ?

the text is aligned left in the composer when your paragraphe is LTR. When you type with the arabic keyboard, it is right-aligned. See:
sms_subject_rtl

Yes, it is the expected behaviour (alignment per paragraphe in the composer).

apps/sms/views/inbox/style/inbox.css
autra: Or we can simply delete these `background-position`, as they are already defined in lists.css (that has been RTL-refactored).

julienw:
> Yes, it is the expected behaviour (alignment per paragraphe in the composer).

But why don't we do the same thing in the subject then ?

apps/sms/views/inbox/style/inbox.css
julienw: if it works I don't mind :)

autra:
>>    Yes, it is the expected behaviour (alignment per paragraphe in the composer).

>But why don't we do the same thing in the subject then ?

We could. 

We force the alignment of one-line fields for visual consistency, but that's just because they usually are in a form with other labels, dropdown menu etc... Here it is not that much the case. There's still the "MMS" label, but I think we don't really care.

So probably it'll look a bit better if we let the natural alignment on the `subject` field. OTOH, we'll loose consistency with the email app, that will most surely force the alignment, as they have other fields (to, cc...) that they'd need to force-align (probably because of icons and labels). Another cons is that we'll need to explain that when bugs will be open about it.

So, there's pros and cons. I don't know. Your call :-)

apps/sms/views/new_message/style/recipients.css
autra: Ok I'll try to be a bit more verbose. OTH, I thought we were very verbose before, just to say something like "We cannot use start/end properties with dir auto".

I can elaborate on comment line 85, and refer to it in other parts, WDYT?

For the comment that was line 169, I cannot reproduce any more, even with the screenreader. I think the general
```css
html[dir="ltr"] [dir] { text-align: left; }
html[dir="rtl"] [dir] { text-align: right; }
```
fixed it anyway.

And yes, I have been a bit quick with the comment that was line 146, I should put it back.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b8f536c62241c7eaadedf118c3e9ad6a1b61113c

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b1137e4fca1ca26bff555ddd61e142896ca078ad

julienw:
> So, there's pros and cons. I don't know. Your call :-)

yeah, not really my call :) Do we have dedicated UX for RTL work ?

autra:
Yes, and the specs are here: https://mozilla.app.box.com/s/sp6lzaf2aimv4i7juafelthg74e6pgys/1/4635667922/39239281605/1

MMhh just checked. The specs says that every text inputs should be right-aligned in RTL and do not give any precisions on multi-line input texts. Too bad I didn't see that when reviewing them :-( 

I'll ask for it to be mentioned then.

autra:
Actually, it was not in the specs because Rob was waiting for someone to object. As nobody has, I've just asked him to update them. I would suggest to land this as is, and I'll do a follow-up if we change our mind. How does that sound to you?

autra:
Ok, I've realised that I've switched topic. Let me rephrase:
- the Subject field must be right-aligned in RTL, no debate on this.
- the specs says that text in the Composer input area should also be right aligned in RTL, but the Phoxygen team proposed to have a per-paragraph alignment (as it is now btw). This proposal have received no objection yet, so *should* be accepted soon.

julienw:
good for me if UX is ok :)

apps/sms/views/new_message/style/recipients.css
```diff
@@ -79,16 +81,16 @@ Do not remove.
   overflow-y: scroll;
 }
 
+/**
+ * BiDi note: all [contenteditable] blocks must carry a `dir="auto"` attribute,
+ * so we cannot use *-inline-{start,end} properties on them, because the
+ * direction of a dir="auto" element (and therefore the position of start/end)
+ * depends on its content. We want our margin/padding to depend on the document
+ * direction, not element's content.
```
julienw: note that for margin, logical properties use the element's parent's direction, not the element's direction, which makes a lot more sense.
(this was not working like this when we were using -moz-margin-start/end, but they changed it for margin-inline-start/end).

See https://drafts.csswg.org/css-logical-props/#propdef-margin-inline-start. I'm quite sure Gecko implements this.

So maybe we can use the logical props after all ?

julienw: note, for padding, this is still using the element's direction though.

apps/sms/views/new_message/style/recipients.css
```diff
@@ -96,19 +98,21 @@ Do not remove.
...
+  background: #fff url(images/icons/icon_sms_compose_email.png) no-repeat;
+  background-size: 1.7rem;
+}
+/* see BiDi note above */
+html[dir="ltr"] .recipient[contenteditable].email {
   padding-left: 3.8rem;
-  background: #fff url(images/icons/icon_sms_compose_email.png) left 1.5rem
-              center / 1.7rem no-repeat;
+  background-position: left 1.5rem center;
+}
```
julienw: nit: please add an empty line

apps/sms/views/new_message/style/recipients.css
```diff
@@ -96,19 +98,21 @@ Do not remove.
   height: auto;
   width: auto;
 }
```
julienw: nit: please add an empty line

julienw: mmm after all nope, I understand why you're doing this.

apps/sms/views/new_message/style/recipients.css
julienw: but I'm fine with keeping this like this for now, and looking at this in a follow-up. I'd rather see this in sooner than later.

apps/sms/views/new_message/style/recipients.css
```diff
@@ -138,61 +142,39 @@ Do not remove.
...
-  margin-right: 0;
-
-  /**
-   * We need "text-align: right" only for the cases when currently editable
-   * recipient looses focus but not yet assimilated: case №1 - when we tap on
-   * thread-container - maybe fixed soon, case №2 - when we use screen reader,
-   * eg. single touch on message input steals focus from currently editable
-   * recipient, but doesn't yet give it to the message input. As the result
-   * flex-grow (see workaround for bug 1103011) returns and because of dir=auto
-   * LTR\RTL name would be always left\right aligned in both LTR and RTL modes.
```
julienw: the screen reader is too crashy so I can't check this currently... I guess we can leave it for now and come back once we can check it.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9e138731ff6559b9468a28a0977cb9595c008444

PR #32418, opened 2015-10-13
Bug 1151847 - [Network Alerts][Refactoring] Use notification "data" attribute instead of icon URL query string parameters

Ilphrin:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=cfa7e52f833ad98e936c15cb51a2976ebde549b3

apps/network-alerts/js/attention/attention.js
```diff
@@ -42,7 +42,10 @@ function sendNotification() {
         navigator.mozL10n.get(title), {
           body: body,
           tag: '' + Date.now(), // needs to be unique
-          icon: NotificationHelper.getIconURI(app) + '?titleID=' + title
+          icon: NotificationHelper.getIconURI(app),
+          data: {
+            "title": title
```
azasypkin: __nit:__ we don't use quotes for object properties unless it's JSON file, so I'd rather have ```data: { title }``` - notice new ES6 shorthand syntax [1] ;)

[1] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer

apps/network-alerts/js/attention/attention.js
```diff
@@ -42,7 +42,10 @@ function sendNotification() {
         navigator.mozL10n.get(title), {
           body: body,
           tag: '' + Date.now(), // needs to be unique
-          icon: NotificationHelper.getIconURI(app) + '?titleID=' + title
+          icon: NotificationHelper.getIconURI(app),
+          data: {
```
azasypkin: __issue:__ It's good start! So let's learn few more things :)

* (simplified) We have Treeherder that we can use to see results of unit/integration and all other types of tests that are run with your patch, you can find more info about it at [1]. So once you create your PR, Autolander [2] runs tests with your PR and pastes the link to the test results as a comment to your PR [3]. When we say ```Treeherder is green``` that means all tests are passing there. Take a look at [3]  we have two jobs that are failing because of your PR:
  * ```GU12``` - GU stands for "gaia unit tests" - you need to fix the unit test :) Use [4] as a guide and ask question if needed.
  * ```JSHINT``` - it's just jshint check that falied, it's not always easy to quickly understand if it's related to your patch, but you can click on the job, then click on "Open the raw log" button and figure out what exactly the reason.

* So you've changed the way we generate notification, good, but you should also change the way we use to parse this data back when user clicks on notification - I haven't tested, but the title of the message will be wrong when you tap on notification (via notification tray) to see network alert once again. Check this code [5] - when user taps on notification in notification tray, special "notification" system message is generated (argument is the notification you've created previously) that will wake up application and display alert message once again [6] - the change here will likely require related unit test change as well.

[1] https://developer.mozilla.org/en-US/docs/Mozilla/Firefox_OS/Automated_testing/Treeherder
[2] https://github.com/mozilla/autolander
[3] https://github.com/mozilla-b2g/gaia/pull/32418#issuecomment-147797476
[4] https://developer.mozilla.org/en-US/Firefox_OS/Automated_testing/Gaia_unit_tests
[5] https://github.com/mozilla-b2g/gaia/blob/master/apps/network-alerts/js/notification/notification.js#L15
[6] https://developer.mozilla.org/en-US/docs/Web/API/Navigator/mozSetMessageHandler

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=62db699b4341c1645b6e8d11d6ec41404dd88a11

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=43ae9e7e71caa34a6264123c9b54e800d6b68fad

apps/network-alerts/test/unit/attention/attention_test.js
```diff
@@ -107,7 +107,6 @@ suite('Network Alerts - Attention Screen', function() {
 
       assert.equal(MockNotifications[0].title, localizedTitle);
       assert.equal(MockNotifications[0].body, body);
-      assert.ok(MockNotifications[0].icon.endsWith('titleID=' + title));
```
azasypkin: __nit:__ please add assert that will verify that we passed ```title``` as data property as well

apps/network-alerts/js/notification/notification.js
```diff
@@ -18,12 +18,17 @@ function onNotification(message) {
     return;
   }
 
-  var title = Utils.parseParams(message.imageURL).titleID;
+  var title = null;
+  if (message.data) {
+    title = message.data.title;
+  }
+  else {
+    title = Utils.parseParams(message.imageURL).titleID;
```
azasypkin: __question:__ I think we don't need this anymore, why do you think we still need it? If it's just for the backward compatibility - don't worry and just keep ```message.data.title`` :)

apps/network-alerts/js/notification/notification.js
```diff
@@ -18,12 +18,17 @@ function onNotification(message) {
...
+  }
 
   var url = [
-    'attention.html?title=',
```
azasypkin: __issue:__ I believe you should not remove this URL parameter, it's passed to ```attention.html``` -> ```attention.js``` and then used to display emergency alert title - you can see how it look like when you open network alert once again via tapping on notification in notification tray.

apps/network-alerts/test/unit/notification/notification_test.js
```diff
@@ -42,7 +42,6 @@ suite('Network Alerts - Notification handling', function() {
 
     var expectedUrl = [
       'attention.html?',
-      'title=titleID&',
```
azasypkin: And once you return ```title``` back you won't need this change :)

apps/network-alerts/js/notification/notification.js
Ilphrin: Because when I don't keep the backward compatibility sometime notification-test.js fails because message.data.title doesn't exist. With this it was working well ^^

Ilphrin:
Ok! I'll work on it =)

apps/network-alerts/js/notification/notification.js
azasypkin: Hmm, interesting, it should not! :) If something is failing, we need to fix it :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=acb298f72f0caa961b683e0dd9acb7e4cdd3ddc3

apps/network-alerts/js/notification/notification.js
```diff
@@ -18,12 +18,12 @@ function onNotification(message) {
     return;
   }
 
-  var title = Utils.parseParams(message.imageURL).titleID;
+  var title = message.data.title;
 
   var url = [
     'attention.html?title=',
     encodeURIComponent(title),
-    '&body=',
```
azasypkin: accidentally removed ```&```?

apps/network-alerts/test/unit/attention/attention_test.js
```diff
@@ -107,7 +107,7 @@ suite('Network Alerts - Attention Screen', function() {
...
+      assert.ok(MockNotifications[0].data.title !== undefined);
```
azasypkin: __nit:__ better ```assert.equal(MockNotifications[0].data.title, title)```

apps/network-alerts/test/unit/notification/notification_test.js
```diff
@@ -34,7 +34,8 @@ suite('Network Alerts - Notification handling', function() {
     var message = {
       title: 'Some title',
       body: 'Some body',
-      clicked: true
+      clicked: true,
```
azasypkin: __nit:__ you don't need the piece of code above now since you don't parse URL anymore (hence the change in the test will be needed, will leave it to you to figure out what exactly :))
```
this.sinon.stub(Utils, 'parseParams').returns({
       titleID: 'titleID'
     });
```

apps/network-alerts/js/notification/notification.js
```diff
@@ -18,12 +18,12 @@ function onNotification(message) {
     return;
   }
 
-  var title = Utils.parseParams(message.imageURL).titleID;
+  var title = message.data.title;
```
azasypkin: __nit:__ if you don't use ```Utils``` in this file you won't need jshint directive as well ``` /* global Utils */``` (see top of the file) - same in tests. Moreover you don't need it in ```notification.html``` - clear performance win!

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b71ed0f0fe63b9043d4f6cc8881336039a6ffa53

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=53d63dd465831eb878a33d0e0d41e8edd9b56b80

apps/network-alerts/test/unit/notification/notification_test.js
azasypkin: This only nit is left :) Since you don't use ```Utils``` inside ```notification.js``` you don't need __anything__ related to ```Utils``` in test as well (in jshint global, in ```require(....mock_util)```, no need in ```MocksHelper``` since it has been used only to mock ```Utils``` and ```this.sinon.stub(Utils, 'parseParams')``` as well )

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=70c476d6e4dbb215127ade4cc585aa0f54700d7e

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b726e1e23b3dbe7ae00f6dc299528a5244fc66e9

PR #32430, opened 2015-10-14
Bug 1169154 - force update timezone string to prevent shows a blank area, r=yzen

gasolin:
when the automatic timezone is the same as user specified timezone, the timezone change event is not being triggered, so we need force update it.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8c23e4e4be92f6d392140e0b58de1c12fb9865bf

PR #32432, opened 2015-10-14
Bug 1197231 - Increase in SMS app memory consumption. Introduce switch for the SharedWorker services.

azasypkin:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d4ebb95d4b8ef626d9103b2cd561e68ec4be51a1

apps/sms/views/shared/js/startup.js
```diff
@@ -94,7 +97,20 @@ var Startup = exports.Startup = {
...
+      // Fetch mmsSizeLimitation and max concat
+      Settings.init();
+
+      var serviceEndpoint;
+      const servicePath = '/services/js/messaging/messaging_service.js';
+      if (Settings.useSharedWorkerServices) {
+        serviceEndpoint = new SharedWorker(servicePath);
+      } else {
+        serviceEndpoint = exports;
+        LazyLoader.load(servicePath).then(() => MessagingService.init());
```
azasypkin: We also can lazy load service file all the time and then make ```delete window.MessagingService``` as well, I don't like both methods, so didn't bother much :)

julienw: yeah, my only fear is this can delay launch time. Did you measure this ?

azasypkin: Yeah, I think it was about +15ms for fullyLoaded - could be a margin of error though, but I totally agree I'd not do it for 1mb win, or maybe just switch v2.5 totally to window context service without any ifs

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=95b7d811d7f43750ba8670c9ecde3ad2b9170b9e

PR #32446, opened 2015-10-14
Bug 1214509 - Template interpolate method with default parameters

sroucheray:
- Set default values to Template.prototype.interpolate parameters
- Test for undefined data properties and throw errors
- Swap safe.indexOf with safe.includes and invert the logic of the ternary operator

### DIFF
--- a/apps/sharedtest/test/unit/template_test.js
+++ b/apps/sharedtest/test/unit/template_test.js
@@ -61,8 +61,10 @@ suite('Template', function() {
 
   suite('interpolate', function() {
     var html = document.createElement('div');
+    var htmlNoIntrpltn = document.createElement('div');
     var css = document.createElement('div');
     html.appendChild(document.createComment('<span>${str}</span>'));
+    htmlNoIntrpltn.appendChild(document.createComment('<span>Nothing</span>'));
     css.appendChild(document.createComment('#foo { height: ${height}px; }'));
 
     test('interpolate(data) => html', function() {
@@ -81,6 +83,39 @@ suite('Template', function() {
       });
       assert.equal(typeof interpolated, 'string');
       assert.equal(interpolated, '#foo { height: 100px; }');
+    });
+
+    test('interpolate() => html', function() {
+      var tmpl = Template(htmlNoIntrpltn);
+      var interpolated = tmpl.interpolate();
+      assert.equal(typeof interpolated, 'string');
+      assert.equal(interpolated, '<span>Nothing</span>');
+    });
+
+    test('interpolate(data, options.safe[bad type]) => throws', function() {
+      var tmpl = Template(html);
+      var pseudoArray = { length: 1 };
+
+      assert.throw(() => tmpl.interpolate({}, { safe: pseudoArray }),
+        '"options.safe" must be an Array, not object');
+      assert.throw(() => tmpl.interpolate({}, { safe: 'badType' }),
+        '"options.safe" must be an Array, not string');
+    });
+
+    test('interpolate(data.undefined) => throws', function() {
+      var tmpl = Template(html);
+      assert.throw(tmpl.interpolate.bind(tmpl),
+        'No value provided for template identifier "str"');
+    });
+
+   test('interpolate(data.property[not a string]) => throws', function() {
+      var tmpl = Template(html);
+      assert.throw(() => tmpl.interpolate({ str: 666 }),
+        '"data.str" must be a String, not number');
+      assert.throw(() => tmpl.interpolate({ str: [] }),
+        '"data.str" must be a String, not object');
+      assert.throw(() => tmpl.interpolate({ str: {} }),
+        '"data.str" must be a String, not object');
     });
   });
 
--- a/apps/sms/views/conversation/js/recipients.js
+++ b/apps/sms/views/conversation/js/recipients.js
@@ -20,7 +20,7 @@
     var number;
 
     opts = opts || {};
-    this.name = opts.name || opts.number || '';
+    this.name = String(opts.name || opts.number || '');
     this.number = (opts.number || '') + '';
     this.email = opts.email || '';
     this.editable = opts.editable || 'true';
--- a/apps/sms/views/inbox/js/inbox.js
+++ b/apps/sms/views/inbox/js/inbox.js
@@ -786,7 +786,7 @@ var InboxView = {
     li.innerHTML = this.tmpl.thread.interpolate({
       hash: isDraft ? '#/composer' : '#/thread?id=' + id,
       mode: isDraft ? 'drafts' : 'threads',
-      id: isDraft ? draft.id : id,
+      id: String(isDraft ? draft.id : id),
       timestamp: String(timestamp)
     }, {
       safe: ['id']
--- a/shared/js/template.js
+++ b/shared/js/template.js
@@ -103,35 +103,56 @@
   /**
    * template.interpolate
    *
-   * Interpolate template string with values provided by
-   * data object. Optionally allow properties to retain
-   * HTML that is known to be safe.
+   * Interpolate template string with values provided by data object.
+   * Optionally allow properties to retain HTML that is known to be safe.
    *
-   * @param {Object} data     properties correspond to substitution.
-   *                          - identifiers in template string.
-   * @param {Object} options  optional.
-   *                          - safe, a list of properties that contain
-   *                          HTML that is known and are
-   *                          "known" to ignore.
-   */
-  Template.prototype.interpolate = function(data, options) {
-    // This _should_ be rewritten to use Firefox's support for ES6
-    // default parameters:
-    // ... = function(data, options = { safe: [] }) {
-    //
-    options = options || {};
-    options.safe = options.safe || [];
+   * @param {Object} data   - key/value properties correspond to substitution,
+   *                        with keys as identifiers in template string and
+   *                        values as replacement strings.
+   *
+   * @param {Object} [options]      - optional
+   * @param {Array}  options.safe   - array of identifiers in the data parameter
+   *                                that contain HTML that is known and
+   *                                are "known" to ignore.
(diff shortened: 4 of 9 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=95d4192b8b140c322cdd7ce198f6c3b48f3f8ff5

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8a2747dc8ba7fe7930b6b46554a402ed2494d998

shared/js/template.js
```diff
@@ -114,24 +114,25 @@
    *                          HTML that is known and are
    *                          "known" to ignore.
    */
-  Template.prototype.interpolate = function(data, options) {
-    // This _should_ be rewritten to use Firefox's support for ES6
-    // default parameters:
-    // ... = function(data, options = { safe: [] }) {
-    //
-    options = options || {};
-    options.safe = options.safe || [];
```
julienw: We still need this in case an `options` object is given without a `safe` property. So it's better to just have `{}` as default value for `options` and keep the second line here.

sroucheray: If we want to prevent unhandled errors related to the safe property we may replace its auto-initialization by this :
`options.safe = Array.isArray(options.safe) ? options.safe : [];`

or even creating a variable to use in the ternary operator :
```javascript
  Template.prototype.interpolate = function(data = {}, options = {}) {
    var hasSafeProperties = Array.isArray(options.safe) && options.safe.length;

    return this.extract().replace(rmatcher, function(match, property) {
      property = property.trim();

      var isPropertySafe = hasSafeProperties && options.safe.includes(property);
// [...]
      return isPropertySafe ? unsafeStr : Template.escape(unsafeStr);
    });
  };
```

what do you think ?

julienw: If we want to check if `options.safe` is correct, I'd prefer to throw if it's present but is not an array. No quiet ignore :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6f6c4fa5561ec2325bebe0d63c3105c1ab4d9d82

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4032fa85e35fcef753c1df9b5e4df9b94cd01634

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c8004ef84ed593b1c4291f06c16affce060a17d2

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=bdaf89477fe07a91e49d20f3b4682de527ee2fb2

shared/js/template.js
```diff
@@ -113,25 +113,43 @@
...
-        // to be treated as suspicious
-        Template.escape(data[property]) :
-        // Otherwise, return the string of rendered markup
-        data[property];
+
+      var isPropertySafe = hasSafeProperties && safeList.includes(property);
+
+      if (!(property in data)) {
+        throw new
+          Error(`No value provided for template identifier "${property}"'`);
```
julienw: nit: I think there is a `'` that should not be here.

also, if you need to go to the next line, please do something like:
```
throw new Error(
  '.....'
);
```

shared/js/template.js
```diff
@@ -113,25 +113,43 @@
...
-    // This _should_ be rewritten to use Firefox's support for ES6
-    // default parameters:
-    // ... = function(data, options = { safe: [] }) {
-    //
-    options = options || {};
-    options.safe = options.safe || [];
+  */
+  Template.prototype.interpolate = function(data = {}, options = { safe: [] }) {
+    var safeList = options.safe;
+    var hasSafeList = Boolean(safeList);
```
julienw: nit: we usually use `!!safeList` for this. But to be honest I don't think we need this boolean :) you can just use `safeList` instead

shared/js/template.js
```diff
@@ -113,25 +113,43 @@
...
+
+    if (hasSafeList && !Array.isArray(safeList)) {
+      throw new TypeError(
+        `"options.safe" must be an Array, not ${typeof options.safe}`
```
julienw: nit: add "if present," at the start of the message

shared/js/template.js
```diff
@@ -113,25 +113,43 @@
...
+      );
+    }
+
+    var hasSafeProperties = hasSafeList && safeList.length;
```
julienw: in the previous code we were using an empty array, just wondering why you changed this. For performance reason ?

julienw:
sorry I haven't looked really deep yet, just some early comments. I'll look deeper tonight or tomorrow.

apps/sms/views/conversation/js/recipients.js
```diff
@@ -20,7 +20,7 @@
     var number;
 
     opts = opts || {};
-    this.name = opts.name || opts.number || '';
+    this.name = String(opts.name || opts.number || '');
```
julienw: just wondering why you did this ? did you see an issue with the existing code ?

shared/js/template.js
```diff
@@ -113,25 +113,43 @@
...
-   */
-  Template.prototype.interpolate = function(data, options) {
-    // This _should_ be rewritten to use Firefox's support for ES6
-    // default parameters:
-    // ... = function(data, options = { safe: [] }) {
-    //
-    options = options || {};
-    options.safe = options.safe || [];
+  */
+  Template.prototype.interpolate = function(data = {}, options = { safe: [] }) {
```
julienw: I still think it's useless to have `safe` in the default argument... an empty object works just as well, even given the following code :)

shared/js/template.js
```diff
@@ -113,25 +113,43 @@
...
-      // options.safe is an array of properties that can be ignored
-      // by the "suspicious" html strategy.
-      return options.safe.indexOf(property) === -1 ?
-        // Any field that is not explicitly listed as "safe" is
-        // to be treated as suspicious
-        Template.escape(data[property]) :
-        // Otherwise, return the string of rendered markup
-        data[property];
+
+      var isPropertySafe = hasSafeProperties && safeList.includes(property);
```
julienw: nit: I'd move this line after the 2 checks

shared/js/template.js
sroucheray: Actually I did not really change it. The empty Array is still in the default parameters. I just throw an error if the type of `safe` is forced to something else than an Array.

shared/js/template.js
sroucheray: Agree, but it gives developers a better insight on the `interpolate` method signature, plus it is inlined with previous code where an empty Array is created if not given.
BTW I can remove it if you prefer.

apps/sms/views/conversation/js/recipients.js
sroucheray: Some unit tests pass Number instead of String to the Recipient `update` method which in turns use the Recipient constructor.

https://github.com/mozilla-b2g/gaia/blob/41e13d5dd4788a69c28d2696cc60b2b7455e4369/apps/sms/views/conversation/test/unit/recipients_test.js#L272-L285

Maybe you would prefer to fix the tests ?

shared/js/template.js
julienw: yeah I think it's more readable. (but it's debatable :) )
You can improve the jsdoc to make it more obvious that `safe` is an array

shared/js/template.js
julienw: anyway it works for me

apps/sms/views/conversation/js/recipients.js
julienw: ok; no I think it's good to make it sure we have a string here, let's keep this change.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3da41cfd5feb9f91a27cd708facdfc54914b42a1

apps/sms/views/inbox/js/inbox.js
```diff
@@ -765,8 +765,8 @@ var InboxView = {
     li.innerHTML = this.tmpl.thread.interpolate({
       hash: isDraft ? '#/composer' : '#/thread?id=' + id,
       mode: isDraft ? 'drafts' : 'threads',
-      id: isDraft ? draft.id : id,
-      number: number,
+      id: String(isDraft ? draft.id : id),
```
julienw: sorry, this has a conflict with current master; you just need to rebase abd add this change back to the new code (`number` is gone :) )

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e530514044d102fe67e2a69462061260e430142e

apps/sms/views/inbox/js/inbox.js
sroucheray: @julienw The branch is up-to-date with master now !

apps/sharedtest/test/unit/template_test.js
```diff
@@ -61,8 +61,10 @@ suite('Template', function() {
 
   suite('interpolate', function() {
     var html = document.createElement('div');
+    var htmlNoIntrpltn = document.createElement('div');
     var css = document.createElement('div');
     html.appendChild(document.createComment('<span>${str}</span>'));
+    htmlNoIntrpltn.appendChild(document.createComment('<span>Nothing</span>'));
     css.appendChild(document.createComment('#foo { height: ${height}px; }'));
```
julienw: It's quite disturbing all this is not in a setup() (which means they would be renegerated again for each test, thus we're sure we have a clean state). I know it was like this before already so if you prefer to keep it like this it's also fine to me.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a1530c5161762d902b526f33faa1600ad2cd78a7

PR #32449, opened 2015-10-14
Bug 1207985 - [music][nga] Support sharing via NFC

jimporter:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c5aa834eb1a3535abc9e9b5f11a43bbb936cad6e

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3dabc2dc3bfe42715f37929cac55b4256edd98c3

apps/music/js/app.js
```diff
@@ -175,8 +175,11 @@ viewStack.addEventListener('change', (evt) => {
 
   document.body.dataset.activeViewUrl = viewUrl;
 
-  playerButton.hidden = viewUrl === VIEWS.PLAYER.URL || !isPlaying;
-  activityDoneButton.hidden = viewUrl !== VIEWS.PLAYER.URL || !activity;
+  var showingPlayer = viewUrl === VIEWS.PLAYER.URL;
+  playerButton.hidden = showingPlayer || !isPlaying;
+  activityDoneButton.hidden = !showingPlayer || !activity;
+  client.method('enableNFC', showingPlayer);
```
justindarc: Both app.js and endpoint.js are loaded in the same context. You can just directly call `enableNFC()` from here instead of going across the bridge.

apps/music/js/endpoint.js
```diff
@@ -511,4 +515,16 @@ function open(blob) {
   });
 }
 
+function enableNFC(enabled) {
+  // This function is a no-op until we enable NFC once.
+  if (!nfcEverEnabled && !enabled) {
```
justindarc: For consistency with the queue and remote, we should instead have a function called `loadNFCShare` that follows the same pattern as `loadQueueSettings()` and `loadRemote()` where it sets a global variable called `nfcShare` to the `NFCShare` global once lazy loaded. Then in this function you can just do `loadNFCShare().then(() => nfcShare.enabled = enabled);`.

apps/music/js/app.js
justindarc: Errrr.. Disregard this. What you're doing here is probably preferred since it allows us to separate the endpoint into the SW later.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2443c5fc522f5276317bc7e133ade933e24b2862

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=507db64fe3b7dc23490dc8075ece544e0ff91fad

PR #32459, opened 2015-10-15
Bug 1214990 - Allow settings to override; skip step when 0 apps

sfoster:
Fixing the skip-step-when-no-apps uncovered a cascade of bugs around the settings override. I've refactored so the actual mobileconnection-sourced info is combined/overridden with properties from the latecustomization.operatorInfo setting value, so you can still get a result while testing even if you don't have one of the magic SIMs.

### DIFF
--- a/apps/ftu/js/late_customization.js
+++ b/apps/ftu/js/late_customization.js
@@ -29,15 +29,25 @@
     _customizationManifestUrl: '',
 
     get enabled() {
-      return Boolean(this._customizationManifestUrl && this._operatorInfo);
+      return Boolean(this._customizationManifestUrl &&
+                     (this._operatorInfo || this._predefinedOperatorInfo));
+    },
+
+    get operatorInfo() {
+      var info = {};
+      Object.assign(info,
+                    this._operatorInfo || this.getInfoFromMobileConnection(),
+                    this._predefinedOperatorInfo);
+      return info;
     },
 
     get carrierName() {
-      return this._operatorInfo && this._operatorInfo.carrier;
+      var info = this.operatorInfo;
+      return info.carrier || info.operator;
     },
 
     get region() {
-      return this._operatorInfo ? this._operatorInfo.region : '';
+      return this.operatorInfo.region || '';
     },
 
     start: function() {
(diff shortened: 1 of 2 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0e227b035da28a487f39752f15ead7425a91dbf6

apps/ftu/js/late_customization.js
```diff
@@ -205,19 +220,18 @@
...
-      for (var i=0, conn; i<connections.length; i++) {
-        conn = connections[i];
-        if(conn && conn.voice && conn.voice.connected) {
-          var operatorInfos = MobileOperator.userFacingInfo(conn);
-          if (operatorInfos.operator) {
-            operatorInfos.mcc = conn.voice.network.mcc;
-            operatorInfos.mnc = conn.voice.network.mnc;
-            if (!operatorInfos.carrier) {
-              operatorInfos.carrier = operatorInfos.operator;
+      if (connections) {
```
arcturus: nit:
I'm  a great fan of early return, to avoid an extra indentation level, but feel free to ignore this comment :)

apps/ftu/js/late_customization.js
```diff
@@ -46,20 +56,17 @@
...
-            info.carrier = info.operator;
-          }
+        if (predefinedInfo) {
+          // store operator/params separate from
+          // bona-fide mobileconnection details
+          this._predefinedOperatorInfo = predefinedInfo;
         }
-        // skip waiting for operator info from mobileConnections if we got
-        // it from settings
-        return this._operatorInfo || this.waitForOperatorInfo();
```
arcturus: Now we are not returning anything

apps/ftu/js/late_customization.js
```diff
@@ -46,20 +56,17 @@
...
       }).then((results) => {
```
arcturus: This results variable now should be undefined, we are not using it, before we were just waiting for the promise to finish.

Which means the previous `then` block is already synchronous, so we could potentially group the two `then` blocks into a single one, not having to wait for the next tick.

apps/ftu/js/late_customization.js
```diff
@@ -71,6 +78,14 @@
           this.init();
         } else {
           this.debug('Not enabled; not initializing');
+          if (!this._operatorInfo) {
+            // keep watching in case we can init later
+            this.waitForOperatorInfo().then((info) => {
```
arcturus: Oh, i see we passed the sync condition to this block :)

apps/ftu/js/late_customization.js
sfoster: Don't fear the indents! Usually I'm an early returner but with a couple conditions and returns in there already, the if seemed clearer.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5f2f9a12aeb1f8c13fa30f0853531d46eff876d7

PR #32505, opened 2015-10-16
Bug 1213946 - Use |now.toString()| instead of |new Date(now)| in mont…

jandem:
…h_day_agenda_test.js. r=julienw

In bug 1187233 we're changing |new Date(now)| to create a copy (which will preserve milliseconds).


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=922255f45eeacd9940fd00ea7b13483943b94159

PR #32532, opened 2015-10-18
Bug 941763 - [B2G][SMS] Sent and recieved SMS containing an invalid U…

yvtheja:
…RL like '.Website.com' are highlighted and function as hyperlinks

### DIFF
--- a/apps/sms/views/conversation/js/link_helper.js
+++ b/apps/sms/views/conversation/js/link_helper.js
@@ -58,7 +58,7 @@ function checkDomain(domain) {
 }
 
 // defines things that can match right before to be a "safe" link
-var safeStart = /[\s,:;\(>\u0080-\uFFFF]/;
+var safeStart = /[\s.,:;\(>\u0080-\uFFFF]/;
 
 const MINIMUM_DIGITS_IN_PHONE_NUMBER = 5;
 const LEADING_DIGIT_BREAKPOINT = 7;
@@ -114,7 +114,7 @@ var LINK_TYPES = {
       // {1} match the protocol https?:// or rtsp:// (optional)
       '(https?://|rtsp://)?',
       // {2} match "server name": . must be followed by at least one letter
-      '((?:\\.?[-\\w]){1,256})',
+      '(\\w(?:\\.?[-\\w]){0,255})',
       // {3} match a . followed by one or more domain valid chars
       '(\\.\\w{1,10})',
       // optional :port
--- a/apps/sms/views/conversation/test/unit/link_helper_test.js
+++ b/apps/sms/views/conversation/test/unit/link_helper_test.js
@@ -153,12 +153,14 @@ suite('link_helper_test.js', function() {
       test('non ascii prefix', function() {
         testURLMatch('àmozilla.org', 'mozilla.org', true);
       });
+      test('Bug 941763: Add dot as URL prefix', function() {
+        testURLMatch('.website.com', 'website.com', true);
+        testURLMatch('https://.website.com', 'website.com', true);
+        testURLMatch('htt://www.mozilla.org', 'mozilla.org', true);
+      });
     });
 
     suite('Failures', function() {
-      test('Simple invalid URL', function() {
-        testURLNOK('htt://www.mozilla.org');
-      });
       test('Simple invalid URL with slashes', function() {
         testURLNOK('http://www.a/b/d.com');
       });


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e869b5b67a8a56a567ee4066691c7c0e560e29c5

apps/sms/views/conversation/js/link_helper.js
```diff
@@ -150,6 +150,10 @@ var LINK_TYPES = {
         return false;
       }
 
+      if(url[0] === '.') {
```
julienw: with this change, you miss to filter something like `http://.website.com` which is also incorrect.
So I'd suggest 2 possibilities:

* use `if (match[2][0] === '.')` (or maybe `match[2].startsWith('.')` is more readable).
* change the regexp for {2} to `'([\\w](?:\\[.][-\\w]){0,255})',`

I'd rather change the regexp (if this works properly ;) ). Can you add a test for "http://.website.com" too ? And test various cases in Firefox or on a device ?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4650402a33c14fb8425e52f517da41abd626d643

ythej:
@julienw, I tried couple of RegEx but couldn't get the correct one !! I couldn't get why your's is wrong.

Please let me know if I need to update this PR.

julienw:
I think this regexp works: `(\w(?:\.?[-\w]){0,255})`
can you try it ?

note: I found https://regex101.com/#javascript which is really nice to help analyzing a regexp :)

julienw:
Note: you need to double all `\` in the previous regexp because we're defining a regexp inside a string here.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c0df4cdc14de9dfd95b59cd005a96f7ef4bc6266

apps/sms/views/conversation/test/unit/link_helper_test.js
```diff
@@ -192,6 +192,11 @@ suite('link_helper_test.js', function() {
         testURLOK('http://whatever.stuff');
         testURLOK('https://whatever.stuff');
       });
+
+      test('Bug 941763', function() {
+        testURLNOK('.website.com');
```
julienw: I think we wanted to link the part after `.` ? Maybe add `.` to the safeStart regexp ?

ythej: @julienw , yes, my blind mistake. I was excited and created PR !! :P

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=36beba407c569e49d5fe59177706f35d56dcae3a

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=36beba407c569e49d5fe59177706f35d56dcae3a

julienw:
can you please rebase to latest master so that we can have a proper try run ?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b93ae96bb5ebbb358c5c036ae7a22c693f4eb967

ythej:
@julienw, done !! :)

Between, I forgot to ask ! Why are we using safeStart ? 

Thank you !

julienw:
sorry it's still not rebased, you need to run something like `git fetch upstream && git rebase upstream/master` and hten `git push -f origin HEAD` (assuming `origin` is your remote and `upstream` is mozilla's remote)

julienw:
the goal of `safeStart` is to not start looking for a URL where it's not appropriate.

Actually  I'm not so sure that adding `.` is something that we should do... because a `.` can be inside an URL or phone numbers. So we'll see, and that's why I especially want to have a proper try run :)

ythej:
@julienw, I did this to rebase -
1> git checkout master 
2> git fetch upstream
3> git pull upstream master
4> git checkout bug-941763
5> git rebase -i master
6> git push -f origin bug-941763

I don't know where it went wrong !! Will try it as you told ! :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=914ac3e4cdc84965883d12c2e7f59dc1cb6c91bf

ythej:
@julienw, I repeated the same process I did and now I see status "4 commits ahead of mozilla-b2g:master". Is it correct ? Please let me know if I need to do something else.

Thank you !

julienw:
2 unit tests are failing, see https://treeherder.mozilla.org/logviewer.html#?job_id=2986494&repo=gaia

As you see, these are expected failures (with your change, now we find more links where before we didn't find them), so you need to change the tests :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=bcf2a9fa6e1984bbd537c4e84f55db5cf0d6fd76

apps/sms/views/conversation/test/unit/link_helper_test.js
```diff
@@ -153,12 +153,14 @@ suite('link_helper_test.js', function() {
       test('non ascii prefix', function() {
         testURLMatch('àmozilla.org', 'mozilla.org', true);
       });
+      test('Bug 941763', function() {
```
steveck-chung: Please add more description for this test (full stop prefix).

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c33c32aced036e7a2b270d638c943e1e613a12a4

apps/sms/views/conversation/test/unit/link_helper_test.js
kumarrishav: IMO bug number is not important to write in test description. Instead, give better/detail description :)

ythej: @kumarrishav , I am not sure. But I think it's just means of including reference to more details to the particular test. A new contributor can go to the bug page and find more details. But I know it's not consistent! :P

apps/sms/views/conversation/test/unit/link_helper_test.js
```diff
@@ -153,12 +153,14 @@ suite('link_helper_test.js', function() {
       test('non ascii prefix', function() {
         testURLMatch('àmozilla.org', 'mozilla.org', true);
       });
+      test('Bug 941763: Add dot as URL prefix', function() {
```
kumarrishav: Then in that way code will full of bug no. Also new contributor doesn't look much on old bug number. Mentors help them to get required info. Also it's not consistent. Also, IMO for new contributor, more important is..what piece of code is doing instead of how and from where it came from.
We do write bug no. as comments in case of big/critical patch or patch where more works need to be done in other Bug. Writing Bug no. In code looks bit weird. Commit details is already there is to show bug no.

steveck-chung: Yeah it's good point that we actually don't need the bug # in this case. I think it's a bad practice from other test cases in this script that only left bug # in the title, so I suggest @ythej to add more details for the title.  So here is some ideas: We create another bug for polishing and remove all the unnecessary bug number in this test, or simply fix right in this patch. @ythej wdyt?

ythej: @steveck-Chung , I think it would be better if we file a new bug for filtering all the test titles which include bug numbers. Anyways, we may proceed to change the present test title in this patch itself.

steveck-chung: Ok please file another bug for this.

PR #32545, opened 2015-10-19
Bug 1216066 - input_areas: alignment depends on doc dir

peppsac:
Using the input[dir] was too restrictive as some elements use
unicode-bidi: -moz-plaintext to handle bidi text and we still
want them to be aligned according to the document direction.

[Review on Reviewable](https://reviewable.io/reviews/mozilla-b2g/gaia/32545)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b8e24faf913e5d2b287dcdb1f25d6a9dc52815ef

PR #32575, opened 2015-10-19
Bug 1214157 - [Video] Handle interruption from audio channel to set correct play/pause button state

punamdahiya:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4d6dae23a16edff9ad795290aa91a5ac2b5d126e

apps/video/js/video.js
```diff
@@ -136,6 +136,22 @@ document.addEventListener('visibilitychange', function visibilityChange() {
...
+// play/pause button. See Bug 121457
+dom.player.addEventListener('mozinterruptbegin', function() {
+  if (playing) {
+    setButtonPaused(true);
+    playing = false;
+  }
+});
+
+dom.player.addEventListener('mozinterruptend', function() {
+  if (!playing) {
```
russnicoletti: Nit: it seems we get 'mozinterruptend' only when a video is not playing so this check appears to be redundant. With that said, it does make the code more readable.

apps/video/js/video.js
```diff
@@ -136,6 +136,22 @@ document.addEventListener('visibilitychange', function visibilityChange() {
   }
 });
 
+// Handle interruption from audio channel and set correct
+// play/pause button. See Bug 121457
+dom.player.addEventListener('mozinterruptbegin', function() {
+  if (playing) {
```
russnicoletti: Nit: it seems we get 'mozinterruptbegin' only when a video is playing so this check appears to be redundant. With that said, it does make the code more readable.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=76f26cb947184bee3d08e06c8a42edb2dc2148c8

apps/video/js/video.js
punamdahiya: On further investigation playing variable is not a correct reflection of interrupted state and that's why I have removed it in the updated patch to keep patch simple to changing button play/pause icon ,  See https://bugzilla.mozilla.org/show_bug.cgi?id=1214157#c19.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ed734659cede19a5a9ca65eb433da7bbb75090ac

apps/video/js/video.js
```diff
@@ -108,6 +108,27 @@ var loadingChecker =
...
+  lastPlayTime = window.performance.now();
+});
+
+dom.player.addEventListener('mozinterruptbegin', function() {
+  // Ignore mozinterruptbegin event if it immediately follows play event
+  if ((window.performance.now() - lastPlayTime) > 50) {
+    // Pause video when interrupted by competing audio.
+    pause();
+  }
+  lastPlayTime = 0;
```
russnicoletti: It seems to me 'lastPlayTime' should only get reset when the code concludes there has been an interruption due to competing audio. The way the code is, when 'play' is pressed, 'lastPlayTime' will get reset. I don't think this will cause a problem because when 'mozinterruptbegin' is fired in response to competing audio, 'now - lastPlayTime' is expected to be greater than 50ms even if 'lastPlayTime' didn't get reset. 

It just seems from a purely logical perspective, 'lastPlayTime' isn't being reset in the best place. 

With all that said, I consider this a nit because I don't see any problem with the functionality and this code is intended to be temporary.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a5d6c0b9e9969c85b985bf528b8b66d36247e19c

apps/video/js/video.js
punamdahiya: I was using lastPlayTime flag more as indication of getting inetrruptbegin event immediately after play event and handling it versus the actual last play event received time.
Having said that when 'mozinterruptbegin' is fired in response to competing audio, 'now - lastPlayTime' difference is expected to be greater than 50ms even if 'lastPlayTime' didn't get reset, so I have updated patch to reset lastPlayTime only when interrupted i.e.  interruptbegin from audio competing is received.

apps/video/js/view.js
```diff
@@ -206,6 +208,25 @@
...
+    // If performance.now() - lastPlayTime is small we ignore interruptbegin
+    // and pause video playback only when interrupted by competing audio.
+    // This workaround should be removed with the fix of Bug 1206581
+    dom.player.addEventListener('play', function() {
+      lastPlayTime = window.performance.now();
+    });
+
+    dom.player.addEventListener('mozinterruptbegin', function() {
+      // Ignore mozinterruptbegin event if it immediately follows play event
+      if ((window.performance.now() - lastPlayTime) > 500) {
```
punamdahiya: While testing updated patch, for view activity I noticed that when we first call showPlayer - andl play() inside it, we are receiving events in sequence
https://github.com/mozilla-b2g/gaia/blob/master/apps/video/js/view.js#L342
1) play
2) interrupt begin
3) inetrrupt end
4) interrupt begin
5) interrupt end
I have checked and this happen only when play is called from inside showPlayer and not for subsequent dom.player.play call.
The second inetrruptbegin call is getting called in time frame > 50 ms and was pausing video when we first open a video in view activity.  
I propose two solutions
a) Increased the time to 500 ms to ignore this second interruptbegin call and fix this bug for view activity
b) Since this bug in view activity is an edge case and we don't update view activity with this work around.

russnicoletti: I hesitate to add a third option, since this is an alternate proposal that ideally would have been suggested earlier and this is an edge case. In any event, what if instead of using the time between 'play' and 'interrupt' we use a bool: waitingForInterruptFromPlay (or something like that), that is set to true in the 'play' event handler. In the 'interruptbegin' handler, if waitingForInterruptFromPlay is true, set it to false and return. If waitingForInterruptFromPlay is false, invoke pause. What do you think?

punamdahiya: Tried and this third alternative won't work for view.js where we are receiving two interrupt-begin event after play, first interrupt-begin will set bool variable to false and second interrupt-begin event will check for false and pause video which is not desired.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=24a6a0d6190567115898377e2680ca09ef802a52

PR #32577, opened 2015-10-20
Bug 1207471 - reorganized shims in alphabet order, r=yzen

gasolin:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ae429f80c6d0a546fd7fa316c6514356301da679

gasolin:
`simcard_lock` is removed so there are extra 3 lines deleted

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=370184e2d4627290407f8855e2b7373c7b38227b

apps/settings/build/settings.build.jslike
```diff
@@ -9,40 +9,41 @@
   // which makes the objects inaccessible by reference.
   // If the inquiries to the object are all performed by requirejs, we can
   // remove the path of the object from the following list.
+  // organize in alphabet order
```
yzen: nit: here and the other file
organize -> organized

yzen:
I suspect it's unrelated but while stepping into panels when testing i get the following:
TypeError: iccCard is null carrier.js:174:9 - stepping into Cellular & Data with no SIM

apps/settings/build/settings.build.jslike
```diff
@@ -9,40 +9,41 @@
...
   paths: {
     'vendor/jszip': 'empty:',
```
yzen: A suggestion here, lets also alphabetize here based on the first character of the parent folder (not just file name). So for example modules/... will go before venoder/.... In fact vendor should go closer to the end.

gasolin: It definitely a bug, thanks for catching it!

gasolin:
`iccCard is null` is unrelated bug its a bug
https://github.com/mozilla-b2g/gaia/blob/master/apps/settings/js/carrier.js#L174

filed in bug 1216847

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=40c6f86a5829f0893c90f6e3403ed81c244a6f4f

PR #32583, opened 2015-10-20
Bug 1216435 - remove RootPanelHandler from startup.js, r=yzen

gasolin:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=37b7cc81fe48204336f0bdfe39711aa179c2337e

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6986444f32de69a2be862b3484295e9c1c987e59

apps/settings/js/activity_handler.js
```diff
@@ -77,13 +77,6 @@
             document.body.dataset.filterBy =
               activitySource.data.filterBy || 'all';
             break;
-          case 'call':
```
gasolin: The pendingTargetPanel name will be checked again in Main.js, so we can safely remove the check from here
https://github.com/mozilla-b2g/gaia/blob/master/apps/settings/js/main.js#L58

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=941087f0f62d9af8b0443e5e0d74a4defd286420

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6a6ed592dd6801773084ea21c99afe236ec19ad6

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=691ffd161ebcf19a84cd8230cac153dffd77c1c5

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b7204bd54175cca3305c9d17dd813ec22fbddbba

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=20bfc0e991f2a45fb9f08de0cf74aa73be74f83c

apps/settings/js/panels/root/root.js
```diff
@@ -27,6 +28,7 @@ define(function(require) {
        */
       LazyLoader.load([
         'js/firefox_accounts/menu_loader.js',
+        'js/dsds_settings.js',
```
gasolin: remove the usage of DsdsSettings in root panel, make DsdsSettings could be lazyloaded later

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2e4fe0f179ddd4c249aae9ceab0b340842e17c44

apps/settings/js/panels/root/root.js
```diff
@@ -38,9 +40,65 @@ define(function(require) {
...
+        return Promise.resolve();
+      }
+    },
+
+    init: function root_init(panel) {
+      this._panel = panel;
+      // Load the necessary scripts after the UI update
       setTimeout(this._loadScripts);
+
+      // Show NFC panel when support
```
yzen: nit: support -> supported

apps/settings/js/panels/root/root.js
```diff
@@ -38,9 +40,65 @@ define(function(require) {
...
+    },
+    /**
+     * Update the developer menu item based on the preference.
+     */
+    _showDeveloperMenuItem: function root_showDeveloperMenuItem() {
+      var developerItem = this._panel.querySelector(
+        '[data-show-name="developer.menu.enabled"]');
+      if (developerItem && navigator.mozSettings) {
+        return navigator.mozSettings.createLock()
+          .get('developer.menu.enabled').then(
```
yzen: Ah, interesting, didn't know that DOMRequest object supports Promise interface

apps/settings/js/startup.js
```diff
@@ -29,7 +29,8 @@
 
     if (this._settings) {
       this._init();
-      if (typeof customPanelHandlerFunc !== 'undefined') {
+      if (customPanelHandlerFunc &&
+          typeof customPanelHandlerFunc !== 'undefined') {
```
yzen: The second part is now redundant, I believe, if customPanelHandlerFunc is truthy, it will never have a type of undefined.

apps/settings/js/startup.js
```diff
@@ -474,16 +383,9 @@
...
-        // XXX: This is an optimization for the root panel to avoid reflow that
-        //      could be observed by users.
-        var customPanelHandler;
-        if (initialPanelId === 'root') {
-          customPanelHandler = RootPanelHandler;
-        }
-
         var initialPanelHandler =
           InitialPanelHandler(document.getElementById(initialPanelId),
-            customPanelHandler);
```
yzen: Looks like customPanelHandlerFunc argument to InitialPanelHandler is not used any more, should we remove all that logic?

apps/settings/test/unit/app_starter_test.js
```diff
@@ -83,7 +78,7 @@ suite('AppStarter', function() {
       }).then(done, done);
     });
 
-    test('should use RootPanelHandler if the initial panel id is root',
+    test('should pass null customPanelHandler if the initial panel id is root',
```
yzen: Same here if we don't use that argument, and can remove that logic, lets do that and this test would be unnecessary

apps/settings/test/unit/panels/root/root_test.js
```diff
@@ -8,47 +15,178 @@ suite('Root', function() {
...
+        'developer.menu.enabled': true
+      });
+      panel.init(document.body);
+      developerMenuItem =
+        document.querySelector('[data-show-name="developer.menu.enabled"]');
+      panel._showDeveloperMenuItem().then(function() {
+        assert.isFalse(developerMenuItem.hidden);
+      }, function() {
+        // This function does not reject.
+        assert.isTrue(false);
```
yzen: lets add second argument describing what the failure means

apps/settings/test/unit/panels/root/root_test.js
```diff
@@ -8,47 +15,178 @@ suite('Root', function() {
...
+        'developer.menu.enabled': false
+      });
+      panel.init(document.body);
+      developerMenuItem =
+        document.querySelector('[data-show-name="developer.menu.enabled"]');
+      panel._showDeveloperMenuItem().then(function() {
+        assert.isTrue(developerMenuItem.hidden);
+      }, function() {
+        // This function does not reject.
+        assert.isTrue(false);
```
yzen: lets add second argument describing what the failure means

apps/settings/js/panels/root/root.js
```diff
@@ -38,9 +40,65 @@ define(function(require) {
...
+
+      // Show NFC panel when support
+      var nfcItem = panel.querySelector('.nfc-settings');
+      nfcItem.hidden = !navigator.mozNfc;
+
+      // Show proper SIM panel
+      this._showSimItems();
+
+      // Show developer panel when necessary
+      this._showDeveloperMenuItem();
```
yzen: Should we maybe return something like:

```
return Promise.all([
  this._showSimItems(), 
  this._showDeveloperMenuItem()
]);
```

This way these tests that test that dev menu is shown or not can be done in then call after init, instead of explicitly calling _showDeveloperMenuItem. What do you think?

apps/settings/js/startup.js
gasolin: good catch!

apps/settings/js/panels/root/root.js
gasolin: The origin cause of RootPanelHandler in startup.js is to reduce the visible repaint at startup.

To delay the show/hide in root panel will cause extra screen reflow, it can be spotted by Toggle on `Developer>Flash Repainted Area`.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=99a87df84a0233fe54e2daf21319bf06557f29e0

apps/settings/js/panels/root/root.js
gasolin: Ya, since I just move it from startup.js, it does work @_@

PR #32588, opened 2015-10-20
Bug 1192263 - [Messages] We load Inbox before going to the notificati…

steveck-chung:
…on conversation when app is run via notification click, r=julienw

### DIFF
--- a/apps/sms/views/shared/js/startup.js
+++ b/apps/sms/views/shared/js/startup.js
@@ -17,6 +17,7 @@
          Navigation,
          Settings,
          SilentSms,
+         StartupManager,
          SystemMessageHandler,
          TimeHeaders,
          Utils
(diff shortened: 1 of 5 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0027f9ed996459a582837e53cda597aba114868b

apps/sms/views/shared/js/startup_manager.js
```diff
@@ -0,0 +1,26 @@
...
+
+const priv = {
+  taskQueue: Symbol('taskQueue')
+};
+
+
+var StartupManager = {
+  init(task) {
+    this[priv.taskQueue] = new TaskRunner();
+    this[priv.taskQueue].push(task);
```
julienw: not sure this is necessary to have a task enqueuing here (but that's a detail)

apps/sms/views/shared/js/startup.js
```diff
@@ -126,7 +127,7 @@ var Startup = exports.Startup = {
 
       MessageManager.init();
       InboxView.init();
-      Navigation.init();
+      StartupManager.init(Navigation.init());
```
julienw: maybe the issue comes from here: this should be `() => Navigation.init()` as argument

PR #32599, opened 2015-10-20
Bug 1178451 - Cache the last overflow count result and try to reuse it in the following call. r=mhenretty

gabrielesvelto:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0805a982fd987d35773bb5d8acac37f450216568

shared/js/font_size_utils.js
```diff
@@ -208,12 +216,22 @@
       var resultWidth;
       var overflowCount = -1;
 
+      if (string.length > this._overflowCountCache) {
```
mikehenrty: Let's add a comment here like above explaining what we are doing with the cache, just like you did above with the `_overflowCountCache`.

shared/js/font_size_utils.js
```diff
@@ -208,12 +216,22 @@
...
+        substring = string.substr(0, string.length - this._overflowCountCache);
+        resultWidth = this.getFontWidth(substring, fontSize, fontFamily);
```
mikehenrty: Now that repeat this logic twice, perhaps we should abstract it into a function:
```
resultWidth = this.getSubstringFontWidth(string, endIndex, fontSize, fontFamily);

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=da6895a289739cfcb281149f7a7840c326aa42d8

PR #32601, opened 2015-10-20
Bug 1216129 - Fix popups causing rocketbar to become collapsible. r=etienne

Cwiiis:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5425ee670b02974e7c7d0fcaf829fb5f65961d90

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9db544ab8449b8dc83e1baa01c928964a71b31c4

apps/system/js/app_chrome.js
```diff
@@ -791,6 +791,12 @@
     };
 
   AppChrome.prototype.handleScrollAreaChanged = function(evt) {
+    // Make sure the scroll-area-changed is coming from the right element.
+    if (evt && (!this.app || !this.app.browser ||
```
etiennesegonzac: this `!this.app` is probably overcautious since the `AppChrome` constructor expects it as a param.

Cwiiis: That's what I thought, but I saw a null this.app while testing this with web activities, so... I dunno what's going on there but I'd rather keep it in :)

etiennesegonzac: hum, we might want to file a bug for this then :) but yeah feel free to leave the check

PR #32607, opened 2015-10-20
Bug 1216231 - Fix failure in test_ftu_skip_tour.py and remove test_ftu_skip_tour_desktopb2g.py

mwargers:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4b6a36a5792b931278edad71605e59fd03f257b5

tests/python/gaia-ui-tests/gaiatest/tests/functional/ftu/test_ftu_skip_tour.py
```diff
@@ -108,23 +108,19 @@ def test_ftu_skip_tour(self):
...
-
-        # Verify that internal state matches UI state
-        self.assertEqual(self.data_layer.get_setting(
-            'debug.performance_data.shared'), self.ftu.is_share_data_enabled,
-            msg='Send Data internal state should match UI state')
-
-        self.ftu.disable_share_data()
-        self.assertEqual(self.data_layer.get_setting(
-            'debug.performance_data.shared'), not share_data_default)
+        self.assertTrue(self.ftu.is_metrics_basic_enabled)
```
JohanLorenzo: We could reduce the number of assertions, in order to make the test less repetitive to read:
```py
self.assertTrue(self.ftu.is_only_metrics_basic_enabled)
self.ftu.enable_metrics_enhanced()
self.assertTrue(self.ftu.is_only_metrics_enhanced_enabled)
# and so on
```
``` py
def is_only_metrics_basic_enabled(self):
    return self._metrics_basic_radio.is_checked and not self._metrics_enhanced_radio.is_checked and not self._metrics_none_radio.is_checked

# and so on
```

A better solution would be to create a `RadioGroup` in widget, but that'll require some more work. We can do it in a follow up

PR #32623, opened 2015-10-21
Bug 1159343 - [Video] Use seek status to throttle seek requests.

jhlin:
Per jya's comment [1], image won't be updated  before the end of dragging. Here is a proposal to address it by throttling seek requests.

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1159343#c38

### DIFF
--- a/apps/video/js/video.js
+++ b/apps/video/js/video.js
@@ -108,6 +108,9 @@ var loadingChecker =
   new VideoLoadingChecker(dom.player, dom.inUseOverlay, dom.inUseOverlayTitle,
                           dom.inUseOverlayText);
 
+// Store seek target before previous one is done.
+var pendingSeekPosition = null;
+
 var lastPlayTime = 0;
 
 // Workaround to fix issue reported in Bug 1214157
@@ -272,7 +275,7 @@ function initPlayerControls() {
 
   // handle video player
   dom.player.addEventListener('timeupdate', timeUpdated);
-  dom.player.addEventListener('seeked', updateVideoControlSlider);
+  dom.player.addEventListener('seeked', seekEnded);
   dom.player.addEventListener('ended', playerEnded);
 
   // handle user tapping events
@@ -1311,6 +1314,22 @@ function handleSliderTouchEnd(event) {
   }
 }
 
+function seekEnded() {
+  updateVideoControlSlider();
+  if (pendingSeekPosition) {
+    requestSeek(pendingSeekPosition);
+    pendingSeekPosition = null;
+  }
+}
+
+function requestSeek(target) {
+  if (dom.player.seeking) {
+    pendingSeekPosition = target;
+  } else {
+    dom.player.fastSeek(target);
+  }
+}
+
 function handleSliderTouchMove(event) {
   if (!dragging) {
     return;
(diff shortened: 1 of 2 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0978cfe539d26e0aff0feccbe4befeb5f387742a

apps/video/js/video.js
```diff
@@ -108,6 +108,9 @@ var loadingChecker =
   new VideoLoadingChecker(dom.player, dom.inUseOverlay, dom.inUseOverlayTitle,
                           dom.inUseOverlayText);
 
+// Store seek target before previous one is done.
+var pendingSeekPosition = NaN;
+
```
davidflanagan: I'd use null here rather than NaN.  null seems more idiomatic JS, since we don't have to worry about type safety.  It could be that NaN is more optimizable, I suppose.

apps/video/js/video.js
```diff
@@ -1290,6 +1293,14 @@ function handleSliderTouchEnd(event) {
   }
 }
 
+function seekEnded(event) {
+  updateVideoControlSlider(event);
```
davidflanagan: no argument is expected, so don't pass the event argument

apps/video/js/video.js
```diff
@@ -1320,7 +1331,11 @@ function handleSliderTouchMove(event) {
   dom.playHead.classList.add('active');
   movePlayHead(percent);
   dom.elapsedTime.style.width = percent;
-  dom.player.fastSeek(dom.player.duration * pos);
+  if (dom.player.seeking) {
+    pendingSeekPosition = dom.player.duration * pos;
+  } else {
+    dom.player.fastSeek(dom.player.duration * pos);
+  }
```
davidflanagan: I'd suggest you write a fastseek() utility function that you can use here.  You should also use it below in handleSliderKeypress()

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b06978635f7feb9048ce8e136130e1a68b920f36

apps/video/js/view.js
```diff
@@ -504,6 +504,22 @@
     }
   }
 
+  function seekEnded() {
+    updateSlider();
+    if (pendingSeekPosition) {
```
davidflanagan: pendingSeekPosition is undeclared here.  For stupid historical reasons, the video app has two completely separate video player implementaions.  The main entry point is index.html and that uses the video player in video.js.  But for the view activity, the entry point is view.html, which loads view.js but does not load video.js.  So you've got to declare this variable.

There should have been a lint warning about the undeclared variable.  If you've run 'npm install' and have the linter code installed, then you should not have been able to commit this code, because it would have failed the linter unless you either declared the variable or added it to the /* global */ comment at the top...

jhlin: My bad! Thanks for the review and info about the linter. I'll install these modules and update the patch ASAP.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7b53a60b5554f0130d1eecb67be72fd335bae177

jhlin:
Changes has been cherry-picked in PR#32759

PR #32625, opened 2015-10-21
Bug 1210869 - Fix app duplicates when installing via WebIDE. r=gmarty

Cwiiis:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e94ded0eb3e0e55996cac8b0f9745021ef73067f

apps/homescreen/js/app.js
```diff
@@ -918,6 +918,21 @@
...
       case 'install':
+        // Check if the app already exists, and if so, update it.
+        // This happens when reinstalling an app via WebIDE.
+        var existing = false;
+        for (child of this.icons.children) {
+          icon = child.firstElementChild;
+          if (icon.app && icon.app.manifestURL === e.application.manifestURL) {
+            icon.app = e.application;
+            icon.refresh();
+            existing = true;
```
gmarty: Way too late for a nit, but can you add a `break` here to avoid useless iteration over items. Maybe add it in a subsequent/unrelated patch? Sorry for that.

Cwiiis: Just spoke on IRC, but to add here for record - I don't break here in case the app has multiple entry points. Note that this still doesn't handle entry points quite correctly, but I heard that they'd be disappearing at some point, so I'm loathe to complicate the code for something that isn't really used much. We can deal with the issues if they get raised.

PR #32639, opened 2015-10-21
Bug 1215654 - Enable HUD when FTU choice is enhanced metrics. r=sfoster.

tamarahills:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=edc454bd84ea8f5f198c01ee19113f56e3665c68

apps/ftu/js/ui.js
```diff
@@ -431,6 +431,9 @@ var UIManager = {
       case 'metrics-basic':
       case 'metrics-enhanced':
       case 'metrics-none':
+        this.updateSetting('hud.hide', (event.target.value === 'Enhanced'));
+        this.updateSetting('devtools.overlay',
```
russnicoletti: This is a nit, I think. I'm wondering if we should only update these settings when the level is enhanced. If ever devtools was enabled by default, we would disable it here if the level != enhanced. Now that I'm thinking about it, how about moving the 'enhanced' case to be first in the list and within the 'enhanced' block set 'hud.hide' and 'devtools.overlay' and then fall-through to the other cases where the metrics level is set.

tamarahills: I think that makes sense, but I'm not sure it really matters as it's FTU.  But I can add that as a safeguard.  I think the only time it would matter if if we turned on HUD via the Dogfood flag.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2585221c408ed27a9dd30fce7280d48f61eba198

apps/ftu/test/unit/ui_manager_test.js
```diff
@@ -516,4 +516,64 @@ suite('UI Manager > ', function() {
...
   });
 
+  suite('Metrics Section', function() {
+    var MockSettings;
+
+    setup(function() {
+      MockNavigatorSettings.mSetup();
+      MockSettings = MockNavigatorSettings.mSettings;
+      UIManager.init();
+      Navigation.currentStep = 7;
```
sfoster: This should be currentStepIndex, currentStep is expected to be an object reference with the hash etc.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=75832ad74a09f46595563b397900b02f63893bfd

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5df0f5a7ab3e2d3e2c2f5ca57e924d4aa21c1c2d

PR #32690, opened 2015-10-23
Bug 1217075 - [Messages] Both message and draft are preserved in the main app instance in case message is sent from another activity instance, r=azasypkin

steveck-chung:


apps/sms/views/shared/js/inter_instance_event_dispatcher.js
```diff
@@ -13,7 +13,7 @@
   var channel = null;
 
   const BROADCAST_CHANNEL_NAME = 'messages-channel';
-  const ALLOWED_EVENTS = ['drafts-changed'];
+  const ALLOWED_EVENTS = ['drafts-saved', 'drafts-deleted'];
```
azasypkin: __optional nit:__ maybe singular form will be better since now we fire these events only for one draft only ```draft-saved``` and ```draft-deleted```?

apps/sms/services/js/drafts.js
```diff
@@ -58,6 +58,7 @@
       draftIndex.set(draft.threadId, drafts);
 
       this.emit('saved', draft);
+      InterInstanceEventDispatcher.emit('drafts-saved', draft);
```
azasypkin: __question:__ my initial idea was to fire event only when draft is actually stored and ```Drafts``` should have compare old and new draft sets and fire corresponding events. But I'm totally fine with your approach as my is more complex and we actually don't need it right now since ```DraftsService``` will replace this soon anyway - so no waste here :)

The only concern is batch deletion - could you please make sure that when we remove big chunk of conversations our memory numbers aren't going crazy (we issue postMessage and read IndexedDB)? I guess we need one active main and one activity instance to verify :)

apps/sms/views/inbox/js/inbox.js
```diff
@@ -118,8 +118,21 @@ var InboxView = {
     Drafts.on('saved', this.onDraftSaved.bind(this));
 
     InterInstanceEventDispatcher.on(
-      'drafts-changed',
-      this.renderDrafts.bind(this, true /* force update */)
```
azasypkin: __question:__ mm how about ```init``` method in ```Drafts``` instead that will allow us to keep ```InterInstanceEventDispatcher``` dependency only inside ```Drafts``` and other pieces won't have to do this as well (if/when they need it)?
```
init() {
  InterInstanceEventDispatcher.on('drafts-saved', (draft) => {
    this.request(true).then(() => this.emit('saved', draft));
  });

  InterInstanceEventDispatcher.on('drafts-deleted', (draft) => {
    this.request(true).then(() => this.emit('deleted', draft));
  });
}
```

apps/sms/views/shared/js/inter_instance_event_dispatcher.js
steveck-chung: ah yeah, you are right

apps/sms/views/inbox/js/inbox.js
steveck-chung: hmm, good idea!

apps/sms/services/js/drafts.js
steveck-chung: I agree with you about the draft update mechanism and we should apply you idea eventually, but maybe not a safe option since it's only 5 day left before RA. And just like you said we will rewrite this part in draft service in the future.

About the memory usage, I was not aware of that because user seldom perform batch action under more than 1 message instance case, which means they need to pause their new/share activity and switch back to original message app for further actions. Anyway I just profiled with b2g-info but found that the memory didn't increase even batch removing the threads with drafts. I thought the memory should somehow increased and dropped later since it will force triggering the drafts request for new drafts map every time. But the fact is the number of memory PSS/USS didn't change while deleting.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9e5763368b06a8cec5a2c4e421c3986423cf548b

apps/sms/views/inbox/js/inbox.js
```diff
@@ -114,14 +113,10 @@ var InboxView = {
     MessageManager.on('message-received', this.onMessageReceived.bind(this));
     MessageManager.on('threads-deleted', this.onThreadsDeleted.bind(this));
 
+    Drafts.init();
```
azasypkin: What about moving ```Drafts.init``` to startup.js like we do with the rest of things?

apps/sms/views/inbox/js/inbox.js
```diff
@@ -114,14 +113,10 @@ var InboxView = {
...
     Drafts.on('deleted', this.onDraftDeleted.bind(this));
     Drafts.on('saved', this.onDraftSaved.bind(this));
 
-    InterInstanceEventDispatcher.on(
-      'drafts-changed',
-      this.renderDrafts.bind(this, true /* force update */)
```
azasypkin: __nit:__ I think we can remove ```force``` parameter support from ```renderDrafts``` (and unit tests if we have any for this parameter) to keep code simpler if we don't use it anymore, what do you think?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=fc888b973b40f3456697f46f1137f71676f36b61

apps/sms/views/inbox/js/inbox.js
steveck-chung: Sounds good, and I found that we never have a test for the force === true actually... :p

apps/sms/views/inbox/test/unit/inbox_test.js
```diff
@@ -47,7 +46,6 @@ var mocksHelperForInboxView = new MocksHelper([
   'OptionMenu',
   'StickyHeader',
   'Navigation',
-  'InterInstanceEventDispatcher',
```
azasypkin: __nit:__ please remove ```require``` with this mock above

apps/sms/views/inbox/js/startup.js
```diff
@@ -39,6 +40,7 @@
       ConversationClient.init(App.instanceId);
       MessageManager.init();
       Navigation.init();
+      Drafts.init();
```
azasypkin: __optional nit:__ not critical right now, but maybe we can also add this call to coversation/js/startup.js in case we add more code to ```Drafts.init``` that will be needed by Conversation, but it's optional

apps/sms/views/inbox/js/inbox.js
azasypkin: Ha-ha, great, now we can safely get rid of it :)

apps/sms/views/inbox/js/startup.js
steveck-chung: I might not add this right now since we don't handle draft change in conversation view, but It's interesting question about handling the draft change in conversation. If we open the same conversation view in another instance, should we sync the draft status as well? It looks like the instance opened in another instance will also load the saved draft correctly, so maybe it's a good idea to sync the changes to original instance.

azasypkin: Yeah, I think we discussed this some time ago - it can be quite tricky especially for the conversation drafts - both instances can have different drafts and we don't want to discard one of them if another has been changed, but if one instance doesn't have it's own draft maybe we can do it...

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=216adf3d18591e08483500be22c3506281a0c4be

apps/sms/views/inbox/js/startup.js
steveck-chung: Yeah I think so. Maybe I can left some comment in bug 1080504, I think it's the most appropriate bug for addressing this issue.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0d4f52c3168490059af33d1f6fefb79587c422b9

PR #32693, opened 2015-10-23
Bug 875371 - a sound when locking the device r=albertopq

prateekjadhwani:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7ce60b153f8247c29e1785c9318133ca414cfd90

apps/system/lockscreen/js/lockscreen.js
```diff
@@ -541,6 +539,16 @@
     this._unlockingMessage = {};
   };
 
+  /**
+   *  Plays unlock sound if required
+   */
+  LockScreen.prototype.playUnlockSound = function(soundFileUrl) {
```
snowmantw: I think it's weird when you are **locking** and you call a method named as 'play**Unlock**Sound', so I suggest to make it as two methods like `playUnlockSound` and `playLockSound` to specify which sound you're playing. Another advantage is you can encapsulate the sound file URL inside the method, rather to pass it as an argument while it's actually necessary to be that.

apps/system/lockscreen/js/lockscreen.js
```diff
@@ -560,6 +568,7 @@
       this._lastUnlockedInterval = now - this._lastUnlockedTimeStamp;
       this._lastLockedTimeStamp = now;
 
+      this.playUnlockSound('/resources/sounds/lock.opus');
```
snowmantw: While we check if the sound is enabled to play the unlocking sound, I think for the locking one it needs to follow the same method. Of course this need an UX' opinion, but I think if the spec didn't change a lot it should be done like that.

prateekjadhwani: Currently, There is no `details` object passed when the user clicks the `Lock Button`. Should I try to make it work in a similar way that `lockscreen_state_unlock.js` does it?

snowmantw: Well to change it like that will be a complicated patch, although that's indeed better. I hope this can still follow the original (although it's legacy) flow similar with how we play unlocking sound in this file. If the `details` is missing in some cases, I think it should be fixed.

PR #32697, opened 2015-10-23
Bug 1216496 - [email] RTL CSS refacto

autra:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4612b8bc69a7bcc40f7f18b846a207a082b1c28e

apps/email/index.html
```diff
@@ -32,6 +32,7 @@
   <link rel="stylesheet" type="text/css" href="shared/style/progress_activity.css" >
   <link rel="stylesheet" type="text/css" href="shared/style/tabs.css" >
   <link rel="stylesheet" type="text/css" href="shared/style/value_selector.css" >
+  <link rel="stylesheet" type="text/css" href="shared/elements/gaia-icons/gaia-icons.css" >
```
autra: This is needed for the 'back' buttons that needs mirroring icons in RTL.

This can be costly for just one icon, but I figured that maybe we want to replace a lot of them by gaia-icons. I'm thinking about the attach, send, trash, back, forward icons... basically most icons in style/images/icons have a corresponding one in gaia-icons. And we could add the one that are not yet there.

I see several advantages of doing so:
- gaia-icons.ttf is cached at the system level to speed up the startup of the application (that's what Wilson Page told me at least). We would benefit from the speed gain if we use them.
- It guarantees we stay visually consistent with the rest of gaia.

WDYT? I can open a bug for that if you want.

apps/email/js/cards/fld/account_item.html
```diff
@@ -1,4 +1,4 @@
 <a class="fld-account-item">
   <span class="selected-indicator"></span>
-  <span dir="auto" class="fld-account-name"></span>
-</a>
\ No newline at end of file
+  <span class="fld-account-name"></span>
```
autra: this is the email adress. `dir="auto"' or `direction: ltr` in the css works equally well. I prefer the last one because:
- it's a pure css solution
- we cannot use start/end css properties with any on them, but if the direction is in the css, it's a lot nearer to any margin/padding... that might be affected, making it more obvious why we can't.

apps/email/style/common.css
```diff
@@ -330,33 +337,34 @@ the list of folders.
...
 .email-value-selector[role="dialog"][data-type="value-selector"] [role="listbox"] .fld-folder-depth3,
 .fld-folder-depth3 {
-  -moz-padding-start: 6rem;
+  padding-inline-start: 6rem;
 }
 .email-value-selector[role="dialog"][data-type="value-selector"] [role="listbox"] .fld-folder-depth4,
+
 .fld-folder-depth4 {
-  -moz-padding-start: 6.5rem;
+  padding-inline-start: 7.5rem;
```
autra: I've opened a bug some time about folder depth: https://bugzilla.mozilla.org/show_bug.cgi?id=1169589

Part of it got corrected, but comment 12 asked me to open another bug for the css part. That's the fix for this bug I've never opened :-)

For the reference, here are some screenshots

**Before**
LTR - RTL 

drawer_02

**After**
LTR - RTL

drawer_02

apps/email/style/compose_cards.css
```diff
@@ -217,6 +216,8 @@ html:-moz-dir(rtl) .cmp-autocomplete-item-details {
   outline: 0;
   white-space: pre-wrap;
   word-wrap: break-word;
+  unicode-bidi: -moz-plaintext;
+  text-align: initial;
```
autra: Nearly every time we have a `unicode-bidi: plaintext`, I think we need a `text-align: initial. It is mainly for mail contents.

As in a browser with webpages, it is more natural to have an english email to be left-aligned, even if the user interface is in arabic. The reverse is also true.

plaintext also allows a per-paragraphe alignment: if you receive a mail with one English paragraph and one Arabic, each one will be aligned according to its logical `start`, ie left for English and right for Arabic. 

These two things are an exception to the rule 
> everything should be right-aligned in RTL, left-aligned in LTR

Some screenshots to show this:

## before
The arabic text in the mail is left-aligned in LTR, which does not feel ok.
LTR - RTL
message_01

Same problem in RTL for an English mail.
LTR - RTL
message_02


## After
In contrast, when we let the mail align naturally: 
message_01
message_02

Even in the Arabic interface, the English mail is correctly readable.

Notice that the bar indicating the quote DOES mirror, because it is more part of the chrome than the mail (as we parse the `>` and format them). Likewise, the subject is still aligned according to the interface, also because the date juste below is.

apps/email/style/setup_cards.css
```diff
@@ -354,27 +349,30 @@ li.align-center {
...
+  text-rendering: optimizeLegibility;
   opacity: 0.7;
 }
-
-html:-moz-dir(rtl) .item-with-children::after  {
-  left: -0.5rem;
-  right: auto;
-}
+html[dir="ltr"] .item-with-children::after  { content: 'right-light'; }
+html[dir="rtl"] .item-with-children::after  { content: 'left-light'; }
```
autra: If you agree with what I've said about gaia-icons, we could have a 

```css
.icon-after::after {
  content: attr('data-icon-after');
  font-family: 'gaia-icons';
  font-weight: 500;
  font-size: 3rem;
  font-style: normal;
  text-decoration: inherit;
  text-transform: none;
  text-rendering: optimizeLegibility;
}
```
Or something like that, with or without `width`, `height`, `position`... it depends if everything uses the same values.

And in the html, we would just need to put a `data-icon-after="myIconName"` on the element to give it an icon.

autra: But maybe better in a followup though.

apps/email/index.html
jrburke: Let's separate the gaia-icons effort as a separate patch. I expect we may need to evaluate startup performance, and likely need some work to get some of the icons into the font. I would prefer to be able to land the rtl changes though quickly, without needing to wait or be blocked by that work.

If that impacts what you can do for rtl though, please let me know.

apps/email/js/cards/fld/account_item.html
jrburke: dir="auto" was our previous guidance. I'm fine changing that though if the common approach for apps has worked out something better, but to me that implies changing the other uses of dir="auto" in email to all use this same pattern, vs just this one element?

apps/email/style/marquee.css
```diff
@@ -6,21 +6,6 @@
...
-laid out correctly though.
-*/
-#marquee-h-text.ltr-align {
-  text-align: left;
-}
-#marquee-h-text.rtl-align {
-  text-align: right;
-}
-
-
```
jrburke: Since these are no longer needed, we can remove the setting of these styles in marquee.js, the section that uses `document.dir`.

apps/email/index.html
autra: Actually, I didn't explain the whole story here (my bad). What motivated me in the first place was that arrows in settings buttons are broken since quite a long time: the image here does not exist any more: https://github.com/mozilla-b2g/gaia/pull/32697/files#diff-dcbffab9bed0fd786a4cd742c9c06e35L166 (there's several other occurrence of this).

In building blocks button, they have been replaced by gaia-icons if I'm not mistaken. So I think we have several choices:
1. don't bother here, leave it broken and fix in another bug (I can open one).
2. Fix it by replacing the images with a new one in `style/images/icons`, copied from gaia-icons and that will be mirrored in RTL.
3. Actually, the correct behaviour is the current one (no arrows displayed), so we just need to remove these dead codes.

I think I should do either 1 or 4, WDYT? 

I'm opening a bug for the gaia-icons story.

apps/email/js/cards/fld/account_item.html
autra: Actually, I was wrong.
> this is the email adress. dir="auto"' ordirection: ltr` in the css works equally well. 

is only true for old-style ASCII email addresses. But now, they can have arabic letters (actually most of utf-8 chars now) in their local part. In this case, `direction: ltr` and `dir="auto"` does not behave identically, see http://jsfiddle.net/23sw5o2n/.

I still believe the correct form is with `direction: ltr`, but I can NI Ahmed to be sure. (I mean, both are confusing, but one is less than the other imho).

We should use `direction: ltr` only for fields that contains only an email address.

If it contains an url,  both are fine, because the scheme part will make the BiDi algorithm calculates a LTR direction for the whole string anyway (assuming the IANA won't register scheme with non-latin letters).

If it contains other texts not handled by l10n.js (typically user input) then we nearly always want to use `dir="auto"` or `unicode-bidi: -moz-plaintext`. We cannot use the latter everywhere because ellipsis are currently buggy with it (see [bug 883884](https://bugzilla.mozilla.org/show_bug.cgi?id=883884), unlikely to be fixed soon).

TL;DR my current rules of thumb is:
- For string handled by l10n.js, don't do anything, that's l10n.js or loca job.
- For emails (alone in their html element), use `direction: ltr`
- For other string, 
   - don't do anything if we're sure they will always have the same direction as the system one (but this is really rare because if they are not handled by l10n.js, it usually means they're from user input).
   - Use unicode-bidi: -moz-plaintext if you don't need ellipsis (typically for most input fields).
   - Use `dir="auto"` everywhere else.

Does this make sense to you?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6994ef80bb9648d4de9311ef7d18a213c0e89a92

apps/email/js/cards/fld/account_item.html
jrburke: OK, if we go with this route for this section, then it seems like these other places listed below where we show the email address will need to be changed to match. Otherwise, we are likely to get bug reports of a mismatch in style between how the account is shown here (as the first line in fiddle shows) vs these other places (where the layout of the second line in the fiddle is used).

I am fine with whatever approach you think makes sense, as long as we can make a best effort for the display to be uniform across these different areas.

Elements with dir="auto", text-overflow:ellipsis:

* js/cards/tng/account_item.html: .tng-account-item-label. Tap on gear icon to show the "Settings" screen. This is used to show each account name.
* js/cards/settings_account.html: .tng-account-header-label. From "Settings" screen, tap an account to see Account Settings. This is the header.
* js/cards/settings_account_credentials.html: .tng-account-header-label.
* js/cards/settings_account_servers.html: .tng-account-header-label (BUG: missing dir="auto")
* js/cards/setup_account_password.html: .sup-email-display

For the last three, in Account Settings, scroll down to "Server settings" section and tap items labeled "Username", and "IMAP settings", or "SMTP settings" or "ActiveSync settings".

No specific styles on the element, but the parent element has dir="auto", text-overflow:ellipsis, which affects the child:

* js/cards/msg/peep_bubble.html: .msg-peep-content, parent: .msg-peep-bubble. When looking at an email with its body in the Message Reader, the bubble used for the email addresses in the header.
* js/cards/cmp/peep_bubble.html: .cmp-peep-name and .cmp-peep-address could contain account. Parent: .cmp-peep-bubble. When composing a message in Compose view, the bubble shown for email addresses in the header.
* js/marquee.js: . marquee-h-text: parent, marquee-h-wrapper. SPECIAL: the marquee-h-text sets overflow: visible so that we can scroll the whole address via JS. Tap on a bubble in Compose or Message Reader to inspect details of the name, provides actions that can be done with the email address.

apps/email/index.html
jrburke: If UX officially decided to never have the carets for either rtl or ltr cases, then we should do 3. Last I heard, it was a bit undecided, and that is why email is in its current state.

Since this is not clear, at least to me, I filed [bug 1218547](https://bugzilla.mozilla.org/show_bug.cgi?id=1218547) to track finding out from UX, so we can go with option 1 if you like. However, if you know for sure that carets are explicitly out for the LTR and RTL cases, feel free to remove them now (option 3), whichever path works best for you. We can use bug 1218547 to track any follow up in either case.

autra: > However, if you know for sure that carets are explicitly out for the LTR and RTL cases

They should be there and mirrored. This is a regression coming from one of our Phoxygen patch earlier in the RTL refacto process (the one you've linked in this bug). I didn't see this at first. Sorry for that!

So I'm going for 1/ and will make a patch in bug 1218547 to fix it.

apps/email/js/cards/fld/account_item.html
autra: Oh thanks for the list. I'm fixing that now.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=846ad743453b0b0d15478d9db842e9d055ae1ada

PR #32699, opened 2015-10-23
Bug 1160054 - [Messages][refactoring] Migrate mozActivity DOM request to promise base

kumarrishav:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0e94fd175a682db6709f6cf0fd815358a1c2c3c0

apps/sms/views/shared/js/activity_picker.js
```diff
@@ -3,51 +3,36 @@
...
-
-  activity.onerror = typeof onerror === 'function' ? onerror : function(error) {
-    console.warn('Unhandled error spawning activity; ' + error.message);
-  };
-}
-
 var ActivityPicker = {
-  dial: function ap_call(number, onsuccess, onerror) {
-    handleActivity(new MozActivity({
+  dial: (number) => {
```
steveck-chung: If you want to simplify this part, you could:
```
dial(number) {
...
}
```

apps/sms/views/shared/js/activity_picker.js
```diff
@@ -3,51 +3,36 @@
...
+    Promise.resolve(new MozActivity({
```
steveck-chung: We should return the promise here instead of nothing, otherwise the caller could not handler the result of resolve/reject.

apps/sms/views/shared/js/activity_picker.js
steveck-chung: BTW it might be better if we can add js-doc comments and user would know it will return promise.

kumarrishav: Ok, will add js-doc comments. 
I would go for  dial: (number) => { .. 
arrow function looks good to me :smiley:

steveck-chung: Using arrow function didn't give any additional benefit here, so I would suggest you could simply follow es6 object Property definitions(https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#Property_definitions)

steveck-chung:
BTW it's not related to your patch, but we still have a place that handled success/error (https://github.com/mozilla-b2g/gaia/blob/master/apps/sms/views/conversation/js/link_action_handler.js#L41). I think it's a bug and we should remove the this.rest.

steveck-chung:
Not sure about the mock_moz_activity, but I think we should at least rewrite the mock activity picker that return promise for unit test. I wil take some time to think about the the proper mock for it.

kumarrishav:
So, should i consider it as part of this bug ? or should file separate bug ?
Btw, i have updated (https://github.com/mozilla-b2g/gaia/blob/master/apps/sms/views/conversation/js/link_action_handler.js#L41) too 
@steveck-chung

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8dc8dabc6316330cc8b1fa108921840c47020012

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=24637ecba2babe66a2b62d47a1bc4135fa3e3a7f

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=24637ecba2babe66a2b62d47a1bc4135fa3e3a7f

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3f162fbd22143e706fdff1091a4df10c28c1ea20

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e606c1aa3ade9001fd4955f4bf062e36d7cfccba

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d9705f40f19c1a364b24962d7302ec0f67579b23

apps/sms/views/shared/js/activity_picker.js
```diff
@@ -3,87 +3,119 @@
...
+        }
+      }));
+    },
 
-  activity.onerror = typeof onerror === 'function' ? onerror : function(error) {
-    console.warn('Unhandled error spawning activity; ' + error.message);
-  };
-}
+    /**
+     * When user wants to send a new Email. The e-mail app can parse mailto
```
steveck-chung: nit: remove trailing whitespace

steveck-chung:
Still not related to your patch, but I think we don't need the mock_activity_handler in the test, so we can remove it safely.

apps/sms/views/shared/js/activity_picker.js
```diff
@@ -3,87 +3,119 @@
...
+   */
+  var ActivityPicker = {
 
-  if (typeof onsuccess === 'function') {
-    activity.onsuccess = onsuccess;
-  }
+    /**
+     * When user wants to pass a phone call.
+     * @param  {Number} contact number where to dial.
+     * @returns {Promise} DOMRequest promise is always successfully resolved.
```
steveck-chung: We don't return the DOMRequest directly, and it's not always resolved successfully. We wrapped it inside Promise.resolve does not mean it could only handle resolve, it's to force it becomes a real promise instance.

apps/sms/views/shared/js/activity_picker.js
```diff
@@ -3,87 +3,119 @@
...
+        data: {
+          type: 'url',
+          url: url
+        }
+      }));
+    },
+
+    /**
+     * When user wants to create a new contact entry.
+     * @param  {Object} Contact object.
```
steveck-chung: It might be a little bit misleading if we commented that it's a contact object. contact object might be a real contact object returned from MozContacts API. It's only the object that contains contact information (phone or email), and it's different from the object we set for  viewContact.

apps/sms/views/shared/js/activity_picker.js
```diff
@@ -3,87 +3,119 @@
...
+        }
+      }));
+    },
+
+    /**
+     * When user wants to open a contact.
+     * @param  {Object} Contact object.
+     * @returns {Promise} DOMRequest promise is always successfully resolved
+     */
+    viewContact(contactProps) {
```
steveck-chung: Like I said previously, the contactProps here actually contains {id: opt.contactId}

steveck-chung:
I think the unit test change should be introduced in this patch as well since we are not hurry for this activity picker refactoring. About the activity picker test, here are some thoughts:
- We can remove all the mock helper, even the mock activity helper since we can simply override it with `window.MozActivity = () => {};` in the suiteSetup and `this.sinon.stub(window, 'MozActivity').returns(Promise.resolve()); ` or `.returns(Promise.resolve())` for simulate the mozActivity success/error case.
- Once we stub the MozActivity, we can assert the success case like:
```
      ActivityPicker.dial('999').then(() => {
        sinon.assert.calledWithMatch(MozActivity, {
          name: 'dial',
          data: { type: 'webtelephony/number', number: '999' }
        });
      }).then(done, done);
```
and the error case:
```
      var fakeError = new Error('fake-error');
      MozActivity.returns(Promise.reject(fakeError));

      ActivityPicker.dial('999').then(() => {
        done('should not called');
      }, (error) => {
        sinon.assert.calledWithMatch(MozActivity, {
          name: 'dial',
          data: { type: 'webtelephony/number', number: '999' }
        });
        assert.equal(error, fakeError);
      }).then(done, done);
```

steveck-chung:
And for the MockActivityPicker, you can simply set:
```
var MockActivityPicker = {
  email: () => {},
  dial: () => {},
  ...
};
```
And replace the original MockActivityPicker.dial.mSetup()/mTeardown() with this.sinon.stub()/sinon.assert.calledWith()

PR #32700, opened 2015-10-23
Bug 1216415 - Fail to unpin the pinned site r=benfrancis

albertopq:


autolander:

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c5221940189f14afcb66276a95ccf7d9c840bbd3

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3af677ffe4a0ae151fd1d881f6cd8ad21327de9d

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3af677ffe4a0ae151fd1d881f6cd8ad21327de9d

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4fabba68ad407b45f508680234fb0dd4a01e1769

apps/system/js/app_window.js
```diff
@@ -224,14 +224,14 @@
     this.getTopMostWindow()._setActive(enable);
   };
 
-  AppWindow.prototype.inScope = function(scope) {
+  AppWindow.prototype.inScope = function(scope, name) {
```
benfrancis: "In scope" is a formally defined concept in the W3C Web App Manifest specification http://w3c.github.io/manifest/#navigation-scope with an algorithm based entirely on URLs.

This inScope() method taking "name" as an argument effectively makes scope a tuple of URL scope + name. If a URL scope is provided by a web manifest then being in scope shouldn't be dependent on any app name. There could be a combination of a manifest's scope and an application-name meta tag which break this in a non-spec-compliant way.

If we need a function to check for origin + name as a fallback for scope then let's make that a separate function and call it something else. Rather than muddying the concept of "inScope" on an app window I would prefer us to not set the origin as a default value for the scope property (as the spec doesn't say to do this), but instead store a separate .origin property on the siteObject alongside the .name property. We could then have a separate method with a name like "matchesOriginAndName" for the use case of re-using open windows.

How does that sound?

apps/system/js/pin_page_system_dialog.js
```diff
@@ -169,25 +169,16 @@
...
-      .then((isPinned) => {
-        if (isPinned) {
-          this.pinSiteButton.setAttribute('data-l10n-id', 'pinning-unpin-site');
-          this.pinSiteButton.dataset.action = 'unpin-site';
-        } else {
-          this.pinSiteButton.setAttribute('data-l10n-id', 'pinning-pin-site');
-          this.pinSiteButton.dataset.action = 'pin-site';
-        }
-      });
+    BookmarksDatabase.get(data.url).then((site) => {
```
benfrancis: We should still set the button to "unpin site" for all URLs within the scope of a pinned site which specifies a real "scope" in its manifest.

apps/system/js/app_chrome.js
```diff
@@ -671,7 +666,11 @@
 
   AppChrome.prototype.handleBookmarksScopeChange = function (evt) {
     var scope = evt.detail.scope;
-    if (this.app.inScope(scope)) {
+    var inScope = !!(scope && this.app.inScope(scope));
+    var origin = new URL(evt.detail.url).origin;
+    var name = evt.detail.name;
+
+    if (inScope || this.app.matchesOriginAndName(origin, name)) {
```
benfrancis: We probably shouldn't pin/unpin the chrome based on origin + name?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5f90c09c6aa3a9e45aebbd8b51a779c282d3ae51

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=796b3b38ac3643917a4820c7d58342578f16989e

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ded3f92381c4c9c62975c2dbd144e62643177e4e

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=87b408f7b4697d20350edc3f83c953f3d4f53c87

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d0411629dab6a95def061e8f2df90f0c7c1b29cb

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9a26cc98c3d27a817c8e318c2efb1d71255b2324

PR #32725, opened 2015-10-25
Bug 1215469 - Create a Gaia UI test that exercises the first run dialog in the browser

mwargers:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=82c9eba8f735c1d47173a4003b1e74fd63cf0a8c

tests/python/gaia-ui-tests/gaiatest/tests/functional/browser/test_browser_first_run.py
```diff
@@ -0,0 +1,54 @@
...
+        search = Search(self.marionette)
+        search.launch()
+
+        tracking_dialog = TrackingDialog(self.marionette)
+        self.assertTrue(tracking_dialog.is_displayed)
+
+        browser = tracking_dialog.open_learn_more()
+        # It takes a short while before the page is starting to load
+        import time
+        time.sleep(1)
```
npark-mozilla: could we have a start_to_load() method here, where it basically checks Wait(self.marionette, timeout).until(lambda m: self.is_page_loading)?

mwargers: Hrm, actually, this could be part of wait_for_page_to_load() I think. What do you think?

npark-mozilla: yeah actually, why not?  we can first wait until is_page_loading is true, then wait again until it's not true.  makes sense, no?

mwargers: Yes, because I think this function is specifically used for when to wait for a loading page to be finished.

tests/python/gaia-ui-tests/gaiatest/apps/system/regions/fullscreen_dialog.py
```diff
@@ -0,0 +1,68 @@
...
+from gaiatest.apps.base import PageRegion
+from gaiatest.form_controls.binarycontrol import GaiaBinaryControl
+from gaiatest.apps.search.regions.browser import Browser
+
+class FullScreenDialog(PageRegion):
+
+    _fullscreen_dialog_locator = (By.ID, 'fullscreen-dialog-overlay')
+
+    def __init__(self, marionette):
+        Base.__init__(self, marionette)
```
JohanLorenzo: We should call the PageRegion constructor

tests/python/gaia-ui-tests/gaiatest/apps/system/regions/fullscreen_dialog.py
```diff
@@ -0,0 +1,68 @@
...
+        self.marionette.switch_to_frame()
+        self.root_element = Wait(self.marionette).until(
+            expected.element_present(*self._fullscreen_dialog_locator))
+
+    @property
+    def is_empty(self):
+        return self.marionette.execute_script("""
```
JohanLorenzo: What happens if we check for the text to be empty? Is sometimes the dom not empty but no visible text is present ?

tests/python/gaia-ui-tests/gaiatest/tests/functional/browser/test_browser_first_run.py
```diff
@@ -0,0 +1,54 @@
...
+        browser.wait_for_page_to_load()
+        self.assertFalse(tracking_dialog.is_displayed)
+
+        browser.go_to_url(self.test_url)
+
+        self.assertTrue(tracking_dialog.is_displayed)
```
JohanLorenzo: IIRC, GaiaBinaryControl already checks if something is displayed before interacting with it. Hence, this assertion is not necessary and can be removed.

tests/python/gaia-ui-tests/gaiatest/tests/functional/browser/test_browser_first_run.py
```diff
@@ -0,0 +1,54 @@
...
+        self.assertFalse(tracking_dialog.is_tracking_protection_enabled)
```
JohanLorenzo: I think GaiaBinaryControl checks if a button is already enabled. If this button is, while it shouldn't, the wait will fail.

In other words, this assertion is an superflous, as the test will fail even if the assertion is not here. I'd remove it.

tests/python/gaia-ui-tests/gaiatest/tests/functional/browser/test_browser_first_run.py
```diff
@@ -0,0 +1,54 @@
...
+        browser.go_to_url(self.test_url2)
+        browser.wait_for_page_to_load()
+        browser.switch_to_content()
+        element = self.marionette.find_element(By.ID, 'caption')
+        Wait(self.marionette).until(lambda m: element.text == 'You have successfully enabled tracking protection. Please continue!')
+
+        # Check that no tracking protection dialog appears anymore
+        browser.switch_to_chrome()
+        tracking_dialog = FullScreenDialog(self.marionette)
+        self.assertTrue(tracking_dialog.is_empty)
```
JohanLorenzo: All along the test we check the `is_displayed` property. I wonder why we check another property here.

JohanLorenzo: Oh we use FullScreenDialog, here. Instancing a new page class at the end of the test looks a bit confusing.

I think that's a sign that an improvement is needed. The relationship between `TrackingDialog` and `FullScreenDialog` could be clearer: `TrackingDialog` is a type of `FullScreenDialog`. Then, `TrackingDialog` should inherit from the other.

Once that is done, we can have in the test:
``` py
self.assertFalse(tracking_dialog.is_overlay_displayed)
```

What do you think?

tests/python/gaia-ui-tests/gaiatest/tests/functional/browser/test_browser_first_run.py
mwargers: Yes, true.

tests/python/gaia-ui-tests/gaiatest/apps/system/regions/fullscreen_dialog.py
mwargers: No, the tracking protection dialog is removed from the dom after tapping the "Got it" button.
That means self.root_element is not there anymore.
I guess I could use `System(self.marionette).is_element_present(*self._tracking_notice_locator)` and then I could get rid of this rather weird FullScreenDialog class.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c082588f817822c370b441f6f6f7f4169354ee15

PR #32736, opened 2015-10-26
Bug 1218091 - [Messages][Performance] Use optimized Bridge libraries. r=schung

azasypkin:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=52dc6beb3cf360c67706ea5ad72cb8b004cda49e

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a4d1ad20e708438158e421399b6b36bacf232686

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9365d890034ce25833367d66e5615f33faab5029

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7df25d69a4256fa8d6fb8baee1115ebf03874b63

apps/sms/build/build.js
```diff
@@ -36,9 +36,31 @@ function removeDesktopOnlyScripts(appStageDir) {
   });
 }
```
azasypkin: How does it look @steveck-chung ?

apps/sms/lib/bridge/bridge.js
```diff
@@ -32,13 +32,27 @@ module.exports = Client;
 /**
```
steveck-chung: Seems like we don't need the bridge.js anymore?

apps/sms/build/build.js
steveck-chung: Exactly what I thought, thanks!!

apps/sms/lib/bridge/bridge.js
steveck-chung: oops I forgot the activity shim again... So is it possible to replace the bridge with client in startup and replace the bridge with service/client ?

azasypkin: > So is it possible to replace the bridge with client in startup and replace the bridge with service/client ?

You mean lazy load both client and service right in the startup instead of single bridge or ...? If so, I'm a bit hesitating to do this in the patch that can be uplifted to v2.5, since it will likely affect ```fullyLoaded``` a bit more (size of ```bridge.min.js``` is ```15.6KB```, size of both ```client.min.js``` (```12.8KB```) and ```service.min.js```(```12KB```) will be ```24.8KB``` and they contain a bunch of duplicating stuff that should be parsed as well).

What about doing this in the patch that will get rid of ```ActivityShim``` from window (if we land it in v2.6 eventually), @steveck-chung ?

steveck-chung: @azasypkin It didn't need to split the client/service in this patch. I just want to make sure that we only need client in window context after moving `ActivityShim` to iframe. Sorry about the confusion :/

azasypkin: Ah, sorry :) Yeah, once we move ActivityShim to iframe/hidden window we'll need only client.js, you're right :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c8cd9f60694c5579b78530e890380feb7934533a

PR #32749, opened 2015-10-26
Bug 1210305 - Lock screen implementation

TatianaMigacheva:
Lock Screen UI implementation as per
http://r1l8an.axshare.com/#p=2_0_lock_screen

Modified Javascript files :-
- notifications.js
- lockscreen_notifications.js
  As per the UX, similar kind of notifications should be clubbed together.
  Tested by installing an external app and generated local notifications.
- lockscreen_inputpad.js
  Password entered by the user on the lockscreen should be visible for the 1
  second before changing it to dot.
- lockscreen_state_manager.js
  If user clicks on the lockscreen notification and clicks on open after
  unlocking the device , the corresponding app should be launched.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9655cdb57b8264c653fc8eb99a1b4817440a5f64

apps/system/js/notifications.js
```diff
@@ -84,7 +84,7 @@ var NotificationScreen = {
     // screen
     this.externalNotificationsCount = 0;
     this.unreadNotifications = [];
-
+    this.availableNotification = [];
```
davidflanagan: I'd like a comment here explaining what this property is for.  And shouldn't it be plural rather than singular? Do unreadNotifications matter anymore, or can you delete that property? Or can you just use unreadNotifications instead of availableNotifications?

apps/system/js/notifications.js
```diff
@@ -408,9 +408,13 @@ var NotificationScreen = {
     // timesout, so we skip updating it here, by passing a skip bool
     this.addUnreadNotification(detail.id, true);
 
+    var cIndex = this.availableNotification.indexOf(detail.appName);
+    if (cIndex === -1) {
```
davidflanagan: It looks like the closing brace for this if clause is 60 lines later!  And you haven't indented the intervening 60 lines correctly, so the code becomes very hard to read.

(I know I complained on an earlier version about unnecessary changes making the diff too big. But I think those were about accidental, incorrect indentation changes.  In this case, the code must be indented in order to be legible, even if it adds 60 lines to your diff.  (When something like that happens, you can always add a comment here on github letting me know that nothing has changed other than indentation.)

apps/system/js/notifications.js
```diff
@@ -408,9 +408,13 @@ var NotificationScreen = {
...
+    this.availableNotification.push(detail.appName);
```
davidflanagan: I notice that there is no code that ever removes notifications from this array. Is that correct? Does the NotificationsScreen object get torn down and rebuilt each time the lock screen is invoked?  If not, then I'm pretty sure you'll have to manage the array contents more actively.

apps/system/js/notifications.js
```diff
@@ -408,9 +408,13 @@ var NotificationScreen = {
...
     var notificationNode = document.createElement('div');
     notificationNode.classList.add('notification');
     notificationNode.setAttribute('role', 'link');
+    notificationNode.setAttribute('id', detail.appName);
```
davidflanagan: appName is not a good identifier. It changes when locale changes, and there can be multiple apps with the same name.  Use the app's manifestURL as a unique string instead.

apps/system/js/notifications.js
```diff
@@ -426,9 +430,9 @@ var NotificationScreen = {
     notificationNode.dataset.type = type;
     notificationNode.dataset.manifestURL = manifestURL;
 
-    if (detail.icon) {
+    if (detail.appIcon) {
```
davidflanagan: What is this change about? The string appIcon doesn't appear anywhere else in your patch, so it doesn't look like you're changing the code that generate these notification event details.  Was this a bug in the existing code?

apps/system/js/notifications.js
```diff
@@ -460,6 +464,17 @@ var NotificationScreen = {
     messageContent.setAttribute('dir', 'auto');
     message.appendChild(messageContent);
     notificationNode.appendChild(message);
+    var countDiv = document.createElement('div');
+    countDiv.classList.add('detail-count');
```
davidflanagan: So if I'm understanding this code right, only one notification per app id displayed, and it will contain the message of the earliest notifiction, and will then display a count of the number of notifications, but the messages in the subsequent notifications are just lost?  Is that really what UX wants?

apps/system/js/notifications.js
```diff
@@ -460,6 +464,17 @@ var NotificationScreen = {
...
+    countDiv.textContent = "1";
```
davidflanagan: Should we display a count when there is only one? Or should the numbers only appear when there are more than one?

apps/system/js/notifications.js
```diff
@@ -460,6 +464,17 @@ var NotificationScreen = {
...
+    notificationNode.appendChild(countDiv);
+  } else {
+    var notificationNode = document.querySelector('.notification');
+    var count = document.getElementById(detail.appName).querySelector('.detail-count').textContent;
```
davidflanagan: This is a terrible idea.  We can't use the HTML view to store data that belongs in the model. You can presumably figure out the correct count of notifications by searching the unreadNotifications array or something.  Please change the way you store the notification data so that you can figure out the count without refering to the HTML.

apps/system/lockscreen/js/lockscreen_inputpad.js
```diff
@@ -202,14 +220,18 @@
...
 
         if (this.states.padVibrationEnabled) {
           navigator.vibrate(this.configs.padVibrationDuration);
         }
 
         if (this.states.passCodeEntered.length === 4) {
-          this.lockScreen.checkPassCode(this.states.passCodeEntered);
+          (function(value) {
+            setTimeout(function() {
+              this.lockScreen.checkPassCode(value);
```
davidflanagan: I don't think that this line of code can possibly work. You're inside a nested function that you did not call .bind(this) on, so 'this' is undefined here.  If I'm right it means that this code is non-functional and it is not actually possible to unlock the phone!

apps/system/lockscreen/js/lockscreen_inputpad.js
```diff
@@ -202,14 +220,18 @@
...
+            }, CODE_VALIDATION_TIMEOUT);
```
davidflanagan: It seems like terrible UX to make the user wait an extra second after entering their pin before unlocking.  That will be infuriating. And if the UX spec says to do that, then we should get the spec created. UX designers are not gods... they make mistakes too, and when they tell us to do something stupid, we should let them know!  :-)

apps/system/lockscreen/js/lockscreen_inputpad.js
```diff
@@ -163,7 +167,19 @@
...
       if (span) {
         if (this.states.passCodeEntered.length > i) {
+        var value = this.states.passCodeEntered.substr(i, this.states.passCodeEntered.length);
+        if ((this.states.passCodeEntered.length - 1 === i) && (keyCode && (keyCode !== 'b'))) {
+          (function(aspan, currentValue) {
+            aspan.textContent = currentValue;
+            setTimeout(function() {
+              aspan.textContent = '';
+              aspan.dataset.dot = true;
+            }, CODE_VISIBLE_TIMEOUT);
```
davidflanagan: I suspect that 1 second will be too long for this, so it is good that it is in a constant that is easy to change.

I think that we also need to add code to ensure that if the user is entering their pin quickly, all previous digits are replaced with dots as soon as a new key is typed. There should never be more than one digit displayed at a time, but with the current code it looks like there can be.

apps/system/lockscreen/js/lockscreen_inputpad.js
```diff
@@ -163,7 +167,19 @@
...
+          })(span, value);
+        } else {
+          span.textContent = '';
           span.dataset.dot = true;
+        }
```
davidflanagan: The indentation of the entire if clause above is wrong, as should be clear from the fact that there are two close braces at the same column on lines 182 and 183.

apps/system/lockscreen/js/lockscreen_inputpad.js
davidflanagan: Ah, I see now that this is in a loop, so there won't ever be more than one digit displayed.  It would be nice to have some way of not leaving the setTimeout around to fire later, but I guess it is harmless to do it this way.

apps/system/lockscreen/js/lockscreen_inputpad.js
```diff
@@ -163,7 +167,19 @@
       var span = this.passcodeCode.childNodes[i];
       if (span) {
         if (this.states.passCodeEntered.length > i) {
+        var value = this.states.passCodeEntered.substr(i, this.states.passCodeEntered.length);
+        if ((this.states.passCodeEntered.length - 1 === i) && (keyCode && (keyCode !== 'b'))) {
```
davidflanagan: It looks like you're passing the keyCode just as some kind of flag for backspace maybe?  I'm not convinced you need this, but if you do it would be clearer to pass a boolean backspace argument.  Or just handle backspacing somewhere else.

apps/system/lockscreen/js/lockscreen_inputpad.js
davidflanagan: And besides, the user could have pressed backspace and entered a new key in those 1200ms and then you'd be validating the wrong PIN!

apps/system/lockscreen/js/lockscreen_inputpad.js
```diff
@@ -173,6 +189,8 @@
 
   LockScreenInputpad.prototype.handlePassCodeInput =
```
davidflanagan: As far as I can tell, this function is still only called in response to click events. There is no key handler for the actual keypad, and this will not work on a non-touch device. I don't think it is worth landing any of the PIN stuff until you have it working non-touch.

What's the problem there? Are you having trouble with key events or is this something that you just have not started on yet? If you have questions about this, let me know and I might be able to help.

apps/system/lockscreen/js/lockscreen_inputpad.js
davidflanagan: And while I'm on the topic, if I tell you to do something that seems stupid, it might very well be something stupid, and you should feel free to debate it with me...

apps/system/lockscreen/js/lockscreen_notifications.js
```diff
@@ -295,8 +299,12 @@
           node,
           oldLockScreenNode
         );
-      }
-      else {
+      } else if ((this.container.firstElementChild) && (this.container.firstElementChild.id === node.id)) {
+        this.container.replaceChild(
+          node,
+          this.container.firstElementChild
+        );
```
davidflanagan: If you really need this separate else clause in addition to the if clause above, please add a comment explaining why it is needed, and why it is a special case for the first element chlid. I don't understand this, and suspect it is wrong.

apps/system/lockscreen/js/lockscreen_state_manager.js
```diff
@@ -67,7 +67,8 @@
 (function(exports) {
```
davidflanagan: If you read the comment up above this line, it looks like this file is mostly about the keypad and slider animations. And since we won't have either of those in red square, you might consider removing this file entirely.

apps/system/lockscreen/js/lockscreen_state_manager.js
```diff
@@ -415,10 +416,52 @@
         this.onUnlockingApp();
         break;
       case 'lockscreenslide-activate-right':
```
davidflanagan: we're going to be removing the slide for red square, so this event won't ever happen anymore, right?  Probably better to remove it than to add code for it!

apps/system/lockscreen/js/lockscreen_state_manager.js
```diff
@@ -415,10 +416,52 @@
...
+                      index = i;
+                      break;
+                    }
+                  }
+                }
+                if (index != -1) {
+                  setTimeout(function() {
+                    var reqResult = request.result;
+                    reqResult[index].launch();
+                  }, 1000);
```
davidflanagan: Why would you make the user wait a second before launching the app for them?

apps/system/lockscreen/js/lockscreen_state_manager.js
```diff
@@ -415,10 +416,52 @@
...
         this.onActivateUnlock();
+        isFromNotification = true;
+        appID = detail.id;
         break;
       case 'lockscreen-request-unlock':
+      // if app is launched from the lockscreen notification screen then it will
+      // find the that app in system and launch it.
+        if (isFromNotification) {
+          isFromNotification = false;
+          var request = navigator.mozApps.mgmt.getAll();
```
davidflanagan: I don't think any of this code is necessary. The lockscreen can already launch apps when the user clicks on a notification. That is part of now notifications work. You should not have to turn the lockscreen into a window manager in order to respond to launch an app.  I'm not sure where the existing code is, but I suspect it is related to the tap() method at apps/system/js/notification.js:312

davidflanagan: Even if this was the right way to launch apps, isn't there a huge security hole here? What if the user has specified a PIN. Don't you need to check that the pin has been entered before launching anything?

apps/system/lockscreen/lockscreen.html
```diff
@@ -43,9 +43,15 @@
...
           <div id="lockscreen-alt-camera-button" role="button" data-l10n-id="camera-a11y-button" aria-label="Camera"></div>
         </div>
+        <div id="lockscreen-clock-widget">
         <div id="lockscreen-clock-time"></div>
         <div id="lockscreen-date"></div>
+        </div>
         <div id="lockscreen-charging"></div>
+        <div id="lockscreen-enter-code">
+          <img src="lockscreen/style/images/lockscreen_unlock.png" id="lockscreen-enter-code-icon">
+          <p data-l10n-id="enterpincode">Enter PIN code to unlock </p>
```
davidflanagan: Just specify the data-l10n-id. Please don't put any default text in the element. (Our build scripts actually pre-process the HTML and pre-localize the files to the default language)

apps/system/lockscreen/lockscreen_inputpad_frame.html
```diff
@@ -9,6 +9,8 @@
 
   <div id="lockscreen-panel-passcode" class="lockscreen-panel" data-l10n-id="panel-passcode">
     <p id="lockscreen-passcode-code"><span></span><span></span><span></span><span></span></p>
+    <img src="lockscreen/style/images/lockscreen_unlock.png" id="lockscreen-passcode-lock">
+    <p id="lockscreen-pincode-error-message" hidden="" data-l10n-id="incorrectPinCode">Incorrect PIN Code</p>
```
davidflanagan: We don't need the input pad for red square. I'd expect you to be deleting this files, not adding new lines to it.

apps/system/lockscreen/style/lockscreen.css
```diff
@@ -456,3 +460,25 @@ button::-moz-focus-inner {
   left: unset;
   right: 0;
 }
+#lockscreen-clock-parent{
```
davidflanagan: Do you mean #lockscreen-clock-widget? That's the id you put in the html file.

apps/system/js/notifications.js
Ashish-Developer: Will do the indentation.

apps/system/js/notifications.js
Ashish-Developer: Will change this code as suggested.

apps/system/js/notifications.js
Ashish-Developer: As per UX, if a particular notification has not focus a generic message will be displayed like "2 New Messages" and "3 Missed Calls" else messages clubbed in a ',' separated string.

apps/system/js/notifications.js
Ashish-Developer: Make sense, will do it in as suggested.

apps/system/js/notifications.js
Ashish-Developer: Will refactor the code as suggested.

apps/system/lockscreen/js/lockscreen_inputpad.js
Ashish-Developer: Will refactor it.

apps/system/lockscreen/lockscreen_inputpad_frame.html
Ashish-Developer: Will remove all code mapped with touch phone.

apps/system/lockscreen/lockscreen.html
Ashish-Developer: ok

apps/system/js/notifications.js
Ashish-Developer: it was a miss, will add code to remove notification.

apps/system/js/notifications.js
Ashish-Developer: In existing code "detail.icon" was never populated whenever a new notification arrived instead of that "detail.appIcon" getting populated.

shinglyu:
Sorry, I might be missing something. Do we have unit test for these code?

apps/system/js/notifications.js
egorlevichev: @Ashish-Developer it's strange, because when I made changes for homescreen notifications both values ``detail.icon`` & ``detail.appIcon`` were populated at least for missed calls and incoming messages.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=57ba4b520ae3ff1c5933f54b124b7639386a2b33

TatianaMigacheva:
This PR has been closed because RedSquare project has been canceled. All changes which were done by Bug#1210305 will be attached as Bugzilla patch.

PR #32758, opened 2015-10-26
Bug 1210860 - Video app launch time regression from 2.2 release

russnicoletti:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=20911c2838e0f5bb5a45567959cff43098137470

apps/video/index.html
```diff
@@ -51,12 +51,12 @@
     <div id="app-container">
       <div id="two-column-spearator"></div>
       <section role="region" id="thumbnail-views">
-        <gaia-header ignore-dir id="picker-header" action="back" class="thumbnails-list">
+        <gaia-header ignore-dir id="picker-header" action="back" class="thumbnails-list" no-font-fit>
```
russnicoletti: I didn't the add 'title-start' and 'title-end' attributes because I found the header title wasn't being centered properly. I investigated and found the gaia-header logic wasn't using the correct screen width when these attributes were supplied. When they were not supplied, the width was determined properly.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d4817e51f1e0123fe72db1bff32464bca77a4a7f

apps/video/index.html
punamdahiya: I noticed we are missing code to remove the attribute 'no-font-fit' to trigger gaia header font-fit logic after we hit 'visuallyloaded' mark, could that be the reason .

russnicoletti: I investigated the problem with the 'title-start' and 'title-end' by debugging the gaia-header code, so the font-fit logic was being applied at that time. It seems the code to remove the 'no-font-fit' attribute got lost after I did some refactoring.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=95ab2f9ef3fbdc951a07092e8647e4b46fe1c1f2

apps/video/js/db.js
```diff
@@ -122,6 +132,16 @@ function enumerateDB() {
...
+
+      // Remove the attribute to trigger gaia header font-fit logic, defer it
+      // until now to postpone the overhead. Also, do this asynchronously to
+      // avoid interrupting the enumerating of the db.
+      setTimeout(function() {
+        var headers = document.querySelectorAll('gaia-header');
+        for (var i = 0; i < headers.length; ++i) {
+          headers[i].removeAttribute('no-font-fit');
+        }
+      }, 0);
```
punamdahiya: nit: we don't need to explicitly specify 0 as the default is set to 0

apps/video/js/video.js
```diff
@@ -44,6 +44,8 @@ dom.player.mozAudioChannelType = 'content';
 
 function $(id) { return document.getElementById(id); }
 
+init();
```
punamdahiya: nit: for readability, we should move init call after all variables are declared ( Line 114)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=bbc28e00378b21e7365f0002c09280924bed4e89

PR #32778, opened 2015-10-27
Bug 1218805 - Implement home screen 'edit mode'. r=gmarty

Cwiiis:


apps/homescreen/js/app.js
```diff
@@ -726,6 +748,97 @@
...
+      if (!this.selectedIcon) {
+        return;
+      }
+
+      if (this.selectedIcon.app) {
+        if (!this.selectedIcon.app.removable) {
+          return;
+        }
+        navigator.mozApps.mgmt.uninstall(this.selectedIcon.app);
+      } else if (this.selectedIcon.bookmark) {
```
gmarty: Why an `if` statement here? Just a `else` is not enough?

apps/homescreen/js/app.js
```diff
@@ -726,6 +748,97 @@
...
+        return;
+      }
+
+      new MozActivity({
+        name: 'save-bookmark',
+        data: { type: 'url', url: this.selectedIcon.bookmark.id }
+      });
+    },
+
+    iconIsEditable: function(icon) {
```
gmarty: `iconIsEditable` means an icon can be edited (renamed) and removed? The naming is a bit confusing with `editSelectedIcon` above.
I suggest renaming `editSelectedIcon` to `renameSelectedIcon`. Edit will refer to both renaming and removing. What do  you think?

apps/homescreen/js/app.js
Cwiiis: Just covering bases in case somehow and icon with no app or bookmark ends up here... Shouldn't be possible, but I'd rather fail gracefully. If we ever add a third category of selectable icon, it'll be easier to transition I guess too...

apps/homescreen/js/app.js
Cwiiis: Agreed, will make these changes - nice spot :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=59bbd67e256c4ed3047f0a81f6749e96912818f2

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e43e1068566f1a6e28f54a9914c9396081992e29

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1eeb8233a179255ebecad610d5d0930ef54d235a

apps/homescreen/js/app.js
```diff
@@ -726,6 +757,100 @@
...
+        this.selectedIcon.classList.remove('selected');
+        this.selectedIcon.removeEventListener('touchstart', this);
+        this.selectedIcon = null;
+      }
+
+      var selectedEditable = false;
+      var selectedRemovable = false;
+
+      if (icon) {
+        selectedEditable = !!icon.bookmark;
```
gmarty: In order to keep the same wording, this should now be `selectedRenamable`.

apps/homescreen/js/app.js
```diff
@@ -726,6 +757,100 @@
...
+      }
+
+      new MozActivity({
+        name: 'save-bookmark',
+        data: { type: 'url', url: this.selectedIcon.bookmark.id }
+      });
+    },
+
+    iconIsEditable: function(icon) {
+      return (icon.bookmark || (icon.app && icon.app.removable)) ? true : false;
```
gmarty: Why not just:
```js
return (icon.bookmark || (icon.app && icon.app.removable));
```

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=51514eeba02af61c645a166a41730061ef45d66f

apps/homescreen/js/app.js
Cwiiis: Because that ends up returning the objects (you could put a !! before it, which would be similar to what I've done). Most things are fine, but it came up in a unit test and I've had funny things happen from returning things like `undefined` instead of `false`, so decided to be explicit.

apps/homescreen/js/app.js
```diff
@@ -726,6 +757,100 @@
...
+        this.selectedIcon.removeEventListener('touchstart', this);
+        this.selectedIcon = null;
+      }
+
+      var selectedRenameable = false;
+      var selectedRemovable = false;
+
+      if (icon) {
+        selectedRenameable = !!icon.bookmark;
+        selectedRemovable = selectedRenameable || icon.app.removable;
```
albertopq: classList.toggle only works with booleans (it won't remove the class if the evaluation is null or undefined) so probably better to add !!() in here.

Cwiiis: Both selectedRenameable and icon.app.removable are booleans, so unless something goes wrong, this should be fine. Famous last words, perhaps...

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7bdc51cebcbae08afb659d1a1e63db6fd582ee34

PR #32781, opened 2015-10-27
Bug 1218327 - Check for updateManifest when determining appUsage. r=f…

tamarahills:
…abrice, r=marshall

### DIFF
--- a/apps/system/js/app_usage_metrics.js
+++ b/apps/system/js/app_usage_metrics.js
@@ -958,15 +958,31 @@
   };
 
   UsageData.prototype.recordInstall = function(app) {
+    var self = this;
     if (!this.shouldTrackApp(app)) {
       return false;
     }
 
+    // If the app manifest is undefined, this means we need to listen to the
+    // downloadsuccess event fired when the download is finished.  The full
+    // manifest will be available on the downloadsuccess.
+    if (typeof app.manifest === 'undefined') {
+      app.addEventListener('downloadsuccess', function onSuccessfulDownload() {
+        app.removeEventListener('downloadsuccess', onSuccessfulDownload);
+        self.incrementInstallRecord(app);
+      });
+      return false;
+    } else {
+      self.incrementInstallRecord(app);
+    }
+    return true;
+  };
+
+  UsageData.prototype.incrementInstallRecord = function(app) {
     var usage = this.getAppUsage(app);
     usage.installs++;
     this.needsSave = true;
     debug(app.manifestURL, 'installed');
-    return true;
   };
 
   UsageData.prototype.recordUninstall = function(app) {
--- a/apps/system/test/unit/mock_app.js
+++ b/apps/system/test/unit/mock_app.js
@@ -70,6 +70,15 @@ MockApp.prototype.mTriggerDownloadError = function(error) {
     });
   }
 };
+MockApp.prototype.addEventListener = function(evt, callback) {
+  callback();
+};
+
+MockApp.prototype.removeEventListener = function(evt, callback) {
+};
+
+MockApp.prototype.dispatchEvent = function(evt) {
+};
 
 MockApp.prototype.mTriggerDownloadProgress = function(progress) {
   this.progress = progress;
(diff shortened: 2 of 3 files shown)


apps/system/js/app_usage_metrics.js
```diff
@@ -805,8 +805,11 @@
     var addOn = false;
 
     if (app !== null && typeof app === 'object') {
+      // updateManifest is used by packaged apps until they are installed
+      var manifest = app.manifest ? app.manifest : app.updateManifest;
```
fabricedesre: If the manifest you get is the updateManifest, we won't have set the `addon` role yet, since we only know that once we have downloaded the package. So I'm not convinced that this will do what you need, but I haven't check when getAppUsage() is called.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f9a7c76c9b198f28041fbe27fb37fb0f822665de

apps/system/js/app_usage_metrics.js
```diff
@@ -957,11 +962,42 @@
...
+      return false;
+    }
+
+    // If it's not privileged or certified, it must be a hosted app.  The type
+    // attribute is required for privileged and certified, but not for
+    // web (hosted).  For now, hosted is the only type of app we want to track
+    // on the install since we won't get a downloadsuccess event.
+    if (typeof app.manifest.type !== 'undefined') {
+      if (app.manifest.type === 'privileged' ||
+          app.manifest.type === 'certified') {
```
fabricedesre: packaged apps can also be unprivileged, so that's not correct.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a42224c08079a03bea8a8b7d0fa75a9b374b59b3

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ea76f690e66a36e4b585d226415caaf07937cdaa

apps/system/test/unit/mock_app.js
```diff
@@ -70,7 +70,13 @@ MockApp.prototype.mTriggerDownloadError = function(error) {
     });
   }
 };
+MockApp.prototype.addEventListener = function(evt, callback) {
+callback();
```
marshall: in the future it might be better to fire events from the test rather than to just blindly execute `callback` when `addEventListener` is called

apps/system/js/app_install_manager.js
```diff
@@ -498,6 +498,12 @@
 
     handleDownloadSuccess: function ai_handleDownloadSuccess(evt) {
       var app = evt.application;
+      var evtDownload = new CustomEvent('downloadsuccess',
+        { bubbles: true,
+          cancelable: false,
+          detail: { application: app } });
+      window.dispatchEvent(evtDownload);
```
fabricedesre: I don't understand why you need to dispatch this event on |window| since you will add your event listener on the app object itself in |recordInstall|

apps/system/test/unit/mock_app.js
tamarahills: Hi Marshall,

I tried doing this, but the event is attached to the app, vs. the window.  window.dispatchEvent seems to work nicely for all the events in AUM.  However, now, that the event is dispatched on the app, and the app uses this MockApp, I had to add add/removeEventListener and dispatchEvent to the MockApp, so they don't really do much right now.  Maybe there is something I'm missing here.

apps/system/js/app_install_manager.js
tamarahills: Hi, yes, that's correct.  I will fix that.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=30da0b58ee940e3054249533a298f4060dc64bea

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4803a4122d5f051331230c72acfc3738d7d3a6ea

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=94f70908a0e316518a6ba27d2c6ab9db8efa04d8

PR #32806, opened 2015-10-28
Bug 1218720 - Statusbar does not work in RTL r=mhenretty

albertopq:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9d5c779dedbe38cff8d2eff64ac0d5b6374af05a

apps/system/style/statusbar/statusbar.css
```diff
@@ -76,6 +76,11 @@ to be available for moz-elements */
     width: calc(100% - var(--software-home-button-height));
     left: 0;
   }
+
+  html[dir="rtl"] #screen.software-button-enabled #statusbar {
+    width: calc(100% - var(--software-home-button-height));
+    right: var(--software-home-button-height);
+  }
```
autra: Just passing by, because I'm on the team working for RTL for this release and this shows up on our bug list.

Please follow the [CSS RTL guidelines](https://wiki.mozilla.org/Gaia/CSS_Guidelines#Direction_.28RTL.2FLTR_and_BiDi.29) (imo that's also a good reading to know about RTL front-end dev :-) ).

Here that means you should also have a `html[dir="ltr"]` rule containing corresponding css for ltr. The rule without any `html[dir]` selector should be completely direction agnostic. But that's a general remark, you won't need it here I think.

For this case:
- you should not need to specify width in the RTL rule, as it is the same in the general rule.
- It's very odd that you have a right set to a different value in RTL than the left one in LTR. Maybe it's because you didn't set a `left: unset` or `left: auto` in the RTL part. But anyway....
- ... if `#statusbar` does not have any dir="auto|rtl|ltr" you can use [`offset-inline-start`](https://developer.mozilla.org/en-US/docs/Web/CSS/offset-inline-start) instead of `left`, and you should not need any rtl rule:

So TL;DR here you may just need 
```css
#screen.software-button-enabled #statusbar {
    width: calc(100% - var(--software-home-button-height));
    offset-inline-start: 0;
}
```

Thanks!

apps/system/style/statusbar/statusbar.css
```diff
@@ -606,3 +611,11 @@ html[dir="rtl"] #statusbar-call-forwardings > * { float: right; }
 .utility-tray .sb-hide-operator .statusbar > .sb-icon-operator {
   visibility: hidden;
 }
+
+/* RTL */
+
+html[dir="rtl"] .statusbar {
```
autra: The RTL CSS guidelines also specifies that this rule should be near its LTR counterpart. That eases reviews and allows them not to be forgotten. 

But anyway, here you can use `offset-inline-start` again in the original rule and get rid of this :-) Thanks!

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6deb7ab0d62ca027e4d8a6c2cac4eda4f5ef82db

apps/system/style/statusbar/statusbar.css
```diff
@@ -74,7 +74,7 @@ to be available for moz-elements */
 @media (orientation: landscape) {
   #screen.software-button-enabled #statusbar {
     width: calc(100% - var(--software-home-button-height));
-    left: 0;
+    right: var(--software-home-button-height);
```
albertopq: SHB is always shown at the right

PR #32821, opened 2015-10-28
Bug 1219264 - [Messages][NG] Move ActivityShim to shim host iframe. r=schung

azasypkin:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4cea2909ce62135bc9ce39d93af3dcc70ef1e252

apps/sms/services/js/bridge_service_mixin.js
```diff
@@ -55,6 +55,10 @@
         });
       }
 
+      if (this.onConnected) {
```
azasypkin: I found it as the simplest way, other alternative was smth like ```BridgeServiceMixin.ON_CONNECT = Symbol('ON_CONNECT')``` and
```
var service = {
  ...
  [BridgeServiceMixin.ON_CONNECT]() {}
  ...
}
```

What do you think?

apps/sms/views/shared/js/shim_host_bootstrap.js
```diff
@@ -7,6 +8,12 @@
 'use strict';
 
 exports.bootstrap = function(appInstanceId) {
+  // We can't handle system messages inside iframe right now, so we access
+  // parent window here. This should be removed once bug 818000 is resolved.
+  ActivityShim.init(
```
azasypkin: Not sure what do you think about this hack :)

apps/sms/views/shared/js/startup.js
```diff
@@ -105,8 +104,13 @@ var Startup = exports.Startup = {
       // dispatch contentInteractive when all the modules initialized
       SilentSms.init();
 
-      ActivityHandler.init();
-      SystemMessageHandler.init();
+      // When app is run as inline activity we don't want to handle any other
+      // system messages other than "activity" system message.
+      if (navigator.mozHasPendingMessage('activity')) {
```
azasypkin: tbh I think it's better to have all these mozHasPendingMessage in one place in startup.js

apps/sms/services/js/activity/activity_shim.js
```diff
@@ -44,30 +47,45 @@
...
-      });
+    init(appInstanceId, setMessageHandler) {
+      if (!appInstanceId) {
+        throw new Error('AppInstanceId is required!');
+      }
+
+      this[priv.setMessageHandler] = setMessageHandler;
+
+      this.initService(
+        new BroadcastChannel(`${SERVICE_NAME}-channel-${appInstanceId}`)
```
azasypkin: I wanted to use ```parent``` instead of ```BC``` here, but I can't because of https://github.com/gaia-components/bridge/issues/83 :/

apps/sms/views/shared/js/startup.js
steveck-chung: I just added mozHasPendingMessage('notification') in the navigation.init because of Bug 1192263. Do you think we should move this to startup and handling the status in startup instead of navigation?

apps/sms/views/shared/js/shim_host_bootstrap.js
steveck-chung: IIRC that's the main reason why we still leave ActivityShim in window. So back to the original problem about the memory consumption: Will using the optimized client in window could reduce further than optimized bridge in bug 1218091? If the effect is limited, maybe we can still land bug 1218091 first and keep this one with using only client in the window after bug 818000, wdyt?

azasypkin: Well, I'll remeasure today to have numbers, but AFAIK the initial reason to keep this service here was that we didn't have shim host iframe (it was first service we landed) :) But even without numbers I think it's good to get rid of single shim in window for the consistency and remove both bigger bridge.js and bridge service mixin from view. And once system messages mechanism is fixed, it will require single line fix here :)

apps/sms/views/shared/js/startup.js
azasypkin: IMO if we can keep code dealing with mozhaspendingmessage (that should go away eventually) in single place it will be easier to maintain all the tricky cases that are happening on startup (though I haven't read bug 1192263 to understand if it's about startup) :)

apps/sms/views/shared/js/shim_host_bootstrap.js
azasypkin: Yeah, as expected, there is no any noticeable improvement per raptor, do you have any concerns regarding moving ActivityShim to iframe?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=387d4de6180fe8705bcda389a14e51640d41f27e

apps/sms/services/js/activity/activity_shim.js
steveck-chung: Ok, it seems like we should migrate all the direct window - iframe communication to this instead of BroadcastChannel. Maybe we can create a bug and leave a comment here?

apps/sms/services/js/bridge_service_mixin.js
steveck-chung: It's simple and clear like other mixin method(initService), so let's keep it as it is.

apps/sms/views/shared/js/startup.js
steveck-chung: At least is good move to remove the unnecessary `ActivityShim.hasPendingRequest` and dependency with ActivityShim :)

apps/sms/services/js/activity/activity_shim.js
azasypkin: Left comment and filed a bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1223363

apps/sms/services/js/bridge_service_mixin.js
azasypkin: Okay!

apps/sms/views/shared/js/startup.js
azasypkin: Yeah :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=85a9b856caf2052d82fe7bf7901b561f4b612ee9

PR #32822, opened 2015-10-28
Bug 1215068 - Use  Intl formatWithPosition in mozIntl

zbraniecki:
https://bugzilla.mozilla.org/show_bug.cgi?id=1215068

### DIFF
--- a/shared/js/moz_intl.js
+++ b/shared/js/moz_intl.js
@@ -71,71 +71,50 @@ global.mozIntl = {
   DateTimeFormat: function(locales, options) {
     const resolvedOptions = Object.assign({}, options);
 
-    if (resolvedOptions.dayperiod) {
-      if (resolvedOptions.hour === undefined) {
-        resolvedOptions.hour = 'numeric';
-      }
-    }
-
     if (resolvedOptions.hour !== undefined &&
         resolvedOptions.hour12 === undefined) {
       resolvedOptions.hour12 = navigator.mozHour12;
     }
 
-    if (resolvedOptions.dayperiod === undefined &&
-        resolvedOptions.hour12 === true) {
-      resolvedOptions.dayperiod = true;
-    }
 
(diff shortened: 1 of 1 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c096182cde2d2b2946fb5b891b0936a96926c98b

shared/js/moz_intl.js
```diff
@@ -71,71 +71,50 @@ global.mozIntl = {
...
 
-    if (resolvedOptions.dayperiod === undefined &&
-        resolvedOptions.hour12 === true) {
-      resolvedOptions.dayperiod = true;
-    }
 
     var intlFormat = new Intl.DateTimeFormat(locales, resolvedOptions);
 
     resolvedOptions.locale = intlFormat.resolvedOptions().locale;
     resolvedOptions.hour12 = intlFormat.resolvedOptions().hour12;
```
stasm: Why are you overwriting `hour12` here unconditionally?  Should this code check if line 76 was run first?

shared/js/moz_intl.js
```diff
@@ -71,71 +71,50 @@ global.mozIntl = {
...
-          var formatter = global.mozIntl.DateTimeFormat(
-            navigator.languages, localOptions);
-          var tokenString = formatter.format(date);
+          var tokenPosition = result.fieldPosition[token];
+
+          //XXX: In the future we may be smarter and use position to
+          // concatenate new string, but for now, we can just replace
+          // the tokenString using str.replace
+          var tokenString =
+            string.substr(tokenPosition.beginIndex, tokenPosition.endIndex);
```
stasm: Wouldn't it make more sense to use `reduce` on `Object.keys(result.fieldPosition)` here instead of the `for-in` and concatenate, as you suggested in the comment?

PR #32839, opened 2015-10-29
Bug 1218374 - Properly stop the conversation rendering when entering …

julienw:
…another conversation r=azasypkin

### DIFF
--- a/apps/sms/test/marionette/generators/thread.js
+++ b/apps/sms/test/marionette/generators/thread.js
@@ -59,7 +59,7 @@ var ThreadGenerator = {
 
     for (var i = 0; i < parameters.numberOfMessages; i++) {
       var message = null;
-      var messageBody = thread.body;
+      var messageBody = thread.body + ' (threadId: ' + thread.id + ')';
       if (i) {
         messageBody +=
           ' (messageId: ' + (ThreadGenerator.uniqueMessageId + 1) + ')';
--- a/apps/sms/test/marionette/lib/views/conversation/message_accessors.js
+++ b/apps/sms/test/marionette/lib/views/conversation/message_accessors.js
@@ -34,7 +34,7 @@ MessageAccessor.prototype = {
       );
 
       return {
-        id: messageNode.getAttribute('data-message-id'),
+        id: +messageNode.getAttribute('data-message-id'),
         type: messageClass.indexOf('mms') >= 0 ? 'mms' : 'sms',
         content: messageNode.findElement(SELECTORS.content).text(),
         attachments: attachments.map(function(attachment) {
--- a/apps/sms/test/marionette/notification_test.js
+++ b/apps/sms/test/marionette/notification_test.js
@@ -3,6 +3,7 @@
 
 var assert = require('chai').assert;
 
+var ThreadGenerator = require('./generators/thread');
 var Messages = require('./lib/messages.js');
 var Storage = require('./lib/storage.js');
 var InboxView = require('./lib/views/inbox/view');
@@ -226,6 +227,59 @@ marionette('Message notification tests', function() {
       openNotification();
 
       assertMessagesIsInCorrectState();
+    });
+  });
+
+  suite('Display a conversation from a notification', function() {
+    var messagesStorage;
+    var thread1, thread2;
+
+    setup(function() {
+      thread1 = ThreadGenerator.generate({
+        numberOfMessages: 500,
+        participants: ['999']
+      });
+      thread2 = ThreadGenerator.generate({
+        numberOfMessages: 5,
+        participants: ['888']
+      });
+
+      messagesStorage = [ thread1, thread2 ];
+    });
+
+    test('Clicking a notification while a conversation is loading', function() {
+      messagesApp.launch();
+      storage.setMessagesStorage(messagesStorage);
+
+      var inbox = new InboxView(client);
+      var conversation = inbox.goToConversation(thread1.id);
+
+      // Receive 'notification' system message and fill up the storage.
+      messagesApp.sendSystemMessage('notification', {
+        clicked: true,
+        data: { threadId: thread2.id, id: thread2.messages[0].id }
+      });
+
+      var thread2Ids = thread2.messages.map(
+        function(message) { return message.id; }
+      );
+
+      // Our testing strategy is:
+      // 1. Wait that all messages are rendered for thread2.
+      thread2Ids.forEach(function(messageId) {
+        conversation.findMessage(messageId);
+      });
+
+      // 2. Look at all rendered messages and see if one of these is part of
+      // thread1.
+      conversation.messages(
+        { first: thread2Ids.length + 1 }
+      ).forEach(function(message) {
+        assert.include(
+          thread2Ids, message.id,
+          'Message ' + message.id + ' is in thread 2'
+        );
+      });
     });
   });
 });
--- a/apps/sms/views/conversation/js/conversation.js
+++ b/apps/sms/views/conversation/js/conversation.js
@@ -1468,6 +1468,18 @@ var ConversationView = {
     this._stopRenderingNextStep = true;
   },
 
+  /**
+   * This method checks whether we're in the right state to render something for
+   * this threadId.
+   * @param {Number} threadId The threadId we'd like to render.
+   * @returns {Boolean} True if we can render for this threadId.
+   */
+  shouldRenderForThreadId(threadId) {
+    return !this._stopRenderingNextStep &&
+      this.activeThread &&
+      this.activeThread.id === threadId;
+  },
+
   // Method for rendering the first chunk at the beginning
   showFirstChunk: function conv_showFirstChunk() {
     // Show chunk of messages
@@ -1519,13 +1531,13 @@ var ConversationView = {
     }).bind(this);
 
     var onRenderMessage = (function renderMessage(message) {
-      if (this._stopRenderingNextStep) {
+      if (!this.shouldRenderForThreadId(threadId)) {
         // stop the iteration and clear the taskQueue
         taskQueue = null;
         return false;
       }
       taskQueue.push(() => {
-        if (!this._stopRenderingNextStep) {
+        if (this.shouldRenderForThreadId(threadId)) {
           Threads.registerMessage(message);
           return this.appendMessage(message,/*hidden*/ true);
         }
@@ -1538,7 +1550,7 @@ var ConversationView = {
       return true;
     }).bind(this);
 
-    if (this._stopRenderingNextStep) {
+    if (!this.shouldRenderForThreadId(threadId)) {
       // we were already asked to stop rendering, before even starting
       return;
     }
@@ -1790,7 +1802,7 @@ var ConversationView = {
     if (doNotDisplayNotice) {
       return null;
     }
-    
+
     var unit = MONTH;
     var noticeL10nIds = {
       [MONTH]:  'late-arrival-notice-in-months',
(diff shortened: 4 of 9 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=bd127661b347f5a93655f22dc8fa3b129c7a1412

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=acc30e657ff3c1f64c4fcd32a739ef18f66359a2

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4725e3c8e0372e36213c8b9a963ab61d78a6cb86

apps/sms/test/marionette/generators/thread.js
```diff
@@ -59,7 +59,7 @@ var ThreadGenerator = {
 
     for (var i = 0; i < parameters.numberOfMessages; i++) {
       var message = null;
-      var messageBody = thread.body;
+      var messageBody = thread.body + ' (threadId: ' + thread.id + ')';
```
julienw: this makes it easier to see if the message is in the right thread id when debugging

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=fe163080c42b7de902d2ecab1a131bca64cb9351

apps/sms/test/marionette/lib/views/conversation/message_accessors.js
```diff
@@ -34,7 +34,7 @@ MessageAccessor.prototype = {
       );
 
       return {
-        id: messageNode.getAttribute('data-message-id'),
+        id: +messageNode.getAttribute('data-message-id'),
```
julienw: the id is a number :)

apps/sms/test/marionette/notification_test.js
```diff
@@ -28,6 +29,29 @@ marionette('Message notification tests', function() {
...
+    assert.equal(
+      messagesApp.Conversation.headerTitle.text(), smsMessage.sender
+    );
+
+    // When we open thread from notification, message input should not be
+    // focused.
+    assertIsNotFocused(
+      messagesApp.Composer.messageInput, 'Message input should not be focused'
+    );
+  }
```
julienw: I changed this in a previous version of the patch, and then I left it even than I don't need it. Tell me if you'd rather want that I revert the change.

apps/sms/test/marionette/notification_test.js
```diff
@@ -225,7 +226,63 @@ marionette('Message notification tests', function() {
...
+
+    test('Clicking a notification while a conversation is loading', function() {
+      messagesApp.launch();
+      storage.setMessagesStorage(messagesStorage);
+      // Set empty contacts store.
+      storage.setContactsStorage();
+
+      var inbox = new InboxView(client);
+      var conversation = inbox.goToConversation(
+        thread1.id, { shouldWait: false }
```
julienw: we need to go as fast as possible to trigger the bug

julienw: without this change, the test is not failing on master :/

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4c6fc224bfabbe652c78c901f8506ae5d2728568

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1802,9 +1813,9 @@ var ConversationView = {
     if (lateness < limits.IN_HOURS)       { unit = MINUTE; } 
     else if (lateness < limits.IN_DAYS)   { unit = HOUR; } 
     else if (lateness < limits.IN_MONTHS) { unit = DAY; }
-    
+
     lateness = Math.floor(lateness / unit);
-    
+
```
julienw: some white space changes from an earlier bug

apps/sms/views/conversation/test/unit/conversation_test.js
```diff
@@ -160,6 +160,21 @@ suite('conversation.js >', function() {
...
+    Threads.get.withArgs(threadId).returns(mockThread);
+
+    var transitionArgs = {
+      id: strThreadId,
+      meta: { next: { panel: 'thread', id: strThreadId }}
+    };
+    return ConversationView.beforeEnter(transitionArgs).then(
+      () => ConversationView.afterEnter(transitionArgs)
+    );
+  }
```
julienw: quite complex setup to have a proper state

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1468,6 +1468,17 @@ var ConversationView = {
     this._stopRenderingNextStep = true;
   },
 
+  /**
```
azasypkin: __nit (:bike: ):__ It seems would be better to have method description as well (more in description and less for return statement description)

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1519,13 +1530,13 @@ var ConversationView = {
...
     var onRenderMessage = (function renderMessage(message) {
-      if (this._stopRenderingNextStep) {
+      if (!this.shouldRenderForThreadId(threadId)) {
         // stop the iteration and clear the taskQueue
         taskQueue = null;
         return false;
       }
       taskQueue.push(() => {
-        if (!this._stopRenderingNextStep) {
+        if (this.shouldRenderForThreadId(threadId)) {
```
azasypkin: maybe someday our TaskQueue can have ```clear/cancel``` method that will automagically not allow to execute next step and return rejected promise, however there likely be overhead for the wrapper function that will perform this check :)

apps/sms/test/marionette/generators/thread.js
azasypkin: yeah, makes sense to have smth like this

apps/sms/test/marionette/lib/views/conversation/message_accessors.js
azasypkin: ah, nice catch!

apps/sms/test/marionette/lib/views/inbox/view.js
```diff
@@ -49,17 +58,17 @@ InboxView.prototype = {
     return this._createNewMessageView();
   },
 
-  _createNewMessageView: function() {
+  _createNewMessageView: function(shouldWait) {
     var NewMessageView = require('../new-message/view');
     var newMessageView = new NewMessageView(this.client);
-    newMessageView.accessors.waitToAppear();
+    shouldWait && newMessageView.accessors.waitToAppear();
```
azasypkin: __question:__ mm what was the issue with always waiting?

apps/sms/test/marionette/notification_test.js
```diff
@@ -225,7 +226,63 @@ marionette('Message notification tests', function() {
...
+      });
+
+      messagesStorage = [ thread1, thread2 ];
+    });
+
+    test('Clicking a notification while a conversation is loading', function() {
+      messagesApp.launch();
+      storage.setMessagesStorage(messagesStorage);
+      // Set empty contacts store.
+      storage.setContactsStorage();
```
azasypkin: __nit:__ do we really need this? It seems we don't inject contacts mock here

apps/sms/test/marionette/notification_test.js
azasypkin: __question:__ mmmmm, I believe there is a chance we won't have the issue on such _small_ conversations in real life as well :) What about 200-300 messages in the first thread instead? Closer to real use case and I believe it should fail :)

apps/sms/test/marionette/notification_test.js
```diff
@@ -225,7 +226,63 @@ marionette('Message notification tests', function() {
...
+        thread1.id, { shouldWait: false }
+      );
+
+      // Receive 'notification' system message and fill up the storage.
+      messagesApp.sendSystemMessage('notification', {
+        clicked: true,
+        data: { threadId: thread2.id, id: thread2.messages[0].id }
+      });
+
+      conversation.accessors.waitToAppear();
```
azasypkin: __nit:__ looks like this one is also not needed or I'm missing something?

apps/sms/test/marionette/notification_test.js
```diff
@@ -225,7 +226,63 @@ marionette('Message notification tests', function() {
...
+      var thread2Ids = thread2.messages.map(
+        function(message) { return message.id; }
+      );
+
+      // Our testing strategy is:
+      // 1. Wait that all messages are rendered for thread2.
+      thread2Ids.forEach(function(messageId) {
```
azasypkin: Looks like ```findMessage``` doesn't guarantee that message is rendered already so it can return null, how about new property on Conversation view instead?
```
client.waitFor(function() {
  return conversation.numberOfMessages >= thread2.messages.length;
});

// or much more slower option, but that exists already
client.waitFor(function() {
  return conversation.messages.length >= thread2.messages.length;
});
```

apps/sms/test/marionette/notification_test.js
azasypkin: And maybe much less messages for the thread#2

apps/sms/test/marionette/notification_test.js
azasypkin: Less unnecessary changes - less chances to break something ;) But in this case up to you, I don't mind really

julienw: you're right, I'll revert this

apps/sms/test/marionette/notification_test.js
julienw: if I wait neither earlier nor here, I'll have the infamous "crash but can't stackwalk" message :/ Which doesn't help much. From my tests, it seems that we need at least that the window is unhidden (but we don't really care that it's been transformed)

apps/sms/test/marionette/notification_test.js
julienw: are you sure of this ? because it seemed it was waiting for me :/

apps/sms/views/conversation/js/conversation.js
julienw: There has been lengthy debates about cancelable promises also ;)

apps/sms/test/marionette/notification_test.js
azasypkin: > are you sure of this ? because it seemed it was waiting for me :/ 

Nope, but let me check :)

apps/sms/test/marionette/notification_test.js
julienw: ok, I tried with much more messages in the first thread, and you're right it reproduces the problem now. I'll revert the "shouldWait" changes.

apps/sms/test/marionette/notification_test.js
azasypkin: Ah yeah, it waits for the default timeout and fails if it can't find, likely was thinking about ```findElements```, sorry, please ignore ```Looks like findMessage doesn't guarantee that message is rendered already so it can return null```, but I can't recall why I added null check there :/

julienw: I thought it was throwing if it couldn't find the element after some timeout.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4fe3736fdb9c3c00191a57915341a84bde3eb179

apps/sms/test/marionette/notification_test.js
```diff
@@ -228,4 +229,57 @@ marionette('Message notification tests', function() {
...
+        conversation.findMessage(messageId);
+      });
+
+      // 2. Look at all rendered messages and see if one of these is part of
+      // thread1.
+      conversation.messages(
```
azasypkin: __optional nit:__ we can probable fail earlier with additional assert, but it's veeeery optional :)
```
var messages = conversation.messages({ first: thread2Ids.length + 1 });

assert.equal(messages.length, thread2Ids.length);

messages.forEach(.....)
```

julienw: it's basically the same IMO :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=938a85afbe576f9a2af36679343701e75e459f27

PR #32845, opened 2015-10-29
Bug 1219610 - JavaScript error: app://browser.gaiamobile.org/js/setti…

ferjm:
…ngs.js, line 171: TypeError: this.getHomepage is not a function. r=yifan

### DIFF
--- a/tv_apps/browser/js/sync/settings.js
+++ b/tv_apps/browser/js/sync/settings.js
@@ -165,7 +165,7 @@
           // but we only want to do that if its enabled via user action
           // (and not because it is already enabled from a previous run).
           if (this.state === 'enabling') {
-            Settings.show();
+            Settings.show.bind(Settings)();
           }
           this.showScreen(ENABLED);
           this.showUser(message.user);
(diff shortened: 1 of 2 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=51f5b4011f805f2cf1983a0a12b081c93fde76dd

tv_apps/browser/js/sync/settings.js
```diff
@@ -165,7 +165,7 @@
           // but we only want to do that if its enabled via user action
           // (and not because it is already enabled from a previous run).
           if (this.state === 'enabling') {
-            Settings.show();
+            Settings.show.bind(Settings)();
```
begeeben: Comparing to the functionality, the changed code is identical to the original code. Tried to reproduce without this patch but don't know what steps trigger the error. Does this line make a difference?

ferjm: It does :) Not binding Settings here will make `this` at [1] not be Settings and so it will throw.

You have to login into Sync from the toolbar to see this error in the console.

[1] https://github.com/mozilla-b2g/gaia/blob/master/tv_apps/browser/js/settings.js#L171

PR #32854, opened 2015-10-29
Bug 1218928 - Remove invalid scopes from taskgraph.json

djmitche:
These two scopes don't correspond to anything anymore -- wrong provisionerId, wrong workerType.  Keeping them in task.scopes means that they need to be in the role for this repository, even though they are not actually used.


autolander:

PR #32857, opened 2015-10-29
Bug 1217989 - [Gallery] Show resolution only for image files inside info view

punamdahiya:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5e7775f9a5f3e768384750f9fcd4eedf9cee74e3

apps/gallery/js/info.js
```diff
@@ -52,10 +52,30 @@ function showFileInformation(fileinfo) {
...
+      // Hide Resolution for video files. See Bug 1217989
+      setFieldVisibility('info-resolution', fileinfo);
     });
   }
 
   function getFileName(path) {
     return path.split('/').pop();
   }
+
+  function setFieldVisibility(id, fileinfo) {
```
davidflanagan: nit: this would be a cleaner api if you pass a boolean as the second argument, and check fileinfo.metadata.video above where you call it instead of inside this function.

apps/gallery/js/info.js
```diff
@@ -52,10 +52,30 @@ function showFileInformation(fileinfo) {
...
+    // Field label in info view
+    var label = $(id).previousElementSibling;
+
+    // Check for video file and hide respective field label and
+    // its value and remove bottom border from the previous field.
+    if (fileinfo.metadata.video) {
+      $(id).classList.add('hide');
+      label.classList.add('hide');
+      label.previousElementSibling.classList.add('no-border');
```
davidflanagan: Note that you only want the no-border thing if the element you're hiding is the last one in the list.  So the way this is written it only works right for hiding the resolution.  Arguably that is good enough to fix this blocker.  But you might be able to do this with something like `if ($(id).nextElementSibling === null) { /* set the no-border thing */ }`

apps/gallery/style/info.css
```diff
@@ -40,6 +40,14 @@
...
 }
 
+.no-border {
+ border-bottom: none !important;
+}
+
+.hide {
+  display: none !important;
+}
+
```
davidflanagan: This is not a scoped stylesheet, so what you add here applies to the whole app.  I think that these classes should be prefixed with `#info-view > section > dl > dd ` or something like that.

And if you do that, then you might not need the !important and the xfail.list changes.

Alternatively you could avoid the stylesheet changes altogether by explicitly setting inline styles (style.display = 'none';) in the JS file.  That would simplify the patch to just a single file.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1eb22d7f2d1acd1d0fd797578de9b02a9146c09b

apps/gallery/js/info.js
punamdahiya: Good catch, updated patch with the check.

apps/gallery/style/info.css
punamdahiya: Thanks for suggesting, this is definitely better. Keeping no-border class as 
label.previousElementSibling.classList.remove('no-border') is cleaner than inline style
label.previousElementSibling.style.borderBottom = '0.1rem solid rgba(255,255,255,0.1)'

apps/gallery/js/info.js
punamdahiya: Done

PR #32861, opened 2015-10-29
Bug 1218928 - Remove invalid scopes from taskgraph.json

djmitche:
These two scopes don't correspond to anything anymore -- wrong provisionerId, wrong workerType.  Keeping them in task.scopes means that they need to be in the role for this repository, even though they are not actually used.

### DIFF
--- a/taskgraph.json
+++ b/taskgraph.json
@@ -17,10 +17,7 @@
         "scopes": [
           "docker-worker:cache:gaia-tc-vcs",
           "docker-worker:cache:gaia-linux-cache",
-          "docker-worker:cache:gaia-misc-caches",
-          "docker-worker:image:quay.io/mozilla/raptor-tester:latest",
-          "queue:define-task:aws-provisioner/gaia",
-          "queue:create-task:aws-provisioner/gaia"
+          "docker-worker:cache:gaia-misc-caches"
         ],
         "payload": {
           "cache": {


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4b0966e0f99e47259bd7c6ec878150ea0942dcca

taskgraph.json
```diff
@@ -18,9 +18,7 @@
           "docker-worker:cache:gaia-tc-vcs",
           "docker-worker:cache:gaia-linux-cache",
           "docker-worker:cache:gaia-misc-caches",
-          "docker-worker:image:quay.io/mozilla/raptor-tester:latest",
-          "queue:define-task:aws-provisioner/gaia",
-          "queue:create-task:aws-provisioner/gaia"
+          "docker-worker:image:quay.io/mozilla/raptor-tester:latest"
```
gregarndt: I don't even think you need this either.  The raptor tests are no longer run on TC, and that's a public image if I recall which doesn't require scopes.

djmitche: You're right, let me try that.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7efcd4c8c293a4010e647f41a515bd0b3c124249

djmitche:
Indeed, yes.  I guess the question is, now, how do I get this merged?

nullaus:
Hooray for cleaning out old crap we don't need!

PR #32882, opened 2015-10-30
Bug 1171848 - put scrollbar at the right position for textarea

autra:
If you put a dir="auto" on an element, its direction is viewed as LTR
when it is empty, and the scrollbar is on the left. This is bug 1103011.
But here, we already have a unicode-bidi: -moz-plaintext on the element.
So we don't need any dir="auto" anyway.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5ed92005d6e1adfeb2bb262f1a7374feb643a2d2

PR #32903, opened 2015-10-30
Bug 1220089 - History collection checkbox enabled although refresh au…

ferjm:
…th didn't succeed. r=yifan

### DIFF
--- a/tv_apps/browser/js/sync/settings.js
+++ b/tv_apps/browser/js/sync/settings.js
@@ -258,13 +258,16 @@
           onlogout: () => {},
           onready: () => {},
           onerror: error => {
+            this.elements.collectionHistory.checked = false;
             console.error(error);
           }
         });
       } catch(e) {}
 
       navigator.mozId.request({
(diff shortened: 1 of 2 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=38f0591378e6de9324014f213c267060da3f6972

tv_apps/browser/js/sync/settings.js
```diff
@@ -258,13 +258,16 @@
           onlogout: () => {},
           onready: () => {},
           onerror: error => {
+            this.elements.collectionHistory.checked = false;
             console.error(error);
           }
         });
       } catch(e) {}
```
begeeben: Do we need to add a handler in the catch block as well?

ferjm: I added the catch because the mozId API is crappy and it throws when you call .watch more than once. Even if it throws, the API still fires the on* callbacks.

PR #32906, opened 2015-10-30
Bug 1219708 - Should not attempt to sync while offline. r=michielbdej…

ferjm:
…ong, yifan

### DIFF
--- a/tv_apps/smart-system/js/sync_manager.js
+++ b/tv_apps/smart-system/js/sync_manager.js
@@ -402,6 +402,11 @@
       // to do it for a periodic sync.
       this.unregisterSyncRequest();
 
+      if (!navigator.onLine) {
+        SyncStateMachine.success();
+        return;
+      }
+
       var collections = {};
       COLLECTIONS.forEach(name => {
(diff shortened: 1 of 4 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d83a0fc03f36e604ad54851f8b6ee55d53755900

tv_apps/smart-system/js/sync_manager.js
```diff
@@ -402,6 +402,11 @@
       // to do it for a periodic sync.
       this.unregisterSyncRequest();
 
+      if (!navigator.onLine) {
```
begeeben: Should we trigger an offline error here? Or leaving it to the ```Enabled``` state is intended?

ferjm: Yes, it is intended. These sync requests can be triggered automatically by the sync scheduler, so we shouldn't show any error to the user and simply go back to the normal `enabled` state. For manual sync requests we show an error when the server request times out.

PR #32907, opened 2015-10-30
Bug 1210201 - Work around the better event fluffing algorithm r=jrburke r=jlorenzo

julienw:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9db67218cb1850aed63c2f58f9424787519b1207

jrburke:
Neat, event.offsetY and use of .closest(), I have not gotten to use those much before. I think this is the right kind of idea, detect if the tap is really outside the area, and if so, treat it as an md__day tap.

These are the issues I see:

* Events are not uniform height, those using hourHeight seems dangerous.
* The use of closest the 'a' tag, not sure if it would capture more than what was needed. This is more about my lack of full comprehension of what can show up in that area, not fully knowing the layout there.

I did this sort of patch, but not sure yet if it is too specific, I am still low on context for the calendar code. The idea was to be both more specific on targeting what actions might cause fuzzing interference (existing event or add event -- the blue + element, it also seems to trigger the fuzzing issue) and to measure the size of those items as seen in the DOM, instead of relying on hourHeight, since an existing event may span more than one hour/be less than an hour.

It might be a bit cumbersome though, what about something along those lines?

```patch
diff --git a/apps/calendar/js/views/hour_double_tap.js b/apps/calendar/js/views/hour_double_tap.js
index 36cde71..8bb1038 100644
--- a/apps/calendar/js/views/hour_double_tap.js
+++ b/apps/calendar/js/views/hour_double_tap.js
@@ -42,8 +42,29 @@ HourDoubleTap.prototype = {
 
   _onDayTap: function(evt) {
     var target = evt.target;
+
     if (!target.classList.contains('md__day')) {
-      return;
+      // Could be aggressive event fuzzing registering a tap on an existing
+      // event or the "add event" UI, where the tap is really below that item.
+      // See bug 1210201.
+      if ((target.classList.contains('md__event') ||
+           target.classList.contains('md__add-event')) &&
+          // In this case, offsetY is the vertical space between the top of the
+          // element and the actual tap, so if the tap is bigger than the height
+          // of the element, treat it as an md__day tap.
+          evt.offsetY &&
+          evt.offsetY > target.getBoundingClientRect().height) {
+
+          target = target.closest('.md__day');
+
+          // Make sure the event tag's default action does not complete, and
+          // any other handlers see this as an event tap.
+          evt.preventDefault();
+          evt.stopPropagation();
+      } else {
+        // Not interesting, ignore it.
+        return;
+      }
     }
 
     var y = evt.clientY + this.main.scrollTop - this._mainOffset;
```

julienw:
You're right, I forgot an event can last more than 1 hour.

Also I noticed that we could have the button's title as event.target and in that case the classList check and likely the offsetY calculation would be wrong. Maybe we need "pointer-events: none" on the event's children. Or use a <button> instead of a <a> but this is a lot more work than I'm willing to do here :) (and does not integrate properly with the router)

I'll look into this more over the week-end or monday. Thanks a lot for your feedback already !

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=52388cd14a7a69e4299f61dc141cedc4f4c0f941

apps/calendar/js/views/hour_double_tap.js
```diff
@@ -40,12 +40,40 @@ HourDoubleTap.prototype = {
...
     var target = evt.target;
     if (!target.classList.contains('md__day')) {
+      // Could be aggressive event fuzzing registering a tap on an existing
+      // event or the "add event" UI, where the tap is really below that item.
+      // See bug 1210201.
+
+      // In this case, offsetY is the vertical space between the top of the
+      // element and the actual tap, so if the tap is bigger than the height
+      // of the element, treat it as an md__day tap.
+      var tappedOnExistingEvent =
```
jrburke: The last part of this comment can be removed: this function is just focused on if it is an event button.

apps/calendar/js/templates/multi_day.js
```diff
@@ -69,7 +69,7 @@ multiday.event = function() {
     </i>`;
   }
 
-  var className = ['md__event'];
+  var className = ['md__event', 'md__event-like-button'];
```
jrburke: Super small nit: unfortunately facebook has polluted my frame of reference for "like button", so I keep needing to do another pass on this name. I think you could get by without the "like" part of the name. But really up to you, not a big deal.

apps/calendar/js/views/hour_double_tap.js
```diff
@@ -40,12 +40,40 @@ HourDoubleTap.prototype = {
...
+  _onDayTap: function(evt) {
+    if (this._isEventRightOnButton(evt)) {
       return;
     }
 
+    var target = evt.target.closest('.md__day');
+    // Note that until Bug 1091889 gets fixed we'll still get the :active
+    // styles because they're being applied with `touchstart` and we can't
+    // defaultPrevent this event without breaking everything.
+    evt.preventDefault();
```
jrburke: I can see placing the preventDefault as an else for the `if (this._isEventRightOnButton(evt))` block, since it is tied specifically to why it is checked to be a fuzzed event. If the event fuzzing concern goes away, and that if test is no longer needed, it would be good if the preventDefault had a chance to be removed at that time too, indicated by grouping/proximity.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=17838787b900a6b7da41c977e3f328c8a19c318d

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f72a3f740af0f3c06b4acc2828c04e9ed3703f8a

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=bee11743772fd8fddbd61bcf38851a3aab3f6615

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6ffb1938f000eed2fe1085a04d34750de39e30f3

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=aa3933853304e1162a94ef6219670832a437d11b

apps/calendar/js/views/hour_double_tap.js
```diff
@@ -40,12 +40,50 @@ HourDoubleTap.prototype = {
...
+      // See bug 1210201.
+
+      // In this case, offsetY is the vertical space between the top of the
+      // target element and the actual tap, so if the tap is bigger than the
+      // height of the element, or smaller than 0, the tap really happened out
+      // of the element.
+      var tappedOnExistingEvent =
+        target.classList.contains('md__add-event') ||
+        target.classList.contains('md__event') &&
+        evt.offsetY >= 0 &&
```
jrburke: The lower precendence of || compared to && leads to this ending as true for just md__add-event, which results in the fuzzing before the + button registering as a press on the + when it should not. Fixed with () around the || part: `(target.classList.contains('md__add-event') || target.classList.contains('md__event')) && ...`.

julienw: actually I made this on purpose as I explained in https://bugzilla.mozilla.org/show_bug.cgi?id=1210201#c46 :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d751adabf66823dab135aa39852cce018a3cb157

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=115e35f6993759e22b010d42e322ed577ea47f62

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=493d40fe6ea42b32b80e8c213722bc0b74165bb6

apps/calendar/test/marionette/lib/views/week.js
```diff
@@ -77,6 +83,153 @@ Week.prototype = {
     });
   },
 
+  /**
+   * This returns a boolean that is true if the hour square is fully visible.
+   *
+   * @param {Number} hour The hour to locate
+   * @returns {Boolean} Whether the square is fully visible.
+   */
+  _isHourDisplayed: function(hour) {
```
JohanLorenzo: IMO, tests shouldn't require documentation. The main reasons are:
* These tests are not part of a public API. The people who will read them, will only do it though the raw code, not an external website.
* The comments often can be replaced by more explicit names. This will prevent scrolling around the code to understand what the test is supposed to do.

Here, I think we can name this function: `_isHourSquareFullyDisplayed`. What do you think?

apps/calendar/test/marionette/lib/views/week.js
```diff
@@ -77,6 +83,153 @@ Week.prototype = {
...
+
+  /**
+   * This method tries to find out where the hour square is located in the
+   * window.
+   *
+   * @param {Number} hour The hour to locate.
+   * @returns {Number} 0 if it's displayed, or the delta otherwise: negative
+   * if it's above or positive if it's below.
+   */
+  _cmpHourDisplayed: function(hour) {
```
JohanLorenzo: How about something like `_getPixelsForHourToBeFullyDisplayed`?

apps/calendar/test/marionette/lib/views/week.js
```diff
@@ -77,6 +83,153 @@ Week.prototype = {
...
+  },
+
+  /**
+   * This method scrolls slowly to the right hour.
+   *
+   * @param {Number} hour The hour to scroll to.
+   */
+  scrollToHour: function(hour) {
+    this.waitForNoScroll();
+    if (this._isHourDisplayed(hour)) {
```
JohanLorenzo: I wonder if we can can get rid of this check, as it's already in the wait.

apps/calendar/test/marionette/lib/views/week.js
```diff
@@ -77,6 +83,153 @@ Week.prototype = {
...
+      // element is hidden below the visible part.
+      // we need to move it down
+      if (where > 0) {
+        this.actions.flick(
+          this.main,
+          100, moveDelta, 100, 0
+        ).perform();
+      }
+
+      this.waitForNoScroll();
```
JohanLorenzo: If we removed the check before the wait, that line could go at the beginning of the wait.

apps/calendar/test/marionette/lib/views/week.js
```diff
@@ -77,6 +83,153 @@ Week.prototype = {
...
+      var where = this._cmpHourDisplayed(hour);
+      if (where === 0) {
+        return true;
+      }
+
+      var moveDelta = 70;
+
+      // element is hidden above the visible part
+      // we need to move it up
+      if (where < 0) {
```
JohanLorenzo: If I understand _cmpHourDisplayed correctly, `where` contains a value that tells you roughly how far is the hour square. Maybe we could use it, to make this function more dynamic? Something like:
```js
this.waitForNoScroll();

var delta = this._getPixelsForHourToBeFullyDisplayed();
var beginPosition;
var endPosition;
if (delta === 0) {
    return true;
} else if (delta < 0) {
    beginPosition = 0;
    endPosition = -delta
} else {
    beginPosition = delta
    endPosition = 0
}

var middleOfMain = this.main.rect().width / 2

this.actions.flick(this.main, middleOfMain, beginPosition, middleOfMain, endPosition).perform();
return false;
```

apps/calendar/test/marionette/lib/views/week.js
```diff
@@ -77,6 +83,153 @@ Week.prototype = {
...
+  tapDayHour: function(options) {
+    var day = options.day;
+    var hour = options.hour;
+    var left = options.left;
+    var right = options.right;
+    var bottom = options.bottom;
+    var top = options.top;
+
+    if (left === undefined && right === undefined) {
+      left = 25;
```
JohanLorenzo: We can make these values more dynamic, just to be safe, if one day the form factor changes.
```js
var dayElt = this.days[day];
var dayEltWidth = dayElt.rect().width
var dayEltHeight = dayElt.rect().height

if (left === undefined && right === undefined) {
    left = dayEltWidth / 2;
}
if (left === undefined) {
    left = dayEltWidth - right;
}

// and so on
```
How do you feel about it?

apps/calendar/test/marionette/week_view_test.js
```diff
@@ -19,6 +19,13 @@ marionette('week view', function() {
     }
   });
 
+
+  function isShowingAddEventLink() {
```
JohanLorenzo: This function should belong in the week view.

apps/calendar/test/marionette/lib/views/week.js
julienw: I did it in the past (that's why "where" holds this value actually :) ) but in the end I found it less reliable. We'd need a more complex formula. For example if the delta is '1' you don't want to use that but use a min value. And I also used a max value in addition. But even this proved to be unreliable.

So I picked a magic value that seems to scroll by some less than 1 full page and it seems reliable enough :)

apps/calendar/test/marionette/week_view_test.js
```diff
@@ -313,12 +320,31 @@ marionette('week view', function() {
...
-      return week.element.scriptWith(function(el) {
-        return !!el.querySelector('.md__add-event');
-      });
-    }
+  suite('event retargeting tests', function() {
+    test('When tapping around an event we can create another event',
+    function() {
+      week.waitForHourScrollEnd();
+
+      week.waitForNoScroll();
```
JohanLorenzo: Waiting twice for the scroll to be ended looks odd. I wonder if we could change `_waitForScrollEnd()` by `waitForNoScroll()`. I'll try that

apps/calendar/test/marionette/lib/views/week.js
JohanLorenzo: You're right. I just tried that code, and I saw flicking going way above.

apps/calendar/test/marionette/week_view_test.js
```diff
@@ -313,12 +320,31 @@ marionette('week view', function() {
...
+      week.scrollToHour(6);
+      week.tapDayHour({ day: 7, hour: 6 });
+
+      assert.ok(isShowingAddEventLink(), 'The `add-event` link is displayed');
+
+      // The "+" button should work with event retargeting.
+      week.tapDayHour({ day: 7, hour: 6, bottom: -1 });
```
JohanLorenzo: I'd extract that value in a variable. It makes easier to test the scroll function :smiley:

apps/calendar/test/marionette/lib/views/week.js
JohanLorenzo: Following up from our IRL discussion:
```js
  scrollToHour: function(hour) {
    this.client.waitFor(function() {
      this.waitForNoScroll();

      var delta = this._cmpHourDisplayed(hour);
      var mainHeight = this.main.rect().height;

      var beginPosition, endPosition = mainHeight / 2;
      var sizeOfOneElement = mainHeight / 6;

      if (delta === 0) {
          return true;
      } else if (delta < 0) {
          endPosition += sizeOfOneElement;
      } else {
          beginPosition -= sizeOfOneElement;
      }

      var middleOfMain = this.main.rect().width / 2 // I wish flick() was able to get the middle all by itself

      this.actions.flick(this.main, middleOfMain, beginPosition, middleOfMain, endPosition).perform();
      return false;
    }.bind(this));
  },
```

apps/calendar/test/marionette/week_view_test.js
julienw: I removed "waitForNoScroll" because we do it in scrollToHour, and added a comment explaining why we need ot use waitForHourScrollEnd.

apps/calendar/test/marionette/week_view_test.js
julienw: moved

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4ae2f52ed448679bd0f028719ae90fc30b72269e

apps/calendar/test/marionette/lib/views/week.js
```diff
@@ -77,6 +81,153 @@ Week.prototype = {
...
+      if (prevScrollTop !== null && prevScrollTop === currentScrollTop) {
+        return true;
+      }
+      prevScrollTop = currentScrollTop;
+      return false;
+    }.bind(this));
+  },
+
+  /**
+   * This method scrolls slowly to the right hour.
```
JohanLorenzo: IMO, this jsdoc doesn't provide any value compared to the function signature

apps/calendar/test/marionette/lib/views/week.js
```diff
@@ -77,6 +81,153 @@ Week.prototype = {
...
+    return 0;
+  },
+
+  /**
+   * This method waits until we don't scroll.
+   *
+   * Note this doesn't seem to work properly when the view is just displayed.
+   * That's why waitForHourScrollEnd is still useful.
+   */
+  waitForNoScroll: function() {
```
JohanLorenzo: This function can now be private.

apps/calendar/test/marionette/lib/views/week.js
julienw: "slowly" :p OK I'll add some more value to this comment.

apps/calendar/test/marionette/lib/views/week.js
julienw: I feel it could be used in a public way though. But I'd prefer to see something like this, and working in all conditions, in a generic helper.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0d4437d31e8dd8c81d4617d875c7337d858097a4

PR #32908, opened 2015-10-30
Bug 1219853 - [email] Go back to mozAlarms instead of request-sync

jrburke:
This is like a reversion of [bug 1098289](https://bugzilla.mozilla.org/show_bug.cgi?id=1098289), but some things have moved around a bit.

The other major difference: email still listens for request-sync entry points, to catch an email client that was using them then upgrades to this one. In that case, the request-sync registrations are removed and the app just starts up in the background going to the message list. During our normal startup process, that triggers the ensureSync alarm checking and we set the alarms correctly for the next sync.

End result: we miss one sync in the upgrade case, but be set with alarms for the next sync.

An alternative might be to just drop the webapp manifest stuff around request-sync and hope that does not cause a lingering issue with request-sync registrations. Or trace the code to try it, but I do not trust just code reading for this, I would want to test it. It is difficult to test since I am not sure of a good way of updating the webapp manifest permissions stuff on a push, just know how to do a gaia reset to get it the new info updated with b2g.

Plus, by actually waking up with the request-sync message, we get a chance to seamlessly upgrade to the alarms without the user needing to manually open the app first. Not sure if that is worth the cost of keeping request-sync in the manifest though. So that part might need more discussion.

### DIFF
--- a/apps/email/js/html_cache_restore.js
+++ b/apps/email/js/html_cache_restore.js
@@ -36,11 +36,11 @@ window.startupOnModelLoaded = null;
  * exists to help us be efficient on closing the app in the case of notification
  * close events. If the notification system improved so that the app could tell
  * it not to notify us on close events, this could be removed. It is a global
- * so that cronsync-main can set it for request-sync operations. This file does
- * not track request-sync setMozMessageHandler messages directly, since they
- * need to acquire wake locks in the event turn the message is received. Since
- * that is a bit complicated, the back-end handles it, since it also knows more
- * about the sync details.
+ * so that cronsync-main can set it for alarm operations. This file does not
+ * track alarm setMozMessageHandler messages directly, since they need to
+ * acquire wake locks in the event turn the message is received. Since that is a
+ * bit complicated, the back-end handles it, since it also knows more about the
+ * sync details.
  */
 window.appDispatchedMessage = false;
 
@@ -51,7 +51,7 @@ window.appDispatchedMessage = false;
   // for the pending message type is received, this variable is cleared,
   // opening the way for other mozSetMessageHandler messages to be processed.
   // This is only important to track for messages that have UI impact, like
-  // inserting cards. Data messages, like request-sync are fine to pass through
+  // inserting cards. Data messages, like alarm are fine to pass through
   // unblocked.
   var pendingUiMessageType = null;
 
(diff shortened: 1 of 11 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=97151f53c9b2856dfefa3aff50378d860406d6d2

asutherland:
Based on my understanding of the requestsync APIs and the app upgrade flow, I think requestsync would *only* remove our existing syncs if our app was fully uninstalled.  (Like where IndexedDB gets nuked too.)

So I think it's absolutely the right call to do what you've done, and kudos for doing the right thing from the first pull request.

apps/email/js/html_cache_restore.js
```diff
@@ -443,12 +443,29 @@ window.appDispatchedMessage = false;
...
+      // Legacy request-sync trigger. Just clear all request-sync tasks, and
+      // allow the UI to go to a normal view this time. When back end starts up
+      // it will ensure the sync alarms are set properly for the next sync, so
+      // in this legacy case we will just miss one sync pass.
+      var navSync = navigator.sync;
+      if (navSync) {
+        navSync.registrations().then(function(regs) {
+          regs.forEach(function(reg) {
+console.log('*****DEBUG: unregistering request sync!');
+            reg.unregister(reg.task);
```
asutherland: this thred an error for me.  based on the code above, I'm assuming this wants to be navSync.unregister(), not reg.unregister().  I'm going to try this change locally and re-migrate.

jrburke: ugh i think you are right, updating now. looking back at my log, i only saw the unregistering message. i will do that retest too.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3e4f2828cb3617e274c05cbee8ddb6ce7ebd9c05

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d871bb2069f2b4af544497ac14becea7b304df3c

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6b61172dc0da64042572e03e0d95698bf022800d

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a3725ad86d9ac6a88109c45b89a043477fa49aa6

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7869bdda7af90f53498944599a724f55d1e80ceb

PR #32912, opened 2015-10-30
Bug 1216745 - [Flame][Music] When the Music app is loading new files …

jimporter:
…the UI covers up the "Back" button on the player preventing users from changing away from the "now playing" page

### DIFF
--- a/apps/music/js/app.js
+++ b/apps/music/js/app.js
@@ -100,14 +100,14 @@ client.on('databaseReady', () => {
 // If you add a Promise to either of these handlers, be careful!
 
 client.on('scanProgress', (detail) => {
-  scanProgress.show({
+  scanProgress.update({
     value:      detail.count,
     heading:    detail.artist,
     subheading: detail.title
   });
 });
 
-client.on('scanStopped', () => scanProgress.hide());
+client.on('scanStopped', () => scanProgress.clear());
 
 client.connect();
 
@@ -258,6 +258,7 @@ function setHeaderTitle(title) {
 
 function setBackButtonHidden(hidden) {
   header.els.actionButton.style.visibility = hidden ? 'hidden' : 'visible';
+  scanProgress.hidden = !hidden;
 }
 
 function getTabByViewURL(url) {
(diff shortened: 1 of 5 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6dc108bffe413e844b46a2699dc828c3b2fb04a1

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=65ffec9b0d9b6ae708bb54bf2e3ebe091d1fdf71

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=eb33676242c5634a13eee9e4cfef1cb709328150

apps/music/js/app.js
```diff
@@ -258,6 +258,7 @@ function setHeaderTitle(title) {
 
 function setBackButtonHidden(hidden) {
   header.els.actionButton.style.visibility = hidden ? 'hidden' : 'visible';
+  scanProgress.hidden = !hidden;
```
justindarc: Couldn't this inadvertently end up showing `scanProgress` when there's no progress to show?

jimporter: No, because there are two levels of visibility: if `.hidden` is true, it's always hidden. If `.hidden` is false, then we defer to `<music-scan-progress>`'s implementation to determine when to show; if you've called `.clear()`, the element will stay hidden (because the *child* of `<music-scan-progress>` is set to hidden).

I chose to do it this way so that the `.hidden` attribute on the main element is only set by users of the element, since that keeps the logic simple. The internal state that gets determined by the current scan progress is all handled privately (in the shadow DOM).

justindarc: Ahh ok. Thanks for clarifying that. Makes sense.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4fb805eca49ef24932eb1ddedba1cf9847f4d1cf

PR #32915, opened 2015-10-31
Bug 1220354 - [FTU] Remove collections migration

KevinGrandon:
https://bugzilla.mozilla.org/show_bug.cgi?id=1220354

### DIFF
--- a/apps/ftu/test/unit/app_test.js
+++ b/apps/ftu/test/unit/app_test.js
@@ -68,11 +68,11 @@ suite('AppManager >', function() {
     MockNavigatormozApps.mTeardown();
   });
 
-  test('IAC Message > will send message', function() {
+  test('publishes setup event', function(done) {
+    window.addEventListener('ftu-setup', function() {
+      done();
+    });
     AppManager.init();
-    MockNavigatormozApps.mTriggerLastRequestSuccess();
-    assert.equal(MockNavigatormozApps.mLastConnectionKeyword,
-                 'setup');
   });
 
 });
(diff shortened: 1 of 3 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=33208ddaebdcbfdf78014934d351bd982f936ace

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=07b7225fd6d6f5c14644bedfb1c1cb67b26623f8

apps/ftu/test/unit/app_test.js
```diff
@@ -1,78 +0,0 @@
-/* global MocksHelper, MockNavigatorMozMobileConnections, MockL10n,
```
sfoster: its a bit crazy that the only test for app.js is a collection one. But I can put this back easily enough with new coverage so I guess removing it for now is the right thing to do for this patch

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=fb2d7da59ff2d465ca84289153d9e27fffb27055

apps/ftu/test/unit/app_test.js
KevinGrandon: Yeah - it feels like it may be more useful to actually keep this around, so I've refactored the test to just perform a very basic sanity check of AppManager.init - just checking that we trigger an event. Feel free to remove this file if you want in a follow-up though.

PR #32929, opened 2015-11-02
Bug 1200090 - Fix remain shifting for Thai keyboard R=veer66

veer66:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f6cd9be4cbf6c221476962877a7b5f0dcc854fa0

apps/keyboard/build/layout_configurator.js
```diff
@@ -151,6 +151,10 @@ KeyboardLayoutDetail.prototype.load = function(appDir) {
 
       break;
 
+    case 'thai':
```
timdream: You should put this at line 137 before `case 'vietnamese':` instead of of creating another empty `case ... break`.

apps/keyboard/js/layouts/th.js
```diff
@@ -1,6 +1,7 @@
 Keyboards.th = {
   label: 'Thai',
   shortLabel: 'Th',
+	imEngine: 'thai',
```
timdream: Please fix indent here.

apps/keyboard/js/imes/thai/thai.js
```diff
@@ -0,0 +1,22 @@
...
+	var keyboard = null;
+	
+	InputMethods.thai = {
+		init: function(interfaceObject) {
+      keyboard = interfaceObject;
+		},
+
+		click: function(keyCode, isRepeat) {
+			keyboard.sendKey(keyCode, isRepeat);
+			keyboard.setUpperCase({isUpperCase: false});
```
timdream: The call is cheap so noop doesn't really cause any problem I think...

veer66: Excuse me. I don't understand this comment. Should I put //noop?

KaiRo-at:
Note that this has a wrong bug number on it, so autolander commented on an ancient unrelated bug.

timdream:
Thanks :KaiRo, Please also amend the commit message with a right bug number in that case :)

PR #32930, opened 2015-11-02
Bug 1220563 - [System] correct alignment for toaster notifications.

autra:
We don't want the alignment to depend on the dir parameter of the
notification, but rather on the system UI direction.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f09024831c810573970e6a4f509cf93ab6bcb012

PR #32949, opened 2015-11-02
Bug 1217859 - [Messages] It is not possible to navigate to any conversation or create new message until Inbox is fully loaded. Wait for the lazy loading complete instead of document onload. r=julien

azasypkin:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=76edf6f1a89ecd63ce5d065355e73fb0cf3c6648

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0cdfb05eb212eb4406bdfd95b1bcd0739ebba570

apps/sms/views/shared/js/navigation.js
```diff
@@ -491,28 +491,21 @@ function executeNavigationStep(stepName) {
...
+ * @returns {Promise.<void>}
  */
-function attachAfterEnterHandler() {
-  if (document.readyState === 'complete') {
-    return executeNavigationStep('afterEnter').catch(catchStepError);
-  }
-
-  var defer = Utils.Promise.defer();
+function waitForThePanelToBeReady() {
+  var readyPromise = readyDefer && readyDefer.promise || Promise.resolve();
```
azasypkin: Honestly I'd rather remove ```'navigation-transition-end'``` support at all and just use ```readyDefer && readyDefer.promise``` right inside ```init()``` method, but maybe you're ok with that a bit complex method :)

azasypkin: ```readyDefer && readyDefer.promise``` - I see there is a theoretical chance that ```readyDefer``` is nullified before we call this method.

``` || Promise.resolve()``` just for the consistency of returned result.

File: apps/sms/views/shared/js/utils.js [hunk 2]
```diff
@@ -945,24 +945,33 @@
...
+     * @param {EventTarget} eventTarget Event target to listen event on.
+     * @param {String} eventName Name of the event to listen for.
+     * @returns {Promise.<*>} Promise resolved with event object if any.
      */
-    onceDocumentIsVisible() {
-      if (!document.hidden) {
-        return Promise.resolve();
-      }
-
+    onceEventIsFired(eventTarget, eventName) {
```
azasypkin: I'd add optional ```timeout``` support here as well, but we don't need it anywhere yet, so didn't want to bring complexity here. The only downside here is that we'll leak ```'navigation-transition-end'``` event handler with such approach, what do you think?

apps/sms/views/shared/js/navigation.js
```diff
@@ -491,28 +491,21 @@ function executeNavigationStep(stepName) {
...
-    defer.resolve(executeNavigationStep('afterEnter'));
+  if (isUsingOldArchitecture()) {
+    return readyPromise;
   }
 
-  window.addEventListener('navigation-transition-end', onNavigationEnd);
-  window.addEventListener('load', onNavigationEnd); // simulate navigation end
-
-  return defer.promise.catch(catchStepError);
+  return Promise.race([
```
azasypkin: Moreover I'm not really sure if this ```Promise.race``` is the right thing here, I'd rather have something like ```return Utils.onceEventIsFired(window, 'navigation-transition-end', 1500 /* timeout */).```

apps/sms/views/shared/js/utils.js
```diff
@@ -945,24 +945,33 @@
...
     /**
-     * Returns Promise instance that is resolved instantly if document is
-     * visible (document.hidden === false), othewise it's resolved only once
-     * `visibilitychange` event is fired.
-     * @returns {Promise.<void>}
+     * Returns Promise instance that is resolved only once specified event is
+     * fired on the specified event target.
+     * @param {EventTarget} eventTarget Event target to listen event on.
+     * @param {String} eventName Name of the event to listen for.
+     * @returns {Promise.<*>} Promise resolved with event object if any.
```
julienw: nit: I think there is always an event as we use the DOM API ?

apps/sms/views/shared/js/utils.js
julienw: I'd say that even with a timeout you could remove the event listener, if you define the function before calling addEventListener.

But I'm not convinced we need a timeout yet.

azasypkin: Mmm didn't get :) I meant timeout for the case when event isn't fired at all, like if we missed it....

julienw: yeah I got you, but still I don't understand why you couldn't remove the listener ?

azasypkin: Mmm but how Onceeventisfired can understand that that it should remove handler if event isn't fired?

apps/sms/views/shared/js/navigation.js
```diff
@@ -491,28 +491,21 @@ function executeNavigationStep(stepName) {
...
-
-  var defer = Utils.Promise.defer();
+function waitForThePanelToBeReady() {
+  var readyPromise = readyDefer && readyDefer.promise || Promise.resolve();
 
-  function onNavigationEnd() {
-    window.removeEventListener('navigation-transition-end', onNavigationEnd);
-    window.removeEventListener('load', onNavigationEnd);
-    defer.resolve(executeNavigationStep('afterEnter'));
+  if (isUsingOldArchitecture()) {
```
azasypkin: Let's move discussion from bugzilla to here, @julienw :)

I see we wait for the ```waitForReady``` in the ```startNavigation``` but should not it be here instead? It looks like it's needed only when we trying to issue navigation request before app is initialized and ```Navigation.init()``` looks like a good single entry point to resolve this.

Or I'm missing something?

apps/sms/views/shared/js/utils.js
azasypkin: Yeah, it's supposed to work only with [EventTarget](https://developer.mozilla.org/en/docs/Web/API/EventTarget) compatible WebAPI - will replace ```*``` with ```Event```, thanks!

apps/sms/views/shared/js/utils.js
azasypkin: I mean case like this:
```
console.log(document.readyState === 'complete') // 'true'
Utils.onceEventIsFired(window, 'load');
```

In this case ```load``` event will never fire and we'll have event handler forever. So that's the main reason I don't feel 100% strong about this method, either in its description should be said that user will have to make sure that this event will fire eventually. So that it should always be:
```
if (document.readyState !== 'complete') {
  Utils.onceEventIsFired(window, 'load');
}
```

But we don't have ```document.readyState```-like api for the navigation transition events and it's not clear if it's implemented at all :)

apps/sms/views/shared/js/navigation.js
julienw: I think it's better in startNavigation because it's also useful when calling toPanel (from ActivityHandler for example). It's useful for both toPanel and init...

azasypkin: Mm, but you can't/shouldn't call ```toPanel``` before ```init()```? And if so, then we need to wait for this promise only once. All toPanel requests that are made after ```init``` will automatically wait until ```init``` is finished (since ```endNavigation``` will be called only once ```readyDefer``` is resolved). Or there is any case I don't see?

apps/sms/views/shared/js/utils.js
julienw: Actually I'd have integrated the timeout inside "onceEventIsFired" :)

but anyway I agree with you, let's remove the support for this non-existant event. We can ressurect it later when it comes to life.

But I'm not so sure about replacing the "load" with our internal ready (in split views). The semantic is that "afterEnter" should happen after the transition happened, after the window is displayed. But maybe it works after all...

apps/sms/views/shared/js/navigation.js
julienw: But we early reject for some cases in "init" :) and in `startNavigationFromCurrentLocation` too.

I'm quite sure we can call "toPanel" without `init` being finshed (using the general case). Actually Navigation's `init` does a lot more things than the other "init" functions, maybe it's badly named.

IMO we don't want to wait that everything is ready before moving to the first panel, so we won't want to wait for ready... unless the requested view is not loaded yet (and in the future the lazy loading should happen in startNavigation, instead of waiting for ready).

azasypkin: > But we early reject for some cases in "init" :) and in startNavigationFromCurrentLocation too.

Yep, that what I was missing, thanks! :)

> I'm quite sure we can call "toPanel" without init being finshed (using the general case). Actually Navigation's init does a lot more things than the other "init" functions, maybe it's badly named.

Yep, and if init is not rejected, toPanel will wait until navigationTransition promise started by init is finished.

apps/sms/views/shared/js/utils.js
azasypkin: > Actually I'd have integrated the timeout inside "onceEventIsFired" :)

Yep, that what I was talking about in my first comment ;)

> but anyway I agree with you, let's remove the support for this non-existant event. We can ressurect it later when it comes to life.

Okay

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f5afc4f24765d0243e3b13b8b51a0c1b94c46bd0

apps/sms/test/marionette/inbox_test.js
```diff
@@ -0,0 +1,79 @@
+/* global require, marionette, setup, suite, test, __dirname */
```
azasypkin: This test doesn't really test the issue we're fixing as it's not possible currently due to marionette apps plugin limitation (app.launch waits for the "load" event), so it just tests that we can enter new message view while rendering is still in progress. Let me know if you prefer to not include this test in this patch :)

apps/sms/views/shared/test/unit/navigation_test.js
```diff
@@ -657,6 +657,8 @@ suite('navigation >', function() {
 
       window.Navigation = NavigationFactory(fakeWindow);
       navigator.mozHasPendingMessage.withArgs('notification').returns(false);
+
+      Navigation.setReady();
```
azasypkin: Please, tell me if you think we need any other dedicated test for this change, I can't really think of any reasonable test :/

julienw: What we need is to have a test that's failing before your patch and working with your patch.

so it should be possible to fake `document.readyState` to `interactive`, call toPanel or init, and do a `.then(done, done)` at the end (without sending "load" event but calling setReady). This should fail on master and work with your patch.

What do you think ?

azasypkin: > What we need is to have a test that's failing before your patch and working with your patch. so it should be possible to fake document.readyState to interactive, call toPanel or init, and do a .then(done, done) at the end (without sending "load" event but calling setReady). This should fail on master and work with your patch.

> What do you think ?

The only thing I'm concerned about is that such test may by a little bit confusing when we read it in the future,  since I've removed code that was relying on ```document.readyState``` or ```load``` event so we're going to mock something that doesn't have any effect :) But if you think it's fine, I can do it!

julienw: yes I think it should be OK. If you label it correctly, maybe with a comment to this bug, it won't be confusing. Remember unit tests is a tool to prevent old bugs from coming back

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=81b6f51b5e786e2e1483d247f77232738be50959

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b7997f2937a0cb143bd7b37e8dda240eacee9bef

PR #32955, opened 2015-11-02
Bug 942837 - [User Story] Show Alarm Time on Lock Screen R=gweng

prateekjadhwani:
User Story:

As a user I want to view the time of my next alarm while my phone is locked - this enables me to verify that it is set without needing to unlock my phone and navigate to the clock app.

Acceptance Criteria:

When my phone is locked (but not in sleep mode) I can see that my alarm is set, and the time and date of my next alarm.
Only my next alarm is shown, for example if I have an alarm set for 8am every day I can only see tomorrows alarm.
If my alarm is removed, It is no longer shown on the lock screen.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f4d2ea8e11318b517e8eaa1555b86c5f99821c00

apps/system/lockscreen/js/widgets/clock/lockscreen_clock_widget.js
```diff
@@ -50,6 +52,56 @@
...
+        type = 'PM';
+        hour = hour - 12;
+      }
+
+      var fullAlarmTime = hour + ':' + alarmData.minute + type;
+      this.resources.elements.alarmtime.textContent = fullAlarmTime;
+      this.resources.elements.alarm.classList.remove('no-alarms');
+      this.resources.elements.alarm.classList.add('has-alarm');
+      this.logger.debug('Alarm updated to :', fullAlarmTime);
+    }).catch(() => {
```
snowmantw: Don't use `catch` as control flow... As I mentioned in the previous commits.
It's really improper because:

1. You will ignore the real error, since you ignore the argument here.
2. Even you give it a check, and throw the real exception, it's still confused since you fork the semantic meaning of an exception to two types: one is your **expected** result and thus it's **obvious not an exception**, and another is the **real exception**.
3. What I've suggested is to pass an empty or specialized object for the **no alarm data data**, which is what you're checking here. You can do this in the `fetchAlarmData` method as I comment below.

apps/system/lockscreen/js/widgets/clock/lockscreen_clock_widget.js
```diff
@@ -50,6 +52,56 @@
...
+    return new Promise((resolve, reject) => {
+      navigator.getDataStores('alarms')
+      .then((stores) => {
+        return [stores[0].getLength(), stores[0]];
+      }).then(([len, store]) => {
+        if(0 === len) {
+          return;
+        }
+        return store.get(1);
+      }).then((fetchedData) => {
```
snowmantw: Here, if you fetch no data, you can design this `then` callback as:

    .then((fetchedData = null) => {..... resolve(fetchedData)})

So you can check it in line 80, rather than to throw exception and catch it.

snowmantw: And after that, you don't need the out most `Promise` and call its `resolve` and `reject` to represent the result. So you can just design the function as

    LockScreenClockWidget.prototype.fetchAlarmData = function() {
        return navigator.getDataStores('alarms').then(...)
    };

Since the `then` chain here, is already a `Promise`, and you don't need to wrap it again unless you have encountered some unsolvable issues.

apps/system/lockscreen/js/widgets/clock/lockscreen_clock_widget.js
snowmantw: And I suggest you that to add a `console.error` here (before you re-throw the error). To log the real error is a good practise for using `Promise`. Since if you neither do that nor re-throw the error, you'll **silent** the error. This is a real pain point during debugging `Promise`.

(That is, after you change the usage of `catch`).

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f39bc0faefd48ff5ae1a19d7d4808a4f51a7e0e4

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0b942f76e5374cf445edd004db8b87373cdbf691

apps/system/lockscreen/js/widgets/clock/lockscreen_clock_widget.js
```diff
@@ -50,6 +52,56 @@
...
+    return navigator.getDataStores('alarms')
+      .then((stores) => {
+        return [stores[0].getLength(), stores[0]];
+      }).then(([len, store]) => {
+        if(0 === len) {
+          return;
+        }
+        return store.get(1);
+      }).catch(function(e){
+        console.error('Error occured while getting alarm data : ', e);
```
snowmantw: You better re-throw the error after log it, since in this way the following promises concates it could be interrupted by the error, too. And, if you're addressing general errors in the whole chain, you don't need to catch it here, but put the catch at the end of the chain. That is, rather than:

    return getDataStores().then().then().catch().then();

You can write it as:

    return  return getDataStores().then().then().then().catch();

Since to put a `catch` in the middle of the chain will make you miss any error after that. So if you're not intentionally to do that, put it at the end is a convention.

apps/system/test/unit/lockscreen/widgets/clock/lockscreen_clock_widget_test.js
```diff
@@ -49,4 +49,50 @@ suite('LockScreenClockWidget > ', function() {
       mockThis.resources.elements.date.textContent,
       `it doesn't update the date.textContent with the date in locale format`);
   });
+
+  test(`it would update the alarm info (an alarm is set)`, function() {
```
snowmantw: For async test with Promise, you need to write it as:

    test(`....`, function(done) {
        promise....then(/* do something */)
        .then(() => {  assert.isTrue();  }    // assert in one then step
        .then(done)    // the `done` will receive nothing and the runner will know it ends without any error
        .catch(done)  // the `done` will receive an error object and let the test runner know there is an error
    })

snowmantw: Without cooperating `done` with `catch` and `assert`, the test will always pass, since after the promise kicks off, the stack is empty and the test is end: it never wait the test result in the async chain.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4fd15708b9e16ed545d7c9a2d545fb3b4fa6b124

apps/system/lockscreen/js/widgets/clock/lockscreen_clock_widget.js
```diff
@@ -50,6 +52,56 @@
...
+          return;
+        }
+        return store.get(1);
+      }).then((fetchedData = {}) => {
+        if(0 === Object.keys(fetchedData).length) {
+          return null;
+        }
+        return fetchedData;
+      }).catch(function(e){
+        console.error('Error occured while getting alarm data : ', e);
```
snowmantw: Please note what I've commented: you need to re-throw the error if you want the following chains after this method (e.g. `fetchAlarmData().then()`) can be interrupted as well. It's important in common cases since the following chains may do some effects or even unexpected effects while without a correct ascendants. Therefore if you thought it's not what you want, and you silent the error here **intentionally**, please comment it here. Or, please re-throw it after you log the error.

And why can't we just leave the error to the following one without dealing with it here, is because you never know who may concate your Promised methods after 3 months, although currently they're all only maintained and designed by you. So it's a defensive design: even others may screw the code by concating a bad Promise after this method, you can still see the error logs in the console, and you're sure if there are any error in your chain, the following Promises wouldn't get any chance to execute, unless they intentionally silent it by catching it (so the next reviewer should ask them the reason to do that like me).

prateekjadhwani: @snowmantw , do i need to catch the error on line 80 as well and again throw it after I add the throw code here?
Thanks

snowmantw: I suggest doing that (the same logging and throwing). In fact, in a more complicated patch, I will design an `onPromiseError` with logging and re-throwing to reuse it in every `catch`. But since your patch only provides two chains that may have errors, to duplicate some code is OK.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f9a207ffb92c511885a2a9b12615f7c8901b8859

apps/system/lockscreen/js/widgets/clock/lockscreen_clock_widget.js
prateekjadhwani: I see, thank you :)

apps/clock/manifest.webapp
```diff
@@ -15,6 +15,12 @@
     "audio-channel-alarm":{},
     "moz-extremely-unstable-and-will-change-webcomponents":{}
   },
+  "datastores-owned": {
+    "alarms": {
+      "access" : "readwrite",
```
mcav: can we make this readonly? It seems like Lockscreen doesn't need to be able to edit the data.

prateekjadhwani: @mcav so everything is good, right?

mcav: yeah

prateekjadhwani: @mcav, so can we have r+ on the review so the the QAs can start testing once the merge is done.

Thanks :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=94ce45dc36850f60f93fb5b49f60b5491f0d5d04

apps/system/lockscreen/js/widgets/clock/lockscreen_clock_widget.js
```diff
@@ -50,6 +52,60 @@
     this.logger.debug('Clock updated', now);
   };
 
+  LockScreenClockWidget.prototype.updateAlarm =
+	function() {
```
snowmantw: nit: this indentation looks a bit weird.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d85ab54e7ba1e16229b9a41bbdefb345531d8823

PR #32970, opened 2015-11-03
Bug 1220579 - Add homescreen unit tests for idb exceptions. r=gmarty

Cwiiis:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f76bf99fdd4d579a4bd6d10c80f753ba895d6caf

apps/homescreen/js/datastore.js
```diff
@@ -217,7 +218,6 @@
               cursor.continue();
             }
           };
-        txn.oncomplete = () => { resolve(results); };
```
gmarty: Just being curious, why the inversion here?

Cwiiis: Purely for the unit test to pass - otherwise getAll completes before oncomplete is set, so oncomplete is never called (in practice, setting afterwards doesn't matter because it's async, but we make it sync in the unit tests). I guess I really ought to have changed the unit test really, but it'd really complicate the changeset vs. this one small change :)

PR #32974, opened 2015-11-03
Bug 1220282 - do not force notification's inner direction.

autra:
Direction is now applied to content of notification, instead of being
applied to deal with alignment. Alignment is instead dealt properly in
CSS (see bug 1219618 for this work).

So there's no more reason to create a notification with the current
UI direction. We should either:
- let the application force the direction of the content if it wants to
  (in accordance with W3C specs)
- Let the default behavior (dir="auto") apply and localize properly in edge cases (for example the SIM1 position when dual-sim).

### DIFF
--- a/shared/js/notification_helper.js
+++ b/shared/js/notification_helper.js
@@ -35,7 +35,6 @@
             options.body = body;
           }
           options.lang = document.documentElement.getAttribute('lang');
-          options.dir = document.documentElement.getAttribute('dir');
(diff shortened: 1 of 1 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=129d4d185ffccd7f5a596813ebe5f59917591fab

PR #32975, opened 2015-11-03
Bug 1219250 - [NGA] Search field should match latest visual spec

wilsonpage:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f0c263cd701dd4094ccbb0a4465688bc80101b61

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9ea15957e4518f321a38230653a9151d22ac99ae

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=033e3b010ecb3fd7e0ff3da338ca0ced5cf71822

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1708f68f1630a28d7faf11a05f4f3a18173fb623

apps/music/css/view.css
```diff
@@ -15,65 +15,14 @@ body {
   -moz-user-select: none;
 }
 
-a {
-  text-decoration: none;
```
justindarc: Is it ok to remove this? Not sure if its used anywhere else. Tiles? etc.

apps/music/elements/music-search-box.js
```diff
@@ -4,82 +4,42 @@
 
 var proto = Object.create(HTMLElement.prototype);
 
-proto.HEIGHT = 37;
+const margin = 12;
+proto.HEIGHT = 40 + margin;
```
justindarc: Eh.. Just hard-code `52` in here and put a comment like `// 12px margin + 40px height`

apps/music/elements/music-search-box.js
```diff
@@ -4,82 +4,42 @@
...
-    content: '';
-    background-color: #c7c7c7;
-    position: absolute;
-    top: 0.7rem;
-    bottom: 0.7rem;
-    offset-inline-start: -0.1rem; /* left */
-    width: 0.1rem;
-  }
+
+#inner {
```
justindarc: Indent all this CSS by 2 spaces.

apps/music/elements/music-search-box.js
```diff
@@ -4,82 +4,42 @@
...
+
 </style>
-<form id="form" role="search">
-  <input type="search" id="input" x-inputmode="verbatim" data-l10n-id="search-music">
-  <button type="reset" id="clear" data-icon="search"></button>
-  <button type="button" id="close" data-l10n-id="search-close"></button>
-</form>`;
+<div id="inner">
+  <gaia-text-input type="search" clearable id="input"></gaia-text-input>
+  <button id="close">Close</button>
```
justindarc: l10n?

apps/music/elements/music-search-box.js
```diff
@@ -143,54 +90,66 @@ proto.attributeChangedCallback = function(attr, oldVal, newVal) {
...
     this.els.results.close();
   }
+
 };
 
-Object.defineProperty(proto, 'value', {
-  get: function() {
-    return this.els.input.value;
-  },
+proto.emit = function(name, detail) {
```
justindarc: This goes against the pattern in our other custom elements where we just use `dispatchEvent` directly.

apps/music/elements/music-search-box.js
```diff
@@ -88,50 +48,37 @@ proto.createdCallback = function() {
...
+  this.els.close.addEventListener('click', () => this.close());
   this.els.input.addEventListener('input', onSearch);
-  this.els.input.addEventListener('keypress', onSearch);
+  this.addEventListener('touchmove', e => e.preventDefault());
 
-  this.onDOMRetranslated = () => {
-    document.l10n.translateFragment(shadowRoot);
+  this.localize = () => {
+    document.l10n.formatValue('search')
+      .then(result => this.els.input.placeholder = result);
```
justindarc: It would be nice if `<gaia-text-input>` could set the placeholder as an attribute. If so, then we can just do `translateFragment()` like everywhere else. The l20n team preferred this over `formatValue()`.

apps/music/elements/music-search-box.js
```diff
@@ -143,54 +90,66 @@ proto.attributeChangedCallback = function(attr, oldVal, newVal) {
...
+proto.emit = function(name, detail) {
+  var evt = new CustomEvent(name, { detail: detail });
+  this.dispatchEvent(evt);
+};
 
-  set: function(value) {
-    this.els.input.value = value;
-  }
+Object.defineProperty(proto, 'value', {
+  get: function() { return this.els.input.value; },
```
justindarc: Please follow the same whitespace pattern and spacing here as everywhere else.

apps/music/elements/music-search-box.js
```diff
@@ -200,14 +159,4 @@ function debounce(fn, ms) {
   };
 }
 
-try {
```
justindarc: Keep this try/catch since we use it in all our custom elements. Will help catch exceptions in case WC is unsupported in an environment (e.g. running on desktop).

apps/music/elements/music-search-results.js
```diff
@@ -59,105 +58,99 @@ var template =
...
+  return this.els.list.setModel(results);
 };
 
 proto.clearResults = function() {
-  this.els.list.model = [];
+  return this.els.list.setModel([]);
+};
 
-  return this.els.list.model;
+proto.emit = function(name, detail) {
```
justindarc: This goes against the pattern in our other custom elements where we just use `dispatchEvent` directly.

apps/music/elements/music-search-results.js
```diff
@@ -59,105 +58,99 @@ var template =
...
+  var evt = new CustomEvent(name, { detail: detail });
+  this.dispatchEvent(evt);
 };
 
-try {
```
justindarc: Keep this try/catch since we use it in all our custom elements. Will help catch exceptions in case WC is unsupported in an environment (e.g. running on desktop).

apps/music/elements/music-search-results.js
```diff
@@ -59,105 +58,99 @@ var template =
...
+    </a>
   </template>
 </gaia-fast-list>`;
 
 proto.createdCallback = function() {
-  var shadowRoot = this.createShadowRoot();
-  shadowRoot.innerHTML = template;
-
-  var $id = shadowRoot.getElementById.bind(shadowRoot);
+  this.innerHTML = template;
```
justindarc: Why no shadow DOM?

apps/music/elements/music-search-results.js
```diff
@@ -59,105 +58,99 @@ var template =
...
   if (results.length === 0) {
-    document.l10n.formatValue('search-no-result').then((noResult) => {
-      this.els.list.model = [{
-        title: noResult,
-        subtitle: ''
-      }];
-    });
-
-    return;
+    return document.l10n.formatValue('search-no-result')
```
justindarc: nit: I prefer leaving this the way it was :-)

`setModel()` has no return value, but `formatValue()` does, so our `setResults()` should not be conditionally returning a value.

apps/music/js/view.js
```diff
@@ -93,11 +93,41 @@ function View() {
   });
 }
 
-View.prototype.once = function(el, name, callback) {
```
justindarc: Why are you removing this method?

apps/music/js/view.js
```diff
@@ -93,11 +93,41 @@ function View() {
...
+View.prototype.setupSearch = function(params) {
+  this.searchBox.addEventListener('search', (evt) => this.search(evt.detail));
+
+  this.searchResults.addEventListener('open', () => {
+    this.client.method('searchOpen');
+  });
+
+  this.searchResults.addEventListener('close', () => {
+    this.client.method('searchClose');
+    this.list.scrollTop = this.searchBox.HEIGHT;
```
justindarc: I like that you're moving all the search setup into the base class. Might as well move the `HEIGHT` const here too so this doesn't accidentally get set to `undefined`.

justindarc: Ehhhh.. Disregard this comment.

apps/music/views/album-detail/view.js
```diff
@@ -22,22 +22,20 @@ var AlbumDetailView = View.extend(function AlbumDetailView() {
 });
 
 AlbumDetailView.prototype.update = function() {
-  this.getAlbum().then((songs) => {
+  return this.getAlbum().then((songs) => {
```
justindarc: This doesn't need to be returning a value.

apps/music/views/albums/view.js
```diff
@@ -10,66 +10,28 @@ var AlbumsView = View.extend(function AlbumsView() {
...
 
-  this.configureList();
-  this.configureSearch();
+  this.setupSearch();
+  this.setupList();
+
   this.update();
 });
 
-AlbumsView.prototype.configureList = function() {
```
justindarc: Thanks for DRYing this up!

apps/music/views/albums/view.js
```diff
@@ -10,66 +10,28 @@ var AlbumsView = View.extend(function AlbumsView() {
...
-      this.client.method('navigate', link.getAttribute('href'));
-    }
-  });
-
-  this.searchResults.getItemImageSrc = (item) => this.getThumbnail(item.name);
-};
-
 AlbumsView.prototype.update = function() {
-  this.getAlbums().then((albums) => {
+  return this.getAlbums().then((albums) => {
```
justindarc: No return value.

apps/music/views/albums/view.js
```diff
@@ -10,66 +10,28 @@ var AlbumsView = View.extend(function AlbumsView() {
...
-
   View.prototype.destroy.call(this); // super(); // Always call *last*
 };
 
 AlbumsView.prototype.render = function() {
   View.prototype.render.call(this); // super();
-
-  this.list.model = this.albums;
-  this.list.cache();
+  return this.list.setModel(this.albums)
```
justindarc: Why return here?

apps/music/views/artist-detail/view.js
```diff
@@ -17,33 +17,28 @@ var ArtistDetailView = View.extend(function ArtistDetailView() {
   });
 
   this.list.configure({
-    getSectionName: (item) => {
-      return item.section;
-    }
+    getSectionName: (item) => item.section
```
justindarc: nit: When doing a one-param, one-line fat-arrow function, leave out the parenthesis -- `item => item.section`

apps/music/views/artist-detail/view.js
```diff
@@ -17,33 +17,28 @@ var ArtistDetailView = View.extend(function ArtistDetailView() {
...
   });
 
   this.client.on('databaseChange', () => this.update());
-
   this.update();
 });
 
 ArtistDetailView.prototype.update = function() {
-  this.getArtist().then((songs) => {
+  return this.getArtist().then((songs) => {
```
justindarc: No return value.

apps/music/views/artist-detail/view.js
```diff
@@ -17,33 +17,28 @@ var ArtistDetailView = View.extend(function ArtistDetailView() {
...
   this.client.destroy();
-
   View.prototype.destroy.call(this); // super(); // Always call *last*
 };
 
 ArtistDetailView.prototype.render = function() {
   View.prototype.render.call(this); // super();
-
-  this.list.model = this.songs;
+  return this.list.setModel(this.songs);
```
justindarc: No return value.

apps/music/views/home/view.js
```diff
@@ -63,18 +64,21 @@ HomeView.prototype.lazyLoadScripts = function() {
 HomeView.prototype.setupSearch = function() {
   this.searchBox = document.getElementById('search-box');
   this.searchResults = document.getElementById('search-results');
+  this.searchBoxHeight = this.searchBox.HEIGHT + 11;
```
justindarc: why + 11?

apps/music/views/home/view.js
```diff
@@ -156,7 +161,7 @@ HomeView.prototype.destroy = function() {
 HomeView.prototype.render = function() {
   View.prototype.render.call(this); // super();
 
-  document.l10n.formatValues(
+  return document.l10n.formatValues(
```
justindarc: No return value

apps/music/views/playlists/view.js
```diff
@@ -32,8 +32,8 @@ var PlaylistsView = View.extend(function PlaylistsView() {
 
   this.searchResults.getItemImageSrc = (item) => this.getThumbnail(item.name);
 
-  this.list.scrollTop = this.searchBox.HEIGHT;
   this.list.minScrollHeight = `calc(100% + ${this.searchBox.HEIGHT}px)`;
+  this.list.offset = this.searchBox.HEIGHT;
```
justindarc: You're not doing this anywhere else?

apps/music/views/playlists/view.js
```diff
@@ -49,34 +49,32 @@ var PlaylistsView = View.extend(function PlaylistsView() {
 
   this.client.on('databaseChange', () => this.update());
 
-  this.update();
+  this.update()
+    .then(() => this.list.scrollTo(this.searchBox.HEIGHT));
```
justindarc: I can kinda see why you might want to have a promise return from `update()` now. You can ignore my comments regarding return values. However, in this case, I would just put this `scrollTo()` call directly in the `update()` method.

apps/music/elements/music-search-box.js
wilsonpage: Ooops good catch

apps/music/elements/music-search-box.js
wilsonpage: I tried following the original technique, but `l10n.js` chokes on attributes for custom-elements.

justindarc: @stasm @zbraniecki Any insight here? ^^

stasm: Attributes are filtered according to https://github.com/mozilla-b2g/gaia/blob/a7f047368218e921949614ea803418c1ffbdc2cb/shared/js/intl/l20n-client.js#L24, which only allows `placeholder` on `input` elements.

Would it help if you could declare `data-l10n-attrs="+placeholder"` on a custom element to tell l20n that this attribute is fine to apply on it?  @zbraniecki: we talked about it in the past, do you think we could reconsider this idea now?

zbraniecki: I'm ok reconsidering it. I'm a bit worried about whitelist per element since it increases the noise in MutationObserver, but except of that, sounds good.

apps/music/css/view.css
wilsonpage: No visible regressions on home view, but added the property to `.tile` just in case :)

apps/music/elements/music-search-box.js
wilsonpage: I find the `dispatchEvent()` syntax pretty verbose, especially when detail with `detail`. This is more DRY IMO, especially as we're calling it in three separate places. But it's your call :)

apps/music/elements/music-search-box.js
wilsonpage: I can't see what value this adds. You're catching the error and then re-throwing with no additional messaging or logic. In the case of `NotSupportedError`, execution will still be interrupted and we'll see the same Error in the console with or without the `try catch`

apps/music/elements/music-search-results.js
wilsonpage: I mentioned this to you prior on IRC. I found in some cases Gecko was dispatching really broken `scroll` events. I spent a very long time digging into this and eventually found pulling the list out of shadow-dom fixed this :/

apps/music/elements/music-search-results.js
wilsonpage: `list.setModel()` now returns a `Promise` which resolves when rendering is complete. This means that this function will always return a `Promise` that resolves when rendering is done.

In other words: this function will return a `Promise` that resolves when the operation you have called is complete. Is this not desirable?

apps/music/js/view.js
wilsonpage: It's no longer used.

apps/music/views/albums/view.js
wilsonpage: IMO if functions are async, it's good convention to return a `Promise` to indicate when they're done.

apps/music/views/artist-detail/view.js
wilsonpage: Sorry thought I was following a Music convention :stuck_out_tongue:

apps/music/views/playlists/view.js
wilsonpage: Yeah it's being done in the `view.js` `.setupList()` abstraction. We can't use that in this view as there are some subtle differences.

apps/music/views/playlists/view.js
wilsonpage: I only want to `scrollTo()` when the view *first* renders, not every time the list content updates.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=266183e17475f7ab67638ac2ffb10b53d875dd76

apps/music/elements/music-search-box.js
```diff
@@ -142,32 +103,46 @@ proto.attributeChangedCallback = function(attr, oldVal, newVal) {
...
-  window.requestAnimationFrame(() => {
+proto.focus = function() {
+  requestAnimationFrame(() => {
     this.els.input.focus();
   });
+};
 
-  this.dispatchEvent(new CustomEvent('open'));
+proto.open = function() {
+  if (this.isOpen) return;
```
justindarc: nit:

```
if (this.isOpen) {
  return;
}
```

apps/music/elements/music-search-box.js
```diff
@@ -142,32 +103,46 @@ proto.attributeChangedCallback = function(attr, oldVal, newVal) {
...
     this.els.results.open();
   }
+
 };
 
 proto.close = function() {
-  this.els.form.reset();
-
-  this.dispatchEvent(new CustomEvent('close'));
+  if (!this.isOpen) return;
```
justindarc: same nit

apps/music/elements/music-search-box.js
justindarc: I can't remember where this originated now. It was something we found while working on Spark/Customizer.

apps/music/elements/music-search-results.js
```diff
@@ -5,157 +5,149 @@
...
-  #list p {
-    background: none !important;
-    padding-right: 60px;
-  }
-  #list img {
-    right: 0;
-    left: auto !important;
-    object-fit: cover;
-  }
+  this.innerHTML = `<style scoped>
```
justindarc: nit: insert line break between `=` and `

apps/music/elements/music-search-results.js
```diff
@@ -5,157 +5,149 @@
...
+      <p>\${subtitle}</p>
+    </a>
   </template>
 </gaia-fast-list>`;
 
-  var $id = shadowRoot.getElementById.bind(shadowRoot);
-
   this.els = {
-    list: $id('list')
+    list: this.querySelector('#list')
```
justindarc: `this.querySelector('#list')` -> `this.getElementById('list')`

apps/music/elements/music-search-results.js
justindarc: Ah ok. Makes sense now.

apps/music/elements/music-search-box.js
wilsonpage: It was for catching a custom-element being registered twice in the same document.

apps/music/elements/music-search-results.js
wilsonpage: `HTMLElement` doesn't have that method. Only `Document` and `DocumentFragment` IIRC.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2063089792438264d5dff2523619307463ab2b35

apps/music/components/gaia-fast-list/gaia-fast-list.js
```diff
@@ -356,14 +382,24 @@ var GaiaFastListProto = {
...
-      ::content .image ~ p {
-        padding-inline-end: 60px;
+      ::content :-moz-dir(ltr) .image ~ h3,
+      ::content :-moz-dir(ltr) .image ~ p {
+        padding-right: 52px;
+      }
+
+      ::content :-moz-dir(rtl) .image ~ h3,
+      ::content :-moz-dir(rtl) .image ~ p {
+        padding-left: 52px;
```
jimporter: I would double-check that this looks right in RTL for the search results list. I tried doing this in a separate patch, and found that -moz-dir(rtl) doesn't actually work (even with the directionality hacks that gaia-components use) in a web component used by another web component.

justindarc: That may have been because of nested Shadow DOM. But, in this patch, I believe the search results are no longer in Shadow DOM.

apps/music/bower.json
```diff
@@ -10,13 +10,13 @@
   ],
   "dependencies": {
     "bridge": "gaia-components/bridge#master",
-    "gaia-fast-list": "gaia-components/gaia-fast-list#~0.2.2",
-    "gaia-header": "gaia-components/gaia-header#~0.9.5",
+    "gaia-fast-list": "gaia-components/gaia-fast-list#master",
```
jimporter: I'm not super-familiar with bower, but doesn't this mean that every time we update our bower deps, we'll fetch the latest code from gaia-fast-list/master? That doesn't seem like a good idea, since we could end up pulling in a newer gaia-fast-list than we intend and potentially cause new bustage.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b6ebebc25ede28e577dc13628af08e407f0cba3e

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=bfa8c2f373bdee971c57e120276fa15ca5dba969

PR #33000, opened 2015-11-04
Bug 1221650 - [Gallery] LazyLoad gaia-header.js

punamdahiya:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=570b93534339c20de3d3d342ebf876617cdec937

apps/gallery/index.html
```diff
@@ -57,7 +57,6 @@
     <link rel="stylesheet" type="text/css" href="app://theme.gaiamobile.org/shared/elements/gaia-theme/gaia-theme.css">
     <link rel="stylesheet" type="text/css" href="shared/elements/gaia-icons/gaia-icons.css" />
     <link rel="stylesheet" type="text/css" href="shared/elements/gaia-icons/bidi-helper.css" />
-    <script defer src="shared/elements/gaia-header/dist/gaia-header.js"></script>
```
davidflanagan: Don't you need to put this line in a comment to ensure that it gets packaged correctly?

I'm not sure why this patch actually works, unless we're packaging way too much shared/ stuff into the application.zip file...

apps/gallery/js/startup.js
```diff
@@ -273,11 +273,17 @@
...
 
       // Remove the attribute to trigger gaia header font-fit logic, defer it
-      // til now to postpone the overhead.
-      var headers = document.querySelectorAll('gaia-header');
-      for (var i = 0; i < headers.length; ++i) {
-        headers[i].removeAttribute('no-font-fit');
-      }
+      // til now to postpone the overhead. Add setTimeout to lazyload
+      // gaia-header.js aynchronously.
+      setTimeout(function() {
```
davidflanagan: LazyLoader.load() is already async, so do you really need setTimeout() here?  If it works without adding the setTimeout() I think that would be better.

If you tap the select button at the bottom right as soon as the app launches can you get to that screen before the header loads or before it is localized?  Maybe the event handlers for the thumbnails and the select button shouldn't be registered until the header is lazy loaded. If so, then we should do performance.mark('contentInteractive') when the event handlers are registered after the lazy load.

I'm not sure that it matters in practice, but that seems like the right way to do it.

punamdahiya: Thats what I thought , but on double check I found https://bugzilla.mozilla.org/show_bug.cgi?id=926851 still open , and till its resolved we need setTimeout to make it async.

While testing, I tried to tap on select as soon as its visible and still see gaia-header loaded. I guess its too fast to catch the interaction manually. I can  move performance.mark('contentInteractive') after the  select and thumbnail event handler are registered and will post updated numbers.

apps/gallery/index.html
punamdahiya: I am not sure if commenting lazy loaded file in index.html is still a requirement for build e.g. info.js is lazy loaded but its not listed in index.html. But I will definitely verify it and update.

punamdahiya: As per https://developer.mozilla.org/en-US/docs/Mozilla/Firefox_OS/Developing_Gaia/Build_System_Primer#Build_script_for_apps
See 'Build Script for Apps'  utils.copyToStage in build.js copies gallery from app directory to build_stage

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=99b9f3d6ded6b8aafcec36d5769571dde6c30bcd

apps/gallery/index.html
punamdahiya: See https://github.com/mozilla-b2g/gaia/tree/master/shared/elements#build-system, I believe for components assets we don't need to comment inside index.html

punamdahiya: https://github.com/mozilla-b2g/gaia/blob/master/apps/gallery/open.html#L38
phew, this is why the patch is working even when we don't have gaia-header shared asset not commented in index.html.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ea25f094e267e34fbc4c31a80695f221f25545ce

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=09c1ece023cef2725ae023a4cf53ab535df8fde4

apps/gallery/index.html
davidflanagan: Thanks for figuring that out!

apps/gallery/js/startup.js
davidflanagan: I don't really understand that bug. LazyLoader.load() is async, though I guess it could be even more async. Maybe at this point we need to return to the event loop as soon as possible, to get stuff on the screen. If so, then using setTimeout is the right thing to do.

apps/gallery/js/startup.js
```diff
@@ -261,23 +257,34 @@
...
+
+      // Clicking on the select button goes to thumbnail select mode
+      $('thumbnails-select-button').onclick =
+        setView.bind(null, LAYOUT_MODE.select);
+
+      // Handle clicks on the thumbnails
+      thumbnails.addEventListener('click', thumbnailClickHandler);
+      // Tell performance monitors that "above the fold" content
+      // is ready to interact with.
+      window.performance.mark('contentInteractive');
```
davidflanagan: I thought you were going to move this stuff into the lazy load callback.  There doesn't see to be much point in moving it here. According to raptor it makes only about 1ms or less difference in when contentInteractive happens. 

I don't feel strongly about this, but I'd say either leave this code where it was, or do it after the gaia-header code is actually loaded.  In that case, I'd expect to see contentInteractive come 30 to 100ms after visually loaded

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=595a74ecb96153223f2d178858340e2eb7863442

apps/gallery/js/startup.js
punamdahiya: On rethinking gallery app content interactive mark on app launch should be all thumbnail items on first page are interactive which we are ensuring, so I will prefer not to move 'contentInteractive' mark and keep the code as it was.

PR #33021, opened 2015-11-05
Bug 1219176 - Make it possible to dynamically preload external apps to be run with Raptor

azasypkin:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=363031421d1bbcd0fb891035eb1d3ad89ef5b3b9

Makefile
```diff
@@ -850,9 +850,9 @@ caldav-server-install:
 .PHONY: raptor
 raptor: node_modules
 ifneq ($(APP),)
-	RAPTOR=1 PERF_LOGGING=1 DEVICE_DEBUG=1 GAIA_OPTIMIZE=1 NOFTU=1 SCREEN_TIMEOUT=0 APP=$(APP) make install-gaia
+	GAIA_DISTRIBUTION_DIR=distros/raptor RAPTOR=1 PERF_LOGGING=1 DEVICE_DEBUG=1 GAIA_OPTIMIZE=1 APP=$(APP) make install-gaia
```
azasypkin: ```NOFTU=1``` and ```SCREEN_TIMEOUT=0``` are set inside corresponding distro folder via ```settings.json```

build/preferences.js
```diff
@@ -177,6 +177,10 @@ PreferencesBuilder.prototype.preparePref = function() {
   if (this.config.PERF_LOGGING === '1') {
     this.userPrefs['dom.performance.enable_user_timing_logging'] = true;
   }
+  // Turn on Service Workers for the Raptor
+  if (this.config.RAPTOR === '1') {
```
azasypkin: It's temporary, I'm not sure where we should set that. But we need this pref for the app that use Service Workers. Just few ideas that come to my mind:
* Leverage existing ```NGA_SERVICE_WORKERS``` here;
* Temporary (until SW are enabled by default) add ```'dom.serviceWorkers.enabled'``` to [settings.js](http://mxr.mozilla.org/mozilla-central/source/b2g/chrome/content/settings.js#661) so that we can turn it on via ```distros/raptor/settings.json```.

What do you think?

distros/raptor/remote.list
```diff
@@ -0,0 +1,2 @@
+shared-worker-memory,https://raptor-shared-worker-memory.github.io/package.webapp
```
azasypkin: These are examples, still I didn't find any good and non-hacky way to host several apps on the same domain with fixed ```origin``` that Raptor can successfully run - so for now separate subdomains.

tools/preload.py
```diff
@@ -269,7 +269,10 @@ def fetch_webapp(app_url, directory=None):
                 manifest['package_path'],
                 os.path.join(app_dir, filename))
             metadata['manifestURL'] = url.geturl()
-            metadata['packageEtag'] = open_from_url(path).headers['etag']
+
+            package_url = open_from_url(path)
```
azasypkin: I'll get rid of this change from this PR and move it to separate PR for the apppreload-toolkit repo - without this change we couldn't preload apps from the server that doesn't support ```Etag``` HTTP header (e.g GitHub Pages) - drawback is that we'll likely redownload package for every ```make raptor``` call since ```packageEtag``` will be empty.

distros/raptor/apps.list
```diff
@@ -0,0 +1,2 @@
+apps/*
```
azasypkin: We can include/exclude any app we want here

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e183880c02314b67b1d6d12cd97fc9b60307321b

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=07d1bae4caeb07aacf004a40730c430eb5f141a1

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d38091cfe6b253247249b92973cf86c64e6c9a1d

PR #33044, opened 2015-11-06
Bug 1216301 - Silent updates for add-ons

daleharvey:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a9da6974427f74cccedba2003546b97c58f6de0c

apps/system/js/updatable.js
```diff
@@ -70,16 +91,103 @@ AppUpdatable.prototype.clean = function() {
...
-    this.app.ondownloadapplied = this.appliedCallBack.bind(this);
-    this.app.onprogress = this.progressCallBack.bind(this);
+  if (this.app.installState !== 'installed') {
+    return;
   }
+
+  let settings = [
+    setting(AppUpdatable.AUTO_UPDATE),
+    setting(AppUpdatable.NOTIFY_UPDATE)
+  ];
```
mikehenrty: We shouldn't read these settings each time. Better would be to read the setting the first time by adding a SettingsObserver, and then keep the value cached in a static variable.

apps/system/js/updatable.js
```diff
@@ -70,16 +91,103 @@ AppUpdatable.prototype.clean = function() {
...
+    var notifyAddonUpdates = results[1];
+
+    // If we have autoUpdates turned off for addons, they
+    // go via the usual install flow
+    if (this.app.manifest.role === 'addon' && autoAddonUpdates) {
+
+      // We fail silently, the addon will be updated next time
+      // we check for updates
+      this.app.ondownloaderror = err => {
+        console.error('zxc failed to auto update', this.name,
```
mikehenrty: what is zxc?

apps/system/js/updatable.js
```diff
@@ -70,16 +91,103 @@ AppUpdatable.prototype.clean = function() {
...
+      this.app.ondownloadapplied = this.appliedCallBack.bind(this);
+      this.app.onprogress = this.progressCallBack.bind(this);
+    }
+  });
+};
+
+AppUpdatable.prototype.notifyAutoUpdate = function() {
+  IconsHelper.getIcon(this.app.origin, 32, null, this.app).then(iconUrl => {
+    var opts = {
+      tag: 'addonUpdated',
```
mikehenrty: This is a little weird. Do we want separate notifications for each add-on? Right now only the last applied add-on will have the notifications since it will overwrite earlier add-on notifications. Not sure if this is the behavior we want here.

apps/system/js/updatable.js
daleharvey: Yeh I had it as an observer to start with, was worried about the cost of multiple observers but will switch it back

apps/system/js/updatable.js
daleharvey: heh its what I started putting in my log statements so there was less swearing and I can grep, will remove

apps/system/js/updatable.js
daleharvey: ah does the `tag` signify individual addons? if so then good catch, we want a notification per add-on

apps/settings/js/panels/about/update_check.js
```diff
@@ -33,11 +33,35 @@ define(function(require) {
...
     /**
+     * Shows and hides the "notify me" checkbox depending on whether auto
+     * updates are enabled.
+     *
+     * @access private
+     * @memberOf UpdateCheck.prototype
+     */
+    _displayNotifyCheckbox: function uc__displayNotifyCheckBox() {
+      var key = 'addons.auto_update';
+      var request = this._settings.createLock().get(key);
```
yzen: settings lock supports promise interface AFAIK, please use that so we have both then and catch screnarios (console.error on catch please).

apps/settings/locales/settings.en-US.properties
```diff
@@ -1151,6 +1151,8 @@ browser-os-desc = {{brandShortName}} is the free and open source operating syste
 learn-more=Learn More
 reset-head=Reset
 reset-warning-title= CAUTION
+addons-auto-update=Automatic add-on updates
```
yzen: Nit: Update add-ons automatically

yzen: Ah, saw the spec text, though the spec feels too techy.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=24182343ccc81f9fbacbf574a1fcda8e24e6de13

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=12afc1bad82fdca9609ea0398a60dace682552e2

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d3841ec92fca3fe0f6d88da4b651bd699914d40f

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a6cccbf326175bc9cdc3d3264a707512c6ffc511

apps/system/js/updatable.js
```diff
@@ -19,6 +21,49 @@
...
+    // create multiple listeners
+    if (key in values) {
+      return values[key];
+    }
+
+    values[key] = new Promise(resolve => {
+      // We store the value of the setting and the observer is
+      // in charge of keeping it up to date
+      function cacheAndResolve() {
+        values[key] = (key in req.result) ? req.result[key] : undefined;
```
mikehenrty: This needs to be:
```
values[key] = (key in req.result) ? Promise.resolve(req.result[key]) : undefined;
```

apps/system/js/updatable.js
```diff
@@ -19,6 +21,49 @@
  * - cancelDownload() to cancel it
  */
 
+// Wrapper to read a settings value
+var setting = (function() {
+  var values = {};
+  return key => {
+
+    // Cache the promise straight away so multiple callers dont
+    // create multiple listeners
```
mikehenrty: This comment would make more sense if it was moved to where we set the key below.

apps/system/js/updatable.js
```diff
@@ -19,6 +21,49 @@
...
+      // in charge of keeping it up to date
+      function cacheAndResolve() {
+        values[key] = (key in req.result) ? req.result[key] : undefined;
+        resolve(values[key]);
+      }
+
+      var lock = navigator.mozSettings.createLock();
+      var req = lock.get(key);
+
+      req.onerror = cacheAndResolve.bind(this);
```
mikehenrty: nit: no need for bind

mikehenrty: Are you sure you want to hide errors here? Seems better to reject the promise, and handle the rejection with a catch below. Right now we are silently disabling auto updates and add-on notifications. Not sure this is what we want for things like Settings db errros.

apps/system/test/unit/updatable_test.js
```diff
@@ -184,6 +184,7 @@ suite('system/Updatable', function() {
 
     downloadAvailableSuite('app has a download available', function() {
       mockApp.downloadAvailable = true;
+      console.log('mock app is', mockApp);
```
mikehenrty: nit: remove

apps/system/js/updatable.js
```diff
@@ -68,18 +113,122 @@ AppUpdatable.prototype.clean = function() {
...
 
-    // we add these callbacks only now to prevent interfering
-    // with other modules (especially the AppInstallManager)
-    this.app.ondownloaderror = this.errorCallBack.bind(this);
-    this.app.ondownloadsuccess = this.successCallBack.bind(this);
-    this.app.ondownloadapplied = this.appliedCallBack.bind(this);
-    this.app.onprogress = this.progressCallBack.bind(this);
+  // If its not an addon, add it to the update queue
+  if (this.app.manifest.role !== 'addon') {
+    this.queueUpdate();
```
mikehenrty: nit: maybe we should call this `queueAppUpdate` to be more explicit

apps/system/js/updatable.js
daleharvey: Good catch cheers

apps/system/js/updatable.js
daleharvey: This is going to be a fairly expected error that a lot of people will not have these settings defined, and I think it makes sense for missing settings to default to false

apps/system/js/updatable.js
daleharvey: Add-ons also go via this queue iff auto updates are not enabled

mikehenrty: ah nm, i see we use this for add-ons too

apps/system/js/updatable.js
```diff
@@ -19,6 +21,49 @@
...
+        values[key] = Promise.resolve(e.settingValue);
+      });
+    });
+
+    return values[key];
+  };
+})();
+
+// Should we automatically update?
+AppUpdatable.AUTO_UPDATE = 'addons.auto_update';
```
mikehenrty: We probably want this to be on by default? If so, we should change this to something like `addons.disable_auto_update`. That way we don't need to do a version update in the Settings db (in gecko) to pick up a `true` default value for this.

apps/system/test/marionette/update_addon_test.js
```diff
@@ -0,0 +1,79 @@
...
+
+    // Update the manifest.
+    server.setRoot(__dirname + '/fixtures/template_addon_updated');
+
+    // Check for an update.
+    appInstall.stageUpdate(server.webappUpdateURL);
+
+    // We should get a notification on the first time an
+    // add-on is automatically installed
+    notificationList.waitForNotificationCount(2);
```
mikehenrty: It would be more robust if we could check for the update notification explicitly.

apps/system/test/marionette/update_addon_test.js
```diff
@@ -0,0 +1,79 @@
...
+    // Update the manifest.
+    server.setRoot(__dirname + '/fixtures/template_addon_updated');
+
+    // Check for an update.
+    appInstall.stageUpdate(server.webappUpdateURL);
+
+    // Dont wait 30 seconds for the notification to show
+    system.sendSystemUpdateNotification();
+
+    notificationList.waitForNotificationCount(1);
```
mikehenrty: Like above, it would be more robust to make sure none of the notifications are update notifications for add-ons.

apps/system/js/updatable.js
```diff
@@ -19,6 +21,49 @@
...
+
+// Notify the users after we have auto updated?
+AppUpdatable.NOTIFY_UPDATE = 'addons.update_notify';
```
mikehenrty: Are we sure we need a setting for this? I haven't checked the spec, but it seems like we could just always show the notification the first time we auto-update, and then never again. It's what we do for tracking protection i believe.

apps/system/test/marionette/update_addon_test.js
```diff
@@ -0,0 +1,79 @@
...
+  });
+
+});
```
mikehenrty: I think we need a test to make sure we don't show the notification on the second app update too. Is this possible?

mikehenrty:
Seems like we should be adding some defaults for the new settings to common-settings.json

apps/system/js/updatable.js
daleharvey: Yeh this is a seperate notification, there is a notification on the first auto update to happen, similiar to private tracking and that is stored in localstorage. There is a setting to let users be notified after every update and this keeps track of that

apps/system/test/marionette/update_addon_test.js
daleharvey: Heh yeh I spent ages trying to figure out how to actually test the contents, but I will do

apps/system/js/updatable.js
daleharvey: I would really prefer that we dont work around bad missing functionality, its fairly confusing to have negative flags, there is a bug open to fix the settings behaviour (https://bugzilla.mozilla.org/show_bug.cgi?id=1223781) and the worst that happens is people get a notification to click to update.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4e168ab8488022aa8ec9bee5e03db47143c85674

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8bf06a165c3e41a24b06c3586abbf3a8d69f8bf2

apps/system/test/marionette/lib/notification.js
```diff
@@ -110,9 +113,11 @@ NotificationList.prototype = {
 
   // fetch the list of open notifications from system tray
   refresh: function() {
+    console.log('refreshing', this.selectors.items);
```
mikehenrty: nit: remove

apps/system/test/marionette/lib/notification.js
```diff
@@ -110,9 +113,11 @@ NotificationList.prototype = {
...
     this.notifications = this.client.executeScript(
       this._remoteGetNotificationDetails,
       [this.selectors.items]);
+    console.log('refreshing', this.notifications.length, this.notifications);
```
mikehenrty: nit: remove

apps/system/test/marionette/lib/notification.js
```diff
@@ -182,6 +187,18 @@ NotificationList.prototype = {
...
 
+  clear: function() {
+    this.client.executeAsyncScript(function() {
+      var clearAll = document.getElementById('notification-clear');
+      var event = document.createEvent('CustomEvent');
+      event.initCustomEvent('click', true, true, {});
+      var notification;
+      notification = new Notification('title 3');
+      notification.onclose = function(){ marionetteScriptFinished(true); };
+      notification.onshow = function(){ clearAll.dispatchEvent(event); };
```
mikehenrty: probably simpler by just doing a `waitForNotificationCount(0)`

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=afa59b4c903f7af24c6f5e80a34025ea8c93ad33

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a8e4da78d502034b7f6652af96520c332e906d0a

PR #33071, opened 2015-11-09
Bug 1200583 - [Messages][Tests] It is difficult to debug when we forget to set the mock data in an integration test. Storage code is rewritten so that data can be set _before_ app is started. r=julien

azasypkin:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=540d85614fd29f37ba54595ccd5ee849dbeef1ad

apps/sms/test/marionette/composer_test.js
```diff
@@ -37,6 +37,7 @@ marionette('Messages Composer', function() {
 
   var MOCKS = [
     '/mocks/mock_test_storages.js',
+    '/mocks/mock_test_blobs.js',
```
azasypkin: I started to use image generation code in contacts mock as well - so moved it to a separate module.

apps/sms/test/marionette/composer_test.js
```diff
@@ -60,10 +61,6 @@ marionette('Messages Composer', function() {
     setup(function() {
       messagesApp.launch();
 
-      // Set empty stores.
-      storage.setMessagesStorage();
```
azasypkin: No need to set empty storage anymore \0/

apps/sms/test/marionette/conversation_test.js
```diff
@@ -40,11 +41,8 @@ marionette('Conversation Panel Tests', function() {
         thread = ThreadGenerator.generate({
           numberOfMessages: 50
         });
-        messagesApp.launch();
-
         storage.setMessagesStorage([thread], ThreadGenerator.uniqueMessageId);
-        // Set empty contacts store.
-        storage.setContactsStorage();
+        messagesApp.launch();
```
azasypkin: But now we need to set storage __before__ app is started, that's easy to do

apps/sms/test/marionette/lib/storage.js
```diff
@@ -15,40 +15,10 @@
...
-            window.wrappedJSObject.TestStorages.setStorage(STORAGE_NAME, {
-              threads: threadMap,
-              messages: messageMap,
-              recipientToThreadId: recipientToThreadId,
-              uniqueMessageIdCounter: uniqueMessageIdCounter
-            });
-
-            window.wrappedJSObject.TestStorages.setStorageReady(STORAGE_NAME);
-          }, [threads || [], uniqueMessageIdCounter || 0]);
+          this.setStorage('messagesDB', {
```
azasypkin: I moved storage preparation code to mock itself as it's the only mock that can use this data directly. and to clean this code a bit.

apps/sms/test/marionette/lib/storage.js
```diff
@@ -57,15 +27,29 @@
...
+         */
+        setStorage: function(storageName, storage) {
+          client.scope({ context: 'chrome' }).executeScript(
+            function(storage, storageName) {
+              Services.obs.addObserver(function(window) {
+                window.wrappedJSObject.TestStorages.set(
+                  storageName,
+                  Components.utils.cloneInto(storage, window.wrappedJSObject)
+                );
+              }, 'test-storage-ready', false);
```
azasypkin: This is the _heart_ of this PR - once storage mock is initialized it broadcasts special event, so that we can set storage before app is started (well in fact app is started, but no app js or html is parsed yet).

apps/sms/test/marionette/mocks/mock_navigator_moz_contacts.js
```diff
@@ -21,20 +21,33 @@ Services.obs.addObserver(function(document) {
 
   var window = document.defaultView;
 
+  function rehydratePhotos(contacts) {
```
azasypkin: No one uses this functionality yet, but I want to use in the test for the [bug 1217859](https://bugzilla.mozilla.org/show_bug.cgi?id=1217859)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5dc547b210c36efe6ee51e794cfb44b027338c14

apps/sms/test/marionette/lib/storage.js
julienw: so that means we can't set the storage after the event has been sent though ? Not a big deal though.

also can you unregister the observer ?

julienw: unless we don't want this -- it's not clear to me how many times the test-storage-ready can be sent here :)

apps/sms/test/marionette/mocks/mock_test_storages.js
```diff
@@ -11,49 +9,13 @@ Services.obs.addObserver(function(document) {
...
-      storage.deferred.resolve(Cu.waiveXrays(storage.data));
-    }
-  };
 
   Object.defineProperty(window.wrappedJSObject, 'TestStorages', {
     // The property should be writable since mock is inserted/rewritten in
     // setup function that is called for every test in the suite.
     writable: true,
-    value: Cu.cloneInto(TestStorages, window, { cloneFunctions: true })
+    value: Cu.cloneInto(new Map(), window, { cloneFunctions: true })
```
julienw: wondering if `value: new window.Map()` works (or `new window.wrappedJSObject.Map()`)

apps/sms/test/marionette/lib/storage.js
azasypkin: > so that means we can't set the storage after the event has been sent though ? Not a big deal though.

Nope, not currently (and we can not update storage in master as well), can be changed when we need such a feature.

> also can you unregister the observer ?

Not sure, but in this patch I don't really want to do this as a lot of tests require the same storage after app is relaunched. I think it can be part of the "reset/change app storage" if we need this eventually.

azasypkin: > it's not clear to me how many times the test-storage-ready can be sent here :)

I'm planning to improve our mocks further in the future - currently I see our mocks are registered several times - it's ok that they are registered for keyboard and Homescreen, but it's not obvious to me why our mocks are registered twice for main frame and twice for shim host - same will be for this event - it will fire several times and every time storage will be rewritten - it's old problem that I didn't find time to fix yet :)

apps/sms/test/marionette/mocks/mock_test_storages.js
azasypkin: You want to get rid of ```Cu.cloneInto```? :) It will likely work, I can check.

apps/sms/test/marionette/lib/storage.js
julienw: > Not sure, but in this patch I don't really want to do this as a lot of tests require the same storage after app is relaunched. I think it can be part of the "reset/change app storage" if we need this eventually. What do you think?

That's what I'm saying: it's not clear to me how the flow happens. I have 2 concerns here:
1. if we don't unregister we keep the storage in memory.
2. if we call setStorage once again after a relaunch, we will have the handler called twice.

but I don't know if this is a big deal.

apps/sms/test/marionette/mocks/mock_test_storages.js
julienw: yeah, if we can avoid it I think it's cleaner :)

apps/sms/test/marionette/lib/storage.js
azasypkin: >  if we don't unregister we keep the storage in memory.

Well, imo, it's not a big deal for the integration tests - I believe memory should be freed once suite is finished. And as I explain below this event is called for several different frames, so we can safely unsubscribe only when once all frames fired this event and we set storage for all of them.

> if we call setStorage once again after a relaunch, we will have the handler called twice.

Yes and this should not break test as storage will be overridden by the last set storage - we can consider this as  temporary way to override storage on the next startup :) We don't have such cases, but as always - once one need this feature - he'll make it better.

However I agree that the way our injected scripts work currently is confusing and not clear. We just follow the way marionette content script plugin suggests us - register observer for document-element-inserted event. That means we call it for keyboard, homescreen, Messages main frame and shim host - not a big deal and somewhat correct that all apps has the same API output during one test case, but I see that this is called twice for every document element - that what I'd like to investigate and filed bug for that - https://bugzilla.mozilla.org/show_bug.cgi?id=1223350

apps/sms/test/marionette/mocks/mock_test_storages.js
azasypkin: Done, thanks!

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1f89ba31430dde55b81d0d1609759b2af65332e0

PR #33081, opened 2015-11-09
Bug 1164790 - [Settings] Set correct direction authorized number

autra:
An international number should have a dir="auto".
Likewise, the name is an input from the user. It needs a dir="auto"


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=51447aa07ab0be1400a2cd93ee5c115eeacb8362

PR #33091, opened 2015-11-09
Bug 1211433 - [Metrics] Increase thresholds for reflows

russnicoletti:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=35dd760680c92df56192653e48808d8926856e1d

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8efd18198450faffae5e1d729daa661e768b9eed

apps/settings/locales/settings.en-US.properties
```diff
@@ -1233,7 +1233,8 @@ hud-security=Security Issues
 hud-performance=Performance
 hud-reflows=Reflows
 hud-jank=Jank
-jank-threshold=Jank Threshold
+jank-threshold=Jank Threshold (ms)
```
gasolin: please rename the id when you change the string, or the l10n team can't fetch and localize it

apps/settings/elements/developer_hud.html
```diff
@@ -69,6 +69,16 @@ <h2 data-l10n-id="hud-performance"></h2>
           </gaia-checkbox>
         </li>
         <li>
+          <p data-l10n-id="reflow-threshold"></p>
+          <span class="button icon icon-dialog">
+            <select name="devtools.reflowduration.threshold" data-value-type="integer">
+              <option value="1" selected>No Threshold</option>
```
jankeromnes: I think that some incremental reflows could happen in less than a millisecond. Please use `value="0"` as "No Threshold" (also, I'm not sure that the "No Threshold" string here can be localized, so please use "0" instead).

apps/settings/elements/developer_hud.html
```diff
@@ -69,6 +69,16 @@ <h2 data-l10n-id="hud-performance"></h2>
...
+              <option value="16">16</option>
```
jankeromnes: After thinking some more about it, 16ms is indeed the length of a frame at 60 FPS, but if your reflow takes the full 16ms, it might not leave enough time for things like JS and repaint to push the frame out in time (i.e. a 15ms reflow will probably cause a frame drop).

10ms seemed like a good value to me. What about the following thresholds: 0, 10, 20, 50, 100? I think that eventually, we'd like the `jank-threshold` to have the same values, but no need to change that now.

apps/settings/locales/settings.en-US.properties
jankeromnes: Since you need to change this ID, please make it consistent with the other IDs (i.e. something like `hud-jank-threshold`).

apps/settings/locales/settings.en-US.properties
```diff
@@ -1233,7 +1233,8 @@ hud-security=Security Issues
...
+reflow-threshold=Duration Threshold (ms)
```
jankeromnes: If you follow my above suggestion, please change this to `hud-reflow-threshold` as well.

build/config/common-settings.json
```diff
@@ -83,6 +83,7 @@
    "devtools.overlay": false,
    "devtools.overlay.system": false,
    "devtools.eventlooplag.threshold": 100,
+   "devtools.reflowduration.threshold": 16,
```
jankeromnes: Nit: Most of the HUD settings start with the `hud.` prefix instead of `devtools.`.

Add-on nit: Also, the duration-threshold you're creating here is for all the reflow-related metrics. Please call it something like `hud.reflows.threshold`. (I know it's inconsistent with `devtools.eventlooplag.threshold`, but I'd love to rename that as well eventually).

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5d4b41abbd2b745aad0f01bb070358cd655d9a1e

PR #33099, opened 2015-11-10
Bug 1223275 - migrate email app to NotificationHelper

hkasemir:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8e9029841b38c62d478330dcd1d0419f099d5a71

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6266eaeff74be4ad392915b5a66a0ab79161772a

apps/email/js/sync.js
```diff
@@ -77,22 +75,23 @@ define(function(require) {
...
-        // and synthesize an "event" ourselves.
-        notification.onclick = function() {
-          evt.emit('notification', {
-            clicked: true,
-            imageURL: iconUrl,
-            tag: notificationId,
-            data: data
-          });
-        };
+        var titleL10n = title;
```
zbraniecki: You can't just ignore the `notification-no-subject` scenario.
You need to do sth like:

```javascript
var titleL10n = title ? { raw: title } : 'notification-no-subject';
```

apps/email/js/sync.js
```diff
@@ -62,13 +62,11 @@ define(function(require) {
         //TODO: consider setting dir and lang?
         //https://developer.mozilla.org/en-US/docs/Web/API/notification
         var notificationOptions = {
-          body: body,
+          bodyL10n: body,
```
zbraniecki: if `body` is not an L10n Object, it should be turned into one as `{raw: body }`.

zbraniecki: But probably not within this function, but in both calls to this function

apps/email/js/sync.js
```diff
@@ -501,7 +503,7 @@ define(function(require) {
 
           sendNotification(
             BACKGROUND_SEND_NOTIFICATION_ID,
-            mozL10n.get('background-send-error-title'),
+            'background-send-error-title',
```
zbraniecki: so, here you are passing L10n Object, which is proper, but you also have to switch all the `subject = mozL10n.get` to `subject = {id: ..., args... };`.

apps/email/js/sync.js
```diff
@@ -321,7 +320,7 @@ define(function(require) {
               });
             }
 
-            body = newFromNames.join(mozL10n.get('senders-separation-sign'));
+            body = newFromNames.join('senders-separation-sign');
```
zbraniecki: So, here you need to use an async version of `mozL10n.get` called `mozL10n.formatValue` - it returns a promise.

```javascript
navigator.mozL10n.formatValue('senders-separation-sign').then(separator => {
  body = newFromNames.join(separator);
});
```

and then the rest of the logic has to stay inside the Promise callback.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e2491483bc0e9587925f2c42d0bd111150b23603

apps/email/js/sync.js
```diff
@@ -7,7 +7,7 @@ define(function(require) {
       appSelf = require('app_self'),
       evt = require('evt'),
       mozL10n = require('l10n!'),
-      notificationHelper = require('shared/js/notification_helper');
+      NotificationHelper = require('shared/js/notification_helper');
```
zbraniecki: Please, stick to the naming convention - variables start with lower case letter.

apps/email/js/sync.js
```diff
@@ -77,22 +75,23 @@ define(function(require) {
...
-        // and synthesize an "event" ourselves.
-        notification.onclick = function() {
-          evt.emit('notification', {
-            clicked: true,
-            imageURL: iconUrl,
-            tag: notificationId,
-            data: data
-          });
-        };
+        titleL10n = { raw: titleL10n } || 'notification-no-subject';
```
zbraniecki: Instead of this you have to use:

```javacscript
titleL10n = titleL10n || 'notification-no-subject';
```

and then format `titleL10n` in `sendNotification` callers.

apps/email/js/sync.js
```diff
@@ -311,17 +310,23 @@ define(function(require) {
             }
 
             if (model.getAccountCount() === 1) {
-              subject = mozL10n.get('new-emails-notify-one-account', {
-                n: count
-              });
+              subjectL10n = { id: 'new-emails-notify-one-account',
+                              args: { n: count }
+                            };
```
zbraniecki: please, stick to formatting used in this file. Indentation is two spaces.

apps/email/js/sync.js
```diff
@@ -311,17 +310,23 @@ define(function(require) {
...
-              subject = mozL10n.get('new-emails-notify-multiple-accounts', {
-                n: count,
-                accountName: result.address
-              });
+              subjectL10n = { id: 'new-emails-notify-multiple-accounts',
+                              args: {
+                                n: count,
+                                accountName: result.address
+                              }
+                            };
```
zbraniecki: same here.

apps/email/js/sync.js
```diff
@@ -311,17 +310,23 @@ define(function(require) {
...
             }
+            
+            mozL10n.formatValue('senders-separation-sign')
+              .then(separator => {
+                bodyL10n = newFromNames.join(separator);
+            });
```
zbraniecki: you can't just go on with your code after async `formatValue`. Everything behind this code that is used until return has to be within the Promise then callback.

apps/email/js/sync.js
```diff
@@ -501,8 +512,8 @@ define(function(require) {
 
           sendNotification(
             BACKGROUND_SEND_NOTIFICATION_ID,
-            mozL10n.get('background-send-error-title'),
-            data.localizedDescription,
+            { raw: 'background-send-error-title' },
+            { raw: data.localizedDescription },
```
zbraniecki: here you take `{raw: data.localizedDescription}`. 50 lines earlier the code was using `mozL10n.get` to localize the `descId`. You should not use it as `raw`, but intead pass `descId` which is an L10nID.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=35342e850b4db9f4ba329e61809cb522618300c4

apps/email/js/sync.js
```diff
@@ -62,13 +62,11 @@ define(function(require) {
...
         var notificationOptions = {
-          body: body,
+          bodyL10n: bodyL10n,
           icon: iconUrl,
           tag: notificationId,
           data: data,
-          mozbehavior: {
-            noscreen: true
-          }
+          mozbehavior: { noscreen: true }
```
zbraniecki: do not alter the formatting of pieces that you're not working on

apps/email/js/sync.js
```diff
@@ -311,17 +310,23 @@ define(function(require) {
             }
 
             if (model.getAccountCount() === 1) {
-              subject = mozL10n.get('new-emails-notify-one-account', {
-                n: count
-              });
+              subjectL10n = { id: 'new-emails-notify-one-account',
+                args: { n: count }
+              };
```
zbraniecki: it's nice to put id in a separate line to align it with args:

```javascript
subjectL10n = {
  id: ...,
  args: ...
};
```

apps/email/js/sync.js
```diff
@@ -336,28 +341,40 @@ define(function(require) {
...
-            body,
-            iconUrl,
-            dataObject,
-            behavior
-          );
+          if (!bodyL10n.then) {
+            bodyL10n = Promise.resolve(bodyL10n);
+          }
+          
+          bodyL10n.then(() => {
```
zbraniecki: You need to take the result of the promise:
```javascript
bodyL10n.then(body => {
  sendNotification(
    ...,
    body,
    ...,
  );
});
```

otherwise you are passing a Promise.

apps/email/js/sync.js
```diff
@@ -501,8 +513,8 @@ define(function(require) {
 
           sendNotification(
             BACKGROUND_SEND_NOTIFICATION_ID,
-            mozL10n.get('background-send-error-title'),
-            data.localizedDescription,
+            { raw: 'background-send-error-title' },
```
zbraniecki: You don't want to pass l10nId as a raw string.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=56e0de61352c5b190b3d7a8cf565c7b090bec13c

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0505901703d7e6954bbe2a7400be6fc7a1a1ac5a

apps/email/js/sync.js
```diff
@@ -336,28 +345,40 @@ define(function(require) {
...
+          bodyL10n.then(result => {
+            sendNotification(
+              result.id,
+              subjectL10n,
+              result,
+              iconUrl,
+              dataObject,
+              behavior
+            );
+          });
```
jrburke: This bit here is a bit tricky since the Promise then resolution will be an async resolution, after the accountsResults.updates.forEach() iteration has finished, and after the finishSync() after the accountsResults.updates.forEach() has been called.

We want to only call finishSync when we know all the work of sending out the notifications has been done.

Since the issue is that we really need to wait on `mozL10n.formatValue('senders-separation-sign')`'s promise to resolve, to avoid too much nesting of async blocks, let's move the `mozL10n.formatValue('senders-separation-sign')` work into the `getEnv function. It can be the innermost step of getEnv(). This also means the fn that getEnv calls will pass the `separator` variable as a fourth argument.

This this means the `if (!bodyL10n.then) {` block is not needed, and there is no more need for the `bodyL10n.then(result => {` part, the sendNotification can be called directly.

apps/email/js/sync.js
```diff
@@ -311,17 +312,25 @@ define(function(require) {
...
+                  n: count,
+                  accountName: result.address
+                }
+              };
             }
+            
+            bodyL10n = mozL10n.formatValue('senders-separation-sign')
+              .then(separator => {
+                return newFromNames.join(separator);
+            });
```
jrburke: Once the separator is known, and `newFromNames.join(separator)` is done, then I believe bodyL10n should be `{raw: newFromNames.join(separator)}` since it will already be localized.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=83a02ef232e74a84371a8f3706ba95dd5c9e70c9

PR #33103, opened 2015-11-10
Bug 1219678 - Impl number keyboard test suite in gij for keyboard app…

raylin:
1. rather than probing inputContext, test keyboard input type as gip did.
2. remove auto-correction api, will change to leverage keyboard setting panel in follow-up bug.
3. add an assertion to check the presentation of alphabet in number keyboard
4. remove duplicate code

Thanks

…, r=timdream


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2533db51a9517f0a367b95fd4039a6fae9077e14

apps/keyboard/test/marionette/lib/keyboard.js
```diff
@@ -97,28 +137,25 @@ Keyboard.prototype = {
...
-    var shiftKey = this.shiftKey;
+  getCurrentKeyboard: function() {
+    var activeFrame = this.activeKeyboardFrame;
 
-    shiftKey.tap();
-    this.client.waitFor(function() {
-      var expected = upperCase ? 'true' : 'false';
-      return (shiftKey.getAttribute('aria-pressed') === expected);
-    });
+    return activeFrame.getAttribute('data-frame-name');
```
timdream: Please change the comment here

https://github.com/mozilla-b2g/gaia/blob/master/apps/system/js/input_window.js#L35-L36

to be more descriptive like

> // Set the ID onto the element so Gij tests can probe it with Keyboard#getCurrentKeyboard.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=482d9a71948c970e8f0ab937503fab7709b47a14

PR #33110, opened 2015-11-10
Bug 1222741 - 'Delete thread' option is missing from the menu shown when viewing an SMS message

kumarrishav:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=18e4babaddff3f976d4887214ffdc7d2b8a65786

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1898,11 +1898,24 @@ var ConversationView = {
...
-    // If we are on a thread, we can call to SelectMessages
+    /*
+     * If we are on a thread,
+     * we can call Delete Thread, SelectMessages or Message Settings
+     */
     if (Navigation.isCurrentPanel('thread')) {
       params.items.push({
+        l10nId: 'delete-thread',
+        method: () => {
+          InboxView.delete([this.activeThread.id]).then(() => {
```
azasypkin: __issue:__ Please, don't introduce direct dependency to another view, we're going to split all these views in a separate documents and such dependencies make this harder to do. See https://github.com/mozilla-b2g/gaia/blob/master/apps/sms/views/conversation/js/startup.js#L54 where we shim such dependencies for the split views.

You likely need to generalize ```ConversationView.delete``` method to do the job.

Also delete confirmation text is not appropriate here - it will be ```Delete selected message thread?``` but we haven't selected anything :)

kumarrishav:
I also thought about this dependency before, but couldn't figure out. Thanks for  pointing this out.
BTW how to solve this, I lmmlwhat do you mean by making delete method more general. I guess ii w

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=18e4babaddff3f976d4887214ffdc7d2b8a65786

kumarrishav:
I guess if we don't want to rewrite same code, then need some solution to use existing method

kumarrishav:
Sorry for typo, was very sleepy while writing these comments :)

kumarrishav:
@azasypkin

apps/sms/views/conversation/js/conversation.js
azasypkin: > BTW how to solve this, I lmmlwhat do you mean by making delete method more general. I guess if we don't want to rewrite same code, then need some solution to use existing method

Can we make ```ConversationView.delete``` to work for this new case as well?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a280497b4508eb11cc98a73a018365b4e71692b7

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ea2fe0b109a0c91010012411ef5016f4b4720168

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8eccdb94544da07e557cc14762baa132a3757f49

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1910,9 +1910,15 @@ var ConversationView = {
     }
     params.items.push(subjectItem);
 
-    // If we are on a thread, we can call to SelectMessages
+    /*
+     * If we are on a thread,
+     * we can call Delete Thread, SelectMessages or Message Settings
+     */
```
azasypkin: __nit:__ use just ```//``` comment style and use all available space (up to 80 chars in a line)

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1910,9 +1910,15 @@ var ConversationView = {
...
     if (Navigation.isCurrentPanel('thread')) {
       params.items.push({
+        l10nId: 'delete-thread',
+        method: this.delete
```
azasypkin: __issue:__ I don't think it's going to work unless you bind ```delete``` to ```this``` context.

apps/sms/views/conversation/js/conversation.js
```diff
@@ -2001,12 +2007,28 @@ var ConversationView = {
   },
 
   delete: function conv_delete() {
```
azasypkin: __nit:__ I'd rather have something like ```deleteEntireThread``` boolean parameter for this delete method instead, you can easily pass it with ```this.delete.bind(this, true /* deleteEntireThread */)```

apps/sms/views/conversation/js/conversation.js
```diff
@@ -2014,13 +2036,25 @@ var ConversationView = {
...
+        MessageManager.getMessages({
+        // Filter and request all messages with this threadId
+          filter: { threadId: ConversationView.activeThread.id },
+          each: onThreadMessageRetrieved,
+          end: onAllThreadMessagesRetrieved
+        });
+      }
     }).then(() => {
-      ConversationView.cancelEdit();
+      if(ConversationView.inEditMode) {
```
azasypkin: __nit:__ ```cancelEdit``` method already does this check internally.

apps/sms/views/conversation/js/conversation.js
```diff
@@ -2014,13 +2036,25 @@ var ConversationView = {
...
+          (item) => +item
+        );
+        return onAllThreadMessagesRetrieved(messageIdsToDelete);
+      } else {
+        console.log('test1');
+        MessageManager.getMessages({
+        // Filter and request all messages with this threadId
+          filter: { threadId: ConversationView.activeThread.id },
+          each: onThreadMessageRetrieved,
+          end: onAllThreadMessagesRetrieved
```
azasypkin: Mmm, I think we should not call ```deleteMessages``` for the case when we delete entire thread - it does too much useless work, let's have ```deleteConversation``` method that will be called here instead.

And please use ```this.``` instead of ```ConversationView.```` where possible.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=804af59460261ef216121ee6ff1b67632d29946d

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2df69688d64dfb06b0bd52350977236b781fc001

apps/sms/views/conversation/js/conversation.js
kumarrishav: what do you mean by <code>deleteConversation</code> method exactly. Do we have any method like this already? or you mean to create a new method, if yes then what it will do differently. 
And what useless work 'deleteMessages' does ? @azasypkin

azasypkin: > what do you mean by deleteConversation method exactly. Do we have any method like this already?

No we don't, we have very similar one in inbox, so we should have very similar one in Conversation view as well.

>  or you mean to create a new method, if yes then what it will do differently. 

Yes, explained above - see the one in Inbox view - and just take what you need from it - I guess almost only retrieve all message Ids and call ```MessageManager.deleteMessages(messageIdsToDelete);``` - but you need to check and find out for sure :)

>  And what useless work 'deleteMessages' does?

Just take a look at ```ConversationView.deleteMessages``` and you'll see :) When you remove entire thread you __don't need__ to:
* Remove Dom node for every message - just show wait screen and then go back to inbox when deletion is finished;
* Check message list length to decide whether or not you should go Inbox - you know for sure!;
* Do very expensive ```lastDateGroup.querySelector('li:last-child')``` ;
* Do ```InboxView.updateThread``` - it will be removed anyway :)

So as you see almost 99% of useless work for the delete-entire-thread case :)

And when you see three ```if(ConversationView.inEditMode)``` in a single method - it's a strong sign that method should be split - the main rule here is that common staff stays in a single place, very specific staff is moved to separate methods - here you just should use your own judgment on whether it looks good or not - good exercise :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b174339ef880352d114c5b4de0fb8365d432d68b

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d825b7fd45f8b50762493a7a388a4013cee40550

apps/sms/services/js/message_manager.js
```diff
@@ -471,8 +471,9 @@ var MessageManager = {
   },
 
   /**
-   * Deletes messages using the message id provided.
-   * @param {number} id The ID of the message to be deleted.
+   * Deletes messages using the parameter provided.
+   * @param {number|array|object} id <long | MmsMessage | SmsMessage and
```
azasypkin: Maybe something like this would be better:
```
@param {number|SmsMessage|MmmsMessage|Array<number|SmsMessage|MmsMessage>} id
Message id, message instance or list of aforementioned entities used to identify message(s) to delete.
```

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1916,9 +1916,15 @@ var ConversationView = {
     }
     params.items.push(subjectItem);
 
-    // If we are on a thread, we can call to SelectMessages
+    /*
+     * If we are on a thread,
```
azasypkin: __nit:__ mm having ```//``` comment that occupies all available space - up to 80 chars - feels better

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=835dee65dbffdee7ae599a17c2918441c6fcb78c

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3684c65d3d4b295a8f57899ceb1f4187664c7f88

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7f7238b967951a3e4a9464c56259671eb76927e2

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=835dee65dbffdee7ae599a17c2918441c6fcb78c

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c1f78db4f38644f2c6d082fe168e8c18922215b6

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=835dee65dbffdee7ae599a17c2918441c6fcb78c

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2814a8a4aacfe417bc7c960deeb4b727c76a1df2

PR #33133, opened 2015-11-11
Bug 1223725 - apply preset eslint rules, r=julien

gasolin:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a2d06ed3c70f583ea5f3e144d385f0e4cd8e816f

.eslintrc
```diff
@@ -168,6 +174,7 @@
     },
     "env": {
         "browser": true,
-        "es6": true
+        "es6": true,
+        "node": true
```
julienw: I suspect you use "node" for the build system ? maybe it's better to use a dedicated .eslintrc in the build/ directory and in every directory that needs it ? My fear is that we may use in an app a variable name that's a global in node but not elsewhere, and eslint does not warn us...

.eslintrc
```diff
@@ -1,5 +1,8 @@
 {
     "parser": "babel-eslint",
+    "extends": [
+      "defaults/configurations/google"
```
julienw: why the google config over the default eslint config (https://github.com/eslint/eslint/blob/master/conf/eslint.json) ? Also I'm not especially convinced to use a default config, do you think it's better than using our own ?

gasolin: Hmm my previous impression is we use gjslint as well, so we do follow google's lint configs. I don't have strong opinion but I tend to pick a default and tweak a bit for our own needs.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=199ac16e9df01beec983f95b2cdb28a66806537b

.eslintrc
gasolin: seems fine to remove that :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=165705e8ba779a4c95fcb369f26d0b6cd913353d

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b982052f9358413facf60fafc57788095539a60f

.eslintrc
julienw: mmm you're right, we did use gjslint (we don't for a long time now but that's actually our defaults).

I'll have a deeper look on the rules then.

PR #33142, opened 2015-11-11
Bug 1221457 - [Messages]Apply gaia-fast-list for inbox list rendering

steveck-chung:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=07959bdd96b2702b88a18500b611701a1e10e20f

apps/sms/index.html
```diff
@@ -149,6 +156,33 @@ <h1 data-l10n-id="messages"></h1>
...
+          <gaia-fast-list id="threads-list" top="0" bottom="0">
+            <!--<template header>
+              <gaia-sub-header data-time="${daySection}">${section}</gaia-sub-header>
+            </template>-->
+            <template item>
+              <li class="${classList}" data-last-message-type="${type}" data-time="${timestamp}" data-thread-id="${id}" data-draftId="${draftId}" id="thread-${id}">
+                <label class="pack-checkbox negative">
+                  <input type="checkbox" value="${id}" data-mode="${mode}">
+                  <span></span>
+                </label>
```
wilsonpage: I think we can remove this checkbox stuff as we will be building 'edit mode' into gaia-fast-list.

apps/sms/tools/update_nga_libs.sh
```diff
@@ -28,6 +28,17 @@ sww_repo="gaia-components/serviceworkerware#master"
 sww_dist="$bower_path/serviceworkerware/dist/sww.js"
 sww_path="lib/sww.js"
 
+fast_list_repo="gaia-components/gaia-fast-list"
+fast_list_libs=(
+  "$bower_path/dom-scheduler/lib/dom-scheduler.js"
+  "$bower_path/fast-list/fast-list.js"
+  "$bower_path/gaia-fast-list/gaia-fast-list.js"
```
wilsonpage: `"$bower_path/gaia-fast-list/gaia-fast-list.js#master"`

wilsonpage: Make sure we depend on very latest version.

apps/sms/index.html
```diff
@@ -79,6 +79,13 @@
     <script defer src="shared/js/fb/fb_data_reader.js"></script>
     <script defer src="shared/js/fb/fb_reader_utils.js"></script>
     <script defer src="shared/js/event_dispatcher.js"></script>
+    <!-- Load gaia fast list libs -->
+    <script defer src="lib/gaia-fast-list/poplar.js"></script>
+    <script defer src="lib/gaia-fast-list/gaia-component.js"></script>
+    <script defer src="lib/gaia-fast-list/gaia-sub-headers.js"></script>
```
wilsonpage: `<script defer src="lib/gaia-fast-list/gaia-sub-header.js"></script>`

apps/sms/index.html
```diff
@@ -149,6 +156,33 @@ <h1 data-l10n-id="messages"></h1>
           <a id="threads-settings-button" data-icon="settings" data-l10n-id="settingsButton"></a>
         </gaia-header>
         <article id="threads-container" class="view-body edit-container" data-type="list">
+          <gaia-fast-list id="threads-list" top="0" bottom="0">
+            <!--<template header>
+              <gaia-sub-header data-time="${daySection}">${section}</gaia-sub-header>
+            </template>-->
+            <template item>
+              <li class="${classList}" data-last-message-type="${type}" data-time="${timestamp}" data-thread-id="${id}" data-draftId="${draftId}" id="thread-${id}">
```
wilsonpage: Remove `class` here as this is overriding our internal class.

apps/sms/views/inbox/js/inbox.js
```diff
@@ -730,47 +700,43 @@ var InboxView = {
...
     if (draft) {
       // Set the "draft" visual indication
-      li.classList.add('draft');
+      classList.push('draft');
 
       if (isDraft) {
-        li.dataset.draftId = draft.id;
-        li.classList.add('is-draft');
+        classList.push('is-draft');
         iconLabel = 'is-draft';
```
wilsonpage: Replace with localized string.

apps/sms/views/inbox/js/inbox.js
```diff
@@ -639,10 +605,10 @@ var InboxView = {
     }
 
     if (!empty) {
-      TimeHeaders.updateAll('header[data-time-update]');
+      // this.updateContactsInfo();
+      this.list.model = this.listData;
```
wilsonpage: This changes to `.setModel()` once you update to latest verison of gaia-fast-list.

apps/sms/index.html
```diff
@@ -149,6 +156,33 @@ <h1 data-l10n-id="messages"></h1>
...
+                <label class="pack-checkbox negative">
+                  <input type="checkbox" value="${id}" data-mode="${mode}">
+                  <span></span>
+                </label>
+                <a href="${hash}">
+                  <aside role="note" data-l10n-id="${iconLabel}" class="icon icon-unread icon-draft"></aside>
+                  <aside class="pack-end threadlist-item-picture"
+                         role="img" aria-label="Contact photo"
+                         data-l10n-id="contact-photo" data-show-contact="${showContact}" data-contact-photo="${photoUrl}">
+                    <span data-type="img" style="${bgImgStyle}">${titleLength}</span>
```
wilsonpage: Remove all this markup and replace with:

```html
<div class="image"><img/></div>
```
You can use the [image example](https://github.com/gaia-components/gaia-fast-list/tree/master/examples/images) as guidance. Note you have to define a [`getItemImageSrc()` function](https://github.com/gaia-components/gaia-fast-list/blob/master/examples/images/app.js#L5).

wilsonpage: If you don't have an image for a thread, just don't return anything, or return a default image. [Here is Francisco's WIP](https://github.com/arcturus/gaia/blob/contacts-fast-list/dev_apps/contacts/views/new_list/index.html#L32-L34) of the new NGA contacts app that uses initials instead of default image. I think this is a good solution :)

apps/sms/index.html
```diff
@@ -149,6 +156,33 @@ <h1 data-l10n-id="messages"></h1>
...
+                  </aside>
+                  <p class="threadlist-item-title"><bdi class="ellipsis-dir-fix">${title}</bdi></p>
```
wilsonpage: Replace this with `<h3>` and remove custom styling where possible.

apps/sms/index.html
```diff
@@ -149,6 +156,33 @@ <h1 data-l10n-id="messages"></h1>
...
+                  <p class="summary">
+                    <time>${timeTextContent}</time>
+                    <span class="mms-icon" data-l10n-id="mms-label">MMS</span>
```
wilsonpage: Prelocalize `data-l10n-id="mms-label"` and pass in localized string directly.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=fea602a00a6fc27dd6faf3bacb5fb757da597107

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=71d8a677396610b2ac9df70549ef5d9fae64d67f

apps/sms/views/inbox/js/inbox.js
```diff
@@ -2,9 +2,9 @@
 /* vim: set shiftwidth=2 tabstop=2 autoindent cindent expandtab: */
 
 /*global Template, Utils, Threads, Contacts, Threads,
-         WaitingScreen, MessageManager, TimeHeaders,
```
steveck-chung: I removed all the TimeHeader in the inbox, but I think we still need to update the time string in subheader in the end. Not sure if it's a good idea to set the list data to timeheader initially?

apps/sms/views/inbox/js/inbox.js
```diff
@@ -622,18 +551,24 @@ var InboxView = {
...
-    this.container.innerHTML = '';
-    this.renderDrafts();
+    // this.container.innerHTML = '';
+    this.listData = [];
+    this.list.configure({
+      getSectionName: function(item) {
+        return item.dayTextContent;
+      }
+    });
+    return this.prepareDrafts();
```
steveck-chung: I want to make sure the drafts data is ready before we start to render the thread, otherwise some edge case like draft is not ready while rendering data, and once the draft fetched it still could not be rendered because data is not inserted into list.

apps/sms/views/inbox/js/inbox.js
```diff
@@ -914,107 +794,57 @@ var InboxView = {
...
-    var threads = threadsContainer.getElementsByTagName('li');
-    var threadFound = false;
-    for (var i = 0, l = threads.length; i < l; i++) {
-      if (timestamp > threads[i].dataset.time) {
-        threadFound = true;
-        threadsContainer.insertBefore(node, threads[i]);
-        break;
+    return this.createThread(thread).then((data) => {
+      // Async update contact after insert to list
+      this.setContact(data);
```
steveck-chung: I think we could polish this part if we have list cached. Currently you will see the cached list rendered first, data without contacts will be rendered next, and the data with contacts rendered again at last. Maybe we can skip the data without contact rendering step once the cache existed.

apps/sms/index.html
```diff
@@ -148,6 +155,33 @@ <h1 data-l10n-id="messages"></h1>
           <a id="threads-settings-button" data-icon="settings" data-l10n-id="settingsButton"></a>
         </gaia-header>
         <article id="threads-container" class="view-body edit-container" data-type="list">
+          <gaia-fast-list id="threads-list" top="0" bottom="0" caching>
+            <!--<template header>
+              <gaia-sub-header data-time="${daySection}">${section}</gaia-sub-header>
+            </template>-->
+            <template item>
```
steveck-chung: Wilson said that we should not use data-l10n-id in the template and use formatValue for localization in script. But we'll need to relocalize for string inside data manually(and we'll need to keep the string id inside data as well). Any thought about this changes?

apps/sms/views/inbox/js/inbox.js
```diff
@@ -646,10 +581,9 @@ var InboxView = {
     }
 
     if (!empty) {
-      TimeHeaders.updateAll('header[data-time-update]');
+      this.list.setModel(this.listData);
```
azasypkin: __question:__ wondering if fast list has the ability to set some sort of "empty" template?

wilsonpage: No that up to you. If you're model is empty, you can render something else inside or behind the list.

apps/sms/views/inbox/js/inbox.js
```diff
@@ -674,13 +606,16 @@ var InboxView = {
...
-        this.emit('visually-loaded');
-      }
+      this.appendThread(thread).then(() => {
+        // Dispatch visually-loaded when rendered threads could fill up the top
+        // of the visible area.
+        if (--firstPanelCount === 0) {
+          this.list.setModel(this.listData);
+          this.emit('visually-loaded');
+        } else if (this.listData.length % this.FIRST_PANEL_THREAD_COUNT === 0) {
+          this.list.setModel(this.listData);
```
azasypkin: __question:__ does this ```if (this.listData.length % this.FIRST_PANEL_THREAD_COUNT === 0)``` change to render new items on page-by-page basis, not item-by-item?

apps/sms/views/inbox/js/inbox.js
```diff
@@ -806,29 +708,16 @@ var InboxView = {
     Threads.delete(threadId);
 
     // Cleanup the DOM
-    this.removeConversationDOM(threadId);
+    var index = this.listData.findIndex((data) => data.id === threadId);
+    if (index > -1) {
+      this.listData.splice(index, 1);
+      this.list.setModel(this.listData);
```
azasypkin: I see we do a lot of ```setModel``` calls, IMO it's going to be a pain for the developers, wondering if fast list or we should provide some sort of custom ```Datasource``` or ```ObservableList``` from RxJS wrapper around that, so that all these updates are done a bit more automagically?

```
interface IDataSource<T> {
  raw: T[];
  
  get(index: number): T;
  update(index: number, item: T);
  add(item: T);
  remove(index: number);
  clear();

  // Optional for keyed storage, like threadId
  getByKey(key: string): T;

  on(eventName: string, handler: function);
}
.....
let datasource = new DataSource();

datasource.on('change', Utils.throttle(() => this.list.setModel(datasource.raw)));

```

Maybe not now, but just throwing ideas :)

azasypkin: __question:__ also it's still not clear to me, how we're going to handle language change scenario - should we just iterate through entire datasource and re-localize everything and then re-set list model?

apps/sms/views/inbox/js/inbox.js
```diff
@@ -914,107 +794,57 @@ var InboxView = {
...
+    // before listData exists.
+    if (!this.listData) {
       return;
     }
-    // Retrieve all 'li' elements
-    var threads = this.container.getElementsByTagName('li');
-
-    [].forEach.call(threads, this.setContact.bind(this));
+    // Retrieve all the list data
+    this.listData.forEach(this.setContact);
```
azasypkin: It looks like after contact change we update model items, but list re-renders it after some time only when you try to scroll up and down :)

apps/sms/index.html
azasypkin: Wondering why list can't handle that, but anyway would be nice to put such limitations to the fast-list README.md :)

Regarding language-change, although it's quite inconvenient we can probably keep all data required for localization inside model and update everything on language change..... I don't really like it to be honest as I'd rather have data model without any visual-related things, but we can live with it if it's too hard to fix from fast-list side.

apps/sms/views/inbox/js/inbox.js
```diff
@@ -55,7 +46,7 @@ var InboxView = {
       'check-uncheck-all-button','composer-link',
       'delete-button', 'edit-header','options-button',
       'settings-button','edit-mode', 'edit-form',
-      'draft-saved-banner'
+      'draft-saved-banner', 'list'
```
azasypkin: __question:__ does fast list provide us with something like "item-click" event? So that we can get rid of ```node.dataset...``` logic in ```InboxView.handleEvent``` and work directly with model?

apps/sms/index.html
```diff
@@ -80,6 +80,13 @@
     <script defer src="shared/js/fb/fb_data_reader.js"></script>
```
azasypkin: I see ```W/Messages(16783): [JavaScript Warning: "Expected ':' but found '}'.  Declaration dropped." {file: "app://sms.gaiamobile.org/index.html" line: 0 column: 9 source: "undefined"}]``` in the logcat, wondering if you have any clue about this?

apps/sms/views/inbox/js/inbox.js
azasypkin: > Not sure if it's a good idea to set the list data to timeheader initially

mmm, not sure I got it, what do you mean?

apps/sms/index.html
wilsonpage: We can't use `data-l10n-id` as we need to strictly control all dom mutations while scrolling to optimize rendering performance. Do you suggest anything else?

apps/sms/views/inbox/js/inbox.js
wilsonpage: It's up to you to set a `'click'` delegate on the list and read any attributes you may have set to discern the model. The less fast-list does the better.

apps/sms/views/inbox/js/inbox.js
steveck-chung: I mean, we set  this.list/this.listData to TimeHeader and updataAll function could handle all the related string updating, including the fast list. But this behavior will be changed once we're in the split mode and each view could have different way to schedule the string update.

>  I doubt our app will be always visible and we should strive to real-time accuracy
I would be glad that we can have a smarter way to update the time related string instead of struggling in some edge cases. We already update the strings when visibility change, maybe we can just set the actual(but it's questionable that how accurate will it be if the duration is really long and app is always on the foreground that without resetting the timeout) timeout for the next necessary changes, not update the by every 10 sec.

apps/sms/views/inbox/js/inbox.js
steveck-chung: yes because I want to reduce the call of setModel for rendering. Maybe we can compare with  item-by-item rendering to see if calling setModel frequently does bring obvious side effects.

apps/sms/views/inbox/js/inbox.js
steveck-chung: yeah you're right, we'll need to re-localize data source and setModel again. I think Wilson already explained the reason why we don't use l10n-id in fast-list. So that's the overhead that we need to deal with once we confirm to land fast-list.
Sorry I'm not very familiar with RxJS, but the methods provide by datasource seems more powerful than native array, and the key point is you can centralize the data changes and using throttle to handle all the setModel/rendering. It would be more clean and readable. Maybe I can try to find some other native approach to reach this idea.

apps/sms/views/inbox/js/inbox.js
steveck-chung: Maybe I should call setModel here to force list rendered right away :)

apps/sms/index.html
steveck-chung: I have no idea about this warning :/ But I can try to find which script triggered this warning.

apps/sms/index.html
azasypkin: I don't really know how the list works internally, so it's useless to suggest anything :) But do you mean that you want to prevent l10n from mutating the DOM while scrolling or kind of "pause" l10n processing?

One silly idea was to prefix all l10n attributes before inserting element to the list, like ```defer-l10n-id``` and unprefix once list is in idle state starting from elements that are in the viewport - but not sure how ugly or slow it could be :)

apps/sms/views/inbox/js/inbox.js
azasypkin: I see, I was thinking that list items are rendered inside shadow DOM and couldn't/shouldn't be exposed to the consumer.... Looks like I just don't understand how it works :)

apps/sms/views/inbox/js/inbox.js
azasypkin: > Sorry I'm not very familiar with RxJS, but the methods provide by datasource seems more powerful than native array, and the key point is you can centralize the data changes and using throttle to handle all the setModel/rendering. It would be more clean and readable. Maybe I can try to find some other native approach to reach this idea.

Yeah, it just was rough idea inspired by RxJS, but I think we need single place to set model list model that can probably support throttling as well, otherwise we'll have to spread ```setModel``` calls across the entire view

apps/sms/index.html
wilsonpage: I don't understand what advantage this gives us. What's wrong with passing localized strings to the list to render?

wilsonpage: `fxos-fastlist` manually manipulates text-nodes in a single batch for the fastest possible render. No other script should be manipulating nodes under fastlist control. It it does, we cannot guarantee performance.

azasypkin: > I don't understand what advantage this gives us. What's wrong with passing localized strings to the list to render?

Mmm, I see at least two drawbacks of current way:

* Every app should track language change event and re-localize entire data set;
* And if data item contains several localizable strings, we'll have somewhat complex code to prepare data model because of async nature of ```document.l10n.formatValue```, e.g.
```
let rawData = [{ field1, field2 }, { field1, field2 }];

let localizedItemPromises = rawData.map((rawItem) => {
    // It's very naive example, in reality there is always some logic behind localization
    // for every field.
	return Promise.all([
		document.l10n.formatValue(rawItem.field1),
		document.l10n.formatValue(rawItem.field1)
	]).then([localizedField1, localizedField2] => {
		return { localizedField1, localizedField2 };
	});
});

Promise.all(localizedItemPromises).then((localizedItems) => {
	list.setModel(localizedItems);
});

+ Do the same job on language change, edge case but still
```

Or I'm missing something? IMO with such code we're loosing some of the benefits of l10n lib and do heavy lifting on our own :/

wilsonpage: As I said, it's a trade off, performance against developer ergonomics. You can try using `data-l10n-id` in your template, but it will trigger mutation observer callbacks whilst scrolling, which probably won't be good.

Not sure how else fxos-fastlist can help. Is there a new API that would make this easier for you?

azasypkin: @steveck-chung did you have chance to check if ```data-l10n-id``` embedded into template unacceptably affects performance (e.g. with heavy or extra-heavy workload)?

apps/sms/views/shared/style/edit-mode.css
```diff
@@ -16,7 +16,7 @@ form[role="dialog"][data-type="edit"] > menu {
 /* Add a margin at the end of the last thread or message node to avoid overlay
  * with the edit mode menu */
 .edit .edit-container {
-  padding-bottom: 7rem;
+  padding-bottom: 5.5rem;
```
julienw: this is fixing https://bugzilla.mozilla.org/show_bug.cgi?id=1128337 ?

apps/sms/index.html
julienw: I don't see this

apps/sms/index.html
julienw: Normally l20n.js handles data-l10n-id in a Mutation Observer, which should happen before any reflow; however it's right that this could bring frame skips.

I think we should start using them, and see how this looks...

apps/sms/locales/sms.en-US.properties
```diff
@@ -295,9 +295,9 @@ send-button-sim-indication.ariaLabel = using SIM {{n}}
 # Labels for screen reader
 attach-button.ariaLabel = Add attachment
 contact-photo.ariaLabel = Contact photo
-unread-thread.ariaLabel = Unread
-has-draft.ariaLabel = Has draft
-is-draft.ariaLabel = Draft
+unread-thread = Unread
+has-draft = Has draft
+is-draft = Draft
```
julienw: I'd suggest to keep "ariaLabel" in the name; did you check if `formatValue("unread-thread.ariaLabel")` was working ? If not maybe we can ask this to our l20n colleagues?

apps/sms/tools/update_nga_libs.sh
```diff
@@ -21,6 +21,17 @@ sww_repo="gaia-components/serviceworkerware#master"
 sww_dist="$bower_path/serviceworkerware/dist/sww.js"
 sww_path="lib/sww.js"
 
+fast_list_repo="gaia-components/gaia-fast-list#master"
```
julienw: nit: it's `fxos-` instead of `gaia-` now. Same below.

apps/sms/tools/update_nga_libs.sh
```diff
@@ -44,3 +55,9 @@ cp -r "$bridge_plugins" "$bridge_destination"
 
 # ServiceWorkerWare lib preparing.
 cp "$sww_dist" "$sww_path"
+
+# Copy all required fast list libs.
+for lib in "${fast_list_libs[@]}"
+do
+  cp "$lib" "$fast_list_destination"
+done
```
julienw: nit: just use `cp "${fast_lib_libs[@]}" "$fast_list_destination"`

apps/sms/views/inbox/js/inbox.js
julienw: anyway we could be smarter about "repeat" time headers; currently we refresh them every minute, just to check whether we changed the day... Likely we should do this only for the relative and day headers (today, yesterday, monday, etc) and check only once a day.

apps/sms/views/inbox/js/inbox.js
julienw: if drafts are ready after the list started to render, isn't it possible to update the list and ask for a rerender ?

apps/sms/views/inbox/js/inbox.js
julienw: we can ask @etiennesegonzac ;)

apps/sms/views/inbox/js/inbox.js
```diff
@@ -674,13 +606,16 @@ var InboxView = {
...
-      // the visible area.
-      if (--firstPanelCount === 0) {
-        this.emit('visually-loaded');
-      }
+      this.appendThread(thread).then(() => {
+        // Dispatch visually-loaded when rendered threads could fill up the top
+        // of the visible area.
+        if (--firstPanelCount === 0) {
+          this.list.setModel(this.listData);
+          this.emit('visually-loaded');
```
julienw: I think we should somehow emit after the render is really done

apps/sms/views/inbox/js/inbox.js
```diff
@@ -697,106 +632,73 @@ var InboxView = {
...
-    // If the draft is newer than the message, update
-    // the body with the draft content's first string.
-    if (draft && draft.timestamp >= record.timestamp) {
-      body = draft.content.find((content) => typeof content === 'string');
-      type = draft.type;
-    }
+    var labelId = '';
 
-    node.dataset.lastMessageType = type;
+    bodyHTML = Template.escape(bodyHTML || '');
```
julienw: nit: I think I removed this in a previous patch :)

apps/sms/views/inbox/js/inbox.js
julienw: If we want a more clever fast list, we should just the library instead of the component: https://github.com/fxos-components/fastlist

apps/sms/views/inbox/js/inbox.js
```diff
@@ -697,106 +632,73 @@ var InboxView = {
...
+      bodyHTML: bodyHTML,
+      timestamp: String(timestamp),
+      timeTextContent: Utils.getFormattedHour(timestamp),
+      daySection: daySection,
+      dayTextContent: results[0],
+      iconLabel: results[1],
+      status: labelId,
+      mmsTextContent: results[2]
+    }));
   },
```
julienw: nit: I think a lot of the patch in bug 1206727 was reverted here, can you look closer when you'll rebase ?

apps/sms/views/inbox/js/inbox.js
```diff
@@ -697,106 +632,73 @@ var InboxView = {
...
+    var daySection = Utils.getDayDate(timestamp);
 
     var thread = Threads.get(id);
     var isDraft = thread.isDraft;
     var draft = thread.getDraft();
 
-    var body = record.body;
-    var type = record.lastMessageType;
+    var bodyHTML = !draft ? record.body : draft.content.toString();
+    var type = !draft ? record.lastMessageType : draft.type;
```
julienw: nit: reverse the condition

apps/sms/views/inbox/js/inbox.js
```diff
@@ -914,107 +794,57 @@ var InboxView = {
...
-
-    if (!threadFound) {
-      threadsContainer.appendChild(node);
-    }
 
-    if (this.inEditMode) {
-      // Remove the new added thread id from the selection handler
-      this.selectionHandler.unselect(thread.id);
+      if (!threadFound) {
+        this.listData.unshift(data);
```
julienw: isn't it "push" instead ? (not sure)

apps/sms/views/inbox/js/inbox.js
```diff
@@ -175,92 +153,73 @@ var InboxView = {
...
-      var groupTitle = document.createElement('span');
-      var separatorNode = document.createElement('span');
-      separatorNode.setAttribute(
-        'data-l10n-id',
-        'thread-participant-separator'
-      );
-
-      picture.firstElementChild.textContent = numbers.length;
-      picture.classList.add('has-picture', 'group-picture');
+      data.titleLength = numbers.length;
```
julienw: not sure about this

julienw: but I don't really know what titleLength does :)

PR #33147, opened 2015-11-11
Bug 1220311 - Use webapps-manage so that IAC works correctly. r=kgrandon

Cwiiis:
When handling an IAC message, the home screen will be launched as a
regular app and not as a home screen. Because of this, the
homescreen-webapps-manage permission isn't granted and the app fails to
launch with an exception caused by this. Revert to using the old
webapps-manage permission for now.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a05fd7e2bc4ebecaa7cf09800d8c1887b519472d

PR #33151, opened 2015-11-11
Bug 1223931 - Added capability to pair to BT device from test

JonHylands:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=28d27757faf076182b4fd5a484491a1ad4ad4b24

tests/python/gaia-ui-tests/gaiatest/apps/settings/regions/bluetooth.py
```diff
@@ -83,3 +86,16 @@ def tap_update_device_name_ok(self):
         Wait(self.marionette).until(expected.element_displayed(element))
         element.tap()
         Wait(self.marionette).until(expected.element_not_displayed(element))
+
+    def tap_search_for_devices(self):
+        self.root_element.find_element(*self._search_for_devices_locator).tap()
```
npark-mozilla: On this method, it would be better if we wait until the search is completed.
If we wait until the Search for device becomes greyed out, then available again, then we can be sure by the time the method is completed, the searching has completed.

And in that case, the title of the method can be search_for_devices(self): since we're doing the waiting as well.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4c9489a008473d9d931b23b6f73a80703e9c5524

PR #33156, opened 2015-11-11
Bug 1224003 - [Gallery] Use IntlHelper to create date time formatter

punamdahiya:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a14fd99972cb9a5613858fa13d2f5323d56b32cd

apps/gallery/js/thumbnail_item.js
```diff
@@ -32,25 +33,16 @@ function ThumbnailItem(fileData) {
...
 ThumbnailItem.prototype.localize = function() {
   var date = new Date(this.data.date);
+  var formatter = IntlHelper.get('time-stamp');
 
   var descId = !this.data.metadata.video ?
     'imageDated' : 'videoDated';
 
   navigator.mozL10n.setAttributes(this.imgNode, descId, {
-    timeStamp: ThumbnailItem.formatter.format(date)
+    timeStamp: formatter.format(date)
```
zbraniecki: it doesn't seem that you are doing `IntlHelper.observer('time-stamp')` anywhere. How will this code react to this intl object being modified? (locale change, timeformat change, moztimechange are all covered by IntlHelper.observe)

apps/gallery/js/thumbnail_date_group.js
```diff
@@ -59,7 +60,8 @@ function ThumbnailDateGroup(item) {
   this.htmlNode.appendChild(this.header);
   this.htmlNode.appendChild(this.container);
 
-  this.localize();
+  this.header.textContent =
+    IntlHelper.get('date-group').format(new Date(this.date));
```
davidflanagan: Why can't you just call localize() here?  The thumbnails array is empty here in the constructor, so localize() will just do the same as this code, won't it?  I think it would be cleaner to do all localization in the localize method

apps/gallery/js/startup.js
```diff
@@ -297,6 +297,7 @@
       // fullyLoaded marker until we're completely stable and have
       // finished scanning.
       window.performance.mark('mediaEnumerated');
+      navigator.mozL10n.once(Thumbnails.list.localize.bind(Thumbnails.list));
```
davidflanagan: Maybe do the observer() call here after the initial localization pass?

apps/gallery/js/thumbnails.js
```diff
@@ -1,12 +1,25 @@
...
+  });
+
+  IntlHelper.define('time-stamp', 'datetime', {
+    hour: 'numeric',
+    minute: 'numeric',
+    month: 'long',
+    day: 'numeric',
+    year: 'numeric'
+  });
+
```
davidflanagan: Since date-group is used by thumbnail_group.js and time-stamp is used by thumbnail_item.js maybe it would be better to manage these in those files rather than here?

apps/gallery/js/thumbnail_item.js
punamdahiya: That’s right, I am not explicitly specifying IntlHelper.observer('time-stamp’) , which I believe we don’t need as this code is invoked via listener on 'date-group'

Here’s the flow in detail:

IntlHelper data-group observe call below is calling localize method of thumbnailList 
https://github.com/mozilla-b2g/gaia/pull/33156/files#diff-598a8e9249c5971f9c555c58eb7eac6aR35

which iterates and call each ThumbnailDateGroup object localize method
https://github.com/mozilla-b2g/gaia/blob/master/apps/gallery/js/thumbnail_list.js#L112

Inside each ThumbnailDateGroup object localize method we are calling localize on all thumbnail item/ thumbnail.
https://github.com/mozilla-b2g/gaia/pull/33156/files#diff-f597ef3c1ed0182ae894ea21db37491aR142

So here we are using a single listener on date-group to format two formatter objects ‘date-group’ and ’time-stamp’ and it’s appears to be working :) . Do you foresee any issues in using a single observe call here?

Thanks!

PR #33173, opened 2015-11-13
Bug 1224124 - [Camera]The 'Cancel' button on the action menu does not appear as an oval button.

prateekjadhwani:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5e313dd76a8620478e1ce918c8bdb766aed58ab6

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=087586407b121c06b560e917d6189fea931b1ca0

apps/camera/locales/camera.en-US.properties
```diff
@@ -21,6 +21,7 @@ close-button  = Close
 
 delete-photo? = Delete photo?
 delete-video? = Delete video?
+gaia-menu-cancel = Cancel
```
KevinGrandon: Since this is a blocker, it would be better to include the localization file from shared/ so we don't have late l10n changes here. You can see how to do this here: https://github.com/mozilla-b2g/gaia/blob/master/apps/video/index.html#L11

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=52c37224651d5bd71392790099c2ef9968d52eac

apps/camera/locales/camera.en-US.properties
prateekjadhwani: Thanks @KevinGrandon

apps/camera/index.html
```diff
@@ -10,7 +10,10 @@
     <link rel="localization" href="locales/camera.{locale}.properties">
     <link rel="localization" href="/shared/locales/date/date.{locale}.properties">
     <link rel="localization" href="/shared/locales/media/media.{locale}.properties">
+    <link rel="localization" href="/shared/elements/gaia_menu/locales/gaia_menu.{locale}.properties">
     <script defer src="/shared/js/l10n.js"></script>
+    <script defer src="shared/js/component_utils.js"></script>
+    <script defer src="shared/elements/gaia_menu/script.js"></script>
```
davidflanagan: The camera app uses Bower to manage its dependencies on web components, so you'll need to edit bower.json to import the upstream gaia_menu component rather than using the one in shared. I'd be somewhat surprised if component_utils is not already included somehow.

Also, we're very sensitive to the startup time of our apps.  Since there is no menu displayed when the app starts up, we'd need this file to be loaded when it is needed rather than loaded at startup.  The camera app uses requirejs to do its loading.

apps/camera/js/views/preview-gallery.js
```diff
@@ -334,43 +334,34 @@ return View.extend({
...
+    return '<gaia-menu id="options-view" class="js-menu" hidden>' +
       '<header data-l10n-id="options">Options</header>' +
-      '<menu class="js-menu">' +
-      '<button class="js-btn" name="gallery" data-l10n-id="open-gallery">' +
+      '<button class="js-gallery-btn" name="gallery" ' +
+        'data-l10n-id="open-gallery">' +
         'Open Gallery' +
       '</button>' +
-      '<button class="js-btn" name="delete" data-l10n-id="delete">' +
+      '<button class="js-delete-btn" name="delete" data-l10n-id="delete">' +
```
davidflanagan: I don't understand why you are defining new CSS classes here. Where are js-gallery-btn and js-delete-btn defined? Why do you need different classes for buttons that should have identitcal presentation?  This seems like a mistake to me, but maybe I'm missing something here.

apps/camera/style/controls.css
```diff
@@ -421,4 +421,4 @@
 .controls.mode-video.capture-active .capture-button .center:before {
   border-radius: 3px;
   background-color: white;
-}
+}
```
davidflanagan: This is an unnecessary change and makes the patch seem larger that it actually is.  Please add the newline back

PR #33191, opened 2015-11-13
Bug 1224547 - [Aries] test_gallery_view.py fails: AssertionError: 640…

JohanLorenzo:
… != 590

### DIFF
--- a/tests/python/gaia-ui-tests/gaiatest/apps/system/app.py
+++ b/tests/python/gaia-ui-tests/gaiatest/apps/system/app.py
@@ -101,6 +101,16 @@ class System(Base):
         Wait(self.marionette).until(lambda m: element.location['y'] == (0 - element.size['height']))
 
     @property
+    def screen_width(self):
+        self.marionette.switch_to_frame()
+        return self.marionette.find_element(*self._screen_locator).rect['width']
+
+    @property
+    def screen_height_without_software_home_button(self):
+        self.marionette.switch_to_frame()
+        return self.marionette.find_element(*self._screen_locator).rect['height'] - self.software_buttons_height
+
+    @property
     def software_buttons_height(self):
         """
         Gets the height of the software buttons container on the screen.
--- a/tests/python/gaia-ui-tests/gaiatest/apps/gallery/regions/fullscreen_image.py
+++ b/tests/python/gaia-ui-tests/gaiatest/apps/gallery/regions/fullscreen_image.py
@@ -113,7 +113,7 @@ class FullscreenImage(Base):
 
     @property
     def photo_toolbar_width(self):
-        return self.marionette.execute_script('return document.getElementById("fullscreen-toolbar").offsetWidth')
+        return self.marionette.find_element(*self._photos_toolbar_locator).rect['width']
 
     @property
     def photo_toolbar_options(self):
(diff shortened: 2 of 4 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d509d4db55adf21d9be00d92ebfd2e4dcef421e0

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f201bf242571ac23a0414ece74b6b705532b80ca

tests/python/gaia-ui-tests/gaiatest/apps/system/app.py
```diff
@@ -101,6 +101,16 @@ def wait_for_app_update_to_clear(self):
         Wait(self.marionette).until(lambda m: element.location['y'] == (0 - element.size['height']))
 
     @property
+    def screen_width(self):
+        self.marionette.switch_to_frame()
+        return self.marionette.find_element(*self._screen_locator).rect['width']
+
```
npark-mozilla: Wouldn't it be better if we return to the displayed app frame once we're done with finding the element?  This method can be called at any time, and it'd be better if the user doesn't have to explicitly return to the displayed app frame after calling this method.

tests/python/gaia-ui-tests/gaiatest/apps/system/app.py
```diff
@@ -101,6 +101,16 @@ def wait_for_app_update_to_clear(self):
...
+    @property
+    def screen_height_without_software_home_button(self):
+        self.marionette.switch_to_frame()
+        return self.marionette.find_element(*self._screen_locator).rect['height'] - self.software_buttons_height
+
```
npark-mozilla: same as above

tests/python/gaia-ui-tests/gaiatest/apps/system/app.py
JohanLorenzo: I agree. However, no method in this Page class does the switch back. I'd prefer to handle this part in a follow up.

npark-mozilla: huh, you're right, yeah that's a better idea.

mwargers: I guess screen_width also could be done with screen.width * window.devicePixelRatio.
Then the switch_to_frame call is not necessarily needed, perhaps.

tests/python/gaia-ui-tests/gaiatest/apps/system/app.py
```diff
@@ -101,6 +101,16 @@ def wait_for_app_update_to_clear(self):
         Wait(self.marionette).until(lambda m: element.location['y'] == (0 - element.size['height']))
 
     @property
+    def screen_width(self):
```
mwargers: I guess this could be called height as well, since it's the height of the system app, but I guess it doesn't matter.

tests/python/gaia-ui-tests/gaiatest/apps/system/app.py
JohanLorenzo: Well, nearly every method in this class calls `switch_to_frame()`, so I'd prefer to keep the consistency. What do you think @mwargers ?

PR #33201, opened 2015-11-14
Bug 1224395 - Back button arrow has no functionality in Message Settings or in Homescreen Settings r=gasolin

andrenatal:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=07500c773ff38367d5dee370e050af48f22d8886

apps/settings/manifest.webapp
```diff
@@ -44,7 +44,8 @@
         "target": "device"
        },
       "href": "index.html#activity-configure",
-      "disposition": "inline"
+      "disposition": "inline",
+      "returnValue": true
```
julienw: Hey Andre, Fred, I just wonder if it could be more right to use window.close instead of postResult ? Because the "configure" activity is not really returning anything...

andrenatal: Sorry @julienw, but I read this message after flag the patch as checkin-needed. Yes, maybe we can change the calls. Do you know where's the right place postResult is called?

julienw: I see `js/activity_handler.js` has a function `postResult` that could maybe be changed ? It seems to be the only place that calls postResult. It can take a "result" but I see it's always called either without a parameter or with `null`.

So a "right" patch could be to rename this `finishActivity` and change all callers and call window.close instead of activity.postResult. But a simple patch could be to simply add `window.close()` after the `this._currentActivity.postResult` call

PR #33220, opened 2015-11-16
Bug 1219704 - Implement *test_play_ogg_video.py* as an integration te…

russnicoletti:
…st in JavaScript


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f2396d68147a4bac6e54229463f8aa9a35e5648f

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e4286918ccdbdd26773c870ba7066153df414c09

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=032e46eb1d8d65ec4618a1e7fd895eca1e5e1c1d

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=201d86a80315c34391ab8d8a216b4f25219c7a18

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a144e607423846ceb7f1998db11d66c9ebb68b97

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7d987484064b7ae34ca4cc0bc899b2a06046fcc0

apps/video/test/marionette/manifest.ini
```diff
@@ -1,2 +1,3 @@
 [video_list_test.js]
 [overlay_test.js]
+[play_multiple_types_test.js]
```
davidflanagan: This file is actually unused and was just removed by bug 1225552 a couple of days ago. That's presumably why the patch is bitrotted.

apps/video/test/marionette/play_multiple_types_test.js
```diff
@@ -0,0 +1,65 @@
...
+    app = new Video(client);
+  });
+
+  function verifyVideo(type, done) {
+    var title = 'VID_0001';
+    var filename = title + '.' + type;
+
+    client.fileManager.add({
+      type: 'videos',
+      filePath: 'test_media/Movies/' + filename
```
davidflanagan: the test_media/ directory is not really owned by anyone. I'd prefer it if you put all of the videos that these tests depend on inside the video app directory, so that they don't get deleted or renamed by someone who doesn't know what they are there for.  I suggest creating a new directory apps/video/test/videos/ and putting the new files in that.

apps/video/test/marionette/play_multiple_types_test.js
```diff
@@ -0,0 +1,65 @@
...
+
+  setup(function() {
+    // Remove all files in temp device storage.
+    client.fileManager.removeAllFiles();
+    selectors = Video.Selector;
+    app = new Video(client);
+  });
+
+  function verifyVideo(type, done) {
+    var title = 'VID_0001';
```
davidflanagan: How about changing this and the filenames to "test" or "sample".  That might make it more obvious what the files are for.

apps/video/test/marionette/lib/video.js
```diff
@@ -130,14 +133,30 @@ Video.prototype = {
...
 
   /**
    * @return {Marionette.Element} video player element.
    */
   get player() {
-    return this.client.findElement(Video.Selector.player);
+    return this.client.helper.waitForElement(Video.Selector.player);
+  },
+
+  showControls: function() {
```
davidflanagan: I wonder about this method name. If you call it a second time it will actually hide the controls, right?  I think you should either make this something like ensureControlsShowing() . Or you should just get rid of the method and just call app.player.tap() directly in the test.

apps/video/test/marionette/lib/video.js
```diff
@@ -130,14 +133,30 @@ Video.prototype = {
...
+    this.player.tap();
```
davidflanagan: Should you do a waitForElement() here to make sure that the controls are visible before you return?  That would only make sense if you were able to first make sure that they were not already visible before you called tap.

apps/video/test/marionette/play_multiple_types_test.js
```diff
@@ -0,0 +1,65 @@
...
+    // To avoid timing issues with the controls
+    // being auto-hidden when the video is first
+    // played, wait for controls to be hidden, then
+    // show them, by tapping on the video player.
+    setTimeout(function() {
+      app.showControls();
+      // Verify the video has begun playing and the
+      // title has been read properly.
+      var elapsedTime = app.elapsedTime();
+      assert(elapsedTime > '00:00');
```
davidflanagan: This assertion will pass even if the player is displaying 'undefined' or "NaN" instead of an actual time.
I think you want something like: `parseInt(elapsedTime.split(':').pop()) > 0`

apps/video/test/marionette/play_multiple_types_test.js
```diff
@@ -0,0 +1,65 @@
...
+    app.tapThumbnail(0);
+    // To avoid timing issues with the controls
+    // being auto-hidden when the video is first
+    // played, wait for controls to be hidden, then
+    // show them, by tapping on the video player.
+    setTimeout(function() {
+      app.showControls();
+      // Verify the video has begun playing and the
+      // title has been read properly.
+      var elapsedTime = app.elapsedTime();
```
davidflanagan: You have only waited 500ms at this point. If everything runs really quickly, the elapsed time might still be 00:00 at this point.

apps/video/test/marionette/lib/video.js
davidflanagan: Okay, reading the docs, it looks like you can do something like
```
var elapsed = this.client.findElement(Video.Selector.elapsedTime)
if (!elapsed.displayed()) {
   this.player.tap();
   this.client.waitFor(function() { return elapsed.displayed(); }
}
```

I do not understand how marionette works, so I'm not confident that that function does what I think it does, though.

apps/video/test/marionette/lib/video.js
```diff
@@ -130,14 +133,30 @@ Video.prototype = {
...
+  },
+
+  elapsedTime: function() {
+    var timeElem =
+      this.client.helper.waitForElement(Video.Selector.elapsedTime);
```
davidflanagan: If it is not displayed, this will block forever, I think. Maybe better to use findElement here?

I suppose that by doing this, it means that you don't need to have a wait after you call tap(). It would be cleaner to put the tap() in the ensureControlsShowing() function, though.

apps/video/test/marionette/lib/video.js
```diff
@@ -130,14 +133,30 @@ Video.prototype = {
...
+    return timeElem.text();
+  },
+
+  videoTitle: function() {
+    var titleElem =
+      this.client.helper.waitForElement(Video.Selector.videoTitle);
```
davidflanagan: Same issue here. Should you just use findElement()?

apps/video/test/marionette/play_multiple_types_test.js
```diff
@@ -0,0 +1,65 @@
...
+      type: 'videos',
+      filePath: 'test_media/Movies/' + filename
+    });
+    app.launch();
+    app.tapThumbnail(0);
+    // To avoid timing issues with the controls
+    // being auto-hidden when the video is first
+    // played, wait for controls to be hidden, then
+    // show them, by tapping on the video player.
+    setTimeout(function() {
```
davidflanagan: Instead of setTimeout, I think you can do `client.helper.waitForElementToDisappear(selectors.elapsedTime)`

apps/video/test/marionette/play_multiple_types_test.js
```diff
@@ -0,0 +1,65 @@
...
+  function verifyVideo(type, done) {
+    var title = 'VID_0001';
+    var filename = title + '.' + type;
+
+    client.fileManager.add({
+      type: 'videos',
+      filePath: 'test_media/Movies/' + filename
+    });
+    app.launch();
+    app.tapThumbnail(0);
```
davidflanagan: I don't actually understand why this works. The launch() method does not wait for metadata parsing to complete, so when launch() returns there might not be any thumbnails displayed.  tapThumbnail() does a findElements() to find all thumbnails.Does that method wait until at least one match is found? I suppose it probably does. But notice that this code would not work reliably if you tried tapThumbnail(1). Because findElements() might return with an array that only had the first thumbnail.

Anyway, my point here is that it might be nice to have an explicit test here to see if there are any thumbnails. If metadata parsing fails, this line will just timeout after 20 seconds, I think. You'll get a test failure but not an informative error message.   If we're not able to play the video then we're probably not able to create a thumbnail for it.  If you try running the test with a corrupt or unsupported video file, it will probably fail here with a timeout.  

So it would be nice if there was some other way to wait for metadata parsing to be complete and then check to see how many thumbnails were created without blocking.

If you don't want to fix all this, how about just simplifying and commenting:

```
// Wait for a thumbnail to appear and then tap it.
// Note that if this test fails it will probably fail with a timeout on this call
app.thumbnail.tap();
```

apps/video/test/marionette/play_multiple_types_test.js
```diff
@@ -0,0 +1,65 @@
...
+  });
+  test('play ogg video', function(done) {
+    verifyVideo('ogg', done);
+  });
+  test('play ogv video', function(done) {
+    verifyVideo('ogv', done);
+  });
+  test('play webm video', function(done) {
+    verifyVideo('webm', done);
+  });
```
davidflanagan: If you can solve the waiting for thumbnails to appear issue described above, then you could refactor this to run one test with four videos and four thumbnails.  Does marionette do a full reboot between each test?  If you can play four videos in one test that is probably much faster than four separate tests. But I'm only suggesting you do that if it is easy and robust.

apps/video/test/marionette/play_multiple_types_test.js
davidflanagan: Maybe you should check the elapsedTime() value, test that it is in the mm:ss format you expect, and then use a client.waitFor() call to wait for the second value to become larger.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=dc182b518e0fa6f9add27ec33a42bd48cf10ab06

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9caf463e3aa024f7856d4e23d50e2b28588f8cb5

apps/video/test/marionette/lib/video.js
```diff
@@ -130,14 +134,79 @@ Video.prototype = {
...
+
+    // Ensure when we tap on the player view the video controls are not
+    // displayed -- they are displayed when the video first plays and then
+    // they are automatically hidden.
+    //
+    // Wait for controls to appear as the video begins to play. Specify
+    // a timeout to account for the controls having already been displayed
+    // and auto-hidden.
+    this.client.waitFor(function()
+      { return elapsedElem.displayed(); },
```
davidflanagan: nit: This is awkward formatting because the function body looks just like the object literal below it.  If you can fit this on the line above, I think that would be best.  Or at least put the { on the line above

apps/video/test/marionette/lib/video.js
```diff
@@ -130,14 +134,79 @@ Video.prototype = {
...
+    // Tap to display the controls and wait for them to be displayed.
+    this.player.tap();
+    this.client.waitFor(function() { return elapsedElem.displayed(); });
+
+    // Wait for the video to play for at least one second (to allow
+    // for validation checks).
+    var getElapsedTimeSecondsFunc = this.getElapsedTimeSeconds.bind(this);
+    this.client.waitFor(function() {
+      return getElapsedTimeSecondsFunc() > 0;
+    });
```
davidflanagan: It seems a little odd to me to put this wait here in the waitForControls() function rather than in the test itself.

Can we use arrow functions in marionette tests yet? If so, this seems like a perfect place to do so because then you could just use this.getElapsedTimeSeconds() in the function.

apps/video/test/marionette/lib/video.js
```diff
@@ -130,14 +134,79 @@ Video.prototype = {
...
+  },
+
+  /**
+   * Taps 'back' button on gaia-header to return from the video player view
+   * to the thumbnails view.
+   */
+  back: function() {
+    var playerHeader = this.client.findElement(Video.Selector.playerHeader);
+    this.client.switchToShadowRoot(playerHeader);
+    var actionButton = this.client.helper.waitForElement('.action-button');
```
davidflanagan: nit: should this selector be in Video.Selector?

davidflanagan: maybe it shouldn't. Its almost like there should be a gaia-header library for tests or something.  But this code is probably fine as it is for now.

apps/video/test/marionette/lib/video.js
```diff
@@ -147,6 +216,23 @@ Video.prototype = {
...
+  get thumbnails() {
+    return this.client.findElements(Video.Selector.thumbnail);
+  },
+
+  waitForThumbnails: function(num) {
+    var thumbnails = this.thumbnails;
+
+    this.client.helper.waitFor(function() {
+      return thumbnails.length === num;
+    });
```
davidflanagan: You query the thumbnails once, get an array of them, and then wait for the length of that array to change.  As far as I can tell, thumbnails.length is never going to change. If it isn't === num before you call waitFor, I can't see how it is going to change while you wait. Maybe I don't understand marionette, but it seems to me that you need to be calling findElements() inside of the waitFor function.

apps/video/test/marionette/play_multiple_types_test.js
```diff
@@ -0,0 +1,69 @@
...
+        'device.storage.testing': true,
+        'device.storage.prompt.testing': true
+      }
+    }
+  });
+
+  setup(function() {
+    // Remove all files in temp device storage.
+    client.fileManager.removeAllFiles();
+    selectors = Video.Selector;
```
davidflanagan: This is unused

apps/video/test/marionette/play_multiple_types_test.js
```diff
@@ -0,0 +1,69 @@
...
+    });
+
+    app.launch();
+    app.waitForThumbnails(types.length);
+    var thumbnails = app.getThumbnails();
+
+    for (var i = 0; i < thumbnails.length; i++) {
+
+      app.tapThumbnail(i);
+      app.waitForControls();
```
davidflanagan: Earlier I'd said I thought it was weird to have the code that waits for the time to be updated as part of waitForControls(). But now I realize that you could just rename that function to app.waitForVideoToStartPlaying() and it would be fine. The whole waiting for the controls thing is just an implementation detail so you can look at the elapsed time, so how about just renaming this?

apps/video/test/marionette/play_multiple_types_test.js
```diff
@@ -0,0 +1,69 @@
...
+      var expectedTitle = 'Test ' + types[i] + ' video';
+      assert(app.videoTitle(), expectedTitle);
+
+      // Go back to the thumbnail view.
+      app.back();
+
+      // Scroll the thumbnails a bit after each video is played
+      // to ensure the thumbnails are displayed when they are
+      // 'tapped'.
+      actions.flick(thumbnails[i], 0, 0, 0, -30).perform();
```
davidflanagan: Yuck! Too bad this is necessary, but I'm glad you figured out a workaround

apps/video/test/marionette/lib/video.js
russnicoletti: The function is getting the thumbnails element. The element may or may not have been fully constructed when we first get the reference to it. The metadata loading process will add list items to the element so 'thumbnails.length' will change if the metadata loading process is still active after findElement is invoked.

apps/video/test/marionette/play_multiple_types_test.js
russnicoletti: I contemplated other solutions but this seemed the most straightforward, and it was oddly rewarding to learn how to scroll using marionette. I say 'oddly' because I find the whole UI testing -- manipulating a UI remotely -- to be kind of magical.

apps/video/test/marionette/play_multiple_types_test.js
russnicoletti: I like the idea to leave the function as it is and to rename it as you suggest. Thanks for that suggestion.

apps/video/test/marionette/lib/video.js
russnicoletti: I've moved the selector into Video.Selector; I find that better than leaving it hard-coded. If in the future there is a gaia-header test library, we can remove the selector and the 'back' function.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f3bcd6e58897ab3654aa40a9db9ffc756985a6ca

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=61ff1fdc4f6f2987bac496ab0c918f569f68dde8

apps/video/test/marionette/lib/video.js
```diff
@@ -130,14 +135,77 @@ Video.prototype = {
...
+    this.client.switchToShadowRoot();
+  },
+
+  getElapsedTime: function() {
+    var elapsedTimeElem = this.client.findElement(Video.Selector.elapsedTime);
+    return elapsedTimeElem;
+  },
+
+  getElapsedTimeSeconds: function() {
+    return this.getElapsedTime().text().split(':').pop();
```
davidflanagan: I thought you had a parseInt() in here. As it stands you're comparing this string to a number up above on line 179. It works, but only because JS is weird that way. I think it would be better to modify this function to return a number.

apps/video/js/db.js
```diff
@@ -153,6 +153,7 @@ function addVideo(videodata) {
 
   // Create the thumbnail view for this video and insert it at the right spot
   var view = thumbnailList.addItem(videodata);
+
```
davidflanagan: You should probably remove this new line before landing. I assume this is an accidental insertion from adding and removing logging statements or something.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=132ab51937a8a94a861d0e3d729634a0554953ea

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1b89d66a91ec0941e6edbb61b8595b4762ca8c8c

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=66103346b9b9f3856758f0f7b28b889d7d2ebb94

PR #33221, opened 2015-11-17
Bug 1225122 - Notification for emails with no subject shows undefined r=asuth

jrburke:
Since getEnv was getting long with args, switched to a "localized" object that contains the two localized values we need to get async. Then the other change was using this value for the subject when it did not have a value.

Tested on flame device for a no title and and a title case.

### DIFF
--- a/apps/email/js/sync.js
+++ b/apps/email/js/sync.js
@@ -77,12 +77,10 @@ define(function(require) {
           });
         }
 
-        titleL10n = titleL10n || 'notification-no-subject';
-
         notificationHelper.send(titleL10n, notificationOptions)
           .then(function(notification){
             // If the app is open, but in the background, when the notification
-            // comes in, then we do not get notifived via our 
+            // comes in, then we do not get notifived via our
             // mozSetMessageHandler that is set elsewhere. Instead need to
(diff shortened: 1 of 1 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4aec9c86a72bdceca1f7dddd1046022e7359c131

apps/email/js/sync.js
```diff
@@ -152,8 +150,16 @@ define(function(require) {
...
           fetchNotificationsData('sync').then(
             function(existingNotificationsData) {
-              mozL10n.formatValue('senders-separation-sign').then(separator => {
-                fn(app, currentAccount, existingNotificationsData, separator);
+              mozL10n.formatValue('senders-separation-sign')
+              .then(function(separator) {
+                var localized = {
+                  separator
+                };
+                mozL10n.formatValue('notification-no-subject')
```
asutherland: Ahh yes, construction resumes on the pyramid of doom... moohoohahaha!

(Referencing the comment above.  This is fine.  Maybe we can make things happier in convoy.)

jrburke: Perhaps we will call this one a temple... of doom!

PR #33244, opened 2015-11-17
Bug 1224696 - downsample jpegs when creating preview images

davidflanagan:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=924fa032c4f851ca10bac516e32a0ea4284766cb

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0e9e30d5e961e4fec4e7d029a257fa4beffe03c9

apps/gallery/js/MetadataParser.js
```diff
@@ -500,60 +586,67 @@ var metadataParser = (function() {
         canvas.toBlob(function(blob) {
```
punamdahiya: Quick clarification: Shouldn't we call  metadataError('createThumbnailAndPreview: Image failed to load') and return; 
instead of error('createThumbnailAndPreview: Image failed to load'); on Line 507

punamdahiya: Nit: Got it after looking at createThumbnailAndPreview call Line 352 and definition on Line 499 that we are calling metadataerror callback here, I guess for readability, it would have helped to keep names to metadataCallback, metadataError here same as in createJPEGThumbnailAndPreview.

punamdahiya: On debugging, I realised without patch with big images the code was failing at context.drawImage call inside createThumbnailFromElement

https://github.com/mozilla-b2g/gaia/pull/33244/files#diff-f50b6edc68dfe30b26bf9d6977594e04R122

The only error seen in logs is  E/Gallery ( 7152): [JavaScript Error: "Image corrupt or truncated." {file: "blob:app://gallery.gaiamobile.org/deb74ba2-08ac-4c9c-80ae-b0a5877fb17e" line: 0}]

which is not very helpful, We are already using try catch but that fails to catch this error.

davidflanagan: Before 2.5, that drawImage() call would fail with an exception if there was not enough memory, and the catch clause would handle it. But apparently the spec does not allow drawImage() to throw anything. So gecko changed, and now drawImage fails silently, creating those black thumbnails.  Eventually we'll probably want to add code to try to detect that: draw some random pixels into the canvas, call drawImage, then check that the random pixels have been modified. If not, then the drawImage failed.

PR #33269, opened 2015-11-19
Bug 1223677 - Early return in handlePresentation if protocol is not app

ChunMinChang:
If presentation-api is allowed by those apps with presentation permission, I think it's better to ignore non-app protocol in handlePresentation

### DIFF
--- a/tv_apps/smart-system/js/app_window_factory.js
+++ b/tv_apps/smart-system/js/app_window_factory.js
@@ -106,18 +106,20 @@
       var parseUrl = detail.url.split('/');
       var protocol = parseUrl[0].toUpperCase();
       var self = this;
-      var manifestURL;
+      var manifestURL = null;
       var requestId = detail.id;
+
+      // Check the protocol type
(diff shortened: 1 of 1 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6fa0e9da11da36d26778f5a7fdf469c97d4a1a87

tv_apps/smart-system/js/app_window_factory.js
```diff
@@ -106,18 +106,20 @@
       var parseUrl = detail.url.split('/');
       var protocol = parseUrl[0].toUpperCase();
       var self = this;
-      var manifestURL;
+      var manifestURL = null;
       var requestId = detail.id;
 
+      // Check the protocol type
+      if (protocol !== 'APP:') {
+        return;
```
rexboy7: please add `this._sendPresentationDenied(requestId);` before return.

PR #33284, opened 2015-11-19
Bug 1224496 - [Gallery] Show header text after lazy loading gaia-header.js

punamdahiya:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=fea29112248d79099a1e81446471531ea96b71fc

apps/gallery/index.html
```diff
@@ -172,7 +172,7 @@ <h1 id="crop-header" data-l10n-id="cropimage"></h1>
 
     <section role="region" id="edit-view" class="skin-dark">
       <gaia-header ignore-dir id="edit-header" action="close" title-start="50" no-font-fit>
-        <h1 id="edit-title" data-l10n-id="edit"></h1>
+        <h1 id="edit-title" data-l10n-id="edit" hidden></h1>
         <button id="edit-save-button" data-l10n-id="save">Save</button>
         <button id="edit-tool-apply-button" data-l10n-id="apply" hidden>Apply</button>
```
davidflanagan: Hiding the h1 element without hiding the button elements still leaves the header visible, doesn't it?   It seems like it would be cleaner to hide the entire header if you can do that easily.  Or, since the bug is only for the pick header, and that one doesn't have buttons, maybe you could just do the hidden attribute in that case.

apps/gallery/js/startup.js
```diff
@@ -280,6 +280,7 @@
           function() {
             var headers = document.querySelectorAll('gaia-header');
             for (var i = 0; i < headers.length; ++i) {
+              headers[i].firstElementChild.removeAttribute('hidden');
```
davidflanagan: This is a little brittle since it assumes that hte <h1> will always be the first child of <gaia-header>. That's probably true, but it still might be better to do 'headers[i].querySelectorAll('h1').forEach(h => h.hidden = false);` or something like that.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9d96c6b21ff93d6d4cc82f7730b8776e537556c4

PR #33293, opened 2015-11-20
Bug 1226166 - [WiP] Move places model from system app to shared, r=da…

michielbdejong:
…leharvey

PR #33298, opened 2015-11-20
[WiP] Bug 1225809 - Move unpinning to shared places model, r=vnicolas

michielbdejong:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=11dea7e8baef3731877c12b7e58bd0c8bd4a4e60

apps/homescreen/js/pages.js
```diff
@@ -219,14 +219,11 @@
...
-        entry.data.pinned = false;
-        this.pagesStore.datastore.put(entry.data, id).then(() => {},
-          e => {
-            console.error('Error unpinning page:', e);
-          });
-      }, e => {
-        console.error('Error retrieving page to unpin:', e);
+      // TODO: fix LazyLoader mock to support promises
+      return new Promise((resolve, reject) => {
+        LazyLoader.load(['shared/js/places_model.js'], () => {
```
Cwiiis: This is a lot of code to do this small check, I'd rather a utility function was added to apps/homescreen/js/datastore.js to put values in the datastore and do the revision check there.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=406398903df0f66f9c20cc868b27656e15bfa08e

apps/homescreen/js/pages.js
michielbdejong: The homescreen app can be the only one not to use the shared places model, but then we do need a code comment about that, so that if someone changes the code, they do keep into account compatibility with the schema used by all other apps that write to the `places` datastore (namely system, operatorvariant, and sync).

michielbdejong:
Code comment referring to this bug added in https://github.com/mozilla-b2g/gaia/pull/33293/files#diff-9c1ad9a168c7fdb95ae783137f0f2da5R222

apps/homescreen/js/pages.js
Cwiiis: This is true - but then, we already assume all of that in the other code that accesses this datastore. If all places access was replaced, I'd agree, but as it is, it's just one small part of it that could easily just be here and then provide a good example for people looking for datastore examples. I'd still prefer that check to exist in homescreen, or alternatively, for all of the homescreen places access to be replaced by the shared code (and for that shared code to be at least as efficient as what it's replacing).

michielbdejong:
Continued in #33462.

PR #33300, opened 2015-11-20
Bug 1224514 - [Messages][NG] We should shutdown MessagingService whenapp goes to background

azasypkin:
…

### DIFF
--- a/apps/sms/lib/bridge/plugins/dispose/client.js
+++ b/apps/sms/lib/bridge/plugins/dispose/client.js
@@ -1,0 +1,59 @@
+!((define)=>{define((require,exports,module)=>{
+
+  /**
+   * Logger
+   * @type {Function}
+   * @const
+   */
+  const debug = 1 ? function(arg1, ...args) {
+    var type = `[${self.constructor.name}][${location.pathname}]`;
+    console.log(`[DisposableClient]${type} - "${arg1}"`, ...args);
+  } : () => {};
+
+  /**
+   * Exports
+   */
+  module.exports = (client) => {
+    var pendingOperationManager = {
+      operations: new Set(),
+      registerOperation(operation) {
+        operation = operation.then(
+          () => this.operations.delete(operation),
+          () => this.operations.delete(operation)
+        );
+        this.operations.add(operation);
+      }
+    };
+
+    var originalMethod = client.method;
+    client.method = function(...args) {
+      var methodPromise = originalMethod.apply(client, args);
+      pendingOperationManager.registerOperation(methodPromise);
+      return methodPromise;
+    };
+
+    if (client.stream) {
+      var originalStream = client.stream;
+      client.stream = function(...args) {
+        var stream = originalStream.apply(client, args);
+        pendingOperationManager.registerOperation(stream.closed);
+        return stream;
+      };
+    }
+
+    client.whenIdle = function() {
+      debug(
+        'Number of pending operations to complete',
+        pendingOperationManager.operations.size
+      );
+
+      return Promise.all(pendingOperationManager.operations);
+    };
+
+    client.requestServiceDisposal = function() {
+      return client.method('@@requestDisposal', client.id);
+    };
+  };
+});})((typeof define)[0]=='f'&&define.amd?define:((n,w)=>{return(typeof
+  module)[0]=='o'?c=>{c(require,exports,module);}:(c)=>{var m={exports:{}};
+  c(n=>w[n],m.exports,m);w[n]=m.exports;};})('disposableClient',this));
--- a/apps/sms/lib/bridge/plugins/dispose/service.js
+++ b/apps/sms/lib/bridge/plugins/dispose/service.js
@@ -1,0 +1,40 @@
+!((define)=>{define((require,exports,module)=>{
+
+  /**
+   * Mini Logger
+   *
+   * @type {Function}
+   */
+
+  var debug = 1 ? function(arg1, ...args) {
+    var type = `[${self.constructor.name}][${location.pathname}]`;
+    console.log(`[DisposableService]${type} - "${arg1}"`, ...args);
+  } : () => {};
+
+  module.exports = function(service) {
+    debug('attaching plugin');
+
+    service.method('@@requestDisposal', (clientId) => {
+      debug('Client %s is requesting to dispose service', clientId);
+
+      let canBeDisposed = true;
+      service.eachClient((client) => {
+        if (client.id !== clientId) {
+          canBeDisposed = false;
+        }
+      });
+
+      if (canBeDisposed) {
+        // Use setTimeout to allow service to respond.
+        setTimeout(() => service.emit('dispose'), 0);
+      }
+
+      debug('Service can be disposed: %s', canBeDisposed);
+
+      return canBeDisposed;
+    });
+  };
+
+});})((typeof define)[0]=='f'&&define.amd?define:((n,w)=>{return(typeof
+  module)[0]=='o'?c=>{c(require,exports,module);}:(c)=>{var m={exports:{}};
+  c(n=>w[n],m.exports,m);w[n]=m.exports;};})('disposableService',this));
--- a/apps/sms/services/js/client_disposer.js
+++ b/apps/sms/services/js/client_disposer.js
@@ -1,0 +1,110 @@
+(function(exports) {
+  'use strict';
+
+  const debug = 1 ?
+    (arg1, ...args) => console.log(`[ClientDisposer] ${arg1}`, ...args):
+    () => {};
+
+  const priv = Object.freeze({
+    tasks: Symbol('tasks')
+  });
+
+  const ClientDisposer = {
+    [priv.tasks]: new WeakMap(),
+
+    dispose(client, timeout = 0) {
+      debug('Client %s to be disposed with timeout %s', client.id, timeout);
+
+      if (this.isDisposing(client)) {
+        return this[priv.tasks].get(client).promise;
+      }
+
+      let task = {
+        promise: null,
+
+        canBeCancelled: true,
+        cancellationRequested: false
+      };
+
+      this[priv.tasks].set(client, task);
+
+      // 1. Let's wait for the pending operations to complete first.
+      task.promise = client.whenIdle().then(() => {
+        debug('Client is in idle state');
+
+        if (task.cancellationRequested) {
+          throw new Error('Cancellation requested');
+        }
+
+        return new Promise((resolve, reject) => {
+          // 2. Now let's wait for specified amount of time, just in case client
+          // will be requested again soon enough.
+          setTimeout(() => {
+            debug('Disposal timeout fired');
+
+            if (task.cancellationRequested) {
+              reject(new Error('Cancellation requested'));
+            } else {
+              // Until service replies we can't guarantee that disposal can be
+              // cancelled.
+              task.canBeCancelled = false;
+
+              // 3. Now let's ask service to dispose itself.
+              resolve(client.requestServiceDisposal());
+            }
+          }, timeout);
+        });
+      }).then((serviceWillBeDisposed) => {
+        debug('Service replied to disposal request: ', serviceWillBeDisposed);
+
+        // If service can't be disposed (e.g. it has other clients connected)
+        // and disposal cancellation has been requested, we can safely cancel
+        // now.
+        if (task.cancellationRequested && !serviceWillBeDisposed) {
+          throw new Error('Cancellation requested');
+        }
+
+        // 4. Otherwise we can safely destroy client, no matter what service
+        // replied. We don't return here, cause client use the same timeout
+        // for both ordinary method calls and service method calls
+        // (connect, disconnect and etc.) and if timeout is disabled and service
+        // has been destroyed already client.destroy() will be never resolved.
+        client.destroy();
+      }).then(
+        () => debug('Client is successfully disposed'),
+        (e) => {
+          debug('Client disposal is failed: ', e);
+          throw e;
+        }
+      );
+
+      return task.promise;
+    },
+
+    isDisposing(client) {
+      return this[priv.tasks].has(client);
+    },
+
+    cancel(client) {
+      debug('Cancel disposal of client %s', client.id);
+
+      let task = this[priv.tasks].get(client);
+
+      task.cancellationRequested = true;
+
+      // If we are on the disposal stage where we can safely cancel disposal
+      // let's just schedule cancellation and allow consumer to use the same
+      // client instance right away, otherwise we should wait for the disposal
+      // task to complete to know for sure.
+      return (task.canBeCancelled ? Promise.reject() : task.promise).then(
+        () => {
+          debug('Cancellation failed');
+          throw new Error('Cancellation failed');
+        },
+        (e) => debug('Client disposal is successfully cancelled')
+      );
+    }
+  };
+
+  exports.ClientDisposer = ClientDisposer;
+})(self);
--- a/apps/sms/services/js/worker_client.js
+++ b/apps/sms/services/js/worker_client.js
@@ -1,0 +1,215 @@
+/*global bridge,
+ ClientDisposer,
+ disposableClient,
+ streamClient
+ */
+
+(function(exports) {
+  'use strict';
+
+  const debug = 1 ?
+    (arg1, ...args) => console.log(`[WorkerClient] ${arg1}`, ...args):
+    () => {};
+
+  /**
+   * Supported worker types.
+   * @enum {string}
+   */
+  const WORKER_TYPE = Object.freeze({
+    WORKER: 'worker',
+    SHARED_WORKER: 'shared-worker'
+  });
+
+  /**
+   * Supported lifetime strategies or the way we manage workers.
+   * @enum {string}
+   */
+  const LIFETIME_STRATEGY = Object.freeze({
+    ALWAYS_ALIVE: 'always-alive',
+    ALIVE_WHEN_VISIBLE: 'alive-when-visible'
+  });
+
+  /**
+   * Default values for the options supported by WorkerClient.
+   * @type {Object}
+   */
+  const DEFAULT_OPTIONS = Object.freeze({
+    /**
+     * Type of the worker to use to run service.
+     * @type {WORKER_TYPE}
+     */
+    type: WORKER_TYPE.SHARED_WORKER,
+
+    /**
+     * Lifetime strategy to use.
+     * @type {LIFETIME_STRATEGY}
+     */
+    lifetime: LIFETIME_STRATEGY.ALWAYS_ALIVE,
+
+    /**
+     * Indicates that both client and service support "streams", so that we
+     * should load appropriate bridge plugin.
+     * @type {boolean}
+     */
+    supportStreams: false,
+
+    /**
+     * Indicates if we want to run worker with the service immediately,
+     * alternatively service will be run only on the first demand.
+     * @type {boolean}
+     */
+    preRun: false,
+
+    /**
+     * Sets the number of milliseconds we should wait before actually shutting
+     * down client once shut down has been requested.
+     */
+    disposeTimeout: 3000,
+
+    /**
+     * Default client-service response timeout. If "false" - then client request
+     * will never be time outed.
+     * @type {Boolean|Number}
+     */
+    timeout: false
+  });
+
+  const priv = Object.freeze({
+    options: Symbol('options'),
+
+    client: Symbol('client'),
+    endpoint: Symbol('endpoint'),
+
+    create: Symbol('create')
+  });
+
+  const WorkerClient = function(serviceName, servicePath, options = {}) {
+    debug('Initialized');
+
+    if (!serviceName || !servicePath) {
+      throw new Error('Both service name and path are required!');
+    }
+
+    this[priv.options] = Object.assign(
+      { serviceName, servicePath }, DEFAULT_OPTIONS, options
+    );
+
+    if (options.lifetime === LIFETIME_STRATEGY.ALIVE_WHEN_VISIBLE) {
+      document.addEventListener('visibilitychange', () => {
+        if (document.hidden) {
+          debug('Document is hidden');
+
+          //this.dispose();
+        }
+      });
+    }
+  };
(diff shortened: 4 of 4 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=61d60d486757fbf24db918be591ad7531fa76e3d

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b9099ed7f65e9cd493d18d76add65d4c45e71f0f

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4be260b2ffe4ce028effe7478c49fa94b679d239

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0d0f6399ff8d2562287745928c3dd223e13365b5

apps/sms/lib/bridge/plugins/dispose/client.js
```diff
@@ -0,0 +1,59 @@
+!((define)=>{define((require,exports,module)=>{
```
azasypkin: We have an option to keep it as plugin or just implement wrapper around client by ourselves if we wish. Also current plugin name is not the best :)

apps/sms/lib/bridge/plugins/dispose/service.js
```diff
@@ -0,0 +1,40 @@
...
+
+  var debug = 1 ? function(arg1, ...args) {
+    var type = `[${self.constructor.name}][${location.pathname}]`;
+    console.log(`[DisposableService]${type} - "${arg1}"`, ...args);
+  } : () => {};
+
+  module.exports = function(service) {
+    debug('attaching plugin');
+
+    service.method('@@requestDisposal', (clientId) => {
```
azasypkin: Same here - it can be either server part of plugin or handled by ourselves in bridge_service_mixin.

apps/sms/lib/bridge/plugins/dispose/service.js
```diff
@@ -0,0 +1,40 @@
...
+      });
+
+      if (canBeDisposed) {
+        // Use setTimeout to allow service to respond.
+        setTimeout(() => service.emit('dispose'), 0);
+      }
+
+      debug('Service can be disposed: %s', canBeDisposed);
+
+      return canBeDisposed;
```
azasypkin: Client should know if service is going to be destroyed - so that consumer knows that it can't reuse the same client instance anymore.

apps/sms/lib/bridge/plugins/dispose/service.js
```diff
@@ -0,0 +1,40 @@
...
+      let canBeDisposed = true;
+      service.eachClient((client) => {
+        if (client.id !== clientId) {
+          canBeDisposed = false;
+        }
+      });
+
+      if (canBeDisposed) {
+        // Use setTimeout to allow service to respond.
+        setTimeout(() => service.emit('dispose'), 0);
```
azasypkin: We have at least 2 options here:
* This one - service fires "dispose" event and consumer (code that creates service instance) destroys environment by itself;
* we close SharedWorker/Worker right here.

Not sure which one is better, what do you think?

apps/sms/services/js/client_disposer.js
```diff
@@ -0,0 +1,110 @@
...
+
+      let task = this[priv.tasks].get(client);
+
+      task.cancellationRequested = true;
+
+      // If we are on the disposal stage where we can safely cancel disposal
+      // let's just schedule cancellation and allow consumer to use the same
+      // client instance right away, otherwise we should wait for the disposal
+      // task to complete to know for sure.
+      return (task.canBeCancelled ? Promise.reject() : task.promise).then(
```
azasypkin: This part can be a bit confusing as I wanted the following:
* ```dispose()``` returns promise that is __resolved__ when disposal has been performed and __rejected__ if not;
* ```cancel()``` returns promise that is __resolved__ when cancellation has been performed (say disposal __hasn't__ been performed) and __rejected__ if cancellation failed (say disposal has been performed).

apps/sms/services/js/worker_client.js
```diff
@@ -0,0 +1,215 @@
...
+  const WORKER_TYPE = Object.freeze({
+    WORKER: 'worker',
+    SHARED_WORKER: 'shared-worker'
+  });
+
+  /**
+   * Supported lifetime strategies or the way we manage workers.
+   * @enum {string}
+   */
+  const LIFETIME_STRATEGY = Object.freeze({
```
azasypkin: I left this ```enum``` from my previous WIP where I have 3rd strategy - ```PER_CALL``` when client is created for method call/stream and automatically disposed once method is finished/stream is closed. But I removed this for now - so can remove this enum as well if you wish.

apps/sms/services/js/worker_client.js
```diff
@@ -0,0 +1,215 @@
...
+  const priv = Object.freeze({
+    options: Symbol('options'),
+
+    client: Symbol('client'),
+    endpoint: Symbol('endpoint'),
+
+    create: Symbol('create')
+  });
+
+  const WorkerClient = function(serviceName, servicePath, options = {}) {
```
azasypkin: I use ```serviceName``` and ```servicePath``` as dedicated parameters to stress that they're mandatory.

apps/sms/services/js/worker_client.js
```diff
@@ -0,0 +1,215 @@
...
+    this[priv.options] = Object.assign(
+      { serviceName, servicePath }, DEFAULT_OPTIONS, options
+    );
+
+    if (options.lifetime === LIFETIME_STRATEGY.ALIVE_WHEN_VISIBLE) {
+      document.addEventListener('visibilitychange', () => {
+        if (document.hidden) {
+          debug('Document is hidden');
+
+          //this.dispose();
```
azasypkin: Disabled for testing

apps/sms/services/js/worker_client.js
```diff
@@ -0,0 +1,215 @@
...
+    [priv.client]: null,
+
+    /**
+     * Service endpoint instance (either Worker or SharedWorker).
+     * @type {Worker|SharedWorker}
+     * @private
+     */
+    [priv.endpoint]: null,
+
+    method(...args) {
```
azasypkin: Here I have several options as well:
* Current: consumer of ```WorkerClient``` will have Promise-based ```method``` and ```stream``` (I'd likely use Proxy to expose all bridge methods in single way if we decide use this approach) , but ```stream``` on bridge client isn't Promise based. So I don't really like that BridgeClient and WorkerClient will have different API.
* We don't expose these method and consumer does smth like ```WorkerClient.getClient().then((bridgeClient) => bridge.client.method(.....))``` But here I don't like that bridge client is exposed to consumer...

What do you think?

apps/sms/services/js/worker_client.js
```diff
@@ -0,0 +1,215 @@
...
+     * Shuts down client and underlying service endpoint. It waits for
+     * options.shutdownTimeout before actually starting to shut client down, so
+     * that we can save resources if client is requested again soon enough.
+     * @private
+     */
+    dispose() {
+      return ClientDisposer.dispose(
+        this[priv.client], this[priv.options].disposeTimeout
+      ).then(() => {
+        if (this[priv.options].type === WORKER_TYPE.SHARED_WORKER) {
```
azasypkin: Need to check if we still need this, likely yes.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a738115d5683262aa4aa2201b7517e879ffe29c1

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3c3e5de2ec85ed8230ba7486ad6814c0dca813b7

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=76a1ca6e932703123376079e64595cb4e525593e

apps/sms/lib/bridge/plugins/dispose/client.js
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * Logger
+   * @type {Function}
+   * @const
+   */
+  const debug = 1 ? function(arg1, ...args) {
+    var type = `[${self.constructor.name}][${location.pathname}]`;
+    console.log(`[DisposableClient]${type} - "${arg1}"`, ...args);
```
julienw: nit: I'd remove the quotes around arg1 so that we can also not use a format if we don't want to :)

apps/sms/lib/bridge/plugins/dispose/service.js
```diff
@@ -0,0 +1,40 @@
...
+    debug('attaching plugin');
+
+    service.method('@@requestDisposal', (clientId) => {
+      debug('Client %s is requesting to dispose service', clientId);
+
+      let canBeDisposed = true;
+      service.eachClient((client) => {
+        if (client.id !== clientId) {
+          canBeDisposed = false;
+        }
```
julienw: nit: is it possible to stop the loop early ?

apps/sms/services/js/worker_client.js
```diff
@@ -0,0 +1,215 @@
...
+     * @type {boolean}
+     */
+    supportStreams: false,
+
+    /**
+     * Indicates if we want to run worker with the service immediately,
+     * alternatively service will be run only on the first demand.
+     * @type {boolean}
+     */
+    preRun: false,
```
julienw: I don't see where it's used ?

apps/sms/services/js/worker_client.js
```diff
@@ -0,0 +1,215 @@
...
+
+    /**
+     * Sets the number of milliseconds we should wait before actually shutting
+     * down client once shut down has been requested.
+     */
+    disposeTimeout: 3000,
+
+    /**
+     * Default client-service response timeout. If "false" - then client request
+     * will never be time outed.
```
julienw: nit: "timed out" :)

apps/sms/services/js/worker_client.js
```diff
@@ -0,0 +1,215 @@
...
+     * @type {Boolean|Number}
```
julienw: nit: "false|Number" ? I don't think we support "true" ?

apps/sms/services/js/worker_client.js
julienw: :+1:

apps/sms/services/js/worker_client.js
azasypkin: Nowhere :) Just wanted your opinion if we may need something like this :)

apps/sms/lib/bridge/plugins/dispose/client.js
azasypkin: Hmm, it's weird indeed, looks like I copied it from bridge lib :D

apps/sms/lib/bridge/plugins/dispose/service.js
azasypkin: Mmm, nope, but if we decide to move this forward we'll propose better ```eachClient``` to Wilson :)

apps/sms/services/js/client_disposer.js
```diff
@@ -0,0 +1,110 @@
...
+
+      return task.promise;
+    },
+
+    isDisposing(client) {
+      return this[priv.tasks].has(client);
+    },
+
+    cancel(client) {
+      debug('Cancel disposal of client %s', client.id);
```
julienw: wondering if the added complexity of being able to cancel the disposal is worth it

apps/sms/services/js/worker_client.js
azasypkin: Ah, I always make this mistake :) Thanks!

apps/sms/services/js/worker_client.js
azasypkin: Right, that's why I wanted it to be 0 instead of ```false``` like C# WebServices :)

apps/sms/services/js/client_disposer.js
julienw: actually I understand we don't dispose right away, but I don't properly get when "dispose" is called (except when visibilitychange)

apps/sms/lib/bridge/plugins/dispose/client.js
```diff
@@ -0,0 +1,59 @@
...
+      };
+    }
+
+    client.whenIdle = function() {
+      debug(
+        'Number of pending operations to complete',
+        pendingOperationManager.operations.size
+      );
+
+      return Promise.all(pendingOperationManager.operations);
```
julienw: what happens if we have new ones happening while these ones are finishing ?

apps/sms/lib/bridge/plugins/dispose/service.js
julienw: I'd do both: emit the event in case it's useful but still close it right after this (but maybe then there could be issues where the event is not sent ?)

apps/sms/lib/bridge/plugins/dispose/client.js
```diff
@@ -0,0 +1,59 @@
...
+   * Exports
+   */
+  module.exports = (client) => {
+    var pendingOperationManager = {
+      operations: new Set(),
+      registerOperation(operation) {
+        operation = operation.then(
+          () => this.operations.delete(operation),
+          () => this.operations.delete(operation)
+        );
```
julienw: nit: use `.catch(() => {})` to simplify this bit ?

apps/sms/lib/bridge/plugins/dispose/client.js
```diff
@@ -0,0 +1,59 @@
...
+    };
+
+    if (client.stream) {
+      var originalStream = client.stream;
+      client.stream = function(...args) {
+        var stream = originalStream.apply(client, args);
+        pendingOperationManager.registerOperation(stream.closed);
+        return stream;
+      };
+    }
```
julienw: I don't see the case of the service sending an event ?

apps/sms/lib/bridge/plugins/dispose/client.js
julienw: I'd use "disposable" as name :)

apps/sms/services/js/worker_client.js
julienw: good for me to keep it

apps/sms/services/js/worker_client.js
julienw: none of these issues makes me cry :)

apps/sms/services/js/client_disposer.js
azasypkin: It's really good question and I'm not 100% sure either :) About the moment when dispose is called - I was thinking about two cases - 1. on visibility change; 2. on demand - I could imagine that we may want to run service - execute operation and shut it down immediately (with timeout) to freed all resources used by service.

And the main problem that makes this code complex is that new request can come in before service is actually disposed. And I feel like there is a high chance of race here if we don't wait for the old client to be disposed completely. But at the same time I don't want new request to wait a lot, that's why I have "safe cancel" stages (when we wait for operations to complete or timeout to fire) - during these we just allow new request to use the same client.

What do you think?

apps/sms/services/js/worker_client.js
azasypkin: Ha-ha, cool :D

apps/sms/lib/bridge/plugins/dispose/client.js
azasypkin: This new operation will reuse the same client and set ```cancellationRequested = true;``` so that once ```whenIdle``` with the _old_ list of operations resolves disposal will be cancelled.

apps/sms/lib/bridge/plugins/dispose/client.js
azasypkin: Good catch, I'm not sure if need to do anything with this - let me dig to understand this for sure.

apps/sms/lib/bridge/plugins/dispose/service.js
azasypkin: I was talking about emitting event to the service host to do something before shared worker is disposed - so here we should be safe as event handler will be processed synchronously.

apps/sms/lib/bridge/plugins/dispose/client.js
azasypkin: You mean something like this?
```
operation = operation.catch(() => {}).then(
  () => this.operations.delete(operation)
);
```

Yeah, sounds like a good idea

apps/sms/lib/bridge/plugins/dispose/client.js
steveck-chung: keep it as plugin and "disposable"  is fine +1

apps/sms/services/js/client_disposer.js
steveck-chung: Just curious about how long will it take for disposal on real devices. I would prefer to start new client after old client disposed completely if if didn't need much time, and there seems not much cases that we will need to switch on/off the service frequently.

Back to dispose functionality, I'm glad to see this applied for certain services, especially for some proactive service like messaging or draft service that could only be triggered by certain user action, because these services are not needed while in background(or inbox view) nor resume to foreground(or conversation/new message view) immediately. And we should be careful that never dispose a service while it's still not complete yet. Not sure how to achieve this, maybe having a status property for client to represent the service is still running or not? Just think out laud :)

apps/sms/lib/bridge/plugins/dispose/service.js
steveck-chung: But I didn't see any meaningful action could be performed after receiving the  dispose event from server. As a developer who will use the dispose feature in bridge, I would prefer merging  all the scripts into plugin, so closing the worker right here works for me. User just need to know the dispose is success in resolve or error reason in reject.

apps/sms/lib/bridge/plugins/dispose/client.js
steveck-chung: Another alternative solution without Promise.finally? ;)

apps/sms/services/js/worker_client.js
steveck-chung: I might prefer 2 just like you did, also wondering how this part would become once we want to integrate this part into plugin.

apps/sms/lib/bridge/plugins/dispose/client.js
steveck-chung: yeah I doubt that we need do something for dispose event.

apps/sms/lib/bridge/plugins/dispose/service.js
julienw: I wish that like ruby it would be enough to provide an iterator and have automatically all array methods available :)

apps/sms/lib/bridge/plugins/dispose/service.js
julienw: maybe some other part of the code (not the one that calls "dispose") could do something. I can't think of something concrete now but this brings clearly more possibilities.

apps/sms/lib/bridge/plugins/dispose/service.js
julienw: as a client for this service, shouldn't we always stop using the client after calling dispose ? I'm not sure to understand why we need to know this here..

apps/sms/services/js/worker_client.js
julienw: not sure it's useful as a first patch :) we'll see later if we need it.

apps/sms/services/js/worker_client.js
```diff
@@ -0,0 +1,215 @@
...
+        new SharedWorker(options.servicePath);
+
+      let client = bridge.client({
+        service: options.serviceName,
+        endpoint: this[priv.endpoint],
+        timeout: options.timeout
+      });
+
+      if (options.supportStreams) {
+        debug('Client supports streams');
```
julienw: not sure that "support" is the right word; maybe "needStreams" is better ?

apps/sms/services/js/worker_client.js
```diff
@@ -0,0 +1,215 @@
...
+        debug('Client is disposable');
+        client = client.plugin(disposableClient);
+      }
+
+      return (this[priv.client] = client);
+    },
+
+    /**
+     * Returns bridge client instance. It creates a new client if it hasn't been
+     * created yet or shit down previously, and cancels shutdown actions if it
```
julienw: nit: "shut down" :)

apps/sms/services/js/worker_client.js
```diff
@@ -0,0 +1,215 @@
...
+     * perform any bridge client method.
+     * @returns {Promise.<Client>}
+     * @private
+     */
+    [priv.getClient]() {
+      let disposeCancelPromise;
+
+      if (ClientDisposer.isDisposing(this[priv.client])) {
+        debug('Client is in process of disposing. Cancelling...');
+        disposeCancelPromise = ClientDisposer.cancel(this[priv.client]);
```
julienw: what do you do if "cancel" doesn't work ?

apps/sms/lib/bridge/plugins/dispose/client.js
```diff
@@ -0,0 +1,59 @@
...
+  module.exports = (client) => {
+    var pendingOperationManager = {
+      operations: new Set(),
+      registerOperation(operation) {
+        operation = operation.then(
+          () => this.operations.delete(operation),
+          () => this.operations.delete(operation)
+        );
+        this.operations.add(operation);
+      }
```
julienw: with this, we need to take care that all operations do not last longer than their promises. I'm thinking that we cold have an operation that takes longer (using events for example) than their promise resolution. Quite hypothetic right now but I don't think it's that uncommon.

apps/sms/services/js/client_disposer.js
julienw: wondering if we can manage the timeout at the client level instead of the service level. I think this would bring down the complexity while still keeping most of the idea ?

Maybe I'm missing something though..

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6cafde52e818c96d26c165c3d41189bb151b9de2

PR #33314, opened 2015-11-21
Bug 1226647 - Fix failure in test_homescreen_column_layout.py

mwargers:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b5e26fca56faa48319611dd0fab6ad5ce3011983

tests/python/gaia-ui-tests/gaiatest/tests/functional/homescreen/test_homescreen_column_layout.py
```diff
@@ -36,11 +36,8 @@ def test_homescreen_column_layout(self):
         Wait(self.marionette).until(lambda m: initial_number_of_columns != self.homescreen.number_of_columns)
         self.assertEqual(4, self.homescreen.number_of_columns)
 
-        settings.launch()
-        # in 319MB setting, the settings app gets killed sometimes,
-        # in which case homescreen needs to be opened again
-        if settings.current_view == 'root':
-            homescreen_settings = settings.open_homescreen()
+        settings.launch(wait_for_airplane_mode_ready=False)
```
JohanLorenzo: I don't think this is the right way to go. The settings app is already launched, so using this function is maybe not a good idea, after all. 
With this parameter we hack the launch app, so it has 2 purposes. I think 1 function should solve 1 problem. Then here, if we want to make the test more reliable, we can close the settings app and then launch it again. So the launch function will have only 1 purpose: launching Settings when it's freshly starting.

What do you think?

mwargers: I've seen it happen more, using launch method while the app is already open.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4d9b033dbc3e0acaf804555a228c1f20722fc9bb

PR #33360, opened 2015-11-25
Bug 1227991 - client.js sends wrong variable to Marionette server; r=aus

andreastt:
This patch rectifies an obvious mistake in an earlier rebase where the
Marionette client is passing on the unserialised body to the Marionette
server instead of the serialised version stored in the data variable.

This only happens to work right now because the server is not using
protocol level 3 yet.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=09d58fa42fd0db566ee322f29a637153532c1421

nullaus:
:+1:

PR #33394, opened 2015-11-26
Bug 1228352 - [Messages] Reduce the flickering happening in activitie…

julienw:
…s, and streamline animations r=azasypkin

### DIFF
--- a/apps/sms/views/shared/js/navigation.js
+++ b/apps/sms/views/shared/js/navigation.js
@@ -491,7 +491,7 @@ function executeNavigationStep(stepName) {
 }
 
 /**
- * Used in `waitForSlideAnimation` to help solving race conditions.
+ * Used in `waitForPanelAnimation` to help solving race conditions.
  * @type {Defer}
  */
 var previousAnimationDefer;
@@ -502,7 +502,7 @@ var previousAnimationDefer;
  * @param {Element} panelElement The element to watch.
  * @returns {Promise} Resolves when the animation ends.
  */
-function waitForSlideAnimation(panelElement) {
+function waitForPanelAnimation(panelElement) {
   if (previousAnimationDefer) {
     previousAnimationDefer.reject(new Error('A new animation started'));
   }
@@ -570,39 +570,50 @@ function switchPanel() {
     ));
   }
 
+  if (newPanelElement === oldPanelElement) {
+    return Promise.resolve();
+  }
+
   var isGoingBack = oldView && oldView.previous === newView.name;
 
   newPanelElement.style = '';
+  newPanelElement.classList.remove('panel-hidden');
+  newPanelElement.classList.add('panel-active');
 
   var doSlideAnimation =
     from && from.panel &&
     from.panel.behavior !== newView.behavior;
 
+  var doFadeInAnimation = !from || !from.panel;
+
   var animationPromise;
   if (doSlideAnimation) {
-
     var shouldGoRight = isGoingBack ^ (document.dir === 'rtl');
     newPanelElement.style.animationName =
-      shouldGoRight ? 'new-slide-right' : 'new-slide-left';
+      shouldGoRight ? 'left-to-current' : 'right-to-current';
     oldPanelElement.style.animationName =
-      shouldGoRight ? 'old-slide-right' : 'old-slide-left';
+      shouldGoRight ? 'current-to-right' : 'current-to-left';
+  } else if (doFadeInAnimation) {
+    newPanelElement.style.animationName = 'fade-in';
+  }
 
-    animationPromise = waitForSlideAnimation(newPanelElement).catch(
+  if (doFadeInAnimation || doSlideAnimation) {
+    animationPromise = waitForPanelAnimation(newPanelElement).catch(
       () => {}
     ).then(() => {
-      oldPanelElement.style = '';
+      if (oldPanelElement) {
+        oldPanelElement.style = '';
+      }
       newPanelElement.style = '';
     });
   }
 
   return (animationPromise || Promise.resolve()).then(() => {
-    if (oldView) {
+    if (oldPanelElement) {
       oldPanelElement.classList.remove('panel-active');
       oldPanelElement.setAttribute('aria-hidden', 'true');
     }
 
-    newPanelElement.classList.add('panel-active');
-    newPanelElement.classList.remove('panel-hidden');
     newPanelElement.setAttribute('aria-hidden', 'false');
   });
 }
--- a/apps/sms/views/shared/style/panel.css
+++ b/apps/sms/views/shared/style/panel.css
@@ -9,56 +9,17 @@
 
   background: white;
 
-  animation: .2s both linear;
+  animation: var(--transition-duration) both ease;
 }
-html[dir="ltr"] .panel { transform: translate(100%); }
-html[dir="rtl"] .panel { transform: translate(-100%); }
 
-/**
- * adding html[dir] to make the specificity of this rule high enough to
- * overwrite the default translation for element with class panel just above,
- * without needing to put a !important
- */
-html[dir] .panel-active {
+.panel {
+  transform: translate(100%); /* BiDi proof */
+}
+
+.panel-active {
   transform: translate(0);
 }
 
 .panel-hidden {
   display: none;
-}
-
-@keyframes new-slide-left {
-  from {
-    transform: translateX(100%);
-  }
-  to {
-    transform: none;
-  }
-}
-
-@keyframes old-slide-left {
-  from {
-    transform: none;
-  }
-  to {
-    transform: translateX(-100%);
-  }
-}
-
-@keyframes new-slide-right {
-  from {
-    transform: translateX(-100%);
-  }
-  to {
-    transform: none;
-  }
-}
-
-@keyframes old-slide-right {
-  from {
-    transform: none;
-  }
-  to {
-    transform: translateX(100%);
-  }
 }
--- a/apps/sms/views/shared/style/root.css
+++ b/apps/sms/views/shared/style/root.css
@@ -12,7 +12,7 @@ body {
   font-size: 1.6rem;
   line-height: 2.2rem;
   vertical-align: middle;
-  background-color: #FFF;
+  background: var(--header-background);
 }
 
 /* Text conventions */
--- a/apps/sms/views/shared/test/unit/navigation_test.js
+++ b/apps/sms/views/shared/test/unit/navigation_test.js
@@ -80,6 +80,25 @@ suite('navigation >', function() {
     });
   }
 
+  function styleDescriptor(setterStub, propSetterStubs) {
+    return {
+      get: function() {
+        var result = {};
+        var descriptor = {};
+
+        for (var prop in propSetterStubs) {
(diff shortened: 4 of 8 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d19d466a50088537aa3515b0cf1da01ec7323e60

apps/sms/views/shared/js/navigation.js
```diff
@@ -583,24 +583,36 @@ function switchPanel() {
   var isGoingBack = oldView && oldView.previous === newView.name;
 
   newPanelElement.style = '';
+  newPanelElement.classList.remove('panel-hidden');
+  newPanelElement.classList.add('panel-active');
```
julienw: I checked with my workload that with a current gecko this does not provoke big reflows anymore. But if you can double check with a big workload also it would be nice.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=03ea2e80a89efbe05fb6c25012d3685950469acc

apps/sms/views/shared/style/panel.css
```diff
@@ -9,56 +9,17 @@
...
 }
-html[dir="ltr"] .panel { transform: translate(100%); }
-html[dir="rtl"] .panel { transform: translate(-100%); }
 
-/**
- * adding html[dir] to make the specificity of this rule high enough to
- * overwrite the default translation for element with class panel just above,
- * without needing to put a !important
- */
-html[dir] .panel-active {
```
julienw: actually I don't think this was necessary; what's necessary is having the right animations going on depending on the LTR/RTL status. But changing where the panel is hidden is IMO useless. What do you think @autra ?

apps/sms/views/shared/style/panel.css
```diff
@@ -9,56 +9,17 @@
...
-}
-
-@keyframes old-slide-right {
-  from {
-    transform: none;
-  }
-  to {
-    transform: translateX(100%);
-  }
-}
```
julienw: nearly the same ones in animations.css.

apps/sms/views/shared/js/navigation.js
```diff
@@ -583,24 +583,35 @@ function switchPanel() {
...
   var animationPromise;
   if (doSlideAnimation) {
-
     var shouldGoRight = isGoingBack ^ (document.dir === 'rtl');
     newPanelElement.style.animationName =
-      shouldGoRight ? 'new-slide-right' : 'new-slide-left';
+      shouldGoRight ? 'left-to-current' : 'right-to-current';
     oldPanelElement.style.animationName =
-      shouldGoRight ? 'old-slide-right' : 'old-slide-left';
+      shouldGoRight ? 'current-to-right' : 'current-to-left';
```
julienw: note that I still prefer to use animation-name; we could use the classes defined in animations.css too, but I didn't want to change too much things. I could try though, tell me if you think it's worth it.

azasypkin: I'm fine with using the same approach with animation-name.

__issue:__ Seems there is a flaw somewhere here - try to send group message and you'll end up in the broken Report view. Just checked and see that intergration tests caught something like this.

apps/sms/views/shared/js/navigation.js
```diff
@@ -583,24 +583,35 @@ function switchPanel() {
...
+  }
```
azasypkin: __nit:__ IMO having ```else if``` here would make logic clearer as conditions are mutually exclusive.
```
if (doSlideAnimation) {
  ....
} else if (doFadeInAnimation) {
  ....
}
```

apps/sms/views/shared/js/navigation.js
julienw: aah thanks, I shouldn't ask review before a green try.

apps/sms/views/shared/js/navigation.js
julienw: oki

apps/sms/views/shared/style/root.css
```diff
@@ -12,7 +12,7 @@ body {
   font-size: 1.6rem;
   line-height: 2.2rem;
   vertical-align: middle;
-  background-color: #FFF;
+  background: var(--header-background);
```
azasypkin: __question:__ wondering if we should use ```--highlight-color``` variable instead here? In Gaia Theme this variable is in ```Brand Color``` section, the name is confusing to me though.

apps/sms/views/shared/style/panel.css
autra: Oops, I let slipped this one, sorry for the delay. I don't think it's a problem if it's hidden in the same position in both direction. But in this case, you might want to put a `/* BiDi proof */ comment just above for the reader (when RTL-izing an application, I usually execute a big sed marking every place where it needs checking. These comments speeds up this process because I don't need to ask myself why it is not symmetric).

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2255ac54c0bb1ea267b2bd83c049bb963934b209

apps/sms/views/shared/style/root.css
julienw: I don't think so; I think the trick here is especially to use the same color than the header so that we see no flickering :)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=33b6ffb6c5f40a680d1a23d81cd713acd21387b5

apps/sms/views/shared/js/navigation.js
julienw: so there was nothing to do with group messages, simply moving from new message to conversation was showing the issue.

apps/sms/views/shared/js/navigation.js
```diff
@@ -573,36 +573,43 @@ function switchPanel() {
...
+      if (oldPanelElement) {
+        oldPanelElement.style = '';
+      }
       newPanelElement.style = '';
     });
   }
 
   return (animationPromise || Promise.resolve()).then(() => {
-    if (oldView) {
+    if (oldPanelElement && oldPanelElement !== newPanelElement) {
```
julienw: this is the fix

apps/sms/views/shared/style/panel.css
julienw: done

apps/sms/views/shared/test/unit/navigation_test.js
```diff
@@ -644,6 +686,105 @@ suite('navigation >', function() {
...
+              );
+              sinon.assert.calledWith(
+                styleAnimationNameStubs[1], 'current-to-left'
+              );
+              sinon.assert.calledWith(styleStubs[0], '');
+              sinon.assert.calledWith(styleStubs[1], '');
+            }).then(done, done);
+          });
+        });
+      });
```
julienw: I know this looks like a lot of repetitions, I didn't really find a better way.

apps/sms/views/shared/style/root.css
azasypkin: Mmm, correct me if I'm wrong, but when we see white flickering - it's entire app frame is white, header is not even displayed, when we see the header - we basically already see ready app page. That is why I think we should use more generic theme color - the same as splash screen and status bar, rather than header. Am I missing something or we're talking about different flickering? :)

apps/sms/views/shared/js/navigation.js
azasypkin: Hmm, weird I was pretty sure that didn't see this issue when sending simple 1-to-1 message - maybe I just don't remember well :)

apps/sms/views/shared/style/root.css
julienw: I agree we see the full white :) But if we change from "full <something>" to "header <something else> + content", there is still a flickering effect, as compared to "full <header color>" to "header <header color> + content" which is a lot easier to the eyes.

That said, I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1228379 :) Not sure it would be fixed anytime soon...
But see also https://bugzilla.mozilla.org/show_bug.cgi?id=1164123 where we made sure that the header color _is_ the same than the theme color.

apps/sms/views/shared/js/navigation.js
julienw: yeah I'm sure I've seen it with simple messages too :)

apps/sms/views/shared/style/root.css
azasypkin: > That said, I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1228379 :) Not sure it would be fixed anytime soon...

Thanks! Hopefully it will be fixed, otherwise it's really confusing that header color is considered as the "main" color :) Not all apps/views even have a header :)

apps/sms/views/shared/js/navigation.js
azasypkin: Heh, okay :)

apps/sms/views/shared/js/navigation.js
```diff
@@ -573,36 +573,43 @@ function switchPanel() {
...
+      shouldGoRight ? 'left-to-current' : 'right-to-current';
     oldPanelElement.style.animationName =
-      shouldGoRight ? 'old-slide-right' : 'old-slide-left';
+      shouldGoRight ? 'current-to-right' : 'current-to-left';
+  } else if (doFadeInAnimation) {
+    newPanelElement.style.animationName = 'fade-in';
+  }
 
+  if (doFadeInAnimation || doSlideAnimation) {
     animationPromise = waitForSlideAnimation(newPanelElement).catch(
```
azasypkin: __optional :bike: :__ now it's more like ```waitForAnimationEnd``` since we have not only ```slide``` animation - but it's very optional anyway.

apps/sms/views/shared/js/navigation.js
azasypkin: mmm, should not we just do just something like the snippet below before any manipulations with classes and aria attributes (sorry if I'm missing something obvious _again_)?
```
if (newPanelElement === oldPanelElement) {
  return Promise.resolve();
}
```

apps/sms/views/shared/test/unit/navigation_test.js
azasypkin: I think it's ok

apps/sms/views/shared/js/navigation.js
julienw: maybe waitForPanelAnimation ?

azasypkin: Yep, sounds good

apps/sms/views/shared/js/navigation.js
julienw: Mmm why not. And maybe we can also remove the comparison `from.panel.behavior !== newView.behavior` then. I'll look into it.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=47ab40337ecf55ac6a25860d0003d594f20f3bd7

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6cf41aa72d3e7b755267d74e50b93dce454b0199

PR #33430, opened 2015-11-28
Bug 1147860 - Guard test logic using promise so files are fully loaded

KevinGrandon:
https://bugzilla.mozilla.org/show_bug.cgi?id=1147860

### DIFF
--- a/shared/js/contacts/search.js
+++ b/shared/js/contacts/search.js
@@ -84,52 +84,55 @@
 
   var doInit = function doInit() {
     if (initialized) {
-      return;
+      return Promise.resolve();
     }
 
-    utils.listeners.add({
-      '#cancel-search': exitSearchMode,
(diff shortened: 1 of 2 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4f1adc662e945573fabdf7a13208dca54c06f440

shared/js/contacts/search.js
```diff
@@ -87,49 +87,52 @@
       return;
```
arcturus: We will need to return Promise.resolve here, as now we are expecting to return a promise.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=af14c022634ea81ae2a3c5685c1396fb4ca9765f

PR #33433, opened 2015-11-29
Bug 1227424 - [Sms] Use event_safety lib for dialogs

KevinGrandon:
https://bugzilla.mozilla.org/show_bug.cgi?id=1227424

### DIFF
--- a/apps/sms/views/shared/js/dialog.js
+++ b/apps/sms/views/shared/js/dialog.js
@@ -107,7 +107,7 @@ var Dialog = function(params) {
     event.preventDefault();
   });
 
-  this.form.addEventListener('transitionend', function(event) {
+  window.eventSafety(this.form, 'transitionend', function(event) {
     var form = event.target;
     if (!form.classList.contains('visible') && form.parentNode) {
       form.remove();
@@ -117,7 +117,7 @@ var Dialog = function(params) {
     }
 
     document.body.classList.remove('dialog-animating');
-  });
+  }, 500);
 
   menu.addEventListener('click', function(event) {
     var action = handlers.get(event.target);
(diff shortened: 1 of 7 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8fc24ef1c853cd575940dfc3903f066762f6829f

apps/sms/views/shared/js/dialog.js
```diff
@@ -117,7 +117,7 @@ var Dialog = function(params) {
     }
 
     document.body.classList.remove('dialog-animating');
-  });
+  }, 500);
```
KevinGrandon: Just picked an arbitrary value that's greater than the current animation time. Too short, and you may cut-off the animation.

apps/sms/views/shared/js/dialog.js
```diff
@@ -107,7 +107,7 @@ var Dialog = function(params) {
     event.preventDefault();
   });
 
-  this.form.addEventListener('transitionend', function(event) {
+  window.eventSafety(this.form, 'transitionend', function(event) {
     var form = event.target;
```
julienw: I see we  can't used `event` when we use `eventSafety` as it's not always set...

julienw: I think the eventSafety lib should never pass the event, it brings an expectation that's not always fulfilled unfortunately. Or it could create a TransitionEvent.

apps/sms/views/shared/test/unit/dialog_test.js
```diff
@@ -5,6 +5,7 @@
 */
 'use strict';
 
+require('/shared/js/event_safety.js');
```
julienw: I'm not comfortable adding something that's not a mock here, as it will bring timeouts and thus unpredictability into the test.

Ideally I'd like mocking event_safety (either with sinon only, or with an empty mock + sinon (preferred)) and that in the test we test both cases (with/without the event parameter) unless you change the event_safety lib.

PR #33443, opened 2015-11-30
Bug 1229011 - Make async TCP driver protocol 3 compatible; r=aus

andreastt:
The asynchronous TCP driver in
tests/jsmarionette/client/marionette-client/lib/marionette/drivers/abstract.js
is not compatible with Marionette protocol level 3 that will get
introduced in bug 1211489.

This causes
tests/jsmarionette/runner/marionette-js-runner/test/integration/clientasync.js
to fail with bug 1211489 is applied to Gecko.

To fix this issue we apply the same logic from the synchronous TCP
driver to marshal and unmarshal messages when talking to a protocol
level 3 server.

### DIFF
--- a/tests/jsmarionette/client/marionette-client/lib/marionette/drivers/abstract.js
+++ b/tests/jsmarionette/client/marionette-client/lib/marionette/drivers/abstract.js
@@ -2,6 +2,9 @@
 (function(module, ns) {
   'use strict';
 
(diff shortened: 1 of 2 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=db3beecfcc92398f8da12855c32099f29859091d

PR #33461, opened 2015-12-01
Bug 1219640 - Implement *test_default_contact_image.py* as an integra…

arcturus:
…tion test in JavaScript r=jlorenzo

### DIFF
--- a/apps/communications/contacts/test/marionette/list_test.js
+++ b/apps/communications/contacts/test/marionette/list_test.js
@@ -64,10 +64,6 @@ marionette('Contacts > List', function() {
       }
     }
 
-    setup(function() {
-      generateContactsWithImage();
-    });
-
     function firstContactHasImage() {
       return client.executeScript(function() {
         var span = document.querySelector(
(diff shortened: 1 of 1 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=bd4bbe6efcf26230b377544972e48ac20f8f9e9b

apps/communications/contacts/test/marionette/list_test.js
```diff
@@ -79,28 +75,55 @@ marionette('Contacts > List', function() {
...
-      //X
-      actions.press(scrollbar, 15, 350).release().perform();
-      client.helper.waitForElement(Contacts.Selectors.overlay);
-      client.helper.waitForElementToDisappear(Contacts.Selectors.overlay);
-      //A
-      actions.press(scrollbar, 15, 40).release().perform();
-      client.helper.waitForElement(Contacts.Selectors.overlay);
-      client.helper.waitForElementToDisappear(Contacts.Selectors.overlay);
+    function firstContactHasDefaultImage() {
+      return client.executeScript(function() {
```
JohanLorenzo: I don't think you need to inject a script. IMO, this can be done entirely with Marionette's API. Something like:
```js
function firstContactHasDefaultImage() {
    try {
        var span = client.findElement('#contacts-list-A .contact-item span');
        return span.cssProperty('background-image') === 'url("app://communications' +
                      '.gaiamobile.org/contacts/style/images/Imagery.png")';
    } catch (e) {
        return false;
    }
}
```

I'm not exactly sure if `cssProperty` works with `'background-image'`. If it doesn't, we can use [getAttribute](http://mozilla-b2g.github.io/marionette-js-client/api-docs/classes/Marionette.Element.html#method_getAttribute)

Also, I think we should check if the default letter is displayed in addition to the picture. What do you think? In the python test [we did it this way](https://github.com/mozilla-b2g/gaia/blob/master/tests/python/gaia-ui-tests/gaiatest/tests/functional/contacts/test_default_contact_image.py#L31). `image_data_group` [is defined as](https://github.com/mozilla-b2g/gaia/blob/master/tests/python/gaia-ui-tests/gaiatest/apps/contacts/app.py#L140).

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=24f1ba56d9e81372942570433019c0b0d8df0b4a

PR #33462, opened 2015-12-01
[Do not merge] Bug 1225809 - Conflict-safe homescreen unpinning, r=Cwiiis

michielbdejong:
With this change, unpinning a page will fail (instead of leading to data corruption) in the unlikely event that a change for that same URL is imported from Firefox Sync at the same time.

### DIFF
--- a/apps/homescreen/js/pages.js
+++ b/apps/homescreen/js/pages.js
@@ -219,9 +219,10 @@
       // Remove child immediately, datastore operations can be quite slow
       this.removeCard(this.selectedCard);
 
+      var revId = this.pagesStore.datastore.revisionId;
       this.pagesStore.get(id).then(entry => {
         entry.data.pinned = false;
-        this.pagesStore.datastore.put(entry.data, id).then(() => {},
+        this.pagesStore.datastore.put(entry.data, id, revId).then(() => {},
           e => {
             console.error('Error unpinning page:', e);
           });
--- a/apps/homescreen/test/unit/pages_test.js
+++ b/apps/homescreen/test/unit/pages_test.js
@@ -284,9 +284,15 @@ suite('Pages', () => {
           return Promise.resolve(mockEntry);
         });
       storePutStub =
-        sinon.stub(pages.pagesStore.datastore, 'put', (data, id) => {
+        sinon.stub(pages.pagesStore.datastore, 'put', (data, id, revId) => {
           if (callback) {
             callback(data, id);
+          }
+          if (revId && storePutStub._raceCondition) {
+            return Promise.reject({
+              name: 'ConstraintError',
+              message: 'RevisionId is not up-to-date'
+            });
           }
           return Promise.resolve();
         });
@@ -298,6 +304,7 @@ suite('Pages', () => {
       storeGetStub.restore();
       storePutStub.restore();
       pages.selectedCard = null;
+      callback = null;
     });
 
     test('removes card from container', () => {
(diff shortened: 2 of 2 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5629df0bf8255718beccac9ba14f6ca713046f2e

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7ee37647f04b1446201a80d8e7985bcbd4d58811

apps/homescreen/test/unit/pages_test.js
```diff
@@ -298,6 +304,7 @@ suite('Pages', () => {
       storeGetStub.restore();
       storePutStub.restore();
       pages.selectedCard = null;
+      callback = null;
```
michielbdejong: Needed to avoid `done` being called from previous test when running the new test below.

apps/homescreen/test/unit/pages_test.js
```diff
@@ -314,6 +321,34 @@ suite('Pages', () => {
       };
       pages.unpinSelectedCard();
     });
+
+    suite('in case of DataStore conflict', function(done) {
```
michielbdejong: For Mocha handlers, lexically scoped functions are not supported (they don't get the right Mocha context attached), so even though for most things they still work without the proper context, I prefer to use `function` for all Mocha callbacks. Not sure if you want to mix this in one file, I can change it either way.

apps/homescreen/test/unit/pages_test.js
```diff
@@ -284,10 +284,16 @@ suite('Pages', () => {
...
-        sinon.stub(pages.pagesStore.datastore, 'put', (data, id) => {
+        sinon.stub(pages.pagesStore.datastore, 'put', (data, id, revId) => {
           if (callback) {
             callback(data, id);
           }
+          if (revId && storePutStub._raceCondition) {
+            return Promise.reject({
+              name: 'ConstraintError',
+              message: 'RevisionId is not up-to-date'
+            });
```
michielbdejong: In real life this is a `ConstraintError` object which inherits from `Error`, but just ducktyping it here.

apps/homescreen/test/unit/pages_test.js
Cwiiis: Either we need a comment to explain this or this should match the rest of the file and use fat-arrow functions (I prefer the latter, but if things will break, obviously the former :)) - otherwise I could see myself or someone else changing this to be consistent with the rest of the file. Seems like we're working around a bug elsewhere by using `function`.

apps/homescreen/js/pages.js
```diff
@@ -219,9 +219,10 @@
       // Remove child immediately, datastore operations can be quite slow
       this.removeCard(this.selectedCard);
 
+      var revId = this.pagesStore.datastore.revisionId;
       this.pagesStore.get(id).then(entry => {
         entry.data.pinned = false;
-        this.pagesStore.datastore.put(entry.data, id).then(() => {},
+        this.pagesStore.datastore.put(entry.data, id, revId).then(() => {},
```
Cwiiis: It's incredibly unlikely we hit a race here, right? Just wondering if we want to have some kind of retry code, or if it's just fine for us to fail. Note that we've removed the card from the container, so when the homescreen restarts, the pinned page will reappear. If it's unlikely enough that we can expect it never to happen, I'm happy to not do that.

michielbdejong: Firefox Sync updates each record at most once per hour, and only if you visited that same URL with your desktop browser during that last hour. The update should be fast, so you would have to aim very well to hit that time window with your unpin action. I think for most users it will happen zero times in the life time of their phone.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=be2f23b2a1e1add44866008cbeccc9a43fbd3bca

apps/homescreen/test/unit/pages_test.js
michielbdejong: OK, changed it to fat arrow functions to match the rest of the file.

apps/homescreen/js/pages.js
michielbdejong: Sorry, I said that wrong, it doesn't have to be the same URL - any simultaneous DataStore update will cause a versioning conflict. I'll see if I can improve my patch so that it shows a warning to the user ('Unpinning failed, please try again'), and also does not remove the pinned page from the container if it was not really removed.

Cwiiis: Perhaps rather than showing a warning, it could just retry in X time (say 10 seconds) up to Y times (say 6 times)? It seems unlikely that it would fail over the course of a minute I would suppose, but if it did, perhaps we could show the warning then and add the pinned site back to the container? That last bit could be done by instead of immediately removing, setting display:none on the item and running .synchronise on the container - then if it fails, you can remove the display:none and run synchronise again to show it, and actually remove it on success.

michielbdejong:
I think maybe the current behavior (where unpinning at least unpins, even though there is an almost theoretical risk of getting slightly wrong history data) is better than the one with this patch, will try to come up with something better.

michielbdejong:
Cleaning up PRs that are > 1 year old.

PR #33499, opened 2015-12-03
Bug 1183074 - [Messages] Show all recipients when receiving a group M…

martiansideofthemoon:
…MS r=julienw. 

Would something like this work? @julienw

### DIFF
--- a/apps/sms/desktop-mock/navigator_moz_mobilemessage.js
+++ b/apps/sms/desktop-mock/navigator_moz_mobilemessage.js
@@ -751,6 +751,14 @@
         lastMessageType: 'mms',
         timestamp: now - (60000000 * 20),
         unreadCount: 0
+      },
+      {
+        id: 15,
+        participants: ['+33623456789', '+33612345678'],
+        body: 'Helllo!',
+        lastMessageType: 'mms',
+        timestamp: now - (60000000 * 30),
+        unreadCount: 0
       }
     ]
   };
@@ -782,7 +790,7 @@
   ];
 
   // Procedurally generate a large amount of messages for a single thread
-  // cycling around different delays 
+  // cycling around different delays
   for (i = 0; i < 150; i++) {
     messagesDb.messages.push({
       threadId: 5,
@@ -971,6 +979,24 @@
     }],
     type: 'mms',
     timestamp: now - (60000000 * 20)
+  });
+
+  messagesDb.messages.push({
+    type: 'mms',
+    id: messagesDb.id++,
+    threadId: 15,
+    delivery: 'received',
+    deliveryInfo: [{
+      deliveryStatus: 'success',
+      deliveryTimestamp: 0,
+      readStatus: 'not-applicable',
+      readTimestamp: 0,
+      receiver: null
+    }],
+    sender: '+33623456789',
+    receivers: ['+33623456789','+33612345678'],
+    timestamp: now - (60000000 * 30),
+    sentTimestamp: now - (60000000 * 30)
   });
 
   // Internal publisher/subscriber implementation
--- a/apps/sms/views/conversation/js/information.js
+++ b/apps/sms/views/conversation/js/information.js
@@ -208,7 +208,16 @@ function createListWithMsgInfo(message) {
   var list = [];
   if (message.delivery === 'received' ||
       message.delivery === 'not-downloaded') { // received message
-    list.push(message.sender);
+    if (message.type === 'mms') {
+      if (message.receivers) {
+        list.push(...message.receivers);
+      }
+      if (!list.includes(message.sender)) {
+        list.push(message.sender);
+      }
+    } else {
+      list.push(message.sender);
+    }
   } else if (message.type === 'mms') { // sent mms message
     message.deliveryInfo.forEach(function(info) {
       list.push({number: info.receiver,
--- a/apps/sms/views/conversation/test/unit/information_test.js
+++ b/apps/sms/views/conversation/test/unit/information_test.js
@@ -316,7 +316,8 @@ suite('Information view', function() {
         contactTitle,
         sizeHide,
         sizeContent,
-        isIncoming
+        isIncoming,
+        expectedList
       } = opts;
 
       assert.equal(reportView.container.dataset.delivery, delivery);
@@ -367,7 +368,10 @@ suite('Information view', function() {
         );
       }
 
-      sinon.assert.called(reportView.renderContactList);
+      sinon.assert.calledWithMatch(
+        reportView.renderContactList,
+        expectedList
+      );
     }
 
     test('gaia-header is properly set up', function() {
@@ -389,7 +393,8 @@ suite('Information view', function() {
         sentTitle: 'message-sending',
         contactTitle: 'report-to-title',
         sizeHide: true,
-        isIncoming: false
+        isIncoming: false,
+        expectedList: [ sinon.match({ number: 'receiver' }) ]
       });
     });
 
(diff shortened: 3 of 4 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=50257ec86accd85e274c2a44f1e92857318eb242

apps/sms/views/conversation/js/information.js
```diff
@@ -209,6 +209,12 @@ function createListWithMsgInfo(message) {
   if (message.delivery === 'received' ||
       message.delivery === 'not-downloaded') { // received message
     list.push(message.sender);
+    if (message.type === 'mms') {
+      message.deliveryInfo.forEach(function(info) {
+        list.push({number: info.receiver,
+                   infoBlock: createReportDiv(info)});
+      });
+    }
```
julienw: No this wouldn't work. More information on the bug.

martiansideofthemoon: @julienw I'm sorry for not testing it. I'll need to fix up my desktop environment before I proceed

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=045a8068faff0c1668d593f9340c7f396998c4d3

martiansideofthemoon:
@julienw , made some changes, image on the bug

apps/sms/desktop-mock/navigator_moz_mobilemessage.js
```diff
@@ -973,6 +981,24 @@
...
+      deliveryStatus: 'success',
+      deliveryTimestamp: 0,
+      readStatus: 'not-applicable',
+      readTimestamp: 0,
+      receiver: null
+    }],
+    sender: '+33623456789',
+    receivers: ['+33623456789','+33612345678'],
+    timestamp: now - (60000000 * 30),
+    sentTimestamp: now - (60000000 * 30)
```
julienw: the issue you have is that the code in https://github.com/mozilla-b2g/gaia/blob/5c79eeb800161651b22af9461e3de07e527a0141/apps/sms/views/conversation/js/conversation.js#L1674 expect that `attachments` is either null or an array. Here it's `undefined` because it's not present, and as a result we have an error at that line.

I'd suggest to fix the issue by changing `message.attachments === null` to `!message.attachments` in conversations.js.

julienw: then the message will be in error because a MMS should either have an attachment or a subject, but it's still good enough to open the report

apps/sms/views/conversation/js/information.js
```diff
@@ -208,7 +208,17 @@ function createListWithMsgInfo(message) {
...
+      var numbers = message.receivers;
+      if (message.receivers.indexOf(message.sender)==-1) {
+        numbers.push(message.sender);
+      }
+      numbers.forEach(function(number) {
+        list.push(number);
+      });
+    } else {
+      list.push(message.sender);
+    }
```
julienw: I think you can rewrite this in a simpler way like this:
```javascript
if (message.receivers) {
  list.push(...message.receivers);
}

if (!list.includes(message.sender)) {
  list.push(message.sender);
}
```

julienw: see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_operator for more information about the awesome `...` operator.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f574c383b2a3d249104ca840faa3982e7816fc94

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=45089e9de34c8859f3949b662e973393be645150

martiansideofthemoon:
@julienw , I've added changes to the unit tests.

julienw:
I think you need to rebase to latest master so that the try build work !

It's as easy as "git fetch upstream && git rebase upstream/master" (assuming "upstream" is mozilla's remote name). hopefully there will be no conflict, but if there are some it should be easy to fix.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c2abfbf1f0e8bd13be3cf4a058c0e4fa55a6db8f

package.json
```diff
@@ -9,11 +9,12 @@
     "node": "^4 || ^5"
   },
   "devDependencies": {
+    "SMS": "file:./apps/sms",
     "adm-zip": "^0.4.7",
     "app-root-path": "^1.0.0",
     "archiver": "^0.16.0",
     "async": "^1.5.0",
-    "babel-eslint": "^4.1.5",
+    "babel-eslint": "^5.0.0-beta6",
```
julienw: I think you only needed to run "npm install" from the root directory. Can you please try this again instead of upgrading babel-eslint ?
Because you did the change I think it's better (to be sure it's working) if you run `rm -rf node_modules && make node_modules`, to be sure your node_modules directory is in a right state.

apps/sms/views/conversation/test/unit/information_test.js
```diff
@@ -389,7 +396,8 @@ suite('Information view', function() {
         sentTitle: 'message-sending',
         contactTitle: 'report-to-title',
         sizeHide: true,
-        isIncoming: false
+        isIncoming: false,
+        receivers: messageOpts.receivers,
```
julienw: wondering why you add it when it's obviously undefined ?

apps/sms/views/conversation/test/unit/information_test.js
```diff
@@ -366,8 +367,14 @@ suite('Information view', function() {
...
-
-      sinon.assert.called(reportView.renderContactList);
+      if (receivers) {
+        sinon.assert.calledWithMatch(
+          reportView.renderContactList,
+          receivers.concat('sender')
+        );
+      } else {
+        sinon.assert.calledWith(reportView.renderContactList);
+      }
```
julienw: proposal: instead of getting `receivers`, maybe have `expectedList` as parameter ? So that you can really use it in all tests, and call directly `sinon.assert.calledWithMatch(reportView.renderContactList, expectedList)` ?

What do you think ?

martiansideofthemoon: @julienw , I don't think that's going to be simple. As we are not passing any `receivers` in 7 of the 8 tests, I'm getting something like this
```
Error: expected renderContactList to be called with match ["receiver"]
    renderContactList([{
  infoBlock: <div class="network-status" data-delivery-status="delivered"><span class="icon"><[...]</div>,
  number: "receiver"
}])
```
The 8 tests are expecting really different parameter. So I don't think it is going to be easy to generalize.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9f9a229b6a2e66ff5b9a8f45843ff09e17024143

apps/sms/views/conversation/test/unit/information_test.js
julienw: hey, here is what I had in mind: https://github.com/julienw/gaia/commit/593c6b2b81b7ccaf79723add7c0b582b366d950b

I noticed something not very good happening in these tests: renderContactList can actually be called twice, once in the setup (because we call beforeEnter) and one later on (because we call beforeEnter once again). I don't think this is really important but we'll need to fix this eventually.

martiansideofthemoon: @julienw , what exactly does `sinon.match()` do?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=593c6b2b81b7ccaf79723add7c0b582b366d950b

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=80431cd9222215140af57f84013c468cf59778b0

julienw:
BTW, some documentation about sinon matchers: http://sinonjs.org/docs/#matchers

PR #33519, opened 2015-12-04
Bug 1178812 - Add metadata copying to crop and resize.

hfiguiere:
jpeg-exif.js is a direct copy of
https://github.com/mozilla-b2g/exif-parser/blob/c3dc9b024f3e8cbb78b5b905274e645dfd8e01e4/jpeg.js
It is currently ignored by jshint.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4c875e9264c7fa59f764f6814059f645036ab507

apps/sharedtest/test/unit/crop_resize_rotate_test.js
```diff
@@ -4,13 +4,95 @@
...
+  });
+}
+
+
+suite('Metadata copy test', function() {
+  suiteSetup(function(done) {
+
+    // We need the mock lazy loader for jpeg-exif.
+    window.LazyLoader = window.MockLazyLoader;
+    done();
```
davidflanagan: Since this setup is synchronous you probably don't need the done()

apps/sharedtest/test/unit/crop_resize_rotate_test.js
```diff
@@ -4,13 +4,95 @@
...
+              assert.ok(outBlob);
+
+              JPEG.readExifMetaData(outBlob, function (error, metaData) {
+                assert.ok(metaData);
+                // check metaData
+                assert.equal(metaData.Orientation, 1);
+                if (testFile.dims) {
+                  if (testFile.rotated) {
+                    assert.equal(metaData.PixelXDimension, 480);
+                    assert.equal(metaData.PixelYDimension, 640);
```
davidflanagan: I'm confused about this case. Aren't we asking for an output image that is always 640 wide by 480 high? In the case where cropResizeRotate alters the orientation metadata, shouldn't it also swap the PixelX/YDimensions metadata?

shared/js/media/crop_resize_rotate.js
```diff
@@ -1,7 +1,9 @@
 /* exported cropResizeRotate */
-/* global getImageSize */
+/* global getImageSize, LazyLoader, JPEG */
 /* global Downsample */
 
+'use strict';
```
davidflanagan: Since there is only a single function in this file, I prefer to put the use strict inside the function. That way it is guaranteed to be valid.  Otherwise, if this file gets concatenated to another file, the use strict will be ignored.

shared/js/media/crop_resize_rotate.js
```diff
@@ -72,14 +74,13 @@
...
 //
 function cropResizeRotate(blob, cropRegion, outputSize, outputType,
                           metadata, callback)
 {
-  'use strict';
-
-  const JPEG = 'image/jpeg';
-  const PNG = 'image/png';
+  const JPEGType = 'image/jpeg';
+  const PNGType = 'image/png';
```
davidflanagan: I'd rather alter jpeg-exif.js so that it defined something with a clearer name then JPEG.  Then you could leave this constant unmodified.   Maybe call it JPEGParser or EXIFParser instead of just plain JPEG

apps/sharedtest/test/unit/crop_resize_rotate_test.js
hfiguiere: cropResizeRotate set them properly depending on the orientation - at least that's the results I get. The second sample is vertical and therfore set properly. ```testFile.rotated``` is the "expected result".

apps/sharedtest/test/unit/crop_resize_rotate_test.js
hfiguiere: good point

shared/js/media/crop_resize_rotate.js
```diff
@@ -488,8 +489,49 @@ function cropResizeRotate(blob, cropRegion, outputSize, outputType,
...
+                gotEncodedBlob(outputBlob);
+                return;
+              }
+              metaData.Orientation = 1;
+              if (metaData.PixelXDimension) {
+                metaData.PixelXDimension = destWidth;
+              }
+              if (metaData.PixelYDimension) {
+                metaData.PixelYDimension = destHeight;
+              }
```
davidflanagan: Are you going to specify an edited date?  I don't think that is necessary; just wondering.

hfiguiere: I planned on doing so before submitting for review.

shared/js/media/crop_resize_rotate.js
hfiguiere: OK.

hfiguiere: JPEGParser it will be

apps/sharedtest/test/unit/crop_resize_rotate_test.js
```diff
@@ -4,13 +4,95 @@
...
+      if (xhr.status !== 200) {
+        reject(new Error('Failed with status: ' + xhr.status));
+      } else {
+        resolve(this.response);
+      }
+    };
+    xhr.onerror = xhr.ontimeout = function() {
+      reject(new Error('Failed' + xhr.status));
+    };
+    xhr.responseType = 'arraybuffer';
```
davidflanagan: can you just use `responseType = 'blob'` here and skip the fetchBlob() function below?

apps/sharedtest/test/unit/crop_resize_rotate_test.js
```diff
@@ -18,6 +100,10 @@ function runCropResizeRotateTests(imageWidth, imageHeight) {
     const W = imageWidth, H = imageHeight;  // The size of the test image
 
     suiteSetup(function(done) {
+
+      // We need the mock lazy loader for jpeg-exif.
+      window.LazyLoader = window.MockLazyLoader;
```
davidflanagan: You're going to have to make the exif stuff optional in cropResizeRotate because we don't want to do when creating thumbnails, for example. And if you do that, then you may not need to modify this test.

apps/sharedtest/test/unit/crop_resize_rotate_test.js
hfiguiere: interesting. Maybe we should fix that in the music metadata tests too. that's where it comes from.

apps/sharedtest/test/unit/crop_resize_rotate_test.js
hfiguiere: I plan on making the metadata copy an explicit flag... so that by default it ain't. Then I'll need to hook it up in a couple of places then.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8b37678497d580f763a8c2c678a2a4b0712126b4

shared/js/media/crop_resize_rotate.js
```diff
@@ -72,14 +74,13 @@
 //    shared/js/media/image_size.js
 //    shared/js/media/jpeg_metadata_parser.js
 //    shared/js/media/downsample.js
+//    shared/js/media/jpeg-exif.js (with LazyLoader)
 //
 function cropResizeRotate(blob, cropRegion, outputSize, outputType,
                           metadata, callback)
```
davidflanagan: cropResizeRotate() is also used for creating thumbnails and previews for images, and in those cases we don't want to put the EXIF back in. So you're going to have to add another boolean parameter to specify that we want the EXIF. You could put the new parameter after the callback, though that makes it sort of ugly to call.  Better maybe to define a new outputType value "image/jpeg+exif" to specify that a jpeg with exif is wanted.

shared/js/media/crop_resize_rotate.js
```diff
@@ -488,8 +489,49 @@ function cropResizeRotate(blob, cropRegion, outputSize, outputType,
...
+      function copyMetadata(outputBlob) {
+        try {
+          LazyLoader.load(['shared/js/media/jpeg-exif.js'], () => {
+            JPEG.readExifMetaData(inputBlob, (error, metaData) => {
+
+              if (error || !metaData) {
+                gotEncodedBlob(outputBlob);
+                return;
+              }
+              metaData.Orientation = 1;
```
davidflanagan: Does Diego's code define symbolic constants for these property names? Can you do `metaData[JPEG.EXIF.ORIENTATION] = 1` instead?  That might be a bit cleaner, but maybe just too verbose to be worthwhile.

shared/js/media/crop_resize_rotate.js
hfiguiere: output format as EXIF isn't a bad idea at all. Actually so much simpler. (Exif implies JPEG)

shared/js/media/crop_resize_rotate.js
hfiguiere: The way I use it is the way it is defined.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c04f704101102215a0ee188996ca9ec7276d397e

apps/sharedtest/test/unit/crop_resize_rotate_test.js
```diff
@@ -4,13 +4,116 @@
...
+
+  [
+    // this one is horizontal
+    { fileName: '/test-data/blue.jpg', dims: true, rotated: false,
+      targetSize: { width: 640, height: 480 } },
+    { fileName: '/test-data/blue.jpg', dims: true, rotated: false,
+      targetSize: null },
+    // this one is vertical (Exif orientation)
+    { fileName: '/test-data/red.jpg', dims: true, rotated: true,
+      targetSize: { width: 640, height: 480 } },
```
punamdahiya: Even though its rotated image,  please check if we need to pass output size as 480 x 640 for vertical image.

hfiguiere: nope that's not it. :-(

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3a85384d0b6894ecbe7f9db7e2bd706575030e11

apps/sharedtest/test/unit/crop_resize_rotate_test.js
hfiguiere: And I was wrong. This was it in the end. Updating the patch.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=822676f0ef4df3987a1e73aec1219008835818e7

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d005b0c36fa71a87dea0f3ac86c01e91e7cf74b7

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=07d07afafcf8b354183ba3140ca00af13d1549c5

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=53637f4f7ffe7cc339efafe984a4aa99fb48a47f

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a0bd3d19eaec59c3d2ef3cf965d7e89bb9037cdc

.jshintignore
```diff
@@ -24,6 +24,7 @@ apps/clock/js/ext/**
 apps/clock/js/text_builder.js
 shared/js/uuid.js
 shared/js/intl/l20n*.js
+shared/js/media/jpeg-exif.js
```
davidflanagan: How hard would it be to make this file pass jshint? Maybe file a followup bug to do that?

hfiguiere: I'll file a follow up bug for that. I'd love to see it pass, but I need to do the changes upstream. Lots and lots of linting errors (like " instead of ')

hfiguiere: Done https://bugzilla.mozilla.org/show_bug.cgi?id=1233808

apps/gallery/js/pick.js
```diff
@@ -238,9 +240,14 @@ var Pick = (function() {
       pickType = null;   // Return unchanged or convert to JPEG
     }
 
-    if (pickType && pickType !== 'image/jpeg' && pickType !== 'image/png') {
+    if (pickType && pickType !== 'image/jpeg' &&
+        pickType !== 'image/jpeg+exif' && pickType !== 'image/png') {
       pickType = null;   // Return unchanged or convert to JPEG
     }
+    else if (!pickType && pickedFileInfo.type == 'image/jpeg') {
```
davidflanagan: I think maybe you shouldn't have the else on this line because it won't run if you set pickType to null in the if clause above.  But note that it does run in the two other places where pickType is set to null.

apps/gallery/js/pick.js
```diff
@@ -220,12 +220,14 @@ var Pick = (function() {
     // then we use null as the type. This value is passed to
     // cropResizeRotate() and will leave the image unchanged if possible
     // or will use jpeg if changes are needed.
+    // EXIF should be preserved.
```
davidflanagan: You're handling the pick case here, but don't you also need to handle share?  See frames.js:shareSingleItem()
Or file a followup bug for this case.

apps/gallery/js/pick.js
```diff
@@ -238,9 +240,14 @@ var Pick = (function() {
       pickType = null;   // Return unchanged or convert to JPEG
     }
 
-    if (pickType && pickType !== 'image/jpeg' && pickType !== 'image/png') {
+    if (pickType && pickType !== 'image/jpeg' &&
+        pickType !== 'image/jpeg+exif' && pickType !== 'image/png') {
       pickType = null;   // Return unchanged or convert to JPEG
     }
+    else if (!pickType && pickedFileInfo.type == 'image/jpeg') {
+      // if we picked a JPEG image, then explictely force EXIF copy.
```
davidflanagan: nit: spelling

apps/sharedtest/test/unit/crop_resize_rotate_test.js
```diff
@@ -4,13 +4,123 @@
...
+              assert.equal(error, null);
+              assert.ok(outBlob);
+
+              JPEGParser.readExifMetaData(
+                outBlob, function (error, metaData) {
+                  assert.ok(metaData);
+                  // check metaData
+                  assert.equal(metaData.Orientation, 1);
+
+                  if (testFile.dims) {
```
davidflanagan: dims is true in all 6 of your test cases, so I'm not sure if you really need this...

apps/gallery/js/pick.js
hfiguiere: uh. did I miss that case. Looks like it :-/

apps/sharedtest/test/unit/crop_resize_rotate_test.js
```diff
@@ -4,13 +4,123 @@
...
+                        expectedX = 1280;
+                        expectedY = 960;
+                      } else {
+                        expectedX = 960;
+                        expectedY = 1280;
+                      }
+                    }
+
+                    assert.equal(metaData.PixelXDimension, expectedX);
+                    assert.equal(metaData.PixelYDimension, expectedY);
```
davidflanagan: It would be a stronger test if you also verified that some of the other EXIF data is preserved. Date, focal length, flash, etc. Maybe the test case objects in the array above could each have a few pieces of EXIF metadata to check.

apps/gallery/js/pick.js
hfiguiere: It is deliberate. We want to set the picktype to Exif if we requested nothing and picked a JPEG. It is really the else case.

shared/js/media/crop_resize_rotate.js
```diff
@@ -255,7 +259,8 @@ function cropResizeRotate(blob, cropRegion, outputSize, outputType,
     }
 
     // Make sure the outputType is valid, if one was specified
-    if (outputType && outputType !== JPEG && outputType !== PNG) {
+    if (outputType && outputType !== JPEG &&
+        outputType !== PNG && outputType !== EXIF) {
       callback('unsupported outputType: ' + outputType);
       return;
```
davidflanagan: There is a problem a little bit below here at line 274.  It doesn't show in your diff so I can't comment on the line directly.  The problem is that if you pass an image/jpeg blob with outputType = 'image/jpeg+exif' and nothing at all needs to be changed, the code at line 274 will still think that the type is changing and so it will do lots of work and copy image instead of just returning it directly.

I recommend that after verifying the output value here, you add this code below:

```
var preserveEXIF = false;
if (outputType === EXIF) {
  preserveEXIF = true;
  outputType = JPEG;
}
```
Then in any of your code below that has a special case for outputType === EXIF you modify it to use this new preserveEXIF flag.

apps/gallery/js/ImageEditor.js
```diff
@@ -298,10 +300,16 @@ function editPhoto(n) {
     var imagesize = metadata.width * metadata.height;
     var maxsize = CONFIG_MAX_EDIT_PIXEL_SIZE || CONFIG_MAX_IMAGE_PIXEL_SIZE;
 
+    LazyLoader.load(['shared/js/media/jpeg-exif.js'], () => {
```
davidflanagan: The file might not be a jpeg. You should only do this if it is jpeg, I think.
Also, you seem to be speculatively loading the exif with the assumption that it will be read before the user finished editing the image. That is a good idea, but I think you should add a comment making it clear that that is what you're doing.

davidflanagan: Also, maybe you should defer this code until after you've run the code below that might resize the image. Otherwise you might have cropResizeRotate() and this code both parsing exif data at the same time, which would use more memory than desireable.

apps/gallery/js/ImageEditor.js
```diff
@@ -298,10 +300,16 @@ function editPhoto(n) {
...
+      JPEGParser.readExifMetaData(file, (error, metaData) => {
+        editSettings.EXIF = metaData;
+      });
+    });
+
     if (metadata.rotation || metadata.mirrored || imagesize > maxsize) {
       Spinner.show();
       cropResizeRotate(file, null, maxsize || null,
-                       null, metadata,
+                       'image/jpeg+exif', metadata,
```
davidflanagan: I think you only want to pass image/jpeg+exif if the file is already a jpeg. Otherwise use null to leave png files unchanged.  

Note that this code can be triggered for large PNGs on low memory devices. This is not only an EXIF rotation thing.

apps/gallery/js/ImageEditor.js
```diff
@@ -775,6 +783,7 @@ function saveEditedImage() {
  */
 function ImageEditor(imageBlob, container, edits, ready, croponly) {
   this.imageBlob = imageBlob;
+  this.imageExif = edits ? edits.EXIF : null;
```
davidflanagan: I'm sort of confused now. I don't think you need the EXIF until the user has completed their edits. And I thought that the code above that reads the exif was there to be reading in the background. But I think it is likely that it won't be done by the time that this constructor is called.  So can you just wait and look at edits.EXIF later on when you need it instead of using this new variable?

apps/gallery/js/ImageEditor.js
```diff
@@ -1221,8 +1229,51 @@ ImageEditor.prototype.getFullSizeBlob = function(type, done, progress) {
...
+    }
+
+    function updateExif(blob, metaData, callback) {
+      metaData.Orientation = 1;
+      if (metaData.PixelXDimension) {
+        metaData.PixelXDimension = canvas.width;
+      }
+      if (metaData.PixelYDimension) {
+        metaData.PixelYDimension = canvas.height;
+      }
```
davidflanagan: Do you want to set DateTime to the current time and ensure that DateTimeOriginal is set to the original time here?  If we modify the app to sort photos by their EXIF date instead of their file date, that will be helpful to have.  But it doesn't have to be part of this patch.

apps/gallery/js/ImageEditor.js
```diff
@@ -1221,8 +1229,51 @@ ImageEditor.prototype.getFullSizeBlob = function(type, done, progress) {
...
+          if (error) {
+            console.error('Error' + error);
+          }
+          // Process modified file
+          callback(modifiedBlob);
+        });
+    }
+
+    var originalBlob = this.imageBlob;
+    var originalExif = this.imageExif;
```
davidflanagan: If you use this.edits.EXIF here instead of this.imageExif, you might actually have a value ready for use.

apps/sharedtest/test/unit/crop_resize_rotate_test.js
hfiguiere: let's remove it then

apps/sharedtest/test/unit/crop_resize_rotate_test.js
hfiguiere: good idea.

shared/js/media/crop_resize_rotate.js
hfiguiere: good idea.

apps/gallery/js/ImageEditor.js
hfiguiere: I could change it here, but then I need to have code to write date time in exif. I'll file a followup bug instead

apps/gallery/js/pick.js
hfiguiere: Filed  https://bugzilla.mozilla.org/show_bug.cgi?id=1237367

apps/gallery/js/ImageEditor.js
hfiguiere: actually I think I want to pass null here. This is just the resize / rotate before edit. changing.

apps/gallery/js/ImageEditor.js
hfiguiere: edits.EXIF isn't available later when the image is saved.

apps/gallery/js/ImageEditor.js
hfiguiere: yeah it works. removing this.imageExif

apps/gallery/js/ImageEditor.js
hfiguiere: it actually is, see below. removing this.

apps/gallery/js/ImageEditor.js
hfiguiere: I actually forfeit metadata on the cropResizeRotate() now. But I need to capture the Exif anyhow. So we parse it once, and here. And only if the blob is JPEG.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0871d265380e78a0551f80fee8949ae23dbb78d1

apps/gallery/js/ImageEditor.js
```diff
@@ -1221,8 +1230,51 @@ ImageEditor.prototype.getFullSizeBlob = function(type, done, progress) {
...
+          // Process modified file
+          callback(modifiedBlob);
+        });
+    }
+
+    var originalBlob = this.imageBlob;
+    var originalExif = this.edits.EXIF;
+
+    // Finally, convert the edited image to a blob and pass to the callback.
+    canvas.toBlob(function(blob) {
```
davidflanagan: I didn't notice this on my first review, but it would be good to move the code in gotEncodedBlob() that frees the canvas memory and put it here.  Since the EXIF editing will be somewhat memory intensive, we should release as much memory as soon as we can.  So I think that the only code that needs to remain in gotEncodedBlob() is the stuff that updates the progress bar to 1.0.

At the point when toBlob() is called the progress bar is at 95% complete. It might be good UX to modify the code so that it is at 90% when toBlob() is called, then update it to 95% before updating the EXIF.  I haven't tried it yet, so I don't know if there is any kind of noticeable pause in the progress bar before the edit finishes or not.

apps/gallery/js/ImageEditor.js
```diff
@@ -1221,8 +1230,51 @@ ImageEditor.prototype.getFullSizeBlob = function(type, done, progress) {
...
+      if (type == 'image/jpeg') {
+        if (originalExif) {
+          updateExif(blob, originalExif, gotEncodedBlob);
+        } else {
+          LazyLoader.load(['shared/js/media/jpeg-exif.js'], () => {
```
davidflanagan: It might be nice to have a comment here explaining that originalExif is almost certainly already set and that this else clause is only here in case the original call did not return yet.

apps/gallery/js/ImageEditor.js
```diff
@@ -298,6 +300,14 @@ function editPhoto(n) {
     var imagesize = metadata.width * metadata.height;
     var maxsize = CONFIG_MAX_EDIT_PIXEL_SIZE || CONFIG_MAX_IMAGE_PIXEL_SIZE;
 
+    if (file.type === 'image/jpeg') {
+      LazyLoader.load(['shared/js/media/jpeg-exif.js'], () => {
+        JPEGParser.readExifMetaData(file, (error, metaData) => {
+          editSettings.EXIF = metaData;
```
davidflanagan: If there is an error for some reason, we don't want to repeat the call to readExifMetaData in the code below, so you should be sure to set editSettings.EXIF in either case. Maybe you just want:

```
editSettings.EXIF = metaData || {};
```

shared/js/media/crop_resize_rotate.js
```diff
@@ -255,10 +259,16 @@ function cropResizeRotate(blob, cropRegion, outputSize, outputType,
...
     }
+    var preserveExif = false;
+    if (outputType === EXIF) {
+      preserveExif = true;
```
davidflanagan: Actually, it occurs to me now that preserveExif should only be true if `blob.type === JPEG` Otherwise there is no exif to preserve.  We probably don't have any code that calls cropResizeRotate with outputType EXIF if the input blob is not a JPEG, but to be completely correct, we should check for that here so we don't try to read metadata from a PNG later on.

shared/js/media/crop_resize_rotate.js
```diff
@@ -488,8 +498,48 @@ function cropResizeRotate(blob, cropRegion, outputSize, outputType,
...
       }
 
+      var inputBlob = blob; // XXX just use a different parameter name
+
       // Finally, encode the image into a blob
-      canvas.toBlob(gotEncodedBlob, outputType || JPEG);
+
+      var postCallback = preserveExif ? copyMetadata : gotEncodedBlob;
+
+      canvas.toBlob(postCallback, !outputType ? JPEG : outputType);
```
davidflanagan: For the second argument, `outputType || JPEG` seems clearer to me. Isn't that the same as your conditional operator?

apps/camera/index.html
```diff
@@ -29,6 +29,7 @@
     <!-- <script src="/shared/js/media/media_frame.js"></script> -->
     <!-- <script src="/shared/js/media/downsample.js"></script> -->
     <!-- <script src="/shared/js/media/image_size.js"></script> -->
+    <!-- <script src="/shared/js/media/jpeg-exif.js"></script> -->
```
davidflanagan: I don't see any code that modifies the camera pick activity to use image/jpeg+exif. I think that means that this file will never be lazy loaded in the camera app.  So I suspect you can delete this line. 

So I'd suggest that you either:

1) delete this line and file a followup bug for the camera pick activity.

2) modify the camera pick activity to use image/jpeg+exif. I think that change would go in apps/camera/js/lib/resize-image-and-save.js

hfiguiere: uh oh. I'll remove it and file a followup bug.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ccfbfbccb112c2e6607beb60a27a8816d7dde5a5

PR #33531, opened 2015-12-07
Bug 1226228 - moved timer control buttons down a bit to be centered R…

jgw96:
…=Justin Willis

### DIFF
--- a/apps/clock/style/timer.css
+++ b/apps/clock/style/timer.css
@@ -39,7 +39,7 @@
 
 #timer-controls {
   display: flex;
-  margin: 2rem 1rem 0;
+  margin: 7rem 1rem 0;
   padding: 0;
 }
 


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=89ab1848cf3b4d5b7d75ba7f2b18f3b57f833668

julienw:
I haven't tested the patch, but remember the bug is about centering the text _within_ the button, and I don't think ths patch is doing this.

First step is: look why the text is not centered correctly. It _should_ be centerd correctly by default. Then and only then you can try to fix the issue.

julienw:
Ok, I confirm that you tried to fix the wrong bug :)

julienw:
I see that for similar buttons, we have a line-height property: https://github.com/mozilla-b2g/gaia/blob/578c972114e6ad0b3284515b0d29c9d72073961a/apps/clock/style/stopwatch.css#L33

You might want to add the same one in https://github.com/mozilla-b2g/gaia/blob/578c972114e6ad0b3284515b0d29c9d72073961a/apps/clock/style/timer.css#L46

This seems to fix the bug for me. This is really shitty CSS but fixing the whole CSS would be a lot more work and I'm not sure this is worth it currently.

PR #33532, opened 2015-12-07
Bug 1140580 - Fix long custom email tag going off screen R=Justin Willis

jgw96:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b4369aff347555b352f7e3ad467bee304191e0c6

arcturus:
Thanks for the patch!

Tested on the phone and works perfectly, but unfortunately that's a shared file, which means there are more applications using it. So if we add this patch we will be modifying almost all apps using the same style.

We will need to overwrite this values in the specific css files for the contacts app, you can try here:
https://github.com/mozilla-b2g/gaia/blob/master/apps/communications/contacts/style/app.css

Thanks again!

PR #33573, opened 2015-12-15
Bug 1097827 - Move manual setting of aria-label in JS code to use l10nId 

ssaavedra:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1c85bfc65e54d8fab0165d64f0ed1cd60d3d9a19

apps/calendar/js/views/month_day.js
```diff
@@ -76,11 +76,11 @@ MonthDay.prototype = {
     var holder = this.element.querySelector('.busy-indicator');
 
     if (count > 0) {
-      holder.setAttribute('aria-label', navigator.mozL10n.get('busy', {
-        n: count
-      }));
+      holder.setAttribute('data-l10n-id', 'busy');
+      holder.setAttribute('data-l10n-args', JSON.stringify({n: count}));
     } else {
-      holder.removeAttribute('aria-label');
```
yzen: Actually lets keep ```holder.removeAttribute('aria-label');``` I believe the label is not cleared when you delete data-l10n attributes.

yzen:
I think it's best to make changes to apps/pdfjs/... upstream and then grab a version from there once it's tagged. See https://github.com/mozilla-b2g/gaia/blob/master/apps/pdfjs/README.md

PR #33580, opened 2015-12-16
Bug 1219680 - added Gij test: passcode is disabled so it will slide to unlock to homescreen

snowmantw:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=507e78a00356cd265b2da658dca062fef091346c

apps/system/test/marionette/lockscreen_unlock_to_homescreen_test.js
```diff
@@ -0,0 +1,55 @@
+'use strict';
+
+var dirapps = require('path').resolve(__dirname + '/../../../');
+var Settings = require(dirapps + '/settings/test/marionette/app/app'),
+    LockScreen = require('./lib/lockscreen'),
+    LockScreenPasscodeUnlockActions = require(
+      './lib/lockscreen_passcode_unlock_actions'),
+    Promise = require('es6-promise').Promise, // jshint ignore:line
```
timdream: Are we not getting Promise out of box on our node runtime?

snowmantw: Well its an **intermittent** error only happens on TaskCluster. From my previous 30 times test result, it seemed that environments on **some** instances lack of native Promise, which was supposed to be either failed on all instances or passed without any problem. So the best way to work this around without digging into the TaskCluster infrastructure is to require it explicitly.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=97772495f5c0b92f9dee338bc1d906c43620d11c

PR #33591, opened 2015-12-16
Bug 1219636 - Implement *test_import_contacts_from_sdcard.py* as an i…

arcturus:
…ntegration test in JavaScript

### DIFF
--- a/apps/communications/contacts/test/marionette/import_sdcard_test.js
+++ b/apps/communications/contacts/test/marionette/import_sdcard_test.js
@@ -1,0 +1,55 @@
+'use strict';
+
+var Contacts = require('./lib/contacts');
+var assert = require('assert');
+
+marionette('Import from SDCARD', function() {
+  var profile = Contacts.config;
+  // Enable device storage
+  var prefs = ['device.storage.enabled',
+        'device.storage.testing',
+        'device.storage.prompt.testing'];
+  prefs.forEach(pref => {
+    profile[pref] = true;
+  });
+  var client = marionette.client({
+    profile: profile,
+    desiredCapabilities: { raisesAccessibilityExceptions: false }
+  });
+  var contactsApp, selectors;
+
+  setup(function() {
+    client.fileManager.removeAllFiles();
+    client.fileManager.add({
+      type: 'text',
+      filePath: 'apps/communications/contacts/test/marionette/data/vcard_4.vcf'
+    });
+
+    contactsApp = new Contacts(client);
+    contactsApp.launch();
+    selectors = Contacts.Selectors;
+  });
+
+  test('> one contact in one file', function() {
+    // Go to settings
+    client.helper.waitForElement(selectors.settingsButton).tap();
+
+    // Now click on import contacts button
+    client.helper.waitForElement(selectors.importContacts).tap();
+    client.helper.waitForElement(selectors.importSDCard).tap();
+
+    // Get back to the list
+    client.switchToShadowRoot(client.helper
+      .waitForElement(selectors.importHeader));
+    contactsApp.clickOn(client.helper.waitForElement(selectors.actionButton));
+    client.switchToShadowRoot();
+    contactsApp.waitForFadeIn(client.helper
+      .waitForElement(selectors.settingsView));
+    client.helper.waitForElement(selectors.settingsClose).tap();
+    contactsApp.waitForFadeIn(client.helper.waitForElement(selectors.list));
+
+    // Check that we have 1 contact and is the one from the SD,
+    var firstContact = client.findElement(selectors.listContactFirstText);
+    assert.equal(firstContact.text(), 'Forrest Gump');
+  });
+});
--- a/apps/communications/contacts/test/marionette/lib/contacts.js
+++ b/apps/communications/contacts/test/marionette/lib/contacts.js
@@ -8,6 +8,7 @@
 function Contacts(client) {
   this.client = client;
   this.client.setSearchTimeout(10000);
+  this.actions = this.client.loader.getActions();
 }
 
 /**
@@ -121,10 +122,15 @@ Contacts.Selectors = {
   iceGroupOpen: '#section-group-ice',
   iceContact: '#ice-group .contact-item',
 
+  importContacts: '#importContacts',
+  importSDCard: '#import-sd-option button',
+  importHeader: '#import-settings-header',
+
   activityChooser: 'form[data-type="action"]',
   buttonActivityChooser: 'form[data-type="action"] button',
   actionMenu: '#action-menu',
   actionMenuList: '#value-menu',
+  actionButton: '.action-button',
 
   multipleSelectSave: '#save-button',
   multipleSelectStatus: '#statusMsg p',


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d1614bf1af2a5e9fe81b0791d57b30d94f5a0ffc

apps/communications/contacts/test/marionette/import_sdcard_test.js
```diff
@@ -0,0 +1,60 @@
+'use strict';
+
+var Contacts = require('./lib/contacts');
+var assert = require('assert');
+
+marionette('Import from SDCARD', function() {
+  var profile = Contacts.config;
+  // Enable device storage
+  var prefs = ['device.storage.enabled',
```
JohanLorenzo: Defining the prefs here is not needed, there already are under `Contacts.config`.

apps/communications/contacts/test/marionette/import_sdcard_test.js
```diff
@@ -0,0 +1,60 @@
...
+        'device.storage.testing',
+        'device.storage.prompt.testing'];
+  prefs.forEach(pref => {
+    profile[pref] = true;
+  });
+  var client = marionette.client({
+    profile: profile,
+    desiredCapabilities: { raisesAccessibilityExceptions: false }
```
JohanLorenzo: I'd put that under `Contacts.config` as this capabitilities is already defined in 8 other tests.

apps/communications/contacts/test/marionette/import_sdcard_test.js
```diff
@@ -0,0 +1,60 @@
...
+  });
+  var actions, subject, selectors;
+
+  setup(function() {
+    // Remove all files in temp device storage.
+    client.fileManager.removeAllFiles();
```
JohanLorenzo: This line should be in `tearDown()`

apps/communications/contacts/test/marionette/import_sdcard_test.js
```diff
@@ -0,0 +1,60 @@
...
+
+    // Add the default vcard file
```
JohanLorenzo: This comment seems redundant with the command below

apps/communications/contacts/test/marionette/import_sdcard_test.js
```diff
@@ -0,0 +1,60 @@
...
+    subject = new Contacts(client);
+    actions = client.loader.getActions();
+    subject.actions = actions;
+    subject.launch();
+    selectors = Contacts.Selectors;
+  });
+
+  test('> one contact in one file', function() {
+    // Go to settings
+    client.helper.waitForElement(selectors.settingsButton).click();
```
JohanLorenzo: We can make this test easier to read, if we implement the [Page Object pattern](https://wiki.mozilla.org/B2G/QA/Automation/Style_Guide/Best_Practices#The_Page_Object_Pattern). We could have something like:
```js
test('> one contact in one file', function() {
  var settings = contactsApp.goToSettings();
  var importContact = settings.goToImportContacts();
  var importContactFromSdCardModal = importContact.startImportFromSdCard()
  importContact = importContactFromSdCardModal.waitToBeOver();

  settings = importContact.goBack();
  contacts = settings.goBack();
  assert.deepStrictEqual(contactsApp.contacts, ['Forrest Gump']);
});
```
Having these kind of steps in the tests brings us closer to what an end-user intends to do.
Then, the implementation details (aka how a user would actually interact with the UI) should belong in the page class.
For example, in `contacts.js`:
```js
goToSettings: function() {
   this.client.helper.waitForElement(selectors.settingsButton).tap();
   return new Settings(this.client);
}
```

And then, you create another file, that is in charge of the "Settings page":
```js
function Settings(client) {
   this.client = client;
   this.client.helper.waitForElement(selectors.settingsHeader);
}
```

And so on.

Also, there is a transition when you tap on "import contacts". A good way to make sure we wait on it is the following:
```js
function ImportContacts(client) {
   this.client = client;
   var panel = this.client.helper.waitForElement(selectors.importContactsPanel);
   client.helper.waitFor(function() {
      return panel.rect().x == 0
   })
}
```

apps/communications/contacts/test/marionette/import_sdcard_test.js
```diff
@@ -0,0 +1,60 @@
...
+
+    // Now click on import contacts button
+    client.helper.waitForElement(selectors.importContacts).click();
```
JohanLorenzo: For Firefox OS, the convention is usually to use `tap()` instead of `click()`;

apps/communications/contacts/test/marionette/import_sdcard_test.js
```diff
@@ -0,0 +1,60 @@
...
+    client.helper.waitForElement(selectors.importSDCard).click();
+
+    // Get back to the list
+    client.switchToShadowRoot(client.helper
+      .waitForElement(selectors.importHeader));
+    subject.clickOn(client.helper.waitForElement('.action-button'));
```
JohanLorenzo: As a matter of consistency, please move this CSS selector in `lib/contact`

apps/communications/contacts/test/marionette/import_sdcard_test.js
```diff
@@ -0,0 +1,60 @@
...
+
+    // Add the default vcard file
+    client.fileManager.add({
+      type: 'text',
+      filePath: 'apps/communications/contacts/test/marionette/data/vcard_4.vcf'
+    });
+
+    subject = new Contacts(client);
+    actions = client.loader.getActions();
+    subject.actions = actions;
```
JohanLorenzo: Injecting this dependency doesn't look correct, as the `subject` is already constructed with `client`. 

I'd retrieve the action at [this line](https://github.com/mozilla-b2g/gaia/blob/master/apps/communications/contacts/test/marionette/lib/contacts.js#L287) (which is the only one using an action) with directly with:
`var action = this.client.loader.getActions();`

apps/communications/contacts/test/marionette/import_sdcard_test.js
```diff
@@ -0,0 +1,60 @@
...
+    // Remove all files in temp device storage.
+    client.fileManager.removeAllFiles();
+
+    // Add the default vcard file
+    client.fileManager.add({
+      type: 'text',
+      filePath: 'apps/communications/contacts/test/marionette/data/vcard_4.vcf'
+    });
+
+    subject = new Contacts(client);
```
JohanLorenzo: TL;DR: I'd use `contactsApp` instead of `subject`.

I know it's been used at some other places in the tests, and I'm fine if we don't change it.

However, `subject` is not a self-explanatory name. You have to do an indirection (that is to say, either remember, or scroll back where it's defined, or take a look at the file path) to remember that's the contact app you're testing. This can be confusing if you read many tests at the same time. For example, if somebody breaks many tests in many apps, because he/she changed something in `/shared`.

apps/communications/contacts/test/marionette/import_sdcard_test.js
arcturus: I don't necessary agree. I didn't want to modify the rest of the test by changing the configuraiton, so explicitly I'm asking for a specific configuration for this test.

apps/communications/contacts/test/marionette/import_sdcard_test.js
arcturus: Agree to be on the tearDown, but definitely in the setup i want to be sure the first test is not polluted.

apps/communications/contacts/test/marionette/import_sdcard_test.js
arcturus: Oks, thx :)

apps/communications/contacts/test/marionette/import_sdcard_test.js
arcturus: Right.

apps/communications/contacts/test/marionette/import_sdcard_test.js
arcturus: Right.

apps/communications/contacts/test/marionette/import_sdcard_test.js
JohanLorenzo: Well, er, there is no need to change the configuration. The configuration is already the same as here: https://github.com/arcturus/gaia/blob/bug-1219636/apps/communications/contacts/test/marionette/lib/contacts.js#L20-L22

apps/communications/contacts/test/marionette/import_sdcard_test.js
arcturus: I understand, and agree, but the support for gij iin contacts is really old, actually the very first one apps. I would consider to move to Page Object Pattern in a follow up (if we do it), but now prefer to have the tests up and running.

JohanLorenzo: Sounds fine to me.

apps/communications/contacts/test/marionette/import_sdcard_test.js
arcturus: I kind of agree, subject could be identified by the file name, but let's add the proper name so will be easier for everyone.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ed85190ff166b1d3fadf284e01f1bf9b19cc6ff1

PR #33603, opened 2015-12-17
Bug 1233355 - Add settings permissions to Sync app so it can use wind…

ferjm:
…ow.DUMP. r=michielbdejong

### DIFF
--- a/apps/sync/manifest.webapp
+++ b/apps/sync/manifest.webapp
@@ -8,6 +8,9 @@
      "name": "The Gaia Team",
      "url": "https://github.com/mozilla-b2g/gaia"
   },
+  "permissions": {
+    "settings": { "access": "readonly" }
+  },
   "datastores-owned": {
     "sync_bookmarks_store": {
       "access": "readwrite",


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0a6a42659854e8cafdec71389b30b321091ed852

michielbdejong:
Isn't there a more elegant way to do this? IMHO logging is quite basic and should be unrelated to reading/writing settings. If only apps that read/write settings are allowed to write to the log then we should at least document that fact on https://developer.mozilla.org/en-US/Apps/Build/App_permissions#settings?

apps/sync/manifest.webapp
```diff
@@ -8,6 +8,9 @@
      "name": "The Gaia Team",
      "url": "https://github.com/mozilla-b2g/gaia"
   },
+  "permissions": {
+    "settings": { "access": "readwrite" }
```
michielbdejong: please add `"description": "Necessary to see if logging is enabled"` and change to `readonly` as discussed on irc

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=b01299269907c0516689f621a6d8e197a4e79ad7

PR #33613, opened 2015-12-18
Bug 1232363 - system app should have all audio channel permissions.

alastor0325:
In bug1214148, we try to solve the issue of the nested-mozBrowserFrame.

If we're in the nested case, the audio channels of the children need to be the subset of the parent's. Therefore, the system app should have all audio channel permission.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c688c84f51382f3866ba74de02ffc77ebd66a566

PR #33662, opened 2015-12-23
Bug 1179604 - Inbox should emit fully-loaded only when Drafts are loaded as well. r=schung

azasypkin:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c22a6fbb798972fda1845f6f34d41bc9204f8f0e

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c22a6fbb798972fda1845f6f34d41bc9204f8f0e

apps/sms/test/marionette/lib/messages.js
```diff
@@ -113,6 +114,7 @@ var ParticipantsAccessor = require('./views/participants/accessors');
           client.switchToFrame();
 
           client.apps.switchToApp(ORIGIN_URL);
+          client.helper.waitForElement(SELECTORS.appReady);
```
azasypkin: It may be required when we call switch to for the app when it's run as activity

apps/sms/test/marionette/new_activity_test.js
```diff
@@ -102,7 +102,7 @@ marionette('Messages as "new" activity target', function() {
         return !!window.wrappedJSObject.Compose;
       }), 'XSS should not be performed');
 
-      assert.isTrue(messagesApp.Composer.sendButton.enabled());
+      client.waitFor(() => messagesApp.Composer.sendButton.enabled());
```
azasypkin: I've seen a bunch of failures in this file with disabled retry logic - so found several weak places and fixed, it's still possible to see failures with disabled retry logic ~ 1 in 30-40 runs, but it seems related to the Treeherder not sure we can do anything here - something is very slow somewhere.

apps/sms/views/inbox/js/startup.js
```diff
@@ -47,6 +47,10 @@
         initLazyDependencies();
       });
 
+      InboxView.once('fully-loaded', () => {
+        App.setReady();
```
steveck-chung: I think it's not related to the issue, right? It seems just for fixing split mode case.
BTW do you think we'll need unit test for it?

apps/sms/test/marionette/lib/messages.js
```diff
@@ -63,11 +64,11 @@ var ParticipantsAccessor = require('./views/participants/accessors');
 
         Participants: new ParticipantsAccessor(client),
 
-        launch: function() {
+        launch() {
           client.switchToFrame();
           client.apps.launch(ORIGIN_URL);
           client.apps.switchToApp(ORIGIN_URL);
-          client.helper.waitForElement(SELECTORS.main);
+          client.helper.waitForElement(SELECTORS.appReady);
```
steveck-chung: Seems like we can remove SELECTORS.main because we didn't need it for other place?

apps/sms/test/marionette/new_activity_test.js
steveck-chung: Not quite sure about the timing of this `client.waitFor`, maybe we can wait for specific element in `launchAsActivity`?

steveck-chung: Looks like `launchAsActivity` already has `switchTo` inside and it should wait for body element with app ready class. If it still not enough, maybe we could add a class to composer to represent composer status is ready?

apps/sms/test/marionette/lib/messages.js
azasypkin: Well, we'll need it in the upcoming patch that revives Inbox tests, see https://github.com/azasypkin/gaia/commit/7aa2b76b0fbbbda38b4809c69ecb652080f30e26#diff-a3f0ad18f54e1750f7f7ecbbc2356f25R71

apps/sms/views/inbox/js/startup.js
azasypkin: It's not related to this particular issue, right, but it's required for the integration test that we have for split view mode - https://github.com/azasypkin/gaia/commit/7aa2b76b0fbbbda38b4809c69ecb652080f30e26#diff-17bd958c31567384687de6f1e8680881R82

> BTW do you think we'll need unit test for it?

Not sure, I'll check if we have one for non-split mode and do the same here as well :)

apps/sms/test/marionette/new_activity_test.js
azasypkin: > If it still not enough, maybe we could add a class to composer to represent composer status is ready?

No, it's still not enough, when we open Composer with a draft (e.g. in case when app is opened as activity), ```updateSendButton``` method inside Compose.js is called 4/5 times asynchronously:
* beforeEnterComposer ->  Compose.clear() -> ....;
* afterEnterComposer -> handleDraft() -> Compose.clear() -> ...;
* afterEnterComposer -> handleDraft() -> Compose.append() -> ...;
* toFieldInput() (throttled, in FxDesktop called twice) -> Compose.refresh() -> ...

> Not quite sure about the timing of this `client.waitFor`, maybe we can wait for specific element in `launchAsActivity`?

Mmm, honestly I don't have idea what else we can wait without refactoring current Compose, do you have idea?

apps/sms/views/inbox/js/startup.js
azasypkin: Good, we have tests for the split view mode startup.js  - so I'll add one for the case as well, thanks!

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d28865817a66a744bef1dae9929f5c0082dca64f

apps/sms/test/marionette/lib/messages.js
steveck-chung: okey, but can we simply wait for app ready for all the launch scenario? Just curious.

apps/sms/test/marionette/new_activity_test.js
steveck-chung: Thanks for the explanation, I agree with you that there's not much we could do before we change the compose.

apps/sms/test/marionette/lib/messages.js
azasypkin: Right, for all except one - see inbox ```User could navigate without waiting for the app to be fully loaded``` test

apps/sms/test/marionette/new_activity_test.js
steveck-chung: Thanks for the explanation. I think that deciding when the compose is ready might not be difficult, but considering the text input at the same time is much more complex. Current solution is good enough for me.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8a82133c96731647dacde0234a3dee55e012540c

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=912ba753620b9acecf7a0f722da34e7508f43846

PR #33715, opened 2015-12-31
Speculative fix for FxA device registration test failures

philbooth:
Over in [bug 1227527](https://bugzilla.mozilla.org/show_bug.cgi?id=1227527), we implemented device registration for Firefox Accounts. When that patch landed, it [broke some tests](https://treeherder.mozilla.org/logviewer.html#?job_id=6354136&repo=fx-team) from this repo. This PR is a tentative fix for those failures.

**Background:**

The browser patch does a couple of things that are problematic in test contexts.
1. Device registration makes an HTTP request to the FxA auth server.
2. Attempting to generate the default device name fails.

These caused a bunch of indirect breakages all over the place, which we mitigated by introducing a new boolean pref called `identity.fxaccounts.skipDeviceRegistration`. When that is `true`, device registration is a nop. The point of this PR is to re-use the same pref here so the patch can land.

**Questions:**

The biggest problem I have at the moment is that I haven't yet managed to run these tests locally to reproduce the issue. I followed the [instructions on MDN](https://developer.mozilla.org/en-US/Firefox_OS/Developing_Gaia/Testing_Gaia_code_changes) for building and testing Gaia and then ran `bin/gaia-test` using a local build of Nightly that has the device registration patch applied. That runs okay, but I don't see any mention of `fxa_screen_flow_test.js` listed. How should I be running these tests?

Another option I saw for maybe setting the pref was in `build/preferences.js`, but I figured the files changed here are better because they're FxA-specific and already appear to set FxA-specific preferences. Am I on the right track?

### DIFF
--- a/apps/system/fxa/test/marionette/fxa_screen_flow_test.js
+++ b/apps/system/fxa/test/marionette/fxa_screen_flow_test.js
@@ -17,7 +17,8 @@ marionette('Firefox Accounts Screen Flow Test (UITest app)', function() {
             'identity.fxaccounts.auth.uri': 'http://' +
               config.SERVER_HOST + ':' +
               config.SERVER_PORT + '/' +
-              config.SERVER_PATH
+              config.SERVER_PATH,
+            'identity.fxaccounts.skipDeviceRegistration': true
           }
         }
       });
--- a/tv_apps/smart-system/fxa/test/marionette/fxa_screen_flow_test.js
+++ b/tv_apps/smart-system/fxa/test/marionette/fxa_screen_flow_test.js
@@ -17,7 +17,8 @@ marionette('Firefox Accounts Screen Flow Test (UITest app)', function() {
             'identity.fxaccounts.auth.uri': 'http://' +
               config.SERVER_HOST + ':' +
               config.SERVER_PORT + '/' +
-              config.SERVER_PATH
+              config.SERVER_PATH,
+            'identity.fxaccounts.skipDeviceRegistration': true
           }
         }
       });


autolander:

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2a1b7709f88f671179abbbb1ebc14b24d04f6820

begeeben:
1. It's an integration test so you could run it following the instructions in https://developer.mozilla.org/en-US/Firefox_OS/Developing_Gaia/Testing_Gaia_code_changes#Integration_tests .

2. The patch only set the pref for integration tests, I wonder if this feature will break b2g FxAccount features? If yes, would it be better to put the pref in build/config/phone/custom-prefs.js and also in other custom-prefs.js for other device types ?

philbooth:
Hey @begeeben, thanks for the info.

I've built the B2G desktop client locally, with the device registration patch applied to my tree. But when I run the fxa_screen_flow_test, I get a different error. Instead of [the timeout that was reported on treeherder](https://treeherder.mozilla.org/logviewer.html#?job_id=6354136&repo=fx-team), my log shows this:

```
  Firefox Accounts Screen Flow Test (UITest app)
    1) should walk screen flow for new user
    ✓ should walk screen flow for existing user (1177ms)


  1 passing (20s)
  1 failing

  1) Firefox Accounts Screen Flow Test (UITest app) should walk screen flow for new user:


      + expected - actual


  ypeError: Cannot read property 'scriptWith' of null
      at Object.MarionetteHelper.tapSelectOption (node_modules/marionette-helper/index.js:303:19)
      at Object.FxA.enterCOPPANew (tv_apps/smart-system/fxa/test/marionette/lib/fxa.js:94:26)
      at Context.<anonymous> (tv_apps/smart-system/fxa/test/marionette/fxa_screen_flow_test.js:48:9)
      at Test.MarionetteTest.run (node_modules/marionette-js-runner/lib/ui.js:25:31)
      at tv_apps/smart-system/fxa/test/marionette/fxa_screen_flow_test.js:34:7
      at ChildProcess.<anonymous> (tv_apps/smart-system/fxa/test/marionette/lib/server.js:30:9)
      at handleMessage (internal/child_process.js:686:10)
      at Pipe.channel.onread (internal/child_process.js:440:11)
```

Any idea what's up?

ferjm:
It seems that the next button is not being found https://mxr.mozilla.org/gaia/source/tv_apps/smart-system/fxa/test/marionette/lib/fxa.js#131 Probably because there's no COPPA screen for the TV...

PR #33724, opened 2016-01-01
Bug 1236166 - Migrate Gallery and Video to l20n.js

zbraniecki:
https://bugzilla.mozilla.org/show_bug.cgi?id=1236166

### DIFF
--- a/apps/camera/test/unit/controllers/confirm_test.js
+++ b/apps/camera/test/unit/controllers/confirm_test.js
@@ -1,8 +1,6 @@
 /*jshint maxlen:false*/
 'use strict';
 
-require('/shared/js/l10n.js');
-
 suite('controllers/confirm', function() {
   suiteSetup(function(done) {
     var modules = this.modules = {};
--- a/apps/gallery/js/ImageEditor.js
+++ b/apps/gallery/js/ImageEditor.js
@@ -95,7 +95,7 @@ function editPhotoIfCardNotFull(n) {
       editPhoto(n);
     }
     else {
-      navigator.mozL10n.formatValue('memorycardfull').then(alert);
+      document.l10n.formatValue('memorycardfull').then(alert);
     }
   });
 }
@@ -119,7 +119,7 @@ var exposureSlider = (function() {
   thumb.addEventListener('pan', function(e) {
     // Handle delta so that slider moves correct way
     // when user drags it for RTL locales
-    var delta = navigator.mozL10n.language.direction === 'ltr' ?
+    var delta = document.documentElement.dir === 'ltr' ?
                 e.detail.absolute.dx : - e.detail.absolute.dx;
 
     var exposureDelta = delta / parseInt(bar.clientWidth, 10) * 6;
@@ -596,7 +596,7 @@ function setAutoEnhanceState(isEnhanced) {
   }
 
   function showStatus(msgId) {
-    navigator.mozL10n.setAttributes(statusLabel, msgId);
+    document.l10n.setAttributes(statusLabel, msgId);
     banner.hidden = false;
     setTimeout(function() {
       banner.hidden = true;
--- a/apps/gallery/js/gallery.js
+++ b/apps/gallery/js/gallery.js
@@ -325,8 +325,8 @@ function fileCreated(fileinfo) {
         var banner = $('edit-copy-save-banner');
         // Show latest saved file inserted at index insertPosition
         showFile(insertPosition);
-        navigator.mozL10n.setAttributes($('edit-copy-save-status'),
-                                        'edit-copy-saved');
+        document.l10n.setAttributes($('edit-copy-save-status'),
+                                    'edit-copy-saved');
         banner.hidden = false;
         setTimeout(function() { banner.hidden = true; }, 3000);
       } else {
@@ -519,7 +519,7 @@ function clearSelection() {
   selectedFileNamesToBlobs = {};
   $('thumbnails-delete-button').classList.add('disabled');
   $('thumbnails-share-button').classList.add('disabled');
-  navigator.mozL10n.setAttributes(
+  document.l10n.setAttributes(
     $('thumbnails-number-selected'),
     'number-selected2',
     { n: 0 }
@@ -586,7 +586,7 @@ function updateSelection(thumbnail) {
 
   // Now update the UI based on the number of selected thumbnails
   var numSelected = selectedFileNames.length;
-  navigator.mozL10n.setAttributes(
+  document.l10n.setAttributes(
     $('thumbnails-number-selected'),
     'number-selected2',
     { n: numSelected }
@@ -724,7 +724,7 @@ function share(blobs, blobName) {
 
   a.onerror = function(e) {
     if (a.error.name === 'NO_PROVIDER') {
-      navigator.mozL10n.formatValue('share-noprovider').then(alert);
+      document.l10n.formatValue('share-noprovider').then(alert);
     }
     else {
       console.warn('share activity error:', a.error.name);
--- a/apps/gallery/js/thumbnails.js
+++ b/apps/gallery/js/thumbnails.js
@@ -21,7 +21,8 @@
   // Bug 1135256: Having a listener in ThumbnailList prevents them from
   // being garbage collected since ready() holds a strong reference, so
   // we have a singleton listener here.
-  navigator.mozL10n.ready(thumbnails.list.localize.bind(thumbnails.list));
+  document.addEventListener('DOMRetranslated',
+    thumbnails.list.localize.bind(thumbnails.list)); 
 
   // How many thumbnails are visible on a page.
   // Batch sizes are based on this.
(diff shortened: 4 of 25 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=6ee08c5cc5ea103d5cbc97fafcc0100c642fe439

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=fb49c5b2130d4f407b9fdc0e22160a69579cdab4

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=5f60a7210433d01a3bcf80965332ace14ea47acf

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=56c59a585abcdcdddfdc8fa8aafc894530357c50

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=284ce6cfea86cd977c5a99f411bce88e6c0e6b3c

apps/camera/test/unit/controllers/confirm_test.js
```diff
@@ -1,8 +1,6 @@
 /*jshint maxlen:false*/
 'use strict';
 
-require('/shared/js/l10n.js');
```
zbraniecki: I'm just removing three requires in camera because they are not needed, I forgot to remove them in my camera patch, and I don't want to file a separate bug for that so I use the excuse that camera is part of media apps :)

apps/gallery/js/ImageEditor.js
```diff
@@ -119,7 +119,7 @@ var exposureSlider = (function() {
   thumb.addEventListener('pan', function(e) {
     // Handle delta so that slider moves correct way
     // when user drags it for RTL locales
-    var delta = navigator.mozL10n.language.direction === 'ltr' ?
+    var delta = document.documentElement.dir === 'ltr' ?
```
davidflanagan: Does document.documentElement.dir change correctly when the user switches locales?

apps/gallery/js/gallery.js
```diff
@@ -325,7 +325,7 @@ function fileCreated(fileinfo) {
         var banner = $('edit-copy-save-banner');
         // Show latest saved file inserted at index insertPosition
         showFile(insertPosition);
-        navigator.mozL10n.setAttributes($('edit-copy-save-status'),
+        document.documentElement.setAttributes($('edit-copy-save-status'),
```
davidflanagan: This should be document.l10n, not document.documentElement, right?

apps/gallery/js/thumbnails.js
```diff
@@ -21,7 +21,9 @@
   // Bug 1135256: Having a listener in ThumbnailList prevents them from
   // being garbage collected since ready() holds a strong reference, so
   // we have a singleton listener here.
-  navigator.mozL10n.ready(thumbnails.list.localize.bind(thumbnails.list));
+  thumbnails.list.localize();
```
davidflanagan: I'm not sure about doing this unconditionally. What if we are not ready yet?

apps/gallery/js/thumbnails.js
```diff
@@ -21,7 +21,9 @@
...
+  document.addEventListener('DOMRetranslated',
```
davidflanagan: Is DOMRetranslated fired on the first translation or only when the locale changes when the app is already running?

apps/gallery/js/ImageEditor.js
zbraniecki: yup :)

apps/gallery/js/gallery.js
zbraniecki: yup :)

apps/gallery/js/thumbnails.js
zbraniecki: l10n is never "not-ready" anymore, that's the nature of asynchronicity of the library now! if you use `document.l10n.formatValue`, it will resolve when resources are loaded. If you set `data-l10n-id`, it will localize when resources are loaded. No need for `ready`/`once` wrappers anymore!

apps/gallery/js/thumbnails.js
zbraniecki: only when locale changes

apps/video/js/thumbnail_list.js
```diff
@@ -47,7 +47,8 @@ function ThumbnailList(groupClass, container) {
 
   // If the locale changes while we're running, then we need to
   // re-localize the dates in the group headers
-  navigator.mozL10n.ready(this.localize.bind(this));
+  this.localize();
+  document.addEventListener('DOMRetranslated', this.localize.bind(this));
```
davidflanagan: Same issue here as in gallery. I'm not convinced that the new code does the same thing as the old code.

apps/video/js/video.js
```diff
@@ -174,14 +174,15 @@ if (acm) {
...
-  });
+  // reload the thumbnail list title field for tablet which is the app name.
+  var req = navigator.mozApps.getSelf();
+  req.onsuccess = function() {
+    if (!req.result) {
+      return;
+    }
+    var manifest = new ManifestHelper(req.result.manifest);
+    dom.thumbnailListTitle.textContent = manifest.name;
+  };
```
davidflanagan: With l10n, does ready() run only once, or does it get called again when the locale changes? If the latter then this new code does not behave the same.

But this !isPhone code is for the old flatflish tablet and is not supported anymore anyway, so we could also just delete this. It seems absurd to look up the app name like this anyway rather than just using something in the regular locale file anyway.

apps/video/test/unit/video_test.js
```diff
@@ -13,7 +13,8 @@
   updateSelection,setButtonPaused,handleSliderTouchStart,MockMediaDB,
   handleSliderTouchEnd,loadingChecker,MockL10n,VideoUtils,
   updateDialog,LAYOUT_MODE,showPlayer,hidePlayer,MockThumbnailItem,
-  MockThumbnailGroup,MockForwardRewindController,MockIntlHelper,captions */
+  MockThumbnailGroup,MockForwardRewindController,MockIntlHelper,captions,
+  MockManifestHelper */
```
davidflanagan: It would also be okay to just remove the manifest related code from the app, since it is not supported anyway. Then you wouldn't have to add this test.

shared/js/media/media_utils.js
```diff
@@ -31,14 +31,14 @@ var MediaUtils = {
     }
     var sizeDecimal = i < 2 ? Math.round(size) : Math.round(size * 10) / 10;
 
-    return navigator.mozL10n.formatValue('byteUnit-' + units[i]).then(
+    return document.l10n.formatValue('byteUnit-' + units[i]).then(
```
davidflanagan: This shared file is also used by apps/fl/download.html.  So unless you want to add that app to this patch, I think you need something like:
```
(document.l10n || navigator.mozL10n).formatValue(...)
```
That might also be a good idea for the shared dialogs.js file above, even though that is currently only used by gallery and video.

apps/video/js/thumbnail_list.js
zbraniecki: It does. The old code fired the callback when l10n was ready, and then when DOM was retranslated.
We could do:
```
document.l10n.ready.then(this.localize.bind(this));
document.addEventListener('DOMRetranslated', this.localize.bind(this));
```

but since everything that `this.localize` uses is asynchronous, we don't have to. :)

apps/video/js/video.js
zbraniecki: Ok, I'll be happy to remove this `if`.

shared/js/media/media_utils.js
zbraniecki: both, l10n.js and l20n.js have proxies to each other because of this problem. So any app with `l10n.js` will work with `document.l10n` and any app using `l20n.js` will handle `navigator.mozL10n` :)

apps/gallery/js/thumbnails.js
davidflanagan: Actually, I don't think this call to localize() is required at all. The individual thumbnails and the group headers call their own localize() method from their constructors. I don't know when that change happened, but it means that we only need to call this top-level localize() function no DOMRetranslated, I think.  So I think you can just remove this line.

apps/video/js/thumbnail_list.js
davidflanagan: Please remove the this.localize() call. All it does is call localize on the thumbnail groups. But this is the constructor so there aren't any groups yet.

zbraniecki: sweet!

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=293d5dc92f345b989fb4ca6915a39b15aa6cd97a

apps/gallery/js/thumbnails.js
```diff
@@ -21,7 +21,8 @@
   // Bug 1135256: Having a listener in ThumbnailList prevents them from
   // being garbage collected since ready() holds a strong reference, so
   // we have a singleton listener here.
-  navigator.mozL10n.ready(thumbnails.list.localize.bind(thumbnails.list));
+  document.addEventListener('DOMRetranslated',
+    thumbnails.list.localize.bind(thumbnails.list));
```
russnicoletti: Nit: trailing whitespace

PR #33777, opened 2016-01-07
Bug 1236623 - Add UnitFormat to mozIntl

zbraniecki:
https://bugzilla.mozilla.org/show_bug.cgi?id=1236623

### DIFF
--- a/apps/sharedtest/test/unit/moz_intl_test.js
+++ b/apps/sharedtest/test/unit/moz_intl_test.js
@@ -1360,4 +1360,162 @@ suite('MozIntl', function() {
       });
     });
   });
+
+  suite('UnitFormat', function() {
+    setup(function() {
+      this.sinon.stub(navigator.mozL10n, 'formatValue',
+        arg => Promise.resolve(arg));
+    });
+
+    function testUnit(group, unit, style) {
+      var formatter = mozIntl.UnitFormat(navigator.languages, {
+        unit:  unit,
+        style: style
+      });
+      return formatter.format(0).then(result => {
+        assert.equal(result, `${group}-${unit}-${style}`);
+      });
+    }
+
+    test('uses correct unit group for durations', function(done) {
+      var units = ['second', 'minute', 'hour', 'day', 'month'];
+
+      var tests = units.map(unit => testUnit('duration', unit, 'narrow'));
+
+      Promise.all(tests).then(() => done(), done);
+    });
+
+    test('uses correct unit group for durations', function(done) {
+      var units = ['byte', 'kilobyte', 'megabyte', 'gigabyte', 'terabyte'];
+
+      var tests = [];
+
+      for (let unit of units) {
+        tests.push(testUnit('digital', unit, 'short'));
+      }
+
+      Promise.all(tests).then(() => done(), done);
+    });
+
+    test('throws range error for invalid unit', function() {
+      assert.throws(() => {
+        mozIntl.UnitFormat(navigator.languages, {
+          unit: 'foo'
+        });
+      }, RangeError);
+    });
+
+    test('throws range error for invalid style', function() {
+      assert.throws(() => {
+        mozIntl.UnitFormat(navigator.languages, {
+          unit: 'second',
+          style: 'foo'
+        });
+      }, RangeError);
+    });
+  });
+
+
+  /* _gaia suite */
+
+  suite('mozIntl._gaia methods', function() {
+    suite('getFormattedUnit', function() {
+      suite('digital type', function() {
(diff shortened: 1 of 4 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=84a90882133931f2cc09609ab254e637b0e8541b

shared/js/moz_intl.js
```diff
@@ -352,6 +373,34 @@ global.mozIntl = {
         },
       };
     },
+
+    getFormattedUnit: function(type, style, v) {
+      if (!unitFormatData.hasOwnProperty(type)) {
+        throw new RangeError(`invalid type ${type}`);
```
stasm: Why is this a `RangeError`?

shared/js/moz_intl.js
```diff
@@ -491,4 +570,13 @@ function relativeTimeFormatId(x, options) {
   };
 }
 
+function getUnitFormatGroup(unitName) {
+  for (let groupName in unitFormatGroups) {
+    if (unitFormatGroups[groupName].units.includes(unitName)) {
+      return groupName;
```
stasm: Naming nit:  I thought this would return the group, not the name.

shared/locales/intl/unitformat.en-US.properties
```diff
@@ -37,3 +37,43 @@ duration-month-narrow[two]    = {{ value }}mths
 duration-month-narrow[few]    = {{ value }}mths
 duration-month-narrow[many]   = {{ value }}mths
 duration-month-narrow[other]  = {{ value }}mths
+
+digital-byte-short            = {[ plural(value) ]}
```
stasm: I'm slightly confused by this naming.  In particular, by the use of the `groupName` prefix, like `duration-`.  The headers from http://www.unicode.org/cldr/charts/27/summary/pl.html#5556 seem to be unique even without the group names ('pages' as CLDR calls them).

I don't mind putting them in the string id, just want to better understand the strategy for extending this.  For instance, https://github.com/unicode-cldr/cldr-units-full/blob/master/main/en/units.json uses `energy-joule` when the CLDR raw data says [Energy and Power, energy-joule], but then it uses `mass-ton` for [Mass and Weight,	ton].

Then again, this is perhaps not important as these translations are temporary—in the future they will come directly from Intl.

shared/js/moz_intl.js
zbraniecki: That's how https://tc39.github.io/ecma402/ deals with options that are not in range. You can test for yourself:

```
(42).toLocaleString('en-US', {style: 'foo'});
RangeError: invalid value foo for option style
```

apps/sharedtest/test/unit/moz_intl_test.js
```diff
@@ -1360,4 +1360,167 @@ suite('MozIntl', function() {
...
+    test('uses correct unit group for durations', function(done) {
+      var units = ['second', 'minute', 'hour', 'day', 'month'];
+
+      var tests = [];
+
+      for (let unit of units) {
+        tests.push(testUnit('duration', unit, 'narrow'));
+      }
+
+      Promise.all(tests).then(() => done(), done);
```
stasm: Looks like you might want to use `units.map` here?

apps/sharedtest/test/unit/moz_intl_test.js
```diff
@@ -1360,4 +1360,167 @@ suite('MozIntl', function() {
       });
     });
   });
+
+  suite('UnitFormat', function() {
+    setup(function() {
+      this.sinon.stub(navigator.mozL10n, 'formatValue', function(...args) {
+        return Promise.resolve(args[0]);
```
stasm: Shorter: `arg => Promise.resolve(arg)` or simply, `Promise.resolve`.

shared/locales/intl/unitformat.en-US.properties
zbraniecki: I just follow what they're discussing in ecma 402 proposal. My take on the reason for the groupName is to aid in a scenario when two units from different areas have the same name (for example angle second vs. time second).

stasm: Let's use what CLDR uses in http://unicode.org/cldr/trac/browser/trunk/common/main/en.xml and https://github.com/unicode-cldr/cldr-units-full/blob/master/main/en/units.json, i.e. `mass-ton`, `energy-joule`, `power-kilowatt`.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=ad3661ee496e22e7095ecfea13421743367f325f

PR #33778, opened 2016-01-07
Bug 1234732 - [email] HeaderCursor seems to be insantiated multiple times for message_list somehow r=asuth

jrburke:
This issue was caused by two things:
- message_list was not calling .die() on the HeaderCursor it creates.
- header_cursor not properly removing the "folder" listener when it dies.

The end result: whenever a message_list is destroyed in an email app session, there would be a zombie HeaderCursor listening for folder change events and asking for new slices. Adding or removing an account is one of a small set of scenarios where this would occur.

The fix was to call headerCursor.die() in the message_list's die(), and then in header_cursor, remove the model 'folder' listener in its die() method. Previously header_cursor internally called die() to reset the local messagesSlice property. This was extracted as a separate `resetMessageSlice` method and that is called instead, with die() also calling resetMessageSlice().

### DIFF
--- a/apps/email/js/cards/message_list.js
+++ b/apps/email/js/cards/message_list.js
@@ -960,6 +960,10 @@ return [
 
       evt.removeListener('folderPickerClosing', this.onFolderPickerClosing);
 
+      if (this.headerCursor) {
+        this.headerCursor.die();
+      }
+
       var model = this.model;
       model.removeListener('folder', this._folderChanged);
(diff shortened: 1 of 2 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=e88efb92975880012ae904fa801499896fcbe4e4

PR #33839, opened 2016-01-12
Bug 1227527 - Disable FxA device registration

philbooth:
@ferjm, it turns out that refactoring the device name stuff out of Sync in the desktop patch [is more complicated than I realised](https://bugzilla.mozilla.org/show_bug.cgi?id=1227527#c62). So it's easiest if we get that one landed by disabling device registration here and then make it FxOS-friendly in a separate effort.

Can you review this PR for me? Or do you know who I should be asking if not?


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=95fd4ff1bd3dcba07161b1a09bc826fcff8fa40a

PR #33870, opened 2016-01-14
Bug 1235881 - Changed invokeControls, added some delay and a longer test video R=mhenretty

tilmankamp:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f4593e03da5dcabcd630929438104fd6ebb41a31

apps/system/test/marionette/browser_video_test.js
```diff
@@ -76,6 +76,7 @@ marionette('Video', function() {
       return player.isFullScreen();
     });
 
+    client.helper.wait (1000);
```
albertopq: Is there anything else we can wait for instead of an specific time? Looks like a source for future intermitents

apps/system/test/marionette/lib/video_player.js
```diff
@@ -58,13 +58,16 @@ VideoPlayer.prototype = {
   },
 
   invokeControls: function() {
-    if (!this.visibleControls()) {
-      this.rootElement.tap();
-
+    if(this.visibleControls()) {
+      this.rootElement.tap(0, 0);
```
albertopq: Why this tap? Shouldn't we just wait for the controls disappearing?

tilmankamp: It's to enforce them to disappear ASAP - it turned out to have some influence on the test outcome.
However: I'm fine removing it to see if it has some actual influence.

apps/system/test/marionette/browser_video_test.js
tilmankamp: The control panel shows every now and then a strange behavior in that very moment the test switches back from full-screen. It reappears for a very short period of time. In essence this wait-call led to 50+ passing tests in sequence.

apps/system/test/marionette/browser_video_test.js
```diff
@@ -35,7 +35,7 @@ marionette('Video', function() {
 
   // https://moztrap.mozilla.org/manage/case/6073/
   test('Confirm video playback', function() {
-    var videoUrl = server.url('VID_0001.mp4');
+    var videoUrl = server.url('VID_0001.webm');
```
mikehenrty: Are we removing the old `VID_0001.mp4`? It looks like not. It's kind confusing to have two completely different videos `VID_0001.webm` and `VID_0001.mp4`. Let's just rename the new one.

apps/system/test/marionette/browser_video_test.js
mikehenrty: Let's add a comment here explaining what we are waiting for. It sounds like we are essentially waiting to make sure something *doesn't* happen, which can only be achieved with a sleep like this. But if that's the case, let's make that explicit in a comment.

apps/system/test/marionette/lib/video_player.js
mikehenrty: Yeah again the problem here is `visibleControls()` doesn't tell us if the controls are about to disappear, but the problem is this tap can be a toggle and if controls are visible, then disappear, and then we tap, we can get in a state where we just activated the controls. But this is a minefield, so I would say we do whatever makes the test the most robust.

PR #33950, opened 2016-01-25
Bug 1235884 - Slower launching test app and later click event R=mhenretty

tilmankamp:
Seems like the notification click event dispatching has to happen slightly later in the game (while the long launch app is the foreground app). Long launch app's increased launch time adds some reliability to the test's racing setup. It also provides simpler mocking of the mozSetMessageHandler call and some tweaking support for future (re)calibration.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3ed2e04796c9a8666442db238c30b81415796be6

apps/system/test/marionette/notification_events_test.js
```diff
@@ -398,22 +393,24 @@ marionette('Notification events', function() {
         type: 'desktop-notification-click',
         id: node.dataset.notificationId
       });
```
mikehenrty: Now would be a great time to actually perform a real click on the notification container rather than manually dispatching a `mozContentNotificationEvent`. I'm not sure why we didn't do that in the first place TBH.

apps/system/test/apps/longlaunchapp/wait.js
```diff
@@ -0,0 +1,11 @@
...
+(function() {
+  if(new Date().getTime() < window.startTime + window.launchDuration) {
+    var i = 10000;
+    var x;
+    while (i--) { /* wasting some time */ x = Math.sqrt(i); }
+    window.sleepOn();
+  } else {
+    window.sayDone();
+  }
+})();
```
mikehenrty: What are we gaining by delaying the loading complete of this test app?

apps/system/test/apps/longlaunchapp/test.js
```diff
@@ -0,0 +1,35 @@
...
+    window.sayDone = function() {
+      print('DONE!');
+    };
+
+    navigator.mozSetMessageHandler('notification', function(m) {
+      fakeData = m;
+      print('notification: ' + JSON.stringify(m));
+    });
+
+    window.getFakeData = function() {
```
mikehenrty: no need to call this `fakeData` anymore since you are using the real message handler for this. i think it would be more clear to call it something like 'lastNotificationData` or something

apps/system/test/marionette/lib/fake_moz_set_message_handler.js
```diff
@@ -1,19 +0,0 @@
...
-                                                                  function(m) {
-                fakeData = m.data;
-            });
-        };
-    });
-    window.wrappedJSObject.__defineGetter__('__getFakeData', function() {
-        return fakeData;
-    });
-})();
-
```
mikehenrty: :+1:

PR #33965, opened 2016-01-26
Bug 1243087 - Move away from deprecated mozL10n.get in Smart-Home

zbraniecki:
https://bugzilla.mozilla.org/show_bug.cgi?id=1243087

### DIFF
--- a/tv_apps/smart-home/js/home.js
+++ b/tv_apps/smart-home/js/home.js
@@ -230,17 +230,17 @@
     },
 
     initClock: function() {
-      var that = this;
-      navigator.mozL10n.ready(function() {
-        that.clock = new Clock();
-        that.clock.start(that.updateClock.bind(that));
-        // Listen to 'moztimechange'
-        window.addEventListener('moztimechange',
-                                that.restartClock.bind(that));
-        // Listen to 'timeformatchange'
-        window.addEventListener('timeformatchange',
-                                that.restartClock.bind(that));
-      });
+      this.clock = new Clock();
+      this.clock.start(this.updateClock.bind(this));
+      // Listen to 'moztimechange'
+      window.addEventListener('moztimechange',
+                              this.restartClock.bind(this));
+      // Listen to 'timeformatchange'
+      window.addEventListener('timeformatchange',
+                              this.restartClock.bind(this));
+      // Listen to 'DOMRetranslated'
+      document.addEventListener('DOMRetranslated',
+                              this.restartClock.bind(this));
     },
 
     _localizeCardName: function(elem, card) {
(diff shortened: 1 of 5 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=f972dfaa0100ad97ab47d918ca2c7bee0c0591c7

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=cc3af41b1d61142c70ee58e64f32fd01c9605cf9

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=bbaa01fa40952f0d9e950b318ec9f61e9b551cca

tv_apps/smart-home/js/home.js
```diff
@@ -793,28 +793,34 @@
...
-      this.timeElem.dataset.ampm = use12Hour ? f.localeFormat(now, '%p') : '';
+      var timeWithoutDayPeriod = parts.map(({type, value}) => {
+        if (type === 'dayperiod') {
+          dayperiod = value;
+          return '';
+        }
+        return value;
+      }).join('');
+
+      if (this.timeElem) {
```
zbraniecki: I'm using this here because unit test loads the file before creating the DOM elements so the `this.timeElem` is not available in the test unit. Before it was not tested because it was wrapped in `mozL10n.ready` which was not fired, but now it has to be guarded this way until the test is fixed

tv_apps/smart-home/test/unit/home_test.js
```diff
@@ -118,6 +118,16 @@ suite('home', function() {
 
     fakeTimer = this.sinon.useFakeTimers();
     window.MozActivity = MockMozActivity;
+
+    if (!Intl.DateTimeFormat.prototype.formatToParts) {
```
zbraniecki: `Intl.DateTime.prototype.formatToParts` is currently only exposed to B2G certified apps (because we're still in the process of standardizing it in ECMA 402), so it's not available in unit tests.

tv_apps/smart-home/js/home.js
dwi2: Maybe we could put your comment here to let others know why we are doing this. (And they might also want to fix the test by reading this comment!)

rexboy7:
I need to confirm with other TV-gaia developers since this patch would make B2G-Desktop no longer a choice for development. Although B2G-Desktop has been deprecated but some of us are still using it for some development tools issue. If I can confirm everyone use Mulet without problem then we can land it.

zbraniecki:
Why wouldn't B2G-Desktop just mimic what the device is doing? I believe formatToParts should work in B2G-Desktop

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1e689f14896b95128ec4e189fbbf7f4d24496484

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=1e689f14896b95128ec4e189fbbf7f4d24496484

rexboy7:
B2G-desktop is just deprecated and the latest nightly build was about 2 months ago. :-/. 
Theoretically we should move to Mulet for development but some issues prevents us from doing that like 
this https://bugzilla.mozilla.org/show_bug.cgi?id=1059753.
The code itself is ok, I can r+ this patch.. But just want to know does this blocks your follow-up works? I'd prefer some transition time to make sure we can all transfer to Mulet for developing smoothly.

zbraniecki:
It doesn't block me too hard, but there are several cleanups that are waiting for this (like, removal of l10n_date).

I'm ok with transition time - how long do you think you'd want it to be?

Last question - are you sure Mulet doesn't work with `formatToParts`? We use `formatToParts` in many places in B2G now and my guess is that Mulet should work, but I didn't test it.

zbraniecki:
Ok, I confirmed that `formatToParts` works in Mulet, but I think I misunderstood the problem. You're saying that `formatToParts` may not work in B2G Desktop that you're using currently, right?

zbraniecki:
Where can I find B2G Desktop for gecko 46+?

rexboy7:
Since I've got bug 1248907 landed, I think we can just land this patch now.I'll try to push people use Mulet. Thanks for help testing that.

rexboy7:
Let's just transfer to Mulet. Latest B2G-Desktop version is just too old for l20n so don't bother to test that.

PR #33993, opened 2016-01-29
Bug 1243751 - Create marionette js test for video FF/REW control 

nojunpark:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=eda040747cef3f29339ea1fe8f6554248788ba6b

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9013d2cb06339343b93dad5b9e0db78525dcb3ce

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0f5dd63526f5c58f99afe043e23f9e3d62519650

apps/video/test/marionette/lib/video.js
```diff
@@ -192,6 +195,66 @@ Video.prototype = {
...
+   */
+  pausePlayback: function() {
+    this.client.helper.waitForElement(Video.Selector.playpauseButton).click();
+    var currentTime = this.getElapsedTimeSeconds();
+    
+    // wait for 2 seconds
+    this.client.executeAsyncScript(function () 
+      {setTimeout(function(){
+        marionetteScriptFinished();
+      }, 2000);
```
russnicoletti: Did you notice when running the test that it's necessary to wait for a couple of seconds after clicking the button to ensure the video has been paused? In other words, have you noticed that 'getElapsedTimeSeconds' changes from line 203 to line 212?

If it does change, it's not ideal to have to wait for two seconds on each run of the test. Also, two seconds seems arbitrary. If you're up for it, I have a suggestion for possibly removing the need to wait at all. My suggestion is to move this line of code [1] to after line 1266 of video.js. And then, in 'pausePlayback', after line 202, you can wait for the 'play' button 'data-l10n-id' attribute to become 'play-button' (see [2]). Once that happens, we know 'pause' on the video element has been invoked in which case it should be safe to assume the elapsed time will not change.

[1] https://github.com/mozilla-b2g/gaia/blob/master/apps/video/js/video.js#L1257
[2] https://github.com/mozilla-b2g/gaia/blob/master/apps/video/js/video.js#L1233

punamdahiya:
Please squash commits into a single commit

apps/video/test/marionette/control_test.js
```diff
@@ -0,0 +1,63 @@
...
+  var app, selectors, initialTime;
+
+  setup(function() {
+    // Remove all files in temp device storage.
+    client.fileManager.removeAllFiles();
+    
+    // Add file into the videos directory
+    client.fileManager.add({
+      type: 'videos',
+      filePath: 'test_media/Movies/elephants-dream.webm'
```
russnicoletti: I've noticed seeking within this video can be problematic. If you're seeing long delays when seeking within this video, you can try './test_media/samples/Movies/meetthecubs.webm'

apps/video/test/marionette/control_test.js
```diff
@@ -0,0 +1,63 @@
+'use strict';
+
+var Video = require('./lib/video'),
+    assert = require('assert');
+    
+marionette('video info view', function() {
```
punamdahiya: nit: Update text to indicate player view tests

apps/video/test/marionette/control_test.js
```diff
@@ -0,0 +1,63 @@
...
+    app.waitForThumbnails(1);
+    app.tapThumbnail(0);
+    app.waitForVideoToStartPlaying();
+    
+    // pause, then tap forward and check the counter 
+    // that it was increased by 10 seconds
+    app.pausePlayback();
+    initialTime = app.getElapsedTimeSeconds();
+    app.tapForward();
+    assert.ok(app.getElapsedTimeSeconds() == (initialTime + 10));
```
russnicoletti: Nit: To reduce the number of times the test is calling 'getElapsedTimeSeconds', perhaphs this code could be something like:
app.pausePlayback();
initialTime = app.getElapsedTimeSeconds();
app.tapForward();
var nextTime = app.getElapsedTimeSeconds();
assert.ok((nextTime - 10) == initialTime);
initalTime = nextTime;
app.tapRewing();
nextTime = app.getElapsedTimeSeconds();
assert.ok((nextTime + 10) == initialTime);
initialTime = nextTime ...

If you think this is overly complicated, feel free to leave the code as it is.

apps/video/test/marionette/control_test.js
```diff
@@ -0,0 +1,63 @@
...
+    client.fileManager.add({
+      type: 'videos',
+      filePath: 'test_media/Movies/elephants-dream.webm'
+    });
+
+    selectors = Video.Selector;
+    app = new Video(client);
+  });
+
+  test('open video and check video info', function() {
```
punamdahiya: nit: please update text to indicate player controls test

apps/video/test/marionette/lib/video.js
```diff
@@ -192,6 +195,66 @@ Video.prototype = {
...
+
+    });
+    return this.getElapsedTimeSeconds() == currentTime;
```
punamdahiya: @russnicoletti I am confused, if  'getElapsedTimeSeconds' changes from line 203 to line 212 shouldn't this condition fail?

russnicoletti: Yes, it should fail. But why is that line necessary? Why is it necessary to wait two seconds after clicking 'pause' to determine if 'getElapsedTimeSeconds' has changed? Tests using arbitrary delays will lead to unreliable results. My comment was exploring the purpose of the delay and how to remove it.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d0e59cb73424b64b27db17b4b555f523e6d91b56

apps/video/test/marionette/lib/video.js
npark-mozilla: ah, so you're saying that by moving that code, it ensures that by the time the icon is changed the playback is actually paused.  hmm, may be a better idea. will try.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=60d644a59ec92cb4cce9537e6b2886c59bff5482

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4738ae8eeebce2d00318b246abb5ef823ba5ba23

apps/video/test/marionette/lib/video.js
```diff
@@ -192,6 +197,58 @@ Video.prototype = {
...
+  pausePlayback: function() {
+    this.client.helper.waitForElement(Video.Selector.playpauseButton).click();
+
+    // wait for the icon to change by checking for play icon
+    return this.client.helper.waitForElement(Video.Selector.playIcon)
+    .displayed();
+  },
+
+  /**
+   * Taps 'play' button and checks that the counter is progressing.
```
russnicoletti: The function is not checking that the elapsed time is incrementing, although I think that is the proper functionality. Probably best to add the 'waitFor' statement similar to how you've done with 'tapForward'.

apps/video/test/marionette/control_test.js
```diff
@@ -0,0 +1,66 @@
...
+    nextTime = app.getElapsedTimeSeconds();
+    assert.ok(nextTime == (initialTime - 10));
+    initialTime = nextTime;
+    app.tapRewind();
+    nextTime = app.getElapsedTimeSeconds();
+    assert.ok(nextTime == (initialTime - 10));
+
+    // resume play and check it's progressing
+    initialTime = nextTime;
+    assert.ok(app.startPlayback());
```
russnicoletti: To definitively check that the video is progressing, would be best to assert elapsed time has increased.

apps/video/test/marionette/lib/video.js
npark-mozilla: so in this i have to wait for nothing is happening, then I would still need the async call ( where I wait for 2 seconds)  to wait arbitrary length of time and ensure nothing is changing, wouldn't I?  Tapforward() was easy because the time is incrementing, but in this case the time stays the same.. would setting interval: 10 make sure that the first check is done 10 milliseconds after, so it wouldn't pass trivially?

tests/python/gaia-ui-tests/gaiatest/tests/graphics/test_browser_play_video.py
```diff
@@ -22,7 +22,8 @@ def test_play_video(self):
 
         https://moztrap.mozilla.org/manage/case/6073/
         Note: this test case does not reflect the entire test steps of above test case.
-        """
+        """_profile_icon = (By.CSS_SELECTOR, "a[href='/profile/']")
+        _profile_icon = (By.CSS_SELECTOR, "a[href='/profile/']")
```
russnicoletti: I'm wondering how this change is related ...

npark-mozilla: strange, this must be a typo.  I'll correct it

apps/video/test/marionette/control_test.js
npark-mozilla: okay, i added back the wait call. the wait will fail inside startPlayback if the counter is progressing.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=53afd91f9f2e441d00987c25a77544bf4db18100

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3b2dfc3f5fe23d9d342dbc462f35e84e0c24aadb

PR #34004, opened 2016-01-30
Bug 1120572 - Added parenthesis support

adarshsanjeev:


autolander:

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=70fb6f8b598815e9a38b3e70a97083aef71ec9be

apps/sms/views/conversation/js/recipients.js
```diff
@@ -54,6 +54,7 @@
     // (ignoring the presense of a '+'), the input value
     // is questionable and may be invalid.
     number = this.number[0] === '+' ? this.number.slice(1) : this.number;
+    number = number.replace('(','').replace(')','');
```
azasypkin: Hey, you solution works, but I think Julien's suggestion from https://bugzilla.mozilla.org/show_bug.cgi?id=1120572#c7 will be much nicer - could you please follow the suggestion and make it via regex? Your regex should be declared right after ```rdigit``` in this file.

apps/sms/views/conversation/test/unit/recipients_test.js
```diff
@@ -143,6 +143,24 @@ suite('Recipients', function() {
       assertDeepEqual(recipient, fixture);
     });
 
+    test('recipients.add() parenthesis ', function(done) {
```
azasypkin: __nit:__ test code looks misaligned, also please use 2 spaces for indentation like in the rest of the file

apps/sms/views/conversation/test/unit/recipients_test.js
```diff
@@ -143,6 +143,24 @@ suite('Recipients', function() {
       assertDeepEqual(recipient, fixture);
     });
```
azasypkin: __typo:__ no need in trailing space in the test name, also maybe it will be better to have ```..... leading parenthesis ....```

apps/sms/views/conversation/test/unit/recipients_test.js
```diff
@@ -143,6 +143,24 @@ suite('Recipients', function() {
...
 
+    test('recipients.add() parenthesis ', function(done) {
+        recipients.on('add', function(count, added) {
```
azasypkin: __issue:__ Mmm can't you just use the same approach as in the test below (without ```done```, subscribing for the ```add``` event and using ```isValid``` method)? Moreover test doesn't work as expected, as you call ```done``` for the first number and the second one is not even checked :)

So what about improving ```isValid``` method above by adding additional checks for the ```candidate.isQuestionable``` and ```candidate.isValid``` there as well?

apps/sms/views/conversation/test/unit/recipients_test.js
```diff
@@ -143,6 +143,24 @@ suite('Recipients', function() {
...
+          assert.isFalse(added.isQuestionable);
+          recipients.off('add');
+          done();
+        });
+
+        recipients.add({
```
azasypkin: Also it would be great if you can add separate ```test``` for the leading ```+``` as well if it doesn't exist in this file.

Thanks!

apps/sms/views/conversation/test/unit/recipients_test.js
adarshsanjeev: I was following the format of the tests above and below it, which seem to have trailing spaces in the name. Also, where should I put the leading parenthesis?

apps/sms/views/conversation/test/unit/recipients_test.js
adarshsanjeev: I am not sure if what I have written is write. Any chance you could point out anything I am doing wrong? Or explanation on this particular case?

apps/sms/views/conversation/test/unit/recipients_test.js
azasypkin: > I was following the format of the tests above and below it, 

We may want to have trailing space in the ```suite``` name, so that when test results a printed - you'll have space between suite and test names. Since the test is the "most bottom node" you don't need trailing space for it.

>  Also, where should I put the leading parenthesis?

I just meant that your test name should say explicitly that it handles the case for leading parenthesis only. So instead of naming test as ```recipients.add() parenthesis``` you'd better say ```recipients.add() with leading parenthesis```

apps/sms/views/conversation/test/unit/recipients_test.js
azasypkin: > I am not sure if what I have written is write. Any chance you could point out anything I am doing wrong? Or explanation on this particular case?

First of all as I said above, you call ```recipients.add``` twice so recipients ```add``` event handler is supposed to be run twice - but as soon as you call ```done()``` your test will be marked as finished - so the second ```recipients.add``` operation is not even checked - I'd suggest you to open Dev Tools and debug - to see what is really happening here.

Second thing is that you don't even need this ```add``` event handler at all - see the example in the test below - it doesn't deal with event handlers, so code looks cleaner.

Third thing is that you can reuse ```isValid``` method to validate that number is marked as valid (this method is used in the test below as well). But I see that ```isValid``` is too naive and may return ```true``` for the ```questionable``` number, so I'd suggest expand it with two additional checks ```candidate.isQuestionable``` and ```candidate.isValid```. 

Basically you should follow the rule - make test, see that it fails without your changes in code and succeed with them. Hope it helps

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=dc9b33de988aafe7783bf89f3bb35459c5b71634

PR #34039, opened 2016-02-05
Bug 1246092 - [TV] Circle animation during app transition dosen't close properly

Fischer-L:
This patch:
If foreground app being closed during app transition is closed before app transition animation ends, it will causes null exception of foreground app and make app transition animation unable to close properly.
This patch add a check to the existence of foreground app being closed.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=c5eda655cbdb5eb63c198ccb30015b9bed080341

PR #34050, opened 2016-02-10
Bug 1040874 - [Messages] Add the phone number actions to the Thread m…

martiansideofthemoon:
…enu as well. r=julienw

### DIFF
--- a/apps/sms/views/conversation/js/conversation.js
+++ b/apps/sms/views/conversation/js/conversation.js
@@ -1933,13 +1933,34 @@ var ConversationView = {
       });
     }
 
-    // Last item is the Cancel button
-    params.items.push({
-      l10nId: 'cancel',
-      incomplete: true
-    });
-
-    new OptionMenu(params).show();
+    var participants = this.activeThread.participants;
+    // Do I need to do this when only one participant?
+    if (participants) {
+      Contacts.findByAddress(participants[0]).then(function(results) {
+        if (results && results.length) {
+          var contact = results[0];
+          var props = [{ id: contact.id }];
+          params.items.push({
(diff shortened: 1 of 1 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7dfab84d4861e1c028f2a1f1b99a9f3ae03472d7

apps/sms/views/conversation/js/conversation.js
```diff
@@ -1933,13 +1933,34 @@ var ConversationView = {
...
+          var contact = results[0];
+          var props = [{ id: contact.id }];
+          params.items.push({
+            l10nId: 'viewContact',
+            method: () => ActivityPicker.viewContact(props),
+            params: props
+          });
+        }
+        // Last item is the Cancel button
+        params.items.push({
```
kumarrishav: We will cancel button in any case. so no need to write twice. Just add 'cancel' at last.
So, Basically you need to check : 
if (contact is saved) 
add 'view contact' option in option menu.
else
add 'create new contact'  and  'add to an existing contact' 
i guess, you missed these options? 

you can see the expected behaviour by tapping on header in both case (when contact is saved and when it's not).
and you can see it's code for reference and reuse it if u want.
https://github.com/mozilla-b2g/gaia/blob/master/apps/sms/views/conversation/js/conversation.js#L2888
https://github.com/mozilla-b2g/gaia/blob/master/apps/sms/views/conversation/js/conversation.js#L2966

julienw: yes the goal of the bug is to extract the code from prompt() so that it can be reused both here and in prompt().

PR #34075, opened 2016-02-20
Bug 1249256 - [layout](uk) Remove key by moving it to alternative character

artpoli:
Also removing unnecessary string in keyboard_test.js

### DIFF
--- a/apps/keyboard/js/layouts/uk.js
+++ b/apps/keyboard/js/layouts/uk.js
@@ -11,14 +11,15 @@ Keyboards.uk = {
     alt: {
       а: '@',
       г: 'ґ',
+      і: 'ї',
       '.': ',?!;:…’'
     },
-    width: 12,
+    width: 11,
     keys: [
       [
         { value: 'й' }, { value: 'ц' }, { value: 'у' }, { value: 'к' },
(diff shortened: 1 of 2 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=154fea93b0e845214a138df721b0c87aa5754977

apps/keyboard/js/layouts/uk.js
```diff
@@ -11,6 +11,7 @@ Keyboards.uk = {
     alt: {
       а: '@',
       г: 'ґ',
+      і: 'ї',
       '.': ',?!;:…’'
     },
     width: 12,
```
timdream: All of the rows are of 11 keys... I am not entirely sure any effect if you keep the `width` 12. If this is fine we could keep it as-is.

stenox:
@timdream Yes, you're right. I should reduce value to 11. Will change it and reply to the bug.
Thanks!

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=8633318d298bedcf0a6d628277fee2c29b6dc450

PR #34085, opened 2016-02-23
Bug 1242634 - test_logout.py : ElementNotVisibleException: ElementNotVisibleException: Element is not currently visible and may not be manipulated

ubermarklar:
Attempt 2 after having Git related issues. This one line change fixes the test by refining the pointer for the profile icon. It was previously defined so that the script thought it was looking for a hidden element, this change fixes that.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=37693720645f1ab14a4fc337278a0316ff38036d

PR #34216, opened 2016-04-02
Bug 1261549 - Gaia does not build due to manifest error

nhirata:
Removing mochitest resolves the issue.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=cc38a8a5386508dab93a52af58fd69e1640f33eb

nhirata:
Um, so ya, based on the comment, if mochitest was running before, we should keep it running.

PR #34240, opened 2016-04-14
Bug 1263840 - [TV][Home] Move card description text to outside of but…

rexboy7:
…tons

### DIFF
--- a/tv_apps/smart-home/js/card_picker.js
+++ b/tv_apps/smart-home/js/card_picker.js
@@ -13,7 +13,7 @@
     hideCardPickerButton: document.getElementById('hide-cardpicker-button'),
 
     init: function(options) {
-      this.appButtons = [];
+      this.appCardElems = [];
 
       this._cardManager = options.cardManager;
       this._folder = null;
@@ -37,7 +37,8 @@
       });
       this._keyNavigationAdapter.on('enter-keyup', this.onEnter.bind(this));
 
-      this._selectedElements = this.gridView.getElementsByClassName('selected');
+      this._selectedButtons = this.gridView.getElementsByClassName('selected');
+      this._appButtons = this.gridView.getElementsByTagName('smart-button');
 
       this.refresh();
     },
@@ -86,10 +87,10 @@
 
       this._cardManager.getCardList()
         .then(cardList => {
-          this._refreshCardButtons(folderList, cardList);
+          this._refreshCardElements(folderList, cardList);
           this._spatialNavigator.setCollection(
-                            this.appButtons.concat(this.navigableElements));
-          this._spatialNavigator.focus(this.appButtons[0]);
+                      this.navigableElements.concat(Array.from(this.allItems)));
+          this._spatialNavigator.focus(this.allItems[0]);
         });
     },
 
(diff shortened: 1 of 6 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=87cf208b31d5d79c89cf8ab31c91ba5a592ad194

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=216d96f743aeb2b0fcef682b0125ced023538080

tv_apps/smart-home/style/card-picker.css
```diff
@@ -30,21 +37,29 @@
...
+#card-picker smart-button[type="app-button"].focused {
+  transform: scale(1.2);
+}
+
 .card-picker-panel {
   top: 80%;
   position: absolute;
 }
+
+card-picker-grid-view
```
luke-chang: ???

tv_apps/smart-home/js/card_picker.js
```diff
@@ -37,7 +37,8 @@
       });
       this._keyNavigationAdapter.on('enter-keyup', this.onEnter.bind(this));
 
-      this._selectedElements = this.gridView.getElementsByClassName('selected');
+      this._selectedButtons = this.gridView.getElementsByClassName('selected');
+      this._appButtons = this.gridView.getElementsByTagName('smart-button');
 
       this.refresh();
```
luke-chang: Not related to this bug, but ... Do we need to call `refresh` in the `init` stage? I suppose it should be called again when it's about to show, right?

tv_apps/smart-home/js/card_picker.js
```diff
@@ -86,10 +87,10 @@
 
       this._cardManager.getCardList()
         .then(cardList => {
-          this._refreshCardButtons(folderList, cardList);
+          this._refreshCardElements(folderList, cardList);
           this._spatialNavigator.setCollection(
-                            this.appButtons.concat(this.navigableElements));
-          this._spatialNavigator.focus(this.appButtons[0]);
+                      this.navigableElements.concat(Array.from(this.allItems)));
+          this._spatialNavigator.focus(this.allItems[0]);
```
luke-chang: nit: You can simply call `focus()` without arguments if you intend to focus the first element in the collection.

tv_apps/smart-home/js/card_picker.js
```diff
@@ -103,34 +104,36 @@
...
 
       folderList && folderList.forEach(card => {
-        var appButton = createButtonHelper(card);
-        if (appButton) {
-          appButton.dataset.parentType = 'folder';
-          appButton.classList.add('selected');
+        var appCardElem = createCardElemHelper(card);
+        if (appCardElem) {
+          var cardButton = appCardElem.firstElementChild;
+          cardButton.dataset.parentType = 'folder';
```
luke-chang: How about put `dataset.parentType` in `createCardElemHelper` if it is a required filed?

tv_apps/smart-home/js/home.js
```diff
@@ -305,7 +305,8 @@
     onCardUpdated: function(card, idx) {
       var cardButton = this.cardScrollable.getItemFromNode(
                                               this.cardScrollable.getNode(idx));
-      CardUtil.updateCardName(cardButton, card);
+      // nextElementSibling of cardButton is the name span.
+      CardUtil.updateCardName(cardButton.nextElementSibling, card);
```
luke-chang: Why not retrieve the name span by something like `this.cardScrollable.getNode(idx).querySelector('.name')`? It would be more readable than `nextElementSibling`.

tv_apps/smart-home/js/card_util.js
```diff
@@ -187,19 +188,18 @@
...
-    updateCardName: function(cardButton, card) {
-      var spans =
-           SharedUtils.nodeListToArray(cardButton.getElementsByTagName('span'));
-      spans.forEach(span => {
-        if (span.classList.contains('name')) {
-          this._localizeCardName(span, card);
-        }
-      });
+    updateCardName: function(nameElement, card) {
+      if (nameElement.classList.contains('name')) {
```
luke-chang: If everyone call this API correctly, I suppose we don't need to verify its classList. (we may just need to check if it's `null`.)

tv_apps/smart-home/js/card_picker.js
rexboy7: Yeah, I think you're right.

tv_apps/smart-home/js/card_picker.js
rexboy7: I think `allItems[0]` is not the first item of collection here (but in old version it is). But I can change it back to the old version and remove the parameter if you think that's better.

tv_apps/smart-home/js/home.js
rexboy7: Agree. Let's change it.

tv_apps/smart-home/js/card_util.js
rexboy7: I don't completely understand the old code's logic so I just kept that. But I can remove it and go back if we find some isuse.

tv_apps/smart-home/style/card-picker.css
rexboy7: my bad :-/

tv_apps/smart-home/js/card_picker.js
luke-chang: Oh, it's my fault. I misunderstood so you don't have to change here.

tv_apps/smart-home/js/card_util.js
luke-chang: I guess the original logic behind the `spans` depends on `SharedUtils.nodeListToArray` so we need to filter that. Since this function is only called by `home.js` now (changes below), we can make sure `nameElement` should be always valid.

tv_apps/smart-home/js/card_picker.js
rexboy7: For now `parentType` is used only in `updateFolder` to determine if we should move it out of folder (if it was a folder's card), or put it into a folder (if it was a main list's card), or do nothing. This is not used elsewhere since there's no other usecases that need it.

luke-chang: Actually, I meant you can define `function createCardElemHelper(card, parentType)`, then pass `createCardElemHelper(card, 'folder')` here and pass `createCardElemHelper(card, 'empty')` below. However, it's up to you. :-p

PR #34256, opened 2016-04-21
Bug 1257824 - Get the Places database working without using DataStore

benfrancis:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=4e67152b5e6c753c735487f81e8712f661f9f662

apps/system/js/places.js
```diff
@@ -295,20 +379,71 @@
...
+     * Get all pinned sites.
+     *.
+     * @returns {Promise} A promise which resolves with the full set of results.
+     */
+    getPinnedSites: function() {
+      var results = [];
+      var transaction = this.db.transaction(this.SITES_STORE);
+      var objectStore = transaction.objectStore(this.SITES_STORE);
+      return new Promise((function(resolve, reject) {
+        objectStore.openCursor().onsuccess = function(event) {
```
albertopq: Is not this returning all the sites, and not only the pinned ones? If the idea is that all the sites stored in the SITES_STORE are pinned, can we remove the index?

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=d83796af0b94196fc53287dcbe6d6039e8baa969

apps/system/js/places.js
benfrancis: Yes it is, but currently all sites are pinned. It turns out you can't create an IndexedDB index on a boolean property. I've removed the pinned index and just added a conditional to filter on the fly for now instead.

PR #34290, opened 2016-05-05
Bug 1270625 - Add extend task graph scope to decision task r=auswerk

gregarndt:
This change should allow the decision task to work with new versions of our worker, which use a set of temporary credentials per task based on task.scopes.


djmitche:
:thumbsup:

nullaus:
👍

PR #34303, opened 2016-05-11
Bug 1268827 - [TV][Home] Focus remains in homescreen when first time entering rename dialog

Fischer-L:
When pressing down to hide keyboard under the prompt for the 1st time, it would trigger the injection of the value selector element. The value selector element is rendered without the hidden set initially.

In the end of the injection operation, the operation would hide the value selector and call focus on the background app since it is not hidden. As a result, the foreground prompt loses the focus.

From the 2nd time on, because the value selector has already been hidden. the foreground prompt would not loses the focus.

We should call focus on the focusManager not on the background app directly so as to let the focusManager control which element is the right one to gain focus.

### DIFF
--- a/tv_apps/smart-system/js/value_selector/value_selector.js
+++ b/tv_apps/smart-system/js/value_selector/value_selector.js
@@ -1,5 +1,5 @@
 /* global BaseUI, LazyLoader, InputParser, ValueSelector, SpinDatePicker,
-          ValuePicker, Template */
+          ValuePicker, Template, focusManager */
 
 'use strict';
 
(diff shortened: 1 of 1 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=3290894316177e66e7d257213fc9003c1266a510

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a9f315f534ece74f98dfdaa0f4b0bd5e6d5a0cf5

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=a9f315f534ece74f98dfdaa0f4b0bd5e6d5a0cf5

PR #34307, opened 2016-05-12
Bug 1270393 - [TV][Home] Always prompt confirmation dialog if press back key on card selection page of creating folder

Fischer-L:
This patch implements the specs. When adding folder, but pressing Back key without no card selected, we should just go back to home screen directly and no confirming prompt is needed.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=71411cdef6edf12ffaca8a425da2bc63fc083378

PR #34323, opened 2016-05-17
Bug 1273413 - [TV][Home] Navigation in card selection page doesn't su…

luke-chang:
…pport to wrap around in both directions

### DIFF
--- a/tv_apps/smart-home/js/card_picker.js
+++ b/tv_apps/smart-home/js/card_picker.js
@@ -285,6 +285,29 @@
         });
         candidates.forEach(candidate => appendToGridView(candidate.element));
       }
+
+      var cardsPerRow = 0;
+      var firstRowTop;
+
+      this.appCardElems.some(elem => {
+        var rect = elem.getBoundingClientRect();
+        if (firstRowTop && rect.top != firstRowTop) {
+          return true;
+        }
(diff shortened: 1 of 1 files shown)


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=28f2968b837d3334d655d2d6bc06008de83b41ee

tv_apps/smart-home/js/card_picker.js
```diff
@@ -248,6 +250,18 @@
       function appendToGridView(appCardElem) {
         that.gridView.appendChild(appCardElem);
         that.appCardElems.push(appCardElem);
+
+        if (top >= 0) {
```
rexboy7: The logic here is not descriptive enough, including the meaning of `top = -1` and variable name of `cr`.

rexboy7: I would prefer to move this paragraph out of `appendToGridView`, and let it be an independent for-loop. After you found the first row you can just break the for-loop. I think that should be more easier to read.

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=15c42f2c7f006d5eb9004d1222fb170a87cc7c45

PR #34335, opened 2016-05-18
Bug 1273839 - Use WebManifest format for app icons r=Cwiis

albertopq:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=9e30800429cb921c18309a90a63ecec3357aa5fa

shared/js/web_manifest_helper.js
```diff
@@ -160,7 +160,7 @@
           return true;
         }
         size = size.toLowerCase();
-        if (!size.contains('x') || size.indexOf('x') !==
+        if (!size.indexOf('x') || size.indexOf('x') !==
```
Cwiiis: This condition isn't correct anymore. If it was a clever attempt at making the size 'x123' invalid, that's neat, but this also makes a size with no 'x' valid, which I'm guessing isn't valid, and certainly isn't expected by the code that follows.

Really, this should probably just use a regex match, like `if (!size.match(/^\d+x\d+$/))`

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=7bf1a89bc528c59132910406e2d26dc0b4e2680d

PR #34378, opened 2016-06-07
Bug 1277836 - Open webapps on system messages r=gsvelto

albertopq:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2d97b2209c296694c94bf9edbe0ce6c64d260c01

apps/system/js/system_messages.js
```diff
@@ -1,8 +1,17 @@
...
+  var MESSAGES = {
+    'bluetooth-dialer-command': '/dialer/index.html#keyboard-view',
+    'headset-button': '/dialer/index.html#keyboard-view',
+    'notification': '/dialer/index.html#keyboard-view',
+    'telephony-new-call': '/dialer/index.html#keyboard-view',
+    'telephony-call-ended': '/dialer/index.html#keyboard-view',
+    'ussd-received': '/dialer/index.html#keyboard-view'
+  };
+
   var SystemMessages = function(core) {
```
gabrielesvelto: nit: Maybe we should remove this?

apps/system/js/system_messages.js
```diff
@@ -1,8 +1,17 @@
...
+/* global BaseModule, BroadcastChannel, Service */
 'use strict';
 
 (function() {
-  // System Messages placeholder
+  var BASE_URL = 'chrome://gaia/content';
+  var MESSAGES = {
+    'bluetooth-dialer-command': '/dialer/index.html#keyboard-view',
+    'headset-button': '/dialer/index.html#keyboard-view',
+    'notification': '/dialer/index.html#keyboard-view',
```
gabrielesvelto: Right now the dialer is the only app that will post notifications but we'll need a way to figure out how to steer them towards the right app once we get more apps working into chrome. So this works for now but let's keep in mind that it won't cut it once we have more than one app excepting one message.

PR #34405, opened 2016-07-05
Bug 1275076 - convert the manifest of the clock app r=albertopq

TitanNano:


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0a34cb25a95ae678e070c5ded96f980b4bd37435

disabled_apps/clock/manifest.webmanifest
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "Clock",
+  "description": "Gaia Clock",
+  "lang": "en",
+  "short_name": "Clock",
+  "scope": "./",
```
benfrancis: I'm not sure this is a valid scope. If the Clock app is hosted at "http://mozilla-b2g.github.io/gaia/dead_apps/clock/" then the scope would be "/gaia/dead_apps/clock/"

Or you could leave the scope property out

disabled_apps/clock/manifest.webmanifest
```diff
@@ -0,0 +1,34 @@
...
+  "start_url": "./index.html?homescreen=1",
```
benfrancis: What is the ?homescreen=1 for?

disabled_apps/clock/manifest.webmanifest
```diff
@@ -0,0 +1,34 @@
...
+  "launch_path": "/index.html",
```
benfrancis: We shouldn't need both start_url and launch_path. Is start_url not working for something?

disabled_apps/clock/manifest.webmanifest
```diff
@@ -0,0 +1,34 @@
...
+  "developer": {
```
benfrancis: This isn't a valid W3C web app manifest property

disabled_apps/clock/manifest.webmanifest
```diff
@@ -0,0 +1,34 @@
...
+    "name": "The Gaia Team",
+    "url": "https://github.com/mozilla-b2g/gaia"
+  },
+  "default_locale": "en-US",
```
benfrancis: This isn't a valid W3C web app manifest property

disabled_apps/clock/manifest.webmanifest
```diff
@@ -0,0 +1,34 @@
...
+  "icons": [{
+    "sizes": "84x84",
+    "src": "./style/icons/clock_84.png"
```
benfrancis: I think this should be "style/icons/clock_84.png", relative to the directory the manifest is in. But I could be wrong.

disabled_apps/clock/manifest.webmanifest
TitanNano: The scope is resolved against the manifest url, so this should work, since our manifest and our index.html sit in the same directory. I think it's more convenient to keep the scope url relative.

https://www.w3.org/TR/appmanifest/#scope-member

disabled_apps/clock/manifest.webmanifest
TitanNano: @benfrancis we don't need it, but this let's the app detect that it was launched from the homescreen.

disabled_apps/clock/manifest.webmanifest
TitanNano: @benfrancis nope, you are right, I forgot to remove it.

disabled_apps/clock/manifest.webmanifest
TitanNano: well "./style/" and "style/" should behave the same :-)

autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=77b1fef166cd08beff8a27847aef4725ff898dd7

PR #34418, opened 2016-09-12
Bug 123472 - Remove :-moz-full-screen-ancestor from gaia code R=fabrice

asppsa:
Added a FullscreenManager to system core that is responsible for
adding/removing a class from #screen when there is a fullscreen
descendent.

This patch reinstates the soft home button.


autolander:
Running tests
https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=2e148b4200be52c54da7f9e94b8b054f41655e3c

PR #34425, opened 2017-01-10
Bug 811624 lock passcode

Neeraj-Seth:


mozilla-autolander-deprecated:

